PR #67, opened 2017-03-21
Makes the codebase more friendly to multiple meters being queried at once

simonbtomlinson:
Modifies compressed_readings to take an array of meter_ids since this is a common use-case.

Patches the momentjs object to be compatible with pg-promise

Refactors start/end timestamps into a common (across client and server) TimeInterval class


src/server/sql/reading/create_function_get_compressed_readings.sql
```diff
@@ -7,11 +7,11 @@
 	returns a query
  */
 CREATE OR REPLACE FUNCTION compressed_readings(
```
simonbtomlinson: That's something you want to do right before displaying them on the graph.

simonbtomlinson:
This ended up effectively getting merged as part of https://github.com/beloitcollegecomputerscience/OED/pull/69. Probably not the most kosher way to do it, but what's done is done.

PR #79, opened 2017-04-06
Add bidirectional moment and moment.duration support in pg-promise

simonbtomlinson:
This adds bidirectional conversion (to and from the database) for
moment objects <=> timestamps and moment.duration objects <=> intervals

This should mean that all date code can use moment types instead of
naive js Date objects.

Any code in other branches that uses js dates may need to be modified to expect moment
objects instead of js dates.


huss:
Can you explain the advantage of using moment types?

simonbtomlinson:
@huss
We've been using moment since December, so it's not really anything new. We originally added moment because it is really good at parsing strings into datetimes and we need to do that to import basically anything. Since then, we've been using moment because it supports a bunch of​ useful stuff that you'd think js dates would support easily, like adding or subtracting an interval from a date or formatting it in a customizable way. This is especially true in unit tests, where a lot of manual date manipulation occurs to set up test data. Take a look [here](https://github.com/beloitcollegecomputerscience/OED/blob/03393fa6dd742e92cd4ec96b743b75cb00bac843/src/server/test/db/compressedReadingsTests.js#L23).

[This stackoverflow answer sums up the advantages of moment pretty well](http://stackoverflow.com/a/1214753)

Before this pull request, moment dates were in use in almost all of the project, but pg-promise returned normal js dates. This pull request just unifies pg-promise with the rest of the project and adds supposed for moment durations (which we're going to need for bar charts).

package.json
```diff
@@ -38,6 +38,7 @@
     "moment": "^2.15.2",
     "morgan": "~1.7.0",
     "nodemailer": "~2.6.4",
+    "pg": "^5.1.0",
```
johndiiorio: Don't we already have pg-promise? What is the point of adding pg?

simonbtomlinson: We already had `pg` as a transitive dependency from `pg-promise`, but I actually interacted with it to change return types from postgres [here](https://github.com/beloitcollegecomputerscience/OED/pull/79/files#diff-089faff11224ef4b4082159c0a2b9d40). I thought it was a good idea to add it as an explicit dependency since we're explicitly using it.

vitaly-t: This is the wrong way of doing it. One must NOT include `pg` when `pg-promise` is used already. See: [How to access the instance of node-postgres that's used?](https://github.com/vitaly-t/pg-promise/wiki/FAQ#how-to-access-the-instance-of-node-postgres-thats-used).

simonbtomlinson: Oh wow, thanks so much. I didn't see that in the docs.

simonbtomlinson:
The travis checks are failing because travis hasn't finished being set up yet.

package.json
simonbtomlinson: Fixed in e1ab4b0

PR #83, opened 2017-04-13
Export group

kooolbreez1:
This adds a button to the UI that when clicked exports the compressed data from the displayed graph for the currently selected meters as a CSV file.


kooolbreez1:
I linted the file and deleted a blank space, I also accidentally changed a comment in dashboardContainer which has now been fixed.

src/client/app/actions/exportData.js
```diff
@@ -0,0 +1,51 @@
...
+function convertToCSV(items) {
+	let csvOutput = 'id:,readings:,timestamp:\n';
+	items.forEach(set => {
+		const data = set.exportVals;
+		const id = set.id;
+		let isStart = true;
+		data.forEach(reading => {
+			const info = reading.y;
+			const timeStamp = moment(reading.x).format('dddd MMM DD YYYY hh:mm a');
+			if (isStart) {
```
paulmtwomey: What is this conditional for?

src/client/app/actions/exportData.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * A function to download the formatted CSV file to the users computer.
+ * @param inputCSV A String containing the formatted CSV data.
+ */
+function downloadCSV(inputCSV) {
+	const csvContent = `data:text/csv;charset=utf-8,${inputCSV}`;
+	const encodedUri = encodeURI(csvContent);
+	const link = document.createElement('a');
+	let fileName = 'exportedDataOED';
+	fileName += '.csv';
```
paulmtwomey: Can do as one line?

src/client/app/services/exportData.js
```diff
@@ -0,0 +1,48 @@
...
+import moment from 'moment';
+
+/**
+ * Function to converts the compressed meter data into a CSV formatted string.
+ * @param items The compressed meter data.
+ * @returns output A string containing the CSV formatted compressed meter data.
+ */
+
+function convertToCSV(items) {
+	let csvOutput = 'id:,readings:,timestamp:\n';
```
johndiiorio: You should change `timestamp` to `start_timestamp` and `end_timestamp`. You should also remove the colons in the labels.

src/client/app/services/exportData.js
```diff
@@ -0,0 +1,48 @@
...
+	items.forEach(set => {
+		const data = set.exportVals;
+		const id = set.id;
+		data.forEach(reading => {
+			const info = reading.y;
+			const timeStamp = moment(reading.x).format('dddd MMM DD YYYY hh:mm a');
+			csvOutput += `${id},${info} kwh, ${timeStamp} \n`;
```
johndiiorio: In addition to the meter ID, I would add the name of the meter. As meter IDs are internal to our system, someone would have no idea what the meter ID represented.

src/client/app/services/exportData.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param inputCSV A String containing the formatted CSV data.
+ */
+function downloadCSV(inputCSV) {
+	const csvContent = `data:text/csv;charset=utf-8,${inputCSV}`;
+	const encodedUri = encodeURI(csvContent);
+	const link = document.createElement('a');
+	const fileName = 'exportedDataOED.csv';
+	link.setAttribute('href', encodedUri);
+	link.setAttribute('download', fileName);
+	document.body.appendChild(link);
```
johndiiorio: I'm not an expert in this field, but is this the `React` way to do things? Also, after clicking the link, it might be prudent to remove this link.

src/client/app/services/exportData.js
```diff
@@ -0,0 +1,48 @@
...
+}
+/**
+ * Function to download the formatted CSV file to the users computer.
+ * @param inputCSV A String containing the formatted CSV data.
+ */
+function downloadCSV(inputCSV) {
+	const csvContent = `data:text/csv;charset=utf-8,${inputCSV}`;
+	const encodedUri = encodeURI(csvContent);
+	const link = document.createElement('a');
+	const fileName = 'exportedDataOED.csv';
```
johndiiorio: Is there a more descriptive and "dynamic" name that we can give the file name? Perhaps include the start and end timestamps in the filename?

src/client/app/containers/UIOptionsContainer.js
```diff
@@ -14,9 +14,24 @@ import { fetchMetersDataIfNeeded } from '../actions/meters';
  */
 function mapStateToProps(state) {
 	const sortedMeters = _.sortBy(_.values(state.meters.byMeterID).map(meter => ({ id: meter.id, name: meter.name.trim() })), 'name');
+
+	const timeInterval = state.graph.timeInterval;
+	const data = { datasets: [] };
+	for (const meterID of state.graph.selectedMeters) {
+		const readingsData = state.readings.byMeterID[meterID][timeInterval];
```
johndiiorio: As of right now, this is completely correct. However, when the bar chart is added, readings will be retrieved from state.[line OR bar].readings...
Will the intention be to only grab line chart readings, bar chart readings, or both? If both, some additional logic will eventually be required here.

src/client/app/services/exportData.js
simonbtomlinson: This is definitely **not** the react way to do things. This throws an `<a>` element onto the page outside of the react root div, then clicks it. This is definitely not the right way to do things.
I could see *maybe* making an `<a>` element, clicking it, then removing it, but I'm sure there's a better way.

src/server/services/readMetasysData.js
```diff
@@ -10,23 +10,23 @@ const Meter = require('./../models/Meter');
 async function readMetasysData(filePath) {
```
simonbtomlinson: Thanks for cleaning this up a bit while you were doing other things. It slipped through linting somehow. I see there are a few other linting errors in here, but they aren't your code so it's no big deal.

src/client/app/services/exportData.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Function to converts the compressed meter data into a CSV formatted string.
+ * @param items The compressed meter data.
+ * @returns output A string containing the CSV formatted compressed meter data.
+ */
+
+function convertToCSV(items) {
+	let csvOutput = 'id:,readings:,timestamp:\n';
+	items.forEach(set => {
```
simonbtomlinson: You use a `for ... of` loop in `UIOptionsContainer.js`, but `.forEach` here. Why the inconsistency?

src/client/app/containers/UIOptionsContainer.js
```diff
@@ -14,9 +14,24 @@ import { fetchMetersDataIfNeeded } from '../actions/meters';
  */
 function mapStateToProps(state) {
 	const sortedMeters = _.sortBy(_.values(state.meters.byMeterID).map(meter => ({ id: meter.id, name: meter.name.trim() })), 'name');
+
+	const timeInterval = state.graph.timeInterval;
+	const data = { datasets: [] };
```
simonbtomlinson: `data` seems to just be a wrapper around `datasets`. Why does it exist at all? Why not just use the `datasets` list as your variable?

kooolbreez1:
The export component now works with bar charts and the file name now changes based on the timestamp of the first reading. It still only exports each readings start time but as far as I can tell the charts only ever receive the start time of each reading. Maybe I am looking in the wrong place? I have also updated the actual save function to be a bit better but it essentially does the same thing as before.

simonbtomlinson:
@kooolbreez1You can get the current bar duration from `state.graph.barDuration` in redux. Then just add it to start points to get end points.

kooolbreez1:
that would work for the bar chart readings but not the line chart readings

simonbtomlinson:
@kooolbreez1 Oh! Yeah, currently the only way to figure that out is to subtract the start of the second reading from the first one. They'll all have the same duration. We should probably start sending end timestamps to the client to do this more nicely.

src/client/app/components/ExportComponent.jsx
```diff
@@ -0,0 +1,34 @@
...
+	exportReading() {
+		const compressedData = this.props.exportVals.datasets;
+		let time = compressedData[0].exportVals[0].x;
+		time = moment(time).format('ddddMMMDDYYYY');
+		const name = `oedExport${time}.csv`;
+		graphExport(compressedData,	name);
+	}
+
+	render() {
+		return (<button onClick={this.exportReading}>Export!</button>);
```
johndiiorio: Could use change this to a Bootstrap button? Preferably btn-default? Ideally, add the React-Bootstrap module and use it through that.

src/client/app/containers/ExportContainer.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+function mapStateToProps(state) {
+	const timeInterval = state.graph.timeInterval;
+	const data = { datasets: [] };
+	let readingsData;
+	const chart = state.graph.chartToRender;
+	const barDuration = state.graph.barDuration;
+
+	for (const meterID of state.graph.selectedMeters) {
+		if (chart === 'line') {
```
johndiiorio: In reducers/graph.js, there is `chartTypes`. Even though it doesn't really do anything, you should use them instead of the string 'line' and 'bar' so that we don't get confused when we start adding more types of charts.

src/client/app/containers/ExportContainer.js
```diff
@@ -0,0 +1,48 @@
...
+			readingsData = state.readings.line.byMeterID[meterID][timeInterval];
+		}		else { readingsData = state.readings.bar.byMeterID[meterID][timeInterval][barDuration]; }
+		if (readingsData !== undefined && !readingsData.isFetching && chart === 'line') {
+			data.datasets.push({
+				label: state.meters.byMeterID[meterID].name,
+				id: state.meters.byMeterID[meterID].id,
+				timestamp: state.readings.line.byMeterID[meterID][timeInterval].start_timestamp,
+				exportVals: state.readings.line.byMeterID[meterID][timeInterval].readings.map(arr => ({ x: arr[0], y: arr[1] }))
+			});
+		} else if (readingsData !== undefined && !readingsData.isFetching && chart === 'bar') {
```
johndiiorio: See above comment.

src/client/app/components/ExportComponent.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+	/**
+	 * Called when Export button is clicked.
+	 * Passes an object containing the selected meter data to a function for export.
+	 */
+	exportReading() {
+		const compressedData = this.props.exportVals.datasets;
+		let time = compressedData[0].exportVals[0].x;
+		time = moment(time).format('ddddMMMDDYYYY');
+		const name = `oedExport${time}.csv`;
```
johndiiorio: Could you add in whether it is bar or line data in the title?

kooolbreez1:
I believe I have made the changes that you requested @johndiiorio the only thing is that the export button doesn't display in line with the UIoptions panel, rather it appears in the bottom left corner of the screen

src/client/app/components/HomeComponent.jsx
```diff
@@ -15,6 +17,7 @@ export default function HomeComponent() {
 		<div>
 			<HeaderComponent renderLoginButton />
 			<DashboardContainer />
+			<ExportContainer />
```
johndiiorio: Could you move this to be in UIOptionsComponent (at the bottom) instead of here?

kooolbreez1:
@johndiiorio All fixed, it is now formatted nicely.

src/client/app/components/ExportComponent.jsx
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { Button } from 'react-bootstrap';
+import moment from 'moment';
+import graphExport from '../services/exportData';
+
+export default class ExportComponent extends React.Component {
```
simonbtomlinson: This could easily be a functional component. No need for a class.

kooolbreez1:
I believe I have made the changes that @simonbtomlinson wanted.

simonbtomlinson:
@kooolbreez1 I think so too. I'll take a look when I get a chance, tomorrow at the latest.

Thanks!

PR #85, opened 2017-04-17
Barchart setup

johndiiorio:
Create bar chart, remove unused UI options, add project logo.


johndiiorio:
I'm not exactly sure why the one test failed (possibly too many rows for Travis to handle). The test run fine on my machine (in 435 ms).

src/server/services/readMetasysData.js
```diff
@@ -10,38 +10,38 @@ const Meter = require('./../models/Meter');
...
 		let meterReading = row[3];
 		meterReading = meterReading.replace('kWh', '');
 		meterReading = parseInt(meterReading);
 
 		/* We have hourly trend and monthly trend for Metasys data. We just read hourly trend.
 		 * So, we skip one line as we go through the data
 		 */
-		if ((i % 2 != 0)) {
-			const reading = new Reading(meter.id, meterReading, start_timestamp, end_timestamp);
+		if ((i % 2 !== 0)) {
```
jameeters: Redundant parentheses.

README.md
```diff
@@ -16,16 +16,16 @@ Node.js - Javascript runtime environment ([nodejs.org](https://nodejs.org/en/))
 1. Clone this repository.
 1. Run ```npm install``` in the project root directory.
 1. Install PostgreSQL, start the PostgreSQL server, and connect to it via psql.
-1. In psql, run ```CREATE DATABASE environmental_dashboard;``` to create the database.
-1. Still in psql, run ```CREATE DATABASE environmental_dashboard_testing;``` to create a database for automated tests.
+1. In psql, run ```CREATE DATABASE oed;``` to create the database.
+1. Still in psql, run ```CREATE DATABASE oed_testing;``` to create a database for automated tests.
 1. Create a .env file in the root directory of the project with the following, replacing (?) with the desired information: <br>
-&nbsp;&nbsp;&nbsp;&nbsp;SERVER_PORT=? // The port that the server should run on. 3000 is a good default choice<br>
+&nbsp;&nbsp;&nbsp;&nbsp;SERVER_PORT=? // The port that the server should run on. `3000` is a good default choice<br>
```
jameeters: This really has nothing to do with the rest of the pull request, but I like this better as a code block:
```
SERVER_PORT=?              // The port that the server should run on. 3000 is a good default choice
DB_USER=?                  // The user that should be used to connect to postgres
DB_DATABASE=?              // The database you just created, so likely environmental_dashboard
DB_TEST_DATABASE=?         // The test database you just created, so likely environmental_dashboard_testing
DB_PASSWORD=?              // The password for your postgres user
DB_HOST=?                  // The host for your postgres db, likely localhost
DB_PORT=?                  // The port for your postgres db, likely 5432
TOKEN_SECRET=?             // Token for authentication. Generate something secure and random
```
Also I'm pretty sure the `<br>` tags are unnecessary.

src/client/app/components/HeaderComponent.jsx
```diff
@@ -22,16 +22,21 @@ export default function HeaderComponent(props) {
 	};
 	const loginLinkStyle = {
 		// Displays the login button link only if the user is not logged in or is explicitly told to display by the parent component
-		display: localStorage.getItem('token') || props.renderLoginButton === 'false' ? 'none' : 'inline'
+		display: localStorage.getItem('token') || props.renderLoginButton === false ? 'none' : 'inline'
```
jameeters: I think this ternary operator is hard enough to read that it would be better as a call to a function with an `if-else` block.

src/client/app/containers/UIOptionsContainer.js
```diff
@@ -23,7 +23,10 @@ function mapStateToProps(state) {
 function mapDispatchToProps(dispatch) {
 	return {
 		selectMeters: newSelectedMeterIDs => dispatch(changeSelectedMeters(newSelectedMeterIDs)),
-		fetchMetersDataIfNeeded: () => dispatch(fetchMetersDataIfNeeded())
```
jameeters: I think `fetchMetersDetails` is a lot clearer than `fetchMetersData`, which can be confused with the data we get from meters (as in readings).

src/client/app/components/HeaderComponent.jsx
johndiiorio: This is a pure styling thing, which is used heavily in the Redux documentation. Combine d with the comment above, I think it is clear enough.

src/client/app/actions/graph.js
```diff
@@ -4,30 +4,56 @@
...
-import { fetchNeededReadings } from './readings';
+import { fetchAllNeededReadings, fetchBarNeededReadings } from './readings';
 import TimeInterval from '../../../common/TimeInterval';
 
 
 export const UPDATE_SELECTED_METERS = 'UPDATE_SELECTED_METERS';
+export const UPDATE_BAR_DURATION = 'UPDATE_BAR_DURATION';
+export const CHANGE_CHART_TO_RENDER = 'CHANGE_CHART_TO_RENDER';
+export const CHANGE_BAR_STACKING = 'CHANGE_BAR_STACKING';
 export const SET_GRAPH_ZOOM = 'CHANGE_GRAPH_ZOOM';
```
johndiiorio: Fixed in next commit. Good catch, especially on line 15.

src/client/app/containers/UIOptionsContainer.js
johndiiorio: Fixed in next commit, good idea. Actions are complicated enough already, we don't need to make them purposely esoteric.

src/client/app/containers/UIOptionsContainer.js
```diff
@@ -23,7 +23,10 @@ function mapStateToProps(state) {
...
+		changeDuration: barDuration => dispatch(changeBarDuration(barDuration)),
+		fetchMetersDataIfNeeded: () => dispatch(fetchMetersDataIfNeeded()),
+		changeChartType: chartType => dispatch(changeChartToRender(chartType)),
+		toggleBarStacking: () => dispatch(changeBarStacking())
```
johndiiorio: Fixed in next commit. I originally had it changed from changeBarStacking() to toggleBarStacking to differentiate the two functions in the file, which is unnecessary.

src/client/app/reducers/readings.js
```diff
@@ -16,7 +16,12 @@ import * as readingsActions from '../actions/readings';
  * @type {State~Readings}
  */
 const defaultState = {
```
johndiiorio: That's a good point, however, what do we do about actions like fetchAllNeededReadings()? If we have two files: actions/barReadings.js and actions/lineReadings.js, it doesn't seem to be appropriate to place it in either one.

src/server/services/readMetasysData.js
johndiiorio: Begrudgingly fixed in the next commit.

README.md
johndiiorio: Fixed in the next commit. The `<br>` tags are unnecessary.

src/server/services/readMetasysData.js
sandeepacharya464: This file has been almost completely changed in my branch. Anyway thanks for 'begrudgingly' fixing the problem @johndiiorio .  :)

src/client/app/reducers/readings.js
jameeters: Since we aren't showing more than one type of graph at a time, why do we need to fetch more than one type of readings at a time?

johndiiorio: On this branch, when the page loads for the first time, readings for both the line chart and bar chart (default options, 30 day bar chart width, etc.) are called. This lets the user switch toggle between bar and line with no delay. The other actions are called when users request specific non-default readings.

jameeters: In that case you can always keep `actions/readings.js` and just add `actions/barChartReadings.js`, etc. as necessary.

src/client/app/actions/readings.js
```diff
@@ -6,58 +6,125 @@
...
 
 /**
- * Fetches readings for the given meterIDs if they are not already fetched or being fetched
+ * Fetches readings for all charts for the given meterIDs if they are not already fetched or being fetched
  * @param {Array.<int>} meterIDs
- * @param {TimeInterval} timeInterval The time interval to fetch readings for
+ * @param {TimeInterval} timeInterval The time interval to fetch readings for on the line chart
+ * @return {*} An action to fetch the needed readings
+ */
+export function fetchAllNeededReadings(meterIDs, timeInterval) {
```
simonbtomlinson: This seems to duplicate some functionality from the `fetchBarNeededReadings` function. Maybe extract the line readings fetch logic into a `fetchLineNeededReadings` and make this method just call both?

Also I think that `fetch[Line|Bar]NeededReadings` sounds worse than `fetchNeeded[Line|Bar]Readings`. Should we switch them?

src/client/app/components/DashboardComponent.jsx
```diff
@@ -5,16 +5,17 @@
...
  * React component that controls the dashboard
  */
-export default function DashboardComponent() {
+export default function DashboardComponent(props) {
 	return (
 		<div className="container-fluid">
 			<UIOptionsContainer />
 			<div className="col-xs-10">
-				<LineChartContainer />
+				{props.chartToRender === 'line' ? <LineChartContainer /> : <BarChartContainer />}
```
simonbtomlinson: `'line'` really feels like it should be defined as an enum-like variable somewhere instead of just a string literal.

src/client/app/components/HeaderComponent.jsx
simonbtomlinson: Maybe wrap the `||` with some parentheses to make it really obvious, but I begrudgingly agree this is probably ok.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -35,6 +46,24 @@ export default class UIOptionsComponent extends React.Component {
 		this.props.selectMeters(selectedMeters);
 	}
 
+	handleBarDurationChange(value) {
+		this.setState({ barDuration: value });
```
simonbtomlinson: It's not obvious that you're putting this in state to delay redux bar changes until the user lets go of the bar. A comment is probably appropriate.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -35,6 +46,24 @@ export default class UIOptionsComponent extends React.Component {
...
+	}
+
+	handleBarDurationChangeComplete(e) {
+		e.preventDefault();
+		this.props.changeDuration(moment.duration(this.state.barDuration, 'days').toISOString());
+	}
+
+	handleChangeChartType(e) {
+		this.setState({ chartType: e.target.value });
```
simonbtomlinson: This one should be in redux.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -58,40 +87,18 @@ export default class UIOptionsComponent extends React.Component {
...
+						<label><input type="radio" name="chartTypes" value="line" onChange={this.handleChangeChartType} checked={this.state.chartType === 'line'} />Line</label>
 					</div>
 					<div className="radio">
-						<label><input type="radio" name="graphTypes" value="Map" />Map</label>
-					</div>
-					<br />
-					<p style={labelStyle}>Other options:</p>
-					<div className="checkbox">
-						<label><input type="checkbox" value="overlayweather" />Overlay Weather</label>
+						<label><input type="radio" name="chartTypes" value="bar" onChange={this.handleChangeChartType} checked={this.state.chartType === 'bar'} />Bar</label>
```
simonbtomlinson: ChartType being an enum should probably extend here.

src/client/app/containers/BarChartContainer.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {State} state
+ */
+function mapStateToProps(state) {
+	const timeInterval = state.graph.timeInterval;
+	const barDuration = state.graph.barDuration;
+	const data = { datasets: [] };
+	// getColor() cycles through the colors, wrapping around the end to the beginning
+	const colors = ['LightBlue', 'GoldenRod', 'Black', 'OrangeRed', 'LightSeaGreen', 'LightSlateGray', 'Purple'];
+	let colorPointer = 0;
+	function getColor() {
```
simonbtomlinson: You're doing this in the line chart component too. It should probably get refactored out.

src/client/app/containers/BarChartContainer.js
```diff
@@ -0,0 +1,94 @@
...
+			const color = getColor();
+			data.datasets.push({
+				label: state.meters.byMeterID[meterID].name,
+				data: readingsData.readings.map(arr => arr[1]),
+				backgroundColor: color,
+				hoverBackgroundColor: color
+			});
+			// Add only the unique time intervals to the label set
+			for (const element of _.flatten(readingsData.readings.map(arr => arr[0]))) {
+				labelsSet.add(`${moment(element).format('MMM DD, YYYY, hh:mm a')} - ${moment(element).add(moment.duration(barDuration)).format('MMM DD, YYYY, hh:mm a')}`);
```
simonbtomlinson: This is a duplicate of the functionality of `TimeInterval` in the common folder. You should probably change it to use that and then format those right before you display.

src/client/app/containers/LineChartContainer.js
```diff
@@ -25,11 +25,11 @@ function mapStateToProps(state) {
 	}
 
 	for (const meterID of state.graph.selectedMeters) {
-		const readingsData = state.readings.byMeterID[meterID][timeInterval];
+		const readingsData = state.readings.line.byMeterID[meterID][timeInterval];
```
simonbtomlinson: I think this is fine. It reads almost like english, and it should be obvious to someone who understands how the redux state is set up.

src/client/app/containers/UIOptionsContainer.js
```diff
@@ -23,7 +23,10 @@ function mapStateToProps(state) {
 function mapDispatchToProps(dispatch) {
 	return {
 		selectMeters: newSelectedMeterIDs => dispatch(changeSelectedMeters(newSelectedMeterIDs)),
-		fetchMetersDataIfNeeded: () => dispatch(fetchMetersDataIfNeeded())
+		changeDuration: barDuration => dispatch(changeBarDuration(barDuration)),
+		fetchMetersDetailsIfNeeded: () => dispatch(fetchMetersDetailsIfNeeded()),
+		changeChartType: chartType => dispatch(changeChartToRender(chartType)),
+		changeBarStacking: () => dispatch(changeBarStacking())
```
simonbtomlinson: I think changeBarStacking should be called toggleBarStacking since it's actually toggling something.

src/client/app/reducers/graph.js
```diff
@@ -19,7 +19,10 @@ import * as graphActions from '../actions/graph';
  */
 const defaultState = {
 	selectedMeters: [],
-	timeInterval: TimeInterval.unbounded()
+	timeInterval: TimeInterval.unbounded(),
+	barDuration: moment.duration(1, 'month').toISOString(),
```
simonbtomlinson: Why aren't we storing this as a moment duration?

src/client/app/reducers/readings.js
simonbtomlinson: I agree that these need to be split.

src/server/models/Reading.js
```diff
@@ -143,6 +150,32 @@ class Reading {
...
+	 *
+	 * Compressed readings are in kilowatts.
+	 * @param meterIDs an array of ids for meters whose points are being compressed
+	 * @param duration A moment time duration over which to sum the readings
+	 * @param fromTimestamp An optional start point for the beginning of the entire time range.
+	 * @param toTimestamp An optional end point for the end of the entire time range.
+	 * @param conn the connection to use. Defaults to the default database connection.
+	 * @return {Promise<object<int, array<{reading_sum: number, start_timestamp: Date, end_timestamp: Date}>>>}
+	 */
+	static async getAggregateReadings(meterIDs, duration, fromTimestamp = null, toTimestamp = null, conn = db) {
```
simonbtomlinson: I don't think this is named well. It doesn't convey the fact that the readings are for the bar chart.

src/server/sql/reading/create_function_get_aggregate_readings.sql
```diff
@@ -0,0 +1,50 @@
...
+			),
+			agg.start,
+			agg.start + real_duration
+		FROM readings r
+		INNER JOIN unnest(meter_ids) specific_meter(id) ON r.meter_id = specific_meter.id
+			INNER JOIN generate_series(real_start_timestamp, real_end_timestamp, real_duration) agg(start)
+				ON
+					(r.start_timestamp, r.end_timestamp) OVERLAPS (agg.start, agg.start + real_duration)
+					AND
+					(agg.start, agg.start + duration) OVERLAPS (real_start_timestamp, real_end_timestamp)
```
simonbtomlinson: Maybe a comment about why we're using the `AND` to deal with the fact that it's an end-inclusive range.

src/server/test/db/common.js
```diff
@@ -21,7 +21,8 @@ function recreateDB() {
 	return db.none('DROP TABLE IF EXISTS readings')
 		.then(() => db.none('DROP TABLE IF EXISTS meters'))
 		.then(() => db.none('DROP TYPE IF EXISTS meter_type'))
-		.then(() => db.none('DROP FUNCTION IF EXISTS compressed_readings(INTEGER, TIMESTAMP, TIMESTAMP, INTEGER);'))
```
simonbtomlinson: Thanks for catching this.

src/client/app/containers/BarChartContainer.js
johndiiorio: I disagree with you here. It makes perfect sense for this line, but a couple more lines down, where I do `data.labels = Array.from(labelsSet).sort((x, y) => moment(x.split(' - ')[0], 'MMM DD, YYYY, hh:mm a').format('x') - moment(y.split(' - ')[0], 'MMM DD, YYYY, hh:mm a').format('x'));`
, it really complicates things.

johndiiorio: The ES6 Set() class also does not work on on objects.

src/client/app/actions/barReadings.js
```diff
@@ -0,0 +1,66 @@
...
+import axios from 'axios';
+
+export const REQUEST_BAR_READINGS = 'REQUEST_BAR_READINGS';
+export const RECEIVE_BAR_READINGS = 'RECEIVE_BAR_READINGS';
+
+/**
+ * @param {State} state
+ * @param {number} meterID
+ * @param {TimeInterval} timeInterval
+ * @param {Moment duration} barDuration
```
simonbtomlinson: Shouldn't this be `moment.duration`?

src/client/app/actions/graph.js
```diff
@@ -37,23 +36,27 @@ export function updateBarDuration(barDuration) {
 export function changeBarDuration(barDuration) {
 	return (dispatch, state) => {
```
simonbtomlinson: I think this parameter should idiomatically be called `getState()`, not `state`, since it's a function that returns state instead of the actual state. This occurs in a few places.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -20,8 +21,7 @@ export default class UIOptionsComponent extends React.Component {
 		this.handleChangeChartType = this.handleChangeChartType.bind(this);
 		this.handleChangeBarStacking = this.handleChangeBarStacking.bind(this);
 		this.state = {
-			barDuration: 30, // barDuration in days
-			chartType: 'line'
+			barDuration: 30 // barDuration in days
```
simonbtomlinson: Why are you storing this as an integer instead of a moment duration?

src/client/app/reducers/barReadings.js
```diff
@@ -0,0 +1,65 @@
...
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import * as readingsActions from '../actions/barReadings';
+
+/**
+ * @type {State~Readings}
```
simonbtomlinson: Shouldn't this be a new `State~BarReadings` type?

src/client/app/reducers/lineReadings.js
```diff
@@ -0,0 +1,60 @@
...
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import * as readingsActions from '../actions/lineReadings';
+
+/**
+ * @type {State~Readings}
```
simonbtomlinson: Shouldn't this be `State~LineReadings`?

src/client/app/components/UIOptionsComponent.jsx
johndiiorio: This is for the UI slider.

PR #87, opened 2017-04-24
API mystery routing issue fix

paulmtwomey:
Hopefully, this should fix the odd issue where we're receiving index.html in some cases.

### DIFF
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -29,10 +29,6 @@ app.use('/api/readings', readings);
 app.use('/api/login', login);
 app.use('/api/verification', verification);
 
-app.get('*', (req, res) => {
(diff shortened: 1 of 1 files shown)


src/server/app.js
```diff
@@ -46,4 +42,9 @@ app.use((err, req, res) => {
 	else res.send(`<h1>${err.status} Server Error</h1>`);
 });
 
+// This is after the other app.XXX entries as having it before seems to cause issues. See OED issue #86.
```
johndiiorio: Minor thing, but this comment isn't super helpful or descriptive. I would recommend removing it or explaining the issue more.

paulmtwomey: Agreed, removed.

paulmtwomey:
This wasn't the fix we were looking for.

Issue #72, opened 2017-04-05
Client UI options is not responsive on mobile devices


LucasFares:
We need to display properly on multiple types of screens.

PR #91, opened 2017-04-29
Replace sidebar options with modal if window width is small.

johndiiorio:
Fixes https://github.com/beloitcollegecomputerscience/OED/issues/72


src/client/app/components/HeaderComponent.jsx
```diff
@@ -29,7 +31,6 @@ export default function HeaderComponent(props) {
 		display: localStorage.getItem('token') ? 'inline' : 'none'
 	};
 	return (
-
 		<div className="container-fluid">
 			<div className="col-xs-4">
 				<Link to="/"><LogoComponent url="./app/images/logo.png" /></Link>
```
jameeters: You should resize the logo to fit smaller window sizes.

src/client/app/components/HeaderComponent.jsx
```diff
@@ -38,8 +39,9 @@ export default function HeaderComponent(props) {
 				<h1 style={titleStyle}>Open Energy Dashboard</h1>
```
jameeters: The title should also be resized to fit smaller windows, this and the logo take up an exorbitant amount of space in small windows.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -76,35 +76,28 @@ export default class UIOptionsComponent extends React.Component {
...
+				</div>
+				<p style={labelStyle}>Graph Type:</p>
+				<div className="radio">
+					<label><input type="radio" name="chartTypes" value={chartTypes.line} onChange={this.handleChangeChartType} checked={this.props.chartToRender === chartTypes.line} />Line</label>
+				</div>
+				<div className="radio">
+					<label><input type="radio" name="chartTypes" value={chartTypes.bar} onChange={this.handleChangeChartType} checked={this.props.chartToRender === chartTypes.bar} />Bar</label>
+				</div>
+				<div className="checkbox">
+					<label><input type="checkbox" onChange={this.handleChangeBarStacking} />Bar stacking</label>
```
jameeters: I don't think showing the option to toggle stacking in the bar chart makes sense until bar chart is selected.

src/client/app/components/DashboardComponent.jsx
```diff
@@ -12,12 +12,22 @@ import { chartTypes } from '../reducers/graph';
...
+		paddingTop: '35px'
+	};
+	const ChartToRender = (props.chartToRender === chartTypes.line) ? LineChartContainer : BarChartContainer;
 	return (
 		<div className="container-fluid">
-			<UIOptionsContainer />
-			<div className="col-xs-10">
-				{props.chartToRender === chartTypes.line ? <LineChartContainer /> : <BarChartContainer />}
-			</div>
+			{ props.drawUIOptions ? (
```
jameeters: This ternary operator is unreadable. There has to be a better way to do this.

simonbtomlinson:
I think we should move the sidebar to the bottom when the screen gets small, instead of putting it in a modal. I'm not a fan of modals for important functionality like this - they hide buttons and the user might not know to look for them.

I have serious issues with dealing with screen size in redux the way that you do when a simple CSS media query (wrapped in bootstrap, which we're already using!) would work. I think this is the wrong way to do what we want to do here. You dispatch a redux action every time the screen size changes, even mid drag! (Note that I don't think that fixing just the dispatch mid-drag is an acceptable way to deal with this.)

[Here](https://gist.github.com/simonbtomlinson/1c2a045ec9343cc037bb4be6fbb47674)'s an example of using bootstrap to render different HTML for a sidebar and a bottom bar. You don't need any javascript for it, and it's quite simple.

I propose closing this pull request because the work done in it is incompatible with a simpler bootstrap solution.

@jameeters, @LucasFares, @paulmtwomey, what do you think?

LucasFares:
I agree with Simon that the screen sizing should be done differently.  As for if a modal is the best way to go about this, I think that should be discussed in the relevant issue thread.

paulmtwomey:
I agree with y'all on the sizing. Simon: I agree on the button hiding, but I don't know whether I mind it here if presented in a properly distinct manner. A matter for another thread. I support closing this for a CSS fix and further discussion elsewhere.

johndiiorio:
I would like to provide[ this StackOverflow post](http://stackoverflow.com/questions/19014250/reactjs-rerender-on-browser-resize) where I based the core of the resizing. @LucasFares or @jameeters (I can't remember which one) already tried moving the UI options to the bottom, but it did not look great. This way, we already have the functionality set-up in place for resizing. As to whether we use a modal or not, we will have to use something like this no matter what. The modal is only shown if the window size < 1000 (which is pretty small). Most user will never even see the modal unless browsing on mobile, then this is pretty much a requirement.

src/client/app/components/HeaderComponent.jsx
```diff
@@ -38,8 +39,9 @@ export default function HeaderComponent(props) {
...
 			</div>
 			<div style={divRightStyle}>
-				<Link style={loginLinkStyle} to="/login"><button className="btn btn-default">Log in</button></Link>
-				<Link style={adminLinkStyle} to="/admin"><button className="btn btn-default">Admin panel</button></Link>
+				{(props.renderOptionsButton && props.drawUIOptionsButton) ? <UIModalComponent /> : null}
```
johndiiorio: That makes sense, but I'm not sure where else to put it. Don't forget that on mobile (where this is important), you really don't want things to the left or right of the chart, you want it above or below. This places it above with all the other buttons.

src/client/app/components/HeaderComponent.jsx
johndiiorio: This makes sense, I will work on it.

src/client/app/components/UIOptionsComponent.jsx
johndiiorio: Yeah, that's a good catch. I thought about it, but that didn't seem relevant to this pull request. Definitely another one, and perhaps worth a Github issue.

simonbtomlinson:
@johndiiorio, my argument is that we don't need to re-render using react, instead we can just use CSS. The modal / bottom discussion is secondary to the react/css one.

src/client/app/components/DashboardComponent.jsx
johndiiorio: This is identical to how the React docs do it. See https://facebook.github.io/react/docs/conditional-rendering.html, under the sectionInline If with Logical && Operator

johndiiorio:
@simonbtomlinson That does make sense. However, I believe your solution is a short-term solution, as I don't think it is the "React way" to do things. The complexity that this pull request adds is fairly minor, and we now have the window size in Redux, so that we could apply this to other elements as they come up in the project. Feel free to correct me if I'm wrong, I'm not an expert in CSS.

simonbtomlinson:
@johndiiorio We don't need to solve a CSS problem the react way. If we wanted to do something like render a text box with the width of the browser in pixels in it, then I would agree with you, but that's not what we're doing. We're adding mobile support and moving something, and we're already using a CSS framework that supports that. We should embrace bootstrap.

jameeters:
I think that CSS is probably the way to go, and at least needs to be tried, so I'm not going to respond to comments on my requested changes until it's clear that they will remain relevant.

johndiiorio:
@simonbtomlinson @jameeters @paulmtwomey Changed it to use CSS.

simonbtomlinson:
Changing this to CSS is a good step, but we need to have a conversation about what we want this to look like that's outside the scope of a pull request. I'm closing this, and I opened issue #94 to discuss the way we want to implement it.

paulmtwomey:
Let's close this for now and address it after we come to a decision on #94. It may end up being exactly this, but we shouldn't keep around a PR if we don't definitively know it's what we want going in.

simonbtomlinson:
I agree. That's why I closed it. We don't know what the code we want is, so this is better suited for an issue. We don't need to delete the branch, and we can just reopen it or open a new one off this branch when we're ready.

johndiiorio:
I believe #94 is a duplicate of #72. Secondly, especially with the demo coming up, we want clients to be able to view our site on mobile (the modal won't be seen otherwise). I thought that this was fairly important to get done promptly, and if we eventually want to move options to another place later, that's fine.

paulmtwomey:
I agree on the duplication, we need a centralized discussion on this topic. Both work, both could be more specific but they're for our discussion and shouldn't live long once we address it. 
I don't know if this is essential for the demo coming up, we know we can do it and we can tell them that if they ask about mobile. That's just me; if this is important to do for the demo let's get whichever issue we use taken care of and resolved quickly then.

huss:
I think mobile is important but not something we have to have for the demo.
Given the discussion, I move we decide how to do this and then push it out
for others to see.  Lets focus on some pressing issues to get done in the
coming days. Mobile can be integrates after this.

jameeters:
If anyone's wondering why @johndiiorio did that, I asked him to close this PR and open a new one so that we could review the branch without wading through the bickering from last spring. But it turns out Github won't let you open a second PR on the same branch like that, so we're stuck with our own comments.

jameeters:
I've approved this, although my requested changes have not been made, because they should be made soon in other PRs.

PR #93, opened 2017-04-29
Deployment Automation with Docker Compose

LeoTindall:
This adds fully automated deployment with Docker Compose (@simonbtomlinson's suggestion), for both development and production. 

There is one minor issue in that creating the schema can take multiple tries due to the PostgreSQL container taking a long time to spin up. No other operation has this issue, and other than that, everything works very well. I consider that a non-blocking problem because it only has to be done once, but if others don't agree I can probably figure out how to get it working.


SilverWingedSeraph:
Commit 04aa92f adds a script that works around the database init issue while also making setup a lot more convenient; it's now just 

```
docker-compose build
docker-compose run --rm web scripts/init.sh
docker-compose up # for production; for development, use:
# docker-compose run --service-ports --rm web scripts/devstart.sh
```

LeoTindall:
I consider this done if others can verify that it works

LeoTindall:
I prefer this method because it makes that image distributable/cacheable. This way, the database is ready for `scripts/init.sh` as soon as its up. I'm definitely amenable to changing that, though.

simonbtomlinson:
I like the other way because we don't need another dockerfile, but it's really just nit picking. Let's just do it your way. I'll make sure it all works for me later today and then approve the pull request.

LeoTindall:
OK. Added a volume mount for Postgres in 27b270d and tests in ae63aef. Only issue with those tests is that they increase CI time a little bit (because the build now requires sudo). On the other hand, they'll make sure the Docker environment never falls out of parity with the native one.

simonbtomlinson:
Does this move all ci testing to docker? I'm fine with that, but not super familiar with Travis configuration.

simonbtomlinson:
@simonbtomlinson Nevermind, I read it more closely and it does both.

PR #95, opened 2017-05-03
appliance and comparison graph page

zhouyizhouyi:
-appliance page & comparison graph page
-some problems with graph page on changing building(when building and graph type change together)
-graph for comparing daily usage not done


LucasFares:
This was moved into the demo branch.

Issue #86, opened 2017-04-21
Intermittent problem with API requests returning the source code of index.html instead of the correct data


jameeters:
This issue has been seen on one personal computer and on the OED server. 
The problem is requests to the API returning the source of `index.html` instead of the requested data, causing an error in the frontend when `data.map` is not a function.
The cause appears to be some thing to do with the Express routing setup in `src/server/app.js`. In line 32 of `app.js`, we have:
```
app.get('*', (req, res) => {
	res.sendFile(path.resolve(__dirname, '..', 'client', 'index.html'));
});
```
This block is intended to forward routing except API requests to react-router on the client. The issue is that requests that should be sent either to an API or to an error handler are instead being handled by this logic and getting `index.html`. The APIs with which we've had this problem have been registered with the API middleware above this block in `app.js`.

Changing the `*` to a `/` so that the code reads:
```
app.get('/', (req, res) => {
	res.sendFile(path.resolve(__dirname, '..', 'client', 'index.html'));
});
```
appears to fix the problem, but the strange thing is that changing it back to a `*` does NOT cause the issue to reappear.

If you run into this issue pay attention to what's going on and reply to this issue with relevant details.

paulmtwomey:
I got the error again last night, switching the `*` to `/` didn't change anything for me. 

However, moving the block to the end fixed it. This is what I have, with it working (for now).

```// catch 404 and forward to error handler
app.use((req, res, next) => {
	const err = new Error('Not Found');
	err.status = 404;
	next(err);
});

app.use((err, req, res) => {
	res.status(err.status || 500);
	if (err.status === 404) res.send(`<h1>${err.status} Not found</h1>`);
	else res.send(`<h1>${err.status} Server Error</h1>`);
});

app.get('*', (req, res) => {
	res.sendFile(path.resolve(__dirname, '..', 'client', 'index.html'));
});
```

johndiiorio:
I think this is the correct solution. You should make a pull request.

simonbtomlinson:
I don't think we should close this issue until we merge a pull request that has both a unit test that confirms it and a fix.

simonbtomlinson:
My bad, I didn't realize that pr was merged

jameeters:
Well, that #245 doesn't include a unit test, although I'm not sure how we would write one for this.

PR #99, opened 2017-09-07
Fix API wildcard routing issue

LucasFares:
See issue #86


jameeters:
If I remember correctly, this change did not actually fix the issue. It's very possible that I am not remembering correctly, but assuming I am, this pull request should be closed.

johndiiorio:
I also believe that this change did not fix the issue. I agree that this PR should be closed.

johndiiorio:
I am closing this PR for the reasons discussed above.

PR #105, opened 2017-09-14
Fix lodash babel error, add source-maps, add prop-types to package.json

johndiiorio:


simonbtomlinson:
Shouldn't we figure out why lodash broke instead of just removing it?

johndiiorio:
I think  `new LodashModuleReplacementPlugin()` in the webpack plugins array should take care of it.

simonbtomlinson:
Ah, I didn't see that.

Issue #107, opened 2017-09-14
Script to add admin users into the database


johndiiorio:
Add a script so that an admin user can be easily inserted into the database. For example:
`npm run createuser` could prompt the user for both a username and password.

When completed, add the script to the README.md docs.

PR #108, opened 2017-09-14
Add script to create new users

johndiiorio:
Implements #107

### DIFF
--- a/src/scripts/createUser.js
+++ b/src/scripts/createUser.js
@@ -42,9 +42,7 @@ function terminateReadline(message) {
 	try {
 		emailResult = await askEmail();
 	} catch (err) {
-		console.log('Invalid email, no user created');
-		rl.close();
-		process.exit(0);
+		terminateReadline('Invalid email, no user created');
 	}
 	const output = await askPassword(emailResult);
 	const email = output[0];
--- a/src/scripts/createUser.js
+++ b/src/scripts/createUser.js
@@ -1,0 +1,67 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const readline = require('readline');
+const bcrypt = require('bcryptjs');
+const User = require('../server/models/User');
+
+const rl = readline.createInterface({
+	input: process.stdin,
+	output: process.stdout
+});
+
+function askEmail() {
+	return new Promise((resolve, reject) => {
+		rl.question('Email: ', email => {
+			// See https://stackoverflow.com/a/46181/5116950
+			// eslint-disable-next-line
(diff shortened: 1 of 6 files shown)


src/scripts/createUser.js
```diff
@@ -0,0 +1,67 @@
...
+		console.log('Invalid email, no user created');
+		rl.close();
+		process.exit(0);
+	}
+	const output = await askPassword(emailResult);
+	const email = output[0];
+	const password = output[1];
+
+	if (password.length < 8) {
+		terminateReadline('Weak password, no user created');
```
jameeters: You should tell the user beforehand that their password must be at least 8 characters long, and this error message should say that the password must be at least 8 characters long.

src/scripts/createUser.js
```diff
@@ -0,0 +1,67 @@
...
+	}
+	const hasUpperCase = /[A-Z]/.test(password);
+	const hasLowerCase = /[a-z]/.test(password);
+	const hasNumbers = /\d/.test(password);
+	const hasNonAlphas = /\W/.test(password);
+	if (hasUpperCase + hasLowerCase + hasNumbers + hasNonAlphas < 3) {
```
jameeters: Every single bit of modern research on the topic has found that these kinds of rules lead to weaker passwords. It's also insane to reject a password based on these rules without ever printing any kind of message saying that these rules exist. The clear solution here is to remove these rules entirely.

src/server/sql/user/insert_new_user.sql
```diff
@@ -2,4 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-INSERT INTO users(email, password_hash) VALUES (${email}, ${passwordHash})
+INSERT INTO users(email, password_hash)
+VALUES (${email}, ${passwordHash})
+ON CONFLICT (email) DO UPDATE SET email=${email};
```
jameeters: I'm not sure I understand what this is doing. I assumed that it changed their password, but that does not seem to be the case. Is this saying that if a user with that email already exists, then update their email? That doesn't seem sane.

jameeters: Also, this should abort on a collision, the query to insert a new user should fail if the new user already exists.

src/scripts/createUser.js
johndiiorio: Good point on the prompt. I'm not sure if we want to remove the rules entirely, it was simply a check to prevent passwords like `password`

src/server/sql/user/insert_new_user.sql
johndiiorio: Yeah, this is not correct. I'll change it in the next commit.

src/scripts/createUser.js
jameeters: I think the length requirement is fine, and if you want to go any further (which I think is overkill for now) you should follow the [NIST guidelines](https://en.wikipedia.org/wiki/Password_policy#NIST_Guidelines).

src/scripts/createUser.js
```diff
@@ -0,0 +1,64 @@
...
+	rl.close();
+	process.exit(0);
+}
+
+(async () => {
+	let emailResult;
+	try {
+		emailResult = await askEmail();
+	} catch (err) {
+		console.log('Invalid email, no user created');
```
simonbtomlinson: Why not use your `terminateReadline` function here?

PR #112, opened 2017-09-17
Actions and reducers for basic group editing

jameeters:
Add actions and reducers to do the following:
* create a new group with a blank name and no children
* edit the name of the group
* add/remove child meters/groups of the group

### DIFF
--- a/src/client/app/actions/groups.js
+++ b/src/client/app/actions/groups.js
@@ -19,6 +19,7 @@ export const GROUPSUI_CHANGE_DISPLAYED_GROUPS = 'GROUPSUI_CHANGE_DISPLAYED_GROUP
 
 export const CREATE_NEW_GROUP = 'CREATE_NEW_GROUP';
 export const EDIT_GROUP_NAME = 'EDIT_GROUP_NAME';
+export const ADOPT_CHILD_GROUPS = 'ADOPT_CHILD_GROUPS';
 
 
 function requestGroupsDetails() {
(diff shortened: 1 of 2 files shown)


johndiiorio:
Could you add a `groupDisplay` string property that is either 'view', 'edit', or 'create' that will control whether the user is viewing, editing, or creating groups?

src/client/app/actions/groups.js
```diff
@@ -17,6 +17,14 @@ export const GROUPSUI_CHANGE_SELECTED_GROUPS_PER_GROUP = 'GROUPSUI_CHANGE_SELECT
 export const GROUPSUI_CHANGE_SELECTED_METERS_PER_GROUP = 'GROUPSUI_CHANGE_SELECTED_METERS_PER_GROUP';
 export const GROUPSUI_CHANGE_DISPLAYED_GROUPS = 'GROUPSUI_CHANGE_DISPLAYED_GROUPS';
 
+export const CREATE_NEW_GROUP = 'CREATE_NEW_GROUP';
+export const EDIT_GROUP_NAME = 'EDIT_GROUP_NAME';
+export const ADOPT_CHILD_GROUPS = 'ADOPT_CHILD_GROUPS';
+export const DISOWN_CHILD_GROUPS = 'DISOWN_CHILD_GROUPS';
```
johndiiorio: I don't think we need both adopt and disown. How about CHANGE_CHILD_GROUPS.

jameeters:
I added a property called `displayMode` that does exactly what you wanted 'groupDisplay' to do. I just think that name more closely matches what the property is doing.

src/client/app/actions/groups.js
jameeters: Done.

PR #114, opened 2017-09-18
Change logo

LucasFares:
Replaced logo with blue gradient with simple black and white version.


huss:
Can someone explain why the color one is an issue? We have color in many
ways on our site so I wanted to better understand. It does not seem to make
it much harder to read for me.

LucasFares:
While the color one isn't an issue per se, we think the black and white version looks nicer.

johndiiorio:
I would also recommend either removing or updating logo_vertical.png, found in the same directory. I don't believe it is used anywhere, but it could be nice to have.

LucasFares:
As far as I can tell, logo_vertical.png was only used to create the favicon. Consistency would be great. Can someone convert the favicon to grayscale? We don't have access to the black and white version of logo_vertical.png so we can't use that to create a new favicon.

huss:
I want to discuss the logo at the meeting before we finalize this. We need
to make a team decision because we should stick with what we choose.

johndiiorio:
This PR is messed up, and a new one will be made shortly.

PR #116, opened 2017-09-19
Logo switch on mouseover

LeoTindall:
Supersedes #114 by giving the user the ability to view both beautiful logos!

To be clear, I made this branch as a learning project for myself but, I mean, it works, so why not? Feel free to close if too silly.


johndiiorio:
Nice concept, but it is too gimmicky for this project, especially since the logo is often clicked on as a "return to homepage" button.

PR #119, opened 2017-09-20
Filter control for meters

### LINKED ISSUE #118: Groups/meters filter search
Allow a user to filter by a search term. Questions: Should this be fuzzy search? Exact? Allow inter-character variation?

LeoTindall:
Beings to address #118 by allowing the user to filter case-insensitively on the meter's canonical name.

The primary drawback here is that if the user selects a meter and enters a filter term that excludes that meter, then selects another meter, even when holding an inclusive select key (e.g. Ctrl or Shift), the excluded meter will be lost from the selection. An overhaul to meter selection is required to address this.

### DIFF
--- a/src/client/app/actions/meters_filter.js
+++ b/src/client/app/actions/meters_filter.js
@@ -1,0 +1,24 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const METERS_FILTER_MODIFIED = 'METERS_FILTER_MODIFIED';
+export const METERS_FILTER_CLEARED = 'METERS_FILTER_CLEARED';
+
+/**
+ * Returns an action signifying that the meters filter was changed to the
+ * given term.
+ * @param {string} term The term to which the filter is set
+ */
+export function metersFilterModified(term) {
+	return { type: METERS_FILTER_MODIFIED, term: term };
+}
+
+/**
+ * Returns an action signifying that the meters filter was cleared.
+ */
+export function metersFilterCleared() {
+	return { type: METERS_FILTER_CLEARED };
+}
--- a/src/client/app/components/MetersFilterComponent.jsx
+++ b/src/client/app/components/MetersFilterComponent.jsx
@@ -1,0 +1,44 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import React from 'react';
+
+// needs setFilter and clearFilter dispatchers
+// and filterTerm, the text in the box.
+export default class MetersFilterComponent extends React.Component {
+	/**
+	 * Initializes the components state, binds all functions to this.
+	 * @param props The props passed down through the MetersFilterContainer
+	 */
+	constructor(props) {
+		super(props);
+		this.handleChanged = this.handleChanged.bind(this);
+		this.handleClearButtonClicked = this.handleClearButtonClicked.bind(this);
+	}
+
+	handleChanged(event) {
+		this.props.setFilter(event.target.value);
+	}
+
+	handleClearButtonClicked() {
+		this.props.clearFilter();
+		// By default, the form field keeps its own state, so when it's
+		// cleared it needs to be invalidated so the vDOM can redraw it.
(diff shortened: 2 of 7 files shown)


src/client/app/actions/meters_filter.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function metersFilterModified(term) {
+	return { type: METERS_FILTER_MODIFIED, term: term };
+}
+
+/**
+ * Returns an action signifying that the meters filter was cleared.
+ */
+export function metersFilterCleared() {
+	return { type: METERS_FILTER_CLEARED };
```
simonbtomlinson: Why not just return `{ type: METERS_FILTER_MODIFIED, term: '' }`? That makes more sense to me than doing it in the reducer.

src/client/app/components/MetersFilterComponent.jsx
```diff
@@ -0,0 +1,44 @@
...
+		this.forceUpdate();
+	}
+
+	render() {
+		const filterInputStyle = {
+			maxWidth: '75%'
+		};
+		return (
+			<div className="form-inline">
+				<input type="text" className="form-control" placeholder="Filter..." style={filterInputStyle} value={this.props.filterTerm} onInput={this.handleChanged} />
```
simonbtomlinson: Doesn't the `value={this.props.filterTerm}` here make the `this.forceUpdate()` unneeded? It'll update to blank to reflect the redux state when the component is re-rendered by react-redux.

src/client/app/components/MetersFilterComponent.jsx
```diff
@@ -0,0 +1,44 @@
...
+				<button className="form-control btn btn-danger" onClick={this.handleClearButtonClicked}>❌</button>
```
simonbtomlinson: Should we use a glyphicon (the glyphicon-remove one) here?

src/client/app/reducers/meters_filter.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+/**
+ * @param {State~MetersFilter} state
+ * @param action
+ * @return {State~MetersFilter}
+ */
+export default function metersFilter(state = defaultState, action) {
+	switch (action.type) {
+		case metersFilterActions.METERS_FILTER_CLEARED:
```
simonbtomlinson: You can get rid of this once you change the action.

src/client/app/components/MetersFilterComponent.jsx
```diff
@@ -0,0 +1,44 @@
...
+		// cleared it needs to be invalidated so the vDOM can redraw it.
+		this.forceUpdate();
+	}
+
+	render() {
+		const filterInputStyle = {
+			maxWidth: '75%'
+		};
+		return (
+			<div className="form-inline">
```
simonbtomlinson: This should probably be react-bootstrap components instead of divs and such. We haven't converted everything, but we're slowly migrating to it.

src/client/app/actions/meters_filter.js
```diff
@@ -0,0 +1,24 @@
+/*
```
johndiiorio: All other multi-word files use hypens, not underscores for their names

src/client/app/actions/meters_filter.js
```diff
@@ -0,0 +1,24 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const METERS_FILTER_MODIFIED = 'METERS_FILTER_MODIFIED';
```
johndiiorio: Why do we have Redux actions for filter out meters/groups? This could all be done in a single React component.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -85,6 +86,7 @@ export default class UIOptionsComponent extends React.Component {
 			<div className="col-xs-2" style={divPadding}>
 				<div className="col-xs-11">
 					<div>
+						<MetersFilterContainer />
```
simonbtomlinson: I think this entire search thing should be one component with a search bar and a dropdown list (both components themselves), instead of putting both as siblings `UIOptionsComponent`. You also aren't actually using the search term anywhere yet.

simonbtomlinson:
In response to John's review summary (which I can't comment on directly for some reason), I think that we'll eventually want to generalize this, but won't know what to generalize until we do it in the non-general way 2 or 3 times first. Making this specific is fine for now.

LeoTindall:
Closing in favor of #125

PR #123, opened 2017-09-21
Cleanup testing

simonbtomlinson:
Lots of our tests were using the `db.task()` generator pattern because they were written before async / await. This updates them all to async / await, which is simpler and less error-prone.


johndiiorio:
Could you also change all of the tests to use async/await instead of Promises? The other parts of the PR look great.

simonbtomlinson:
@johndiiorio your comment is fixed in d45fa09.

PR #124, opened 2017-09-22
Add logic to submit new groups and edits to groups

jameeters:
Adds redux and routes to handle submission of new groups and group edits.
Cleans up and re-organizes the groups actions and reducers.
More specifics can be found in commit messages.

### DIFF
--- a/src/server/sql/group/create_groups_tables.sql
+++ b/src/server/sql/group/create_groups_tables.sql
@@ -4,7 +4,7 @@
 
 CREATE TABLE groups (
 	id SERIAL PRIMARY KEY NOT NULL,
-	name VARCHAR(50) UNIQUE NOT NULL
+	name VARCHAR(50) UNIQUE NOT NULL CHECK (char_length(name) >= 1)
 );
 
 /*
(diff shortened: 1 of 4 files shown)


src/server/sql/group/create_groups_tables.sql
```diff
@@ -4,7 +4,7 @@
 
 CREATE TABLE groups (
 	id SERIAL PRIMARY KEY NOT NULL,
-	name VARCHAR(50) UNIQUE NOT NULL
+	name VARCHAR(50) UNIQUE NOT NULL CHECK (char_length(name) >= 1)
```
johndiiorio: This seems useful, but shouldn't there also be a server-side check in routes/groups.js that checks the name length (and sanitizes other input such as group and meter ids) and returns the correct status code to the client?

jameeters: This is enough to reject empty group names. I could add a second check, but it would be functionally equivalent. 
At the moment I'm trusting that the provided group and meter IDs are valid. I'm not sure, but I think the queries to add children should fail if the given IDs are invalid.

jameeters:
There's not enough information provided for me to comment on 1 or 2. 
As to 3, they are sorted by ID and I want to keep it that way.
And 4: I don't intend to implement any more complex error messaging at this time, it's not immediately necessary.

Issue #103, opened 2017-09-13
Groups Graphing


LeoTindall:
Parallel to #57. Allow graphing between groups (each group is represented in aggregate) or within groups (each meter in a single group is represented individually). 

Design questions:
 - [x] How should the UI look?
 - [x] How should nested groups work? (graph subgroups or graph meters? Optional? Default?)

Initially, focus should be on comparing groups with each other.

LeoTindall:
In my opinion, the best experience is obtained if groups can be selected as items (like meters are now) and graphed with all meters and sub-groups in aggregate. Sub-meters can be selected individual or in groups.

It would be good to add a way to select all sub-meters; perhaps a button saying "break down into sub-units" or something. This would be a tree operation, not a flat-map, so:

```
A -+
     |
     + - B
     + - C
     + - D
           + - E
           + - F
      + - G
H - +
       |
       + - I
```

Selecting A would graph the aggregate of B C and D (which would in turn be the aggregate of E and F), but not H (or, by extension, I).

Then, selecting "break into sub units" would separately graph B, C, and D (which is the aggregate of E and F). Selecting "break into sub units" _again_ would separately graph B, C, E, and F. The user could manually select A and D again if they wished.

Does this seem like a good idea?

sandeepacharya464:
I like the idea of displaying one layer of sub-units at a time. This makes it lot easier for the user to figure out the immediate parent and children of various groups and sub-groups.

LeoTindall:
@johndiiorio mentioned that we may want groups to be opaque to the user. In that case, I propose the following changes:

* Show groups in the `meterList` along with meters
* Clicking a group in the `meterList` has the same effect as clicking a meter
* Groups can only be managed from the groups management UI

This results in hiding the DAG from the user, which vastly simplifies the UI.

huss:
I propose we discuss this at our team meeting. It is also likely to be an
item where doing agile development by reaching out to end users for ideas
and feedback during development would be useful.

LeoTindall:
@johndiiorio says: Groups have a name field, perhaps add a group description field in a tooltip and an annotation of how many meters.

Also, add an annotation link that allows users to go to group management.

After this, ship to users and see if they need a drill-down button.

johndiiorio:
Implemented in #148

PR #125, opened 2017-09-23
Add searchable data source selector

### LINKED ISSUE #118: Groups/meters filter search

LeoTindall:
Will close (when merged into master) #118 and lays the groundwork for groups graphing (#103)


src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -81,24 +113,32 @@ export default class UIOptionsComponent extends React.Component {
...
+					<p style={labelStyle}>Meters:</p>
+					<div style={divBottomPadding}>
+						<MultiSelect options={selectOptions} placeholder="Select Meters" theme="bootstrap3" style={multiSelectStyle} onValuesChange={this.handleDatasourceSelect} />
 					</div>
 					<p style={labelStyle}>Graph Type:</p>
 					<div className="radio">
 						<label><input type="radio" name="chartTypes" value={chartTypes.line} onChange={this.handleChangeChartType} checked={this.props.chartToRender === chartTypes.line} />Line</label>
-					</div>
-					<div className="radio">
+						&nbsp; &nbsp; { /* For alignment. */ }
```
johndiiorio: Don't use this for alignment reasons. Use something else like CSS's inline-block.

package.json
```diff
@@ -45,9 +45,12 @@
     "react-bootstrap": "^0.30.10",
     "react-chartjs-2": "^2.0.5",
     "react-dom": "^15.4.2",
+    "react-dom-factories": "^1.0.1",
     "react-rangeslider": "^2.0.1",
     "react-redux": "^5.0.2",
     "react-router": "^3.0.2",
+    "react-selectize": "^3.0.1",
+    "react-transition-group": "^1.2.0",
```
johndiiorio: Why is react-transition-group added? How does it benefit us?

package.json
```diff
@@ -45,9 +45,12 @@
     "react-bootstrap": "^0.30.10",
     "react-chartjs-2": "^2.0.5",
     "react-dom": "^15.4.2",
+    "react-dom-factories": "^1.0.1",
```
johndiiorio: Why are we adding this? It also seems deprecated?

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -4,11 +4,42 @@
...
 import Slider from 'react-rangeslider';
+import { MultiSelect } from 'react-selectize';
+import 'react-selectize/themes/bootstrap3.css';
 import moment from 'moment';
 import 'react-rangeslider/lib/index.css';
 import { chartTypes } from '../reducers/graph';
 import ExportContainer from '../containers/ExportContainer';
 
+// Signifies that the containing object represents a meter
+const DATA_TYPE_METER = 'DATA_TYPE_METER';
```
johndiiorio: You can define these in the constructor, and bind it to `this`.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -4,11 +4,42 @@
...
+// Signifies that the containing object represents a group of meters and groups
+const DATA_TYPE_GROUP = 'DATA_TYPE_GROUP';
+
+/**
+ * Put item's id field in tgt if the item specifies a meter
+ * @param {int[]} tgt The array to perhaps insert an item into
+ * @param {{String, int}} item The item being considered
+ * @return {Array} The modified tgt array
+ */
+function metersFilterReduce(tgt, item) {
```
johndiiorio: Don't put functions like this outside the component. If you need a function, put it as a method in the component (and potentially bind it in the constructor like `this.myNewFunction = this.myNewFunction.bind(this);`

However, these two functions are only used once. Consider just inlining the code in `handleDatasourceSelect()`

package.json
LeoTindall: https://github.com/furqanZafar/react-selectize#install

src/client/app/components/UIOptionsComponent.jsx
LeoTindall: These functions aren't in any way related to the component; why would they need to be associated with it? That's not OOP at all.

src/client/app/components/UIOptionsComponent.jsx
LeoTindall: These constants aren't associated with the component at all; I have them in that file because we currently don't have a better place to put generic constants. Making them associated constants would be very un-OOP.

src/client/app/components/UIOptionsComponent.jsx
LeoTindall: Done

LeoTindall:
@johndiiorio is it acceptable to work on a fork and then push to a branch on the upstream repo for PRs only?

package.json
johndiiorio: In general, I'm against adding three dependencies to our project for something this simple (HTML input and selection field). Every dependency increases our bundle size.

jameeters:
@johndiiorio Cloning his repo is not necessary. You can add it as a remote and interact with his branches in exactly the same way you do with origin's.

johndiiorio:
Didn't think of that, good idea. Ignore my previous comments.

package.json
LeoTindall: Writing the component by hand seems like a poor use of time to save a 71 kilobytes (1,322,461 bytes for my bundle vs 1,250,851 bytes for groupUI) especially given that these dependencies also provide single-select fuzzy search controls (for, e.g. groups editing UI).

src/client/app/components/UIOptionsComponent.jsx
LeoTindall: Is `utils/` a more appropriate place for these unassociated constants?

src/client/app/utils/Datasources.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Put item's id field in tgt if the item specifies a group
+ * @param {int[]} tgt The array to perhaps insert an item into
+ * @param {{String, int}} item The item being considered
+ * @return {Array} The modified tgt array
+ */
+export function groupsFilterReduce(tgt, item) {
+	if (item.type === DATA_TYPE_GROUP) {
+		tgt.push(item.value);
+	}
```
johndiiorio: You don't appear to return `tgt`. @simonbtomlinson, do you agree with this pattern?

LeoTindall: Ah, yeah, that function is messed up (because it's not used yet so I didn't test it, oops).

LeoTindall:
Per changes requested by @johndiiorio, I added a bit of custom CSS to make the selector scroll.

johndiiorio:
Scroll looks good, however, removing a single meter is unintuitive. Could you add little x's next to each meter to remove them from the dropdown?

LeoTindall:
Redone with react-select, now works perfectly with little click-to-remove Xes that use universal symbols.

src/client/app/styles/react-selectize-css.css
```diff
@@ -0,0 +1,4 @@
+.react-selectize.root-node .react-selectize-control {
```
johndiiorio: Do we still need this? Didn't we remove react-selectize?

src/client/app/utils/Datasources.js
johndiiorio: You still need to `return tgt` here.

LeoTindall:
@johndiiorio I fixed the issues. I had made the changes, but my rebase screwed them up :( Anyway, it should be working now.

Issue #126, opened 2017-09-25
Top level React notification system


johndiiorio:
We need a top level React notification system that will display a custom notification to the user. I recommend using something like _react-notification-system_ or implementing this ourselves. The top level component should be wired through Redux, so a single action (containing message, urgency, and display time) can be fired from anywhere to display the notification.

This can be used to display 4xx/5xx status codes from API endpoints, notify users of failed login attempts, etc.

PR #129, opened 2017-09-28
Add top level UI notification system

johndiiorio:
Implements #126

### DIFF
--- a/src/client/app/actions/topLevel.js
+++ b/src/client/app/actions/topLevel.js
@@ -1,0 +1,20 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const SEND_NOTIFICATION = 'SEND_NOTIFICATION';
+export const CLEAR_NOTIFICATIONS = 'CLEAR_NOTIFICATIONS';
+
+/**
+ * @param notification object containing { message, level, position, autoDismiss }
+ * @returns {{type: string, notification: *}}
+ */
+export function sendNotification(notification) {
+	return { type: SEND_NOTIFICATION, notification };
+}
+
+export function clearNotifications() {
+	return { type: CLEAR_NOTIFICATIONS };
+}
--- a/src/client/app/components/LoginComponent.jsx
+++ b/src/client/app/components/LoginComponent.jsx
@@ -51,14 +51,19 @@ export default class LoginComponent extends React.Component {
 			localStorage.setItem('token', response.data.token);
 			browserHistory.push('/admin');
 		})
-		.catch(() => {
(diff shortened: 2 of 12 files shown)


src/client/app/components/LoginComponent.jsx
```diff
@@ -51,7 +51,15 @@ export default class LoginComponent extends React.Component {
...
-		.catch(console.error);
+		.catch(() => {
+			this.inputEmail.focus();
+			this.props.sendNotification({
+				message: 'Invalid email/password combination',
+				level: 'error',
+				position: 'tr',
+				autoDismiss: 3
+			});
+		});
```
jameeters: Will this run on any error, no matter the status? It should be limited to whatever status code indicates an invalid password. (I.e., it should not run if the API returns 5xx).

src/client/app/actions/topLevel.js
```diff
@@ -0,0 +1,20 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+export const SEND_NOTIFICATION = 'SEND_NOTIFICATION';
+export const CLEAR_NOTIFICATIONS = 'CLEAR_NOTIFICATIONS';
```
jameeters: I think `show` and `hide` would be better names for these. 

I also think that `topLevel` is a bad name for this file (and the reducers). This file deals with only notifications, and it should be named something that makes that obvious.

LeoTindall: Agreed; either 'show' and 'hide' or 'add' and 'clear'; 'send' implies some network activity.

src/client/app/components/RouteComponent.jsx
```diff
@@ -5,50 +5,69 @@
...
 	}
-	const token = localStorage.getItem('token');
-	// Redirect route to login page if the auth token does not exist
-	if (!token) {
-		redirectRoute();
-		return;
+
+	componentWillReceiveProps(nextProps) {
+		if (!_.isEmpty(nextProps.notification)) {
+			this.notificationSystem.addNotification(nextProps.notification);
```
LeoTindall: Will this display only one notification at a time? Is there a way to display multiple? For example, if two API requests are made and both fail within a short time, it would be useful to show two error messages, because the first might be more important than the second (bad auth vs. a network timeout, for example).

src/client/app/actions/topLevel.js
johndiiorio: I'm thinking about changing it to `show` and keeping `clear` because `hide` doesn't really encapsulate what the action is doing.

src/client/app/components/RouteComponent.jsx
johndiiorio: I believe that `componentWillReceiveProps` will be called twice for two actions, so it is fine.

PR #132, opened 2017-10-04
Reduce the length of the date labels on BarChartContainer.

LeoTindall:
This fixes the issue of the graph looking tiny with lots of labels (small timescale).

### DIFF
--- a/src/client/app/containers/BarChartContainer.js
+++ b/src/client/app/containers/BarChartContainer.js
@@ -32,7 +32,7 @@ function mapStateToProps(state) {
 			});
 			// Add only the unique time intervals to the label set
 			for (const element of _.flatten(readingsData.readings.map(arr => arr[0]))) {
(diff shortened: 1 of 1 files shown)


src/client/app/containers/BarChartContainer.js
```diff
@@ -32,7 +32,7 @@ function mapStateToProps(state) {
 			});
 			// Add only the unique time intervals to the label set
 			for (const element of _.flatten(readingsData.readings.map(arr => arr[0]))) {
-				labelsSet.add(`${moment(element).format('MMM DD, YYYY, hh:mm a')} - ${moment(element).add(barDuration).format('MMM DD, YYYY, hh:mm a')}`);
```
johndiiorio: Why did you remove the year? Years are important!

LeoTindall: This format was approved by @huss during the meeting today, because it preserves the year in the _second_ date. In almost every case this is totally redundant; when wrapping around (e.g. Dec 10 - Jan 10 2017) it's very clear that the first date is in the previous year.

johndiiorio: Didn't see that, sounds good

PR #134, opened 2017-10-04
Fix group editing, add group deletion

jameeters:


LeoTindall:
I'm not going to approve it because I don't have a way to test it, but from looking at the code it seems good to me.

PR #140, opened 2017-10-06
Compare graph

LucasFares:
Adds a way to compare current weeks usage to last week.


src/client/app/actions/barReadings.js
```diff
@@ -64,3 +75,15 @@ export function fetchNeededBarReadings(timeInterval) {
 		return Promise.resolve();
 	};
 }
+
+export function fetchNeededCompareReadings(timeInterval) {
+    return (dispatch, getState) => {
+    	console.log('Called fetchNeededCompareReadings');
```
johndiiorio: Removed console.log here.

src/client/app/containers/BarChartContainer.js
```diff
@@ -74,6 +74,11 @@ function mapStateToProps(state) {
 			callbacks: {
 				label: tooltipItems => `${data.datasets[tooltipItems.datasetIndex].label}: ${tooltipItems.yLabel} kWh`
 			}
+		},
+		plugins: {
+			datalabels: {
+				display: false
```
johndiiorio: Can you set this to global state somewhere in chartjs instead of individual instances?

src/client/app/containers/CompareChartContainer.js
```diff
@@ -0,0 +1,123 @@
...
+					datalabels: {
+						anchor: 'end',
+						align: 'start',
+					}
+				});
+			// sorts the data so that one doesn't cover up the other
+			data.datasets.sort((a, b) => a.data[0] - b.data[0]);
+		}
+	}
+    //
```
johndiiorio: Remove random comment here

src/client/app/containers/CompareChartContainer.js
```diff
@@ -0,0 +1,123 @@
...
+	const soFar = moment().diff(moment().startOf('week'), 'days');
+	for (const meterID of state.graph.selectedMeters) {
+		const readingsData = state.readings.bar.byMeterID[meterID][timeInterval][barDuration];
+		if (readingsData !== undefined && !readingsData.isFetching) {
+			// Calculate cw
+			for (let i = 1; i <= soFar; i++) {
+				cw += readingsData.readings[readingsData.readings.length - i][1];
+			}
+			// Calculate lw
+			for (let i = 0; i < 7; i++) {
```
johndiiorio: Could you add more comments to how you calculate cw, clw, and lw? Also maybe explain these abbreviations?

src/client/app/containers/CompareChartContainer.js
```diff
@@ -0,0 +1,123 @@
...
+
+
+/**
+ * @param {State} state
+ */
+function mapStateToProps(state) {
+	const timeInterval = state.graph.compareTimeInterval;
+	const barDuration = state.graph.compareDuration;
+	const data = { datasets: [] };
+	const labelsSet = new Set();
```
johndiiorio: labelsSet does not need to be a set, just a list. Also, change the variable name to `labels`.

src/client/app/containers/CompareChartContainer.js
```diff
@@ -0,0 +1,123 @@
...
+						anchor: 'end',
+						align: 'start',
+					}
+				});
+			// sorts the data so that one doesn't cover up the other
+			data.datasets.sort((a, b) => a.data[0] - b.data[0]);
+		}
+	}
+    //
+	data.labels = Array.from(labelsSet);
```
johndiiorio: Change this to `data.labels = labels;` if you fix the above comment about labels

johndiiorio:
Few things.

1. Can you add 'kWh' to the labels on the graph.
2. Is there a way you can have more contrast on the prediction color. It looks very similar.
3. You either need to make it work for multiple meters selected or disallow multiple meters selected. Note that in the groupUI branch, we have a new component for selecting multiple meters, so potentially look into using that (or waiting for groupUI to merge into master, then changing the multi-select component to a single select component).

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -95,16 +95,23 @@ export default class UIOptionsComponent extends React.Component {
...
 					<p style={labelStyle}>Bar chart interval (days):</p>
 					<Slider min={1} max={365} value={this.state.barDuration} onChange={this.handleBarDurationChange} onChangeComplete={this.handleBarDurationChangeComplete} />
 				</div>
 				}
+				{this.props.chartToRender === chartTypes.bar &&
+				<div className="checkbox">
+					<label><input type="checkbox" onChange={this.handleChangeBarStacking} />Bar stacking</label>
+				</div>
+				}
+				{this.props.chartToRender !== chartTypes.compare &&
```
johndiiorio: Sort of pedantic, but you change this to be inclusive to bar and line chart instead of disallowing compare chart.

huss:
I think multiple meters will be some effort so I'm fine with just one for
now (let me know if I have this wrong). We can add that soon. If it won't
take much time to disallow multiple meters then do it. If it will take past
the weekend then let's put it off so we can show off the idea.

LucasFares:
I'll look into disallowing multiple meter select but I think that's going to take a while. The rest of the changes have been made so I think this should be merged and a relevant issue opened to deal with disabling multi select.

huss:
I'm okay with this merging. Let's keep this in the development release
since it is functional but still needs work. It can move to demo when it is
cleaned up. Also, I think we should consider allowing mutli-select on the
graph if that can be done reasonably. It is certainly a feature that is
used by people. We can discuss to decide.

simonbtomlinson:
I don't think this should be merged into master until after it either works or disallows multi-select.

huss:
I thought we had a development branch or where using groupUI at this point
for these types of situations.

LucasFares:
I can create a new "Development" branch and merge into that since we don't seem to have a release system. Or I could wait for groupsUI to merge so I can use the multi select component but I don't foresee solving this problem before fall break.

LucasFares:
Getting multiple meters to display properly at once is very hard. Disabling multi select  turned out to be easier than I thought is and has now been done.

johndiiorio:
Looks great, only one bug. If multiple meters are selected on line/bar, after selecting compare, you run into the same issue. I think you solve this by clearing the selected meters when switching to compare.

Issue #142, opened 2017-10-07
master/devel/release + feature branch system


LeoTindall:
I'd like to propose that we adopt a "moving master" system whereby the master branch isn't expected to be 100% functional. To aid this, we should maintain a "beta" or "devel" branch which is mostly stable and a "release" or "gold" branch which is even with the deployed version on the server (that is, when a merge is made into the release branch, we do a deployment). 

New features would still be developed in separate branches (and hopefully merged quickly).

We could then tie, either manually or using the build automation @johndiiorio and I are working on, the beta/devel branch and the gold/release branch to their respective deployments, so any instances on the Beloit College server would be updated automatically.

huss:
At the last meeting I proposed 3 branches: development, demo & release.
Release is the same as mentioned. Development is code that is under
development but we want to test or show to special collaborators. Demo is
one that is fairly stable and available for comment by others. This is the
way I have been showing the code to others on the server. If it isn't too
much extra work I think this is something to consider but am open to other
ideas.

LeoTindall:
It does impose a certain amount of overhead, but I think it's worth it, since we'll be able to make guarantees about what will work for demos.

I think we should also introduce a front-end manual testing checklist that we go through before pushing to Release.

johndiiorio:
I think we should have a system, but I think having a development, demo, and release branch doesn't quite work for us. I propose that we start tagging our master branch with versions, then keeping our servers updated to proper versions. For instance, the "development" server could be a direct clone of master (patch), the "demo" server could be all new features (minor), and finally having a LTS server (major).

LeoTindall:
I like @johndiiorio's proposal. SemVer is a great standard and instantly conveys all the meaning we need about our versioning.

huss:
I don't care how we functionally implement the idea of three development
levels. Since there has not been any counter feelings, I'm happy to do
tagging. I move we do this to begin getting ready for the upcoming
demonstrations of OED.

huss:
Is this ready to be closed?

LeoTindall:
With the create of the release branch this is complete.

PR #146, opened 2017-10-08
Add version adornments.

LeoTindall:
Relevant to #142.

The file "common/version.js" should be updated in the LAST commit before a tag is made for that version. 

The contents of the created version object are displayed on the app and are available as JSON in the API, from /api/version, in case we ever want to make mobile apps or support very long-running displays (e.g. the "wall graph" use case).

### DIFF
--- a/src/client/app/components/HeaderComponent.jsx
+++ b/src/client/app/components/HeaderComponent.jsx
@@ -31,10 +31,6 @@ export default function HeaderComponent(props) {
 		// Displays the admin button link only if the user is logged in (auth token exists)
 		display: localStorage.getItem('token') ? 'inline' : 'none'
 	};
-	const versionLabelStyle = {
-		fontWeight: 'bold',
-		margin: '10px'
-	};
 	return (
 		<div className="container-fluid">
 			<div className="col-xs-4">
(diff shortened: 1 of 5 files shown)


src/client/app/components/HeaderComponent.jsx
```diff
@@ -7,6 +7,7 @@ import { Link } from 'react-router';
 import { Button } from 'react-bootstrap';
 import LogoComponent from './LogoComponent';
 import UIModalComponent from './UIModalComponent';
+import VERSION from '../../../common/version';
```
johndiiorio: Why is this being important in the HeaderComponent?

src/common/version.js
```diff
@@ -0,0 +1,31 @@
...
+	this.major = versionParts[0];
+	this.minor = versionParts[1];
+	this.patch = versionParts[2];
+}
+
+/**
+ * @returns {string} a string representation of the Semver version of the application,
+ * in the form v1.2.3
+ */
+OEDVersion.prototype.toString = function toString() {
```
johndiiorio: What is the point of parsing the package.json, splitting the major, minor, patch, then recombining them? Why not just prepend 'v' to the beginning of the string?

src/server/app.js
```diff
@@ -29,6 +30,13 @@ app.use('/api/readings', readings);
 app.use('/api/login', login);
 app.use('/api/verification', verification);
 
+/**
+ * Returns the version of the application to the client.
+ */
+app.get('/api/version', (req, res) => {
```
johndiiorio: Even though this is really small, I would prefer this to be in its own route.

src/common/version.js
LeoTindall: It has to be an object to return it as JSON.

PR #147, opened 2017-10-09
Group Graphing

LeoTindall:
Here's the implementation of group graphing. It all works - group line graphs, group bar graphs, and group data exports - except that there's no API endpoint for bar graph information, so it "fakes it" by just requesting data for the meter with the same ID as the selected groups. When the API endpoint is added, it's a single line change to switch it over.

The history is seriously messed up here, I'm trying to work out why?


LeoTindall:
Closed while I figure out the history problem.

Issue #128, opened 2017-09-27
Read IPs in CSV format, not just Excel


LeoTindall:
Currently we only read meter IPs in xlsx files. That's weird since everything else is CSV.

PR #154, opened 2017-10-10
Read meter details from csv

jameeters:
This addresses #128. We now read the IPs from a .csv file instead of a .xlsx file.


LeoTindall:
Closes #128

jameeters:
I added logic to filter the input to only lines containing a valid ipv4 addresses.
I was planning on adding a check for ipv6 addresses, but all the regular expressions I've found for ipv6 addresses have been absurd, incorrect, or both. 
I like the idea of allowing ipv6, but I'm not sure it's worth the effort at the moment.

src/scripts/addMamacMeters.js
```diff
@@ -2,25 +2,29 @@
...
-	const worksheet = workbook.Sheets['Sheet1']; // eslint-disable-line dot-notation
-	return XLSX.utils.sheet_to_json(worksheet);
+async function parseCSV(filename) {
+	// returns a list of lists representing the lines of the file
+	const meterInfo = await readCSV(filename);
+	// the headers should be in the first line
+	const headers = meterInfo[0];
+	const meterDataRows = meterInfo.slice(1);
+	const formattedData = meterDataRows.map(row => _.zipObject(headers, row));
+	console.log(formattedData);
```
johndiiorio: Remove unnecessary console.log

PR #166, opened 2017-10-11
Explicit dependency versioning

jameeters:
Change `package.json` to take only patch updates automatically on all dependencies and devDependencies.
I've updated the required versions where appropriate: I incremented patches and minors where the `^` was used. I incremented only patches where `~` was already used. I updated no major versions.

I've tested that the changes cause no problems on this branch (so master is fine), but I have not checked any other branches.


package.json
```diff
@@ -14,63 +14,61 @@
...
     "express": "~4.13.4",
-    "fs": "0.0.1-security",
-    "jsonwebtoken": "^7.2.1",
-    "lodash": "^4.17.4",
-    "lodash-webpack-plugin": "^0.11.0",
-    "moment": "^2.15.2",
+    "jsonwebtoken": "~7.2.1",
+    "lodash": "~4.17.4",
+    "lodash-webpack-plugin": "~0.11.4",
+    "moment": "^2.18.1",
```
johndiiorio: The latest version of moment is patched

johndiiorio: Also getting: 'Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.'

jameeters:
I've updated the version of moment. Dealing with that deprecation warning is firmly outside the scope of what this PR is supposed to be doing. (That is: this PR is intended to establish which versions of our dependencies we are using, not to make the changes necessary to update those dependencies. )

Issue #159, opened 2017-10-11
Don't bail out on meter not responding


LeoTindall:
When reading meters, the system should gracefully handle errors and should still consume data from working meters if one or more are failed.

johndiiorio:
This bug was fixed in #169

johndiiorio:
The system does not handle errors while reading meters properly (such as one line of the CSV is improperly formatted). Thus, I am reopening this issue.

jameeters:
This issue is "Don't bail out on meter not responding". That is, do not bail out on failed network requests.
PR #169 dealt with that problem, so this issue has been addressed and should remain closed.

PR #169, opened 2017-10-14
Throw out failed meter data requests

jameeters:
Alter `updateMeters` to move past failed requests
Throw out failed network requests and insert data from whatever meters
are reachable, instead of stopping the update for all meters whenever any meter is unreachable. 
This addresses #159.

### DIFF
--- a/src/server/services/updateMeters.js
+++ b/src/server/services/updateMeters.js
@@ -23,7 +23,8 @@ async function updateAllMeters() {
 			metersToUpdate
 				.map(readMamacData)
 				.map(p => p.catch(err => {
-					console.error(err);
+					console.error(`ERROR ON REQUEST TO ${err.options.uri}`); // eslint-disable-line no-console
+					console.error(err.message); // eslint-disable-line no-console
 					return null;
 				}))
 		), elem => elem !== null);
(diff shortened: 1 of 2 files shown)


huss:
This is not for right now but I believe I have a way to have a meter that
can be read and made to fail at will without impacting real submeter data.

jameeters:
I've looked at unit testing this and I think it's complex enough that it should be in a separate pull request, meaning that this one should be approved as-is. I'm happy to continue working on unit testing this, but I'd like that to be a separate issue and PR.

src/server/services/updateMeters.js
```diff
@@ -14,9 +15,19 @@ async function updateAllMeters() {
...
-		const readingInsertBatches = await Promise.all(meters.filter(m => m.enabled && m.type === Meter.type.MAMAC).map(readMamacData));
+		const allMeters = await Meter.getAll();
+		const metersToUpdate = allMeters.filter(m => m.enabled && m.type === Meter.type.MAMAC);
+
+		// Do all the network requests in parallel, then throw out any requests that fail after logging the errors.
+		const readingInsertBatches = _.filter(await Promise.all(
+			metersToUpdate
+				.map(readMamacData)
+				.map(p => p.catch(err => {
+					console.error(err);
```
johndiiorio: Instead of printing out the entire error, could we only print out failing meter id and the RequestError.message, i.e. "Error: connect ETIMEDOUT 8.8.8.8:80" if ip address 8.8.8.8 was failing.

So it would look like: Error: connect ETIMEDOUT 8.8.8.8:80 from meter id 14

PR #170, opened 2017-10-15
MPL Header Check

LeoTindall:
This PR adds an MPL header check for all JavaScript, JSX, SQL and shell script files, Dockerfiles, and Y markup files, integrates it into the CI process, and fixes all missing headers in master.


johndiiorio:
Could this also work for CSS and HTML files? The HTML comment syntax is slightly different, though.

LeoTindall:
Turns out the regex I already wrote matches the HTML comment syntax as well as /* */ and #* syntax.

johndiiorio:
Could you also create a unit test, so that running `npm test` will also catch it?

src/scripts/checkheader.sh
```diff
@@ -0,0 +1,34 @@
...
+# *
+# * This Source Code Form is subject to the terms of the Mozilla Public
+# * License, v. 2.0. If a copy of the MPL was not distributed with this
+# * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# *
+
+echo "[MPL] Checking source files for MPL headers..."
+
+# Checks for the existance of the MPL header
+MPL_REGEX=".*This Source Code Form is subject to the terms of the Mozilla Public\s{0,3}.{0,4}License, v\. 2\.0\. If a copy of the MPL was not distributed with this\s{0,3}.{0,4}file, You can obtain one at http:\/\/mozilla\.org\/MPL\/2\.0\/\."
```
simonbtomlinson: This regex requires that the line is split in at most 2 places (after "Public" and after "this"), and that the beginning-of-line comment syntax is between 0 and 4 characters long. Do we really want to be that specific? It also allows any number of characters before the license starts, which seems wrong to me, although I don't have a good solution for files that need something like /bin/bash at the beginning, so we may need to keep that.

Here's an alternative that can split between any word and allows any number of non-word characters between each word to allow for most comment syntaxes. It just uses the pattern `\W+`, which matches one or more non-word characters, between each word in the license. It also seems to be about twice as fast as yours although both are totally negligible (.016 seconds vs .030 seconds to run the script on my machine).

`.*?This\W+Source\W+Code\W+Form\W+is\W+subject\W+to\W+the\W+terms\W+of\W+the\W+Mozilla\W+Public\W+License,\W+v\.\W+2\.0\.\W+If\W+a\W+copy\W+of\W+the\W+MPL\W+was\W+not\W+distributed\W+with\W+this\W+file,\W+You\W+can\W+obtain\W+one\W+at\W+http:\/\/mozilla\.org\/MPL\/2\.0\/\.`

package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "webpack -d --watch --color --progress",
     "build": "cross-env NODE_ENV=production webpack -p",
     "lint": "./node_modules/.bin/eslint src/client/app/**/*.js src/client/app/**/*.jsx src/server/**/*.js src/server/**/*.jsx src/scripts/**/*.js",
-    "test": "./node_modules/.bin/mocha \"src/server/test/**/*.js\"",
+    "test": "./src/scripts/checkheader.sh && ./node_modules/.bin/mocha \"src/server/test/**/*.js\"",
```
simonbtomlinson: Doesn't this make tests not run if `checkheader` fails? We probably want to run both, right?

src/scripts/checkheader.sh
LeoTindall: Regex implemented in 709246e. I put the header check in its own script and set Travis to run that; I don't think we can lint in Travis just yet (there are too many failures).

src/scripts/checkheader.sh
```diff
@@ -0,0 +1,34 @@
...
+# *
+
+echo "[MPL] Checking source files for MPL headers..."
+
+# Checks for the existance of the MPL header
+MPL_REGEX=".*?This\W+Source\W+Code\W+Form\W+is\W+subject\W+to\W+the\W+terms\W+of\W+the\W+Mozilla\W+Public\W+License,\W+v\.\W+2\.0\.\W+If\W+a\W+copy\W+of\W+the\W+MPL\W+was\W+not\W+distributed\W+with\W+this\W+file,\W+You\W+can\W+obtain\W+one\W+at\W+http:\/\/mozilla\.org\/MPL\/2\.0\/\."
+
+# This find commands lists all our source code files, excluding bad directories
+# It prunes gitignore and gitattributes, and everything in the public dir, .git, node_modules, and postgres-data
+FILES=$(find . -type f \( -iname '*.jsx' -o -iname '*.js' -o -iname '*.sh' -o -iname '*.sql' -o -iname "*.y*ml" \
```
johndiiorio: Could this be slightly generalized to a list of files/paths we want to ignore (like a .gitignore)?
It seems tricky to add additional parameters to the regex.

jameeters: I would also prefer this solution. I think it would be nice to have piecemeal control over exactly which files we want excluded from the header check.

LeoTindall: It's not a regex, it's just a find command, which basically says: include everything which matches the first line, which is a file, and which doesn't match anything on the second two lines. 

There's not really a better way to do this without introducing additional dependencies; we could theoretically write the script in JS, I suppose, but I really don't think this is that complex.

.babelrc
```diff
@@ -1,4 +1,5 @@
 {
+  "__mplheader": "This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.",
```
johndiiorio: I don't think this is the proper way to do the MPL header. You should just use comments.

package.json
```diff
@@ -1,11 +1,13 @@
-{
+{ 
+  "__mplheader": "This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/.",
```
johndiiorio: Do not do this. I recommend using package.json's license field for this very reason.

LeoTindall: That's not what the license field is for; its purpose is to specify the _name_ (or, rather SPDX ID) of the license being used for the package, not to license the file itself. https://docs.npmjs.com/files/package.json#license

.babelrc
LeoTindall: Is the Babel RC not JSON?

package.json
johndiiorio: In every single open sourced Javascript project, I have never seen a license header in the package.json. We should use the license field for the SPDX ID, and leave out the header.

PR #171, opened 2017-10-16
Create line/bar color by hashing label into known distinct colors

johndiiorio:


src/client/app/utils/getGraphColor.js
```diff
@@ -0,0 +1,27 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+const colors = [
+	'#000000', '#00FF00', '#0000FF', '#FF0000', '#01FFFE', '#FFA6FE', '#FFDB66', '#006401',
+	'#010067', '#95003A', '#007DB5', '#FF00F6', '#FFEEE8', '#774D00', '#90FB92', '#0076FF',
+	'#D5FF00', '#FF937E', '#6A826C', '#FF029D', '#FE8900', '#7A4782', '#7E2DD2', '#85A900',
+	'#FF0056', '#A42400', '#00AE7E', '#683D3B', '#BDC6FF', '#263400', '#BDD393', '#00B917',
+	'#9E008E', '#001544', '#C28C9F', '#FF74A3',	'#01D0FF', '#004754', '#E56FFE', '#788231',
```
LeoTindall: Super weird indentation here.

src/client/app/utils/getGraphColor.js
```diff
@@ -0,0 +1,27 @@
...
+	'#0E4CA1', '#91D0CB', '#BE9970', '#968AE8', '#BB8800', '#43002C', '#DEFF74', '#00FFC6',
+	'#FFE502', '#620E00', '#008F9C', '#98FF52', '#7544B1', '#B500FF', '#00FF78', '#FF6E41',
+	'#005F39', '#6B6882', '#5FAD4E', '#A75740', '#A5FFD2', '#FFB167', '#009BFF', '#E85EBE'
+];
+
+export default function getGraphColor(label) {
```
LeoTindall: Two things here: documentation for the function would be nice, and I'd also like a unit test to show that this is deterministic on the label. It's the kind of thing where breaking it would be fairly easy and really annoying to a user.

src/client/app/containers/LineChartContainer.js
```diff
@@ -7,24 +7,24 @@
...
 	const data = { datasets: [] };
-	const graphColors = new GraphColors();
 
 	for (const meterID of state.graph.selectedMeters) {
 		const readingsData = state.readings.line.byMeterID[meterID][timeInterval];
 		if (readingsData !== undefined && !readingsData.isFetching) {
+			const label = state.meters.byMeterID[meterID].name;
 			data.datasets.push({
-				label: state.meters.byMeterID[meterID].name,
+				label,
```
LeoTindall: I personally don't like inconsistent use of insert-by-name syntax in object literals, but feel free to keep it if that's the project style.

johndiiorio: We've been using this everywhere.

src/client/app/utils/getGraphColor.js
johndiiorio: I can add documentation to how this works, but we don't unit test anything on the client. I think when we set up testing, we should add a test for this.

LeoTindall: Ok, sounds good.

PR #175, opened 2017-10-24
Add Weekly Change

LucasFares:
Adds percent changes to compare graph to show user if energy is being saved.

### DIFF
--- a/src/client/app/containers/CompareChartContainer.js
+++ b/src/client/app/containers/CompareChartContainer.js
@@ -26,22 +26,11 @@ function mapStateToProps(state) {
 	// Power used up to this point last week
 	let currentLastWeek = 0;
 	const soFar = moment().diff(moment().startOf('week'), 'days');
-	const delta = function (change) {
-		if (isNaN(change)) {
-			return '';
-		}
-		if (change < 0) {
-			return `You have used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% less energy this week.`;
-		}
-		return `You have used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% more energy this week.`;
+	const delta = change => {
+		if (isNaN(change)) return ''; if (change < 0) return `${state.meters.byMeterID[state.graph.selectedMeters].name} has used ${parseInt(change.toFixed(2).replace('.', '').slice(1), 10)}% less energy this week.`;
+		return `${state.meters.byMeterID[state.graph.selectedMeters].name} has used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% more energy this week.`;
 	};
-
-	const colorize = function (change) {
-		if (change < 0) {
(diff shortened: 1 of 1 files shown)


src/client/app/containers/CompareChartContainer.js
```diff
@@ -26,6 +26,22 @@ function mapStateToProps(state) {
 	// Power used up to this point last week
 	let currentLastWeek = 0;
 	const soFar = moment().diff(moment().startOf('week'), 'days');
+	const delta = function (change) {
```
johndiiorio: You should write this as an arrow function

src/client/app/containers/CompareChartContainer.js
```diff
@@ -26,6 +26,22 @@ function mapStateToProps(state) {
...
+		if (isNaN(change)) {
+			return '';
+		}
+		if (change < 0) {
+			return `You have used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% less energy this week.`;
+		}
+		return `You have used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% more energy this week.`;
+	};
+
+	const colorize = function (change) {
```
johndiiorio: Arrow function

src/client/app/containers/CompareChartContainer.js
```diff
@@ -26,6 +26,22 @@ function mapStateToProps(state) {
...
+		if (change < 0) {
```
johndiiorio: Shouldn't this be <= 0? Or is no change not great?

src/client/app/containers/CompareChartContainer.js
```diff
@@ -26,6 +26,22 @@ function mapStateToProps(state) {
 	// Power used up to this point last week
 	let currentLastWeek = 0;
 	const soFar = moment().diff(moment().startOf('week'), 'days');
+	const delta = function (change) {
+		if (isNaN(change)) {
+			return '';
+		}
+		if (change < 0) {
+			return `You have used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% less energy this week.`;
```
johndiiorio: You should replace "You have used" with "{Meter name} has used {value}% less energy this week"

src/client/app/containers/CompareChartContainer.js
```diff
@@ -26,6 +26,22 @@ function mapStateToProps(state) {
...
+		}
+		return `You have used ${parseInt(change.toFixed(2).replace('.', ''), 10)}% more energy this week.`;
```
johndiiorio: You should replace "You have used" with "{Meter name} has used {value}% more energy this week"

src/client/app/containers/CompareChartContainer.js
LucasFares: We want them to improve!

PR #182, opened 2017-10-26
Add MPL header to log module

johndiiorio:
This fixes the failing Travis build on master. We may need to investigate why the Travis build did not fail before merging.


LeoTindall:
The MPL header check wasn't merged at the time the logging PR was checked. It won't happen anymore because all future PRs will be header checked.

PR #184, opened 2017-10-27
Re-factor group graphing actions.

LeoTindall:
Remove the condition-heavy datasource abstract actions in favor of twice as many datasource specific ones.

Compare is broken on this commit.

### DIFF
--- a/src/client/app/reducers/barReadings.js
+++ b/src/client/app/reducers/barReadings.js
@@ -6,7 +6,6 @@
  */
 
 import * as readingsActions from '../actions/barReadings';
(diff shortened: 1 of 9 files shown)


src/client/app/reducers/barReadings.js
```diff
@@ -27,72 +26,88 @@ const defaultState = {
  * @return {State~Readings}
  */
 export default function readings(state = defaultState, action) {
+	console.log(action);
```
johndiiorio: Remove console.log

johndiiorio:
You also have a fair amount of lint errors, mostly from tabs vs spaces. Run `npm run lint` to see all the errors.

PR #185, opened 2017-10-27
Create display component and implement it

johndiiorio:


src/client/app/components/DisplayComponent.jsx
```diff
@@ -0,0 +1,16 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+
+export default function DisplayComponent(props) {
+	return (
+		<select className="form-control" id="meterList" size={props.height || 8}>
```
LeoTindall: The use of `<select>` here is a bad choice. See my PR for more details.

src/client/app/components/DisplayComponent.jsx
```diff
@@ -0,0 +1,16 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+
+export default function DisplayComponent(props) {
```
LeoTindall: `DisplayComponent` is a very ambiguous name. I suggest `ListDisplayComponent`.

PR #186, opened 2017-10-28
Rewrite DisplayComponent

LeoTindall:
First: `DisplayComponent` is a confusing name. It says nothing about _what_ is being displayed. The component is now `ListDisplayComponent`.

Second: Using a `<select>` is not worth it, and is actively bad in a lot of cases. I've re-written the component to just use a `<div>` wrapping a `<ul>` with some styling.

### DIFF
--- a/src/client/app/components/ListDisplayComponent.jsx
+++ b/src/client/app/components/ListDisplayComponent.jsx
@@ -6,11 +6,29 @@
 import React from 'react';
 
 export default function ListDisplayComponent(props) {
+	const heightPx = (props.height || 8) * 18;
+	const listWrapperStyle = {
+		border: '1px solid #cccccc',
+		borderRadius: '2px',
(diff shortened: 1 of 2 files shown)


src/client/app/components/ListDisplayComponent.jsx
```diff
@@ -0,0 +1,34 @@
...
+		paddingTop: '6px',
+		paddingBottom: '6px',
+		paddingLeft: '12px',
+		paddingRight: '12px',
+	};
+	return (
+		<div className="list-wrapper" style={listWrapperStyle} >
+			<ul id="meterList" style={listStyle} >
+				{props.items.map((item, index) =>
+					<li key={index} disabled>{ item.toString() }</li>
```
johndiiorio: Is there any reason `<li/>` is disabled?

PR #187, opened 2017-10-28
Add check to prevent error on error message after failed meter update

jameeters:
Check that `err.options` exists before attempting to print `err.options.uri`.
If `err.options` exists, but `err.options.uri` does not, it will show up as `undefined`.

### DIFF
--- a/src/server/services/updateMeters.js
+++ b/src/server/services/updateMeters.js
@@ -24,7 +24,8 @@ async function updateAllMeters() {
 			metersToUpdate
 				.map(readMamacData)
 				.map(p => p.catch(err => {
-					log(`ERROR ON REQUEST TO ${err.options.uri}, ${err.message}`, 'error');
+					const uri = err.options !== undefined ? err.options.uri : '[NO URI AVAILABLE]';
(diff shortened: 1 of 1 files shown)


src/server/services/updateMeters.js
```diff
@@ -24,7 +24,8 @@ async function updateAllMeters() {
 			metersToUpdate
 				.map(readMamacData)
 				.map(p => p.catch(err => {
-					log(`ERROR ON REQUEST TO ${err.options.uri}, ${err.message}`, 'error');
+					const uri = err.options !== undefined ? err.options.uri : '[NO URI AVAILABLE]';
```
LeoTindall: Please don't use the ternary operator like this. We can't stop John from doing it, but that doesn't mean the rest of us have to join in. It looks weird and also doesn't handle the case where `err.options` is defined but has no `uri` property.

I suggest:

```javascript
let uri = '[NO URI AVAILABLE]';
if (err.options !== undefined && err.options.uri !== undefined) {
    uri = err.options.uri;
}
```

Issue #179, opened 2017-10-25
Unit test for meter data retrieval


jameeters:
We need unit tests for the functions that get meter data. We especially need one to make sure that failures are handled properly. This will require mocking things in a way we currently don't have to anywhere else in the code.

jameeters:
This was addressed in #189.

PR #189, opened 2017-10-30
Unittest meter polling

jameeters:
This adds a unit test for getting mamac meter data, and makes a few other changes related to this. 
This addresses #179.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "mocha": "~3.5.3",
     "mock-require": "~2.0.2",
     "redux-devtools": "~3.4.0",
-	"sinon": "~4.0.2",
+    "sinon": "~4.0.2",
     "style-loader": "~0.19.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,4 +1,4 @@
-{ 
+{
   "name": "Open-Energy-Dashboard",
   "version": "0.0.0",
   "private": true,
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
     "mocha": "~3.5.3",
     "mock-require": "~2.0.2",
     "redux-devtools": "~3.4.0",
+	"sinon": "~4.0.2",
```
johndiiorio: Weird tab here

package.json
```diff
@@ -1,4 +1,4 @@
-{ 
+{
   "name": "Open-Energy-Dashboard",
   "version": "0.0.0",
   "private": true,
```
johndiiorio: Unrelated to the PR, but we might want to get rid of _private: true_

src/server/test/db/updateMetersTests.js
```diff
@@ -0,0 +1,49 @@
...
+		const metersToUpdate = [goodMeter, badMeter];
+
+		const dataReader = sinon.stub();
+		dataReader.withArgs(goodMeter).resolves(new Reading(
+			goodMeter.id,
+			0,
+			moment('1970-01-01 00:00:00'),
+			moment('1970-01-01 01:00:00')
+		));
+		dataReader.withArgs(badMeter).rejects(new Error('Bland error message'));
```
johndiiorio: Perhaps change the error message here to something more descriptive

jameeters: The message isn't really describing anything to begin with, but I thought the original was more descriptive.

PR #194, opened 2017-11-03
TypeScript: Conservative Implementation

LeoTindall:
This is the conservative implementation of TypeScript. It's full of type errors and has no motion toward refactoring actions or the Redux state. It _does_, however, include the TypeScript compiler as the primary (though not exclusive) compiler in the build process, falling back to Babel to build files that are common with the (non-TypeScript) server.

The BIG change to code here is a change from importing things that don't have default imports as 
```
import React from 'react;
```
to
```
```


simonbtomlinson:
So all frontend files are `.ts` now, right?

LeoTindall:
Yes, that's correct. There should be no .js or .jsx files anywhere in `src/client/app/**/`.

simonbtomlinson:
Can we do a long check for that during the transition period?

LeoTindall:
Not a bad plan; want to put a script in CI for it?

simonbtomlinson:
Sure. Do we want to make it work on windows too for people developing? It's so much easier on Linux.

LeoTindall:
Cygwin, WSL, etc make it easy enough to run Bash scripts on Windows, imo, and batch scripts aren't really up to the task... I could try to do it with PowerShell?

simonbtomlinson:
I don't think it's worth it. Just don't do anything for now, or maybe do it in JavaScript?

LeoTindall:
Let's just not do anything. We can catch incorrect extensions in code reviews.

simonbtomlinson:
Sounds good.

simonbtomlinson:
Let's merge groupui first though.

LeoTindall:
Closing as I need to make so many changes it's not really worth having this.

PR #195, opened 2017-11-05
Remove version information from client

LeoTindall:
Removes the version info from the client, to avoid including package.json in the client.

### DIFF
--- a/src/server/version.js
+++ b/src/server/version.js
@@ -2,18 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/**
- * Creates a new VERSION object, which contains a major, minor, and patch release level.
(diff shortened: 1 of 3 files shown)


src/server/version.js
```diff
@@ -2,18 +2,17 @@
...
+/**
+ * Creates a new OEDVersion object, which contains a major, minor, and patch release level.
+ * OED is compliant with Semver. The version data is fetched from the package.json.
+ */
 function OEDVersion() {
 	const versionParts = pkgJson.version.split('.');
 	if (versionParts.length !== 3) {
-		// Cannot use log module as we do not want the server config included on the client
-		console.error('package.json version string is not in semver x.y.z format'); // eslint-disable-line no-console
+		log(`package.json version string "${pkgJson.version}" is not in semver x.y.z format`, 'error', true);
```
johndiiorio: You don't need to fix it, but the third parameter of log defaults to true.

Issue #196, opened 2017-11-06
sanity checking meter data


huss:
Issue #159 that is trying to avoid failure when OED gets bad meter data. This is to know when it happens and properly log it. There may already be some code to do checks in the importing of meter data. We should find a way to note if the timestamp is invalid or if the gap in time varies from other points and/or exceeds a threshold.  Any such detection would be useful to the people working with the meters and avoid bad data being shown in OED. We should think what would work and then implement this.

PR #197, opened 2017-11-06
Throw errors when parseTimestamp or parseInt fails

jameeters:
So now we can skip over these the same way we already do with failed network requests. 
This addresses #196 in that invalid meter data will no longer cause the entire update to fail, but the meter with the bad data will not be updated.

### DIFF
--- a/src/server/services/readMamacData.js
+++ b/src/server/services/readMamacData.js
@@ -11,7 +11,11 @@ const Reading = require('../models/Reading');
 const parseCsv = promisify(csv.parse);
 
 function parseTimestamp(raw) {
-	return moment(raw, 'HH:mm:ss MM/DD/YY');
+	const ts = moment(raw, 'HH:mm:ss MM/DD/YY');
+	if (!ts.isValid()) {
+		throw new Error('raw timestamp does not parse to a valid moment object');
+	}
+	return ts;
 }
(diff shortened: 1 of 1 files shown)


src/server/services/readMamacData.js
```diff
@@ -26,12 +30,19 @@ async function readMamacData(meter) {
 	if (!meter.id) throw new Error(`${meter} doesn't have an id to associate readings with`);
 	const rawReadings = await reqPromise(`http://${meter.ipAddress}/int2.csv`);
 	const parsedReadings = await parseCsv(rawReadings);
-	return parsedReadings.map(raw => new Reading(
+	return parsedReadings.map(raw => {
+		console.log(raw);
```
simonbtomlinson: Log statement left over.

src/server/services/readMamacData.js
```diff
@@ -26,12 +30,19 @@ async function readMamacData(meter) {
...
+		const reading = parseInt(raw[0]);
+		if (isNaN(reading)) {
+			throw new Error(`Meter reading parses to NaN for meter named ${meter.name} with id ${meter.id}`);
```
simonbtomlinson: You should include the raw string that was supposed to be an int in this message.

PR #198, opened 2017-11-07
GroupUI - Update to Master Feature Parity

LeoTindall:
This PR updates GroupUI to be concurrent with master, so it can be merged, as well as re-factoring the UIOptionsComponent so that the compare functionality works.

The only issues here are 

1. when 8 or 9 meters are selected, the display freaks out (intermittent for me)
2. groups cannot be compared


src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -103,46 +88,29 @@ export default class UIOptionsComponent extends React.Component {
...
 		));
 
-		const groupSelectOptions = this.props.groups.map(group => (
-			{
-				label: group.name,
-				value: group.id
-			}
-		));
+		let singleSelectedMeter;
+		if (this.props.selectedMeters[0]) { singleSelectedMeter = this.props.selectedMeters[0]; }
```
johndiiorio: `singleSelectedMeter` is never used. Also some weird formatting here.

src/client/app/components/UIOptionsComponent.jsx
```diff
@@ -103,46 +88,29 @@ export default class UIOptionsComponent extends React.Component {
...
-				</div>
-				<p style={labelStyle}>Graph Type:</p>
-				<div className="radio">
-					<label><input type="radio" name="chartTypes" value={chartTypes.line} onChange={this.handleChangeChartType} checked={this.props.chartToRender === chartTypes.line} />Line</label>
-					<div style={radioButtonInlinePadding} />
-					<label><input type="radio" name="chartTypes" value={chartTypes.bar} onChange={this.handleChangeChartType} checked={this.props.chartToRender === chartTypes.bar} />Bar</label>
-				</div>
+			<div style={divTopPadding}>
+				<ChartSelectContainer />
+				{ /* Controls specific to the bar chart. */}
```
johndiiorio: bar -> compare

src/client/app/components/ChartSelectComponent.jsx
```diff
@@ -0,0 +1,48 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import React from 'react';
+import { ToggleButtonGroup, ToggleButton } from 'react-bootstrap';
+import { chartTypes } from '../reducers/graph';
+
+/**
+ * A component that allows users to select which chart should be displayed.
+ */
+export default class ChartSelectComponent extends React.Component {
```
johndiiorio: Probably cleaner written as a functional component. Just call `onChange={value => props.changeChartType(value)}` in your render method

johndiiorio:
Also getting _TypeError: Cannot read property 'exportVals' of undefined_
Export button displays on compare chart, which it shouldn't.
For the other charts, the export button downloads a CSV file with the wrong timestamp.

PR #200, opened 2017-11-07
Fix UI crashes with ExportContainer

LeoTindall:


src/client/app/actions/barReadings.js
```diff
@@ -153,9 +154,9 @@ export function fetchNeededBarReadings(timeInterval) {
 		);
 		// Fetch data for any missing groups
 		if (groupIDsToFetchForBar.length > 0) {
-			return dispatch(fetchGroupBarReadings(groupIDsToFetchForBar, timeInterval));
+			futures.push(dispatch(fetchGroupBarReadings(groupIDsToFetchForBar, timeInterval)));
 		}
-		return Promise.resolve();
+		return Promise.all(futures);
```
simonbtomlinson: Is this something you can do? I wasn't aware that `dispatch` returned a promise.

LeoTindall: From the `redux-thunk` docs: "Any return value from the inner function will be available as the return value of dispatch itself. This is convenient for orchestrating an asynchronous control flow with thunk action creators dispatching each other and returning Promises to wait for each other’s completion".

Issue #204, opened 2017-11-09
Graphs should be labeled KW not KWh


huss:
We are incorrectly labeling the graphs as KWh when they are actually KW.  We should relabel all graphs (lines, bar & compare) to be KW. We need to do this ASAP and include this when we update all the systems with the new version (0.2). It can be a new subversion if we want.

johndiiorio:
This is addressed in PR #206

PR #206, opened 2017-11-09
Change kWh to kW on graphs/export

johndiiorio:


src/client/app/containers/BarChartContainer.js
```diff
@@ -84,7 +84,7 @@ function mapStateToProps(state) {
 				stacked: state.graph.barStacking,
 				scaleLabel: {
 					display: true,
-					labelString: 'kWh'
+					labelString: 'kW'
```
simonbtomlinson: The bar chart is still kWh

johndiiorio: What about the compare chart?

simonbtomlinson: I'm not sure. My guess is it's kWh, but it depends on where it's getting its data. It probably should be kWh.

huss: Issue #204 says that all should be KW. The rationale is that you will see the same units as you change representations. Also, for compare, it will not cause the bars to change height so it is only a units question. Also, since the bar graph is related to the compare I think it is best if they are the same.

simonbtomlinson: I thought we were displaying a different your of data for the bar chart. I just looked at the SQL and it should all be kW. I was wrong.

PR #207, opened 2017-11-12
Chart hotlinking

johndiiorio:
Allows one user to share a chart link with another user.

Known bug: If the user receiving the link changes part of the query parameter to an invalid meter/group id manually, then UIOptionsContainer throws an error (because it cannot get a name on an invalid meter).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3793,6 +3793,12 @@
         }
       }
     },
+    "diff": {
+      "version": "3.4.0",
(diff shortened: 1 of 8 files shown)


johndiiorio:
I disagree, think of sites like Dropbox and Google drive which generate a shareable link to the content. Users also might not notice that their URL changed.

LeoTindall:
Fair enough; can we do both?

simonbtomlinson:
@LeoTindall both makes a lot of sense to me.

PR #215, opened 2017-11-14
Compare groups

LucasFares:
Integrates groups and compare chart.


LucasFares:
I'm having trouble replicating the error.

LeoTindall:
That's very odd. It repros 100% of the time for me, with the following steps:

1. start app
2. Go to compare chart
3. Select at least 1 meters (the chart for this meter appears)
4. Select a group. The layout correctly resizes, but the chart does not appear, and the error message is printed. Repeat this any number of times to get more error messages and more resizes.
5. Selecting additional meters now has the same effect as selecting groups.

johndiiorio:
@LeoTindall I cannot reproduce this issue

LeoTindall:
OK, must just be my system. Merge it, if there's an issue later we can open a bug.

PR #216, opened 2017-11-14
Fixes compare flicker

LucasFares:
Only displaying 2 graphs per line seems to fix the flickering. I will continue to look into fixing the flickering using a footer but for the sake of a timely feature rollout this should do the trick.

### DIFF
--- a/src/client/app/components/MultiCompareChartComponent.jsx
+++ b/src/client/app/components/MultiCompareChartComponent.jsx
@@ -6,17 +6,20 @@ import React from 'react';
 import CompareChartContainer from '../containers/CompareChartContainer';
 
 export default function MultiCompareChartComponent(props) {
-	// Compute how much space should be used in the bootstrap grid system
-	let size;
-	if (props.selectedMeters.length + props.selectedGroups.length === 1) {
-		size = 12;
-	} else {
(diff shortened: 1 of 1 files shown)


src/client/app/components/MultiCompareChartComponent.jsx
```diff
@@ -6,21 +6,22 @@ import React from 'react';
...
-
 	const centeredStyle = {
 		marginTop: '20%',
 		textAlign: 'center'
 	};
 
+	const flexContainerStyle = {
+		display: 'flex',
+		flexFlow: 'row wrap',
+		flexShrink: '3'
```
johndiiorio: Remove flexShrink, not needed

src/client/app/components/MultiCompareChartComponent.jsx
```diff
@@ -6,21 +6,22 @@ import React from 'react';
...
+	};
+
+	const flexChildStyle = {
+		width: '30%',
+		flexGrow: '1'
```
johndiiorio: Change this line to `flexGrow: props.selectedMeters.length + props.selectedGroups.length > 3 ? '0' : '1'`

PR #224, opened 2017-11-19
Move js scripts to services, add server scripts

johndiiorio:


src/scripts/oed.service
```diff
@@ -0,0 +1,28 @@
+# *
+# * This Source Code Form is subject to the terms of the Mozilla Public
+# * License, v. 2.0. If a copy of the MPL was not distributed with this
+# * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# *
```
jameeters: Files that users are required to edit during setup should not have the MPL header.

src/scripts/updateOED.bash
```diff
@@ -5,11 +5,8 @@
 # * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # *
```
jameeters: This file also has to be edited during setup, and should not have an MPL header. I should have caught that when the change was first made.

Also, I think `updateMamacMetersCron` would be a more accurate name for this file. `updateOED` could easily be understood to update the OED instance itself, rather than fetching new readings.

src/server/services/createUser.js
```diff
@@ -6,21 +6,27 @@
...
 
 const rl = readline.createInterface({
 	input: process.stdin,
 	output: process.stdout
 });
 
+// Verifies that an email is valid
+function checkEmail(email) {
+	// See https://stackoverflow.com/a/46181/5116950
+	// eslint-disable-next-line
```
jameeters: Name the rule you are disabling.

src/scripts/updateOED.bash
jameeters: When we add more functionality, we can rename it, but for now it should be named for what it does, and all it does right now is update mamac meters.

INSTALL.md
```diff
@@ -62,3 +40,39 @@ OED_LOG_FILE=?                 // Path to the log file, defaults to ./log.txt
...
+* Restore a database dump with by first copying the dump into the container with `docker cp /path/to/dump.sql container_name:/dump.sql` and then restoring it into the database with `docker-compose exec database psql -U oed -f /dump.sql`.
+
+### Upgrading the App ###
+
+To upgrade the app:
+1. Stop the app (`systemctl stop oed.service`)
+1. Store your local config changes with `git stash` 
+1. Update with `git pull`. 
+1. Replace your local changes with `git stash pop`
+1. Re-build the app (`docker-compose run --rm web ./src/scripts/init.sh NONE --build`)
```
johndiiorio: I think we need the --no-user flag here. @LeoTindall, is this right?

src/scripts/checkHeader.sh
```diff
@@ -22,24 +22,27 @@ FILES=$(echo "$FILES" | grep -v "src\/client\/app\/images\/.*")
 FILES=$(echo "$FILES" | grep -v "src\/client\/favicon\.ico")
 # Filter out test data
 FILES=$(echo "$FILES" | grep -v "src\/server\/test\/db\/data\/.*")
+# Filter out outside scripts
+FILES=$(echo "$FILES" | grep -v "src\/scripts\/oed\.service")
+FILES=$(echo "$FILES" | grep -v "src\/scripts\/OEDCron\.bash")
```
jameeters: This isn't really related to this PR, but I wonder if it would be nice to add something like `/* no MPL header */` to exclude a file from the MPL header check, rather than having to add them to this script.

jameeters:
My only remaining problem is that I think the name of `OEDCron` should be changed to reflect the fact that the only thing it currently does is update the mamac meters. I know this all but guarantees that we will have to change the name in the future, but I think we should have the name describe the functionality of the file as exactly as possible for now. 
+

PR #225, opened 2017-11-20
fixed a small bug in the Bar intervals

Majidhub:
These added lines help eliminate the dates with no recorded readings from showing up on the bar intervals. It replaces the last date with the date of the latest reading instead of adding the bar duration without checking if we really have readings for all the dates showing up or not.


Majidhub:
Hey John,

Thank you for taking the time to review.

Can I just ask if your line graph shows readings up to 11/21/17? Or does it
also stop at 11/20/17? Because the line graph is where I take the last date
from.

johndiiorio:
@Majidhub, it looks like the line graph only goes up to the previous date. This is probably due to the compressed readings. Do you also encounter this as well?

Majidhub:
Yes I do actually, will it always go to one day before? Because in that
case we'll just add one day at the very end to show on the bar interval to
fix the problem

johndiiorio:
I don't think this is a complete fix. While it works if the user has already fetched line readings for the meter/group to be graphed, it does not work if the user directly starts graphing bar readings. @simonbtomlinson or @jameeters, do you agree with this/have an idea how to proceed with this bug?

PR #234, opened 2017-11-22
Footer

LucasFares:
Adds a footer with contact information across the site. As discussed, it is positioned absolutely and works as intended for all cases except 10+ compare charts and mobile.


src/client/app/components/FooterComponent.jsx
```diff
@@ -0,0 +1,32 @@
...
+		borderTop: '1px #e1e4e8 solid',
+		color: '#999',
+	};
+
+
+	return (
+		<div className="fluid-container">
+			<footer className="footer" style={footerStyle}>
+				<span>Open Energy Dashboard is an open source project coordinated by Beloit College. <a href="mailto:oed@beloit.edu">Contact Us</a> or
+					visit our <a href="https://github.com/beloitcollegecomputerscience/OED">website</a> for more information.</span>
```
johndiiorio: Recommend changing the <a> tag to open in a new tab.

src/client/app/components/FooterComponent.jsx
```diff
@@ -0,0 +1,32 @@
...
+		bottom: '1em',
+		paddingTop: '1em',
+		paddingLeft: '.5em',
+		borderTop: '1px #e1e4e8 solid',
+		color: '#999',
+	};
+
+
+	return (
+		<div className="fluid-container">
```
johndiiorio: Don't you want "container-fluid"?

src/client/app/components/FooterComponent.jsx
```diff
@@ -0,0 +1,32 @@
...
+import React from 'react';
+
+
+/**
+ * React component that controls the footer strip at the bottom of all pages
+ * @param props The props passed down by the parent component
+ * @return JSX to create the footer
+ */
+export default function FooterComponent() {
+	const footerStyle = {
```
johndiiorio: Recommend making this either center or right-aligned, not left-aligned.

johndiiorio:
I'm also against introducing a known bug into our code. I would like to not merge this immediantly, and see if we can come up with a solution to the problem.

LucasFares:
While I understand your concern, it's not so much a bug as a graphical annoyance in a very narrow use case. As per what we talked about this afternoon it's important that we have this contact info on the dashboard so I think this should be merged now.

src/client/app/components/FooterComponent.jsx
```diff
@@ -0,0 +1,35 @@
...
+		textAlign: 'center',
+		width: '98%'
+	};
+
+
+	return (
+		<div className="container-fluid">
+			<footer className="footer" style={footerStyle}>
+				<span>Open Energy Dashboard is an open source project coordinated by Beloit College. <a target="_blank" rel="noopener noreferrer" href="mailto:oed@beloit.edu">Contact us</a> or
+					visit our <a href="https://github.com/beloitcollegecomputerscience/OED">website</a> for more information.</span>
```
johndiiorio: You need target="_blank" on the `<a/>` tag to the OED website

LucasFares:
The link should behave like a normal link and not open itself in a new tab. A user can right click and select the appropriate option like they would with any other link.

huss:
I don't feel strongly about this but since this link takes the user away from the dashboard I am inclined to open it in a new tab/window. I also think that sometimes the user may be reporting an issue and want to keep the current window with the dashboard. I understand a user could make it open a new tab/window but they may not. Just my thoughts.

jameeters:
I was under the impression that part of the goal of the footer was to move our branding out of the header and replace it with institution-specific branding. I also think that we should have a link to our github in the footer, I think most OSS projects do so. I was thinking that we'd be going along the lines of the attached thumbnails. 
footer_thumbnails

johndiiorio:
You also need to add the Footer component to the LoginComponent

PR #235, opened 2017-11-27
Admin panel and admin preferences

### LINKED ISSUE #106: Admin panel
The admin panel should allow admins to import new meters, create/edit/view groups, and any other administrative duties.

There has already been some work done on this, located in the "adminPanel" branch.

### LINKED ISSUE #145: Consistent Error/Exception UX
Currently, the app has very little in the way of error handling. Thanks to #126 we now have a consistent, global mechanism for displaying arbitrary messages to the user. I'd like to have:

1. A consistent doctrine for when we want to spend time preventing users from doing things that are likely to cause errors. There has been some discussion of this re: naming (see #135, for example), but no consensus in the general case.
1. A global exception handling framework, especially implemented for failed API requests, as they are the main point of failure. A user should not ever have to guess at why a chart isn't showing up; error visibility and system status transparency are [major usability features](https://uxplanet.org/golden-rules-of-user-interface-design-19282aeb06b), and it's [important to communicate failure clearly](https://www.nngroup.com/articles/error-message-guidelines/).

(Note also the difference between _errors_ and _exceptions_; errors are things the user does wrong while exceptions are exceptional conditions encountered during normal operation, like network problems).

### LINKED ISSUE #149: allow admin to set name on dashboard
As we roll this out to other organizations, they will want to put their name at the top. They also may want a logo. We need to plan where the OED info will go on the page and implement this.

johndiiorio:
This PR addresses a number of topics:

- Admin panel
- Preferences for graph items (set by admins)
- Improves the HeaderComponent's buttons
- Creates a top-level initialization component
- Improves the CSS on the GroupView page
- Improves the CSS on and around the Header component

This PR does not add every possible user preference. It only implements the default chart type, default bar stacking, and site title as a proof of concept.

Closes #106
Closes #145 
Closes #149


src/client/.eslintrc.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
 		"browser": true
 	},
 	"rules": {
-		"no-console": "off",
+		"no-console": 1,
```
jameeters: This change has nothing to do with the admin panel and should be a separate PR.

src/client/app/actions/admin.js
```diff
@@ -0,0 +1,60 @@
...
+
+function fetchPreferences() {
+	return dispatch => {
+		dispatch(requestPreferences());
+		return axios.get('/api/preferences')
+			.then(response => {
+				dispatch(receivePreferences(response.data));
+				dispatch((dispatch2, getState) => {
+					const state = getState();
+					dispatch2(changeChartToRender(state.admin.defaultChartToRender));
```
jameeters: Why assign `state` at all? Why not use `getState().admin. . . `?

src/client/app/actions/groups.js
```diff
@@ -277,9 +278,14 @@ function submitNewGroup(group) {
...
-			.catch(error => {
+			.catch(() => {
 				dispatch(markGroupInEditingNotSubmitted());
-				console.error(error);
+				dispatch(showNotification({
+					message: 'Failed to create a new group',
+					level: 'error',
+					position: 'tr',
+					autoDismiss: 3
+				}));
```
jameeters: It's fine to show a notification, but it should also log to the console.

src/client/app/actions/groups.js
```diff
@@ -296,9 +302,14 @@ function submitGroupEdits(group) {
...
-			.catch(error => {
+			.catch(() => {
 				dispatch(markGroupInEditingNotSubmitted());
-				console.error(error);
+				dispatch(showNotification({
+					message: 'Failed to edit group',
+					level: 'error',
+					position: 'tr',
+					autoDismiss: 3
+				}));
```
jameeters: It's fine to show a notification, but it should also log to the console.

The changes to this file also have nothing to do with the admin panel, and should not really be in this PR.

src/client/app/actions/groups.js
```diff
@@ -354,6 +365,11 @@ export function deleteGroup() {
 					dispatch2(changeDisplayMode('view'));
 				});
 			})
-			.catch(console.error);
+			.catch(dispatch(showNotification({
+				message: 'Failed to delete group',
+				level: 'error',
+				position: 'tr',
+				autoDismiss: 3
+			})));
```
jameeters: It's fine to show a notification, but it should also log to the console.
This is also the third usage of this exact logic in this file. It should be factored out into a function of its own.

src/client/app/components/HeaderComponent.jsx
```diff
@@ -15,41 +15,71 @@ import getToken from '../utils/getToken';
  * @return JSX to create the header strip
  */
 export default function HeaderComponent(props) {
-	const title = props.title ? props.title : 'Open Energy Dashboard';
+	const urlArr = window.location.href.split('/');
```
jameeters: is `urlArr` short for `urlArray`? 
If so, please just type out the whole word.

src/client/app/reducers/admin.js
```diff
@@ -0,0 +1,48 @@
...
+
+import * as adminActions from '../actions/admin';
+import { chartTypes } from './graph';
+
+const defaultState = {
+	displayTitle: '',
+	defaultChartToRender: chartTypes.line,
+	defaultBarStacking: false,
+	isFetching: false
+};
```
jameeters: I think it would be better to track the current settings and the settings being edited separately, but I don't think it's absolutely necessary.
It is necessary, however, to have a flag to store whether or not there are any changes that have yet to be submitted.

src/client/app/components/AdminComponent.jsx
```diff
@@ -3,15 +3,129 @@
...
+	handleDefaultChartToRenderChange(e) {
+		this.props.updateDefaultGraphType(e.target.value);
+	}
+
+	handleDefaultBarStackingChange() {
+		this.props.toggleDefaultBarStacking();
+	}
+
+	handleSubmitPreferences() {
+		axios.post('/api/preferences',
```
jameeters: As far as I know, this is the only place in the project where we make a request from a react component.
Is there a very good reason for this?
This seems to make it difficult, if not impossible, to properly track whether or not the admin preferences have been submitted. I think this should be moved into an action, and the logic handling submissions made significantly more robust.

src/client/app/components/AdminComponent.jsx
```diff
@@ -3,15 +3,129 @@
...
+					autoDismiss: 3
+				});
+			})
+			.catch(() => {
+				this.props.showNotification({
+					message: 'Failed to submit changes',
+					level: 'error',
+					position: 'tr',
+					autoDismiss: 3
+				});
```
jameeters: You have this 'show notification' logic everywhere, and the only parameters that differ are the level and the message.
These should be factored out into functions that take in a message.

src/client/app/components/AdminComponent.jsx
johndiiorio: There are multiple places this is done, including LoginComponent, AdminComponent, CreateGroup component, etc. It could be moved into Redux, but it ultimately does not touch state at all. Therefore, there is no reason to put it into Redux. Note that the examples above are all POST requests.

src/client/app/reducers/admin.js
johndiiorio: Why is it important to check if there are any changes that have yet to be submitted?

src/client/app/components/AdminComponent.jsx
simonbtomlinson: This is something I'd like to fix eventually. We should refactor all of the network stuff out into its own part of the project and then only call it from thunk actions.

src/client/app/reducers/admin.js
jameeters: Because if there are no changes that need to be submitted, then it should not be possible to send the network requests. It also rules out somehow sending a second request while the first is still in flight. It's also generally good practice to write redux actions defensively --- you never know how they'll be used in the future.

src/client/app/components/AdminComponent.jsx
jameeters: @johndiiorio of the examples you mentioned, in master, only loginComponent makes a post request from the react.* That example is fundamentally different because it involves handling passwords.
This should move to redux. It will need to touch state to toggle a `submitted` flag. 

*If there are versions of these that _do_ make post requests from the react, then I will most likely have exactly the same problems with them.

LeoTindall: I agree with Simon; I think once TypeScript is done (soon!) we can do this pretty easily.

johndiiorio: My philosophy is that Redux should only be for the code that directly interacts with the state. If we want to factor out all the network requests into some file, that's fine. I just disagree with the all Network requests => places in thunks philosophy.

Also, @jameeters, I don't see the need of a submitted flag for submitting admin preferences. Could you elaborate as to why that is necessary?

src/client/.eslintrc.js
johndiiorio: See PR #237

src/client/app/utils/notifications.js
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import store from '../index';
+import { showNotification as showNotificationAction } from '../actions/notifications';
+
+export function showSuccessNotification(message, position = 'tr', autoDismiss = 3) {
```
LeoTindall: To forestall future comments on this technique (exporting store and dispatching on it), this is, as far as I can tell, the best way to create a global dispatcher without having to modify every Container that could possibly have errors.

src/client/app/components/HeaderComponent.jsx
```diff
@@ -14,47 +14,95 @@ import getToken from '../utils/getToken';
...
+	}
+
+	render() {
+		const urlArr = window.location.href.split('/');
+		const page = urlArr[urlArr.length - 1];
+		let renderOptionsButton = false;
+		let renderLoginButton = false;
+		let renderAdminButton = false;
+		let renderGroupsButton = false;
+		const renderLogoutButton = !!getToken();
```
jameeters: Use `Boolean()`.

src/client/app/components/HeaderComponent.jsx
```diff
@@ -14,47 +14,95 @@ import getToken from '../utils/getToken';
...
+
+		switch (page) {
+			case '': // home page
+				renderOptionsButton = true;
+				if (getToken()) {
```
jameeters: Why do you have to cast to boolean earlier, but not here?

src/client/app/reducers/admin.js
```diff
@@ -0,0 +1,54 @@
...
+
+import * as adminActions from '../actions/admin';
+import { chartTypes } from './graph';
+
+const defaultState = {
+	displayTitle: '',
+	defaultChartToRender: chartTypes.line,
+	defaultBarStacking: false,
+	isFetching: false,
+	isSubmitting: false
```
jameeters: I think we generally call this `submitted` in other parts of the project. That would be better.

src/client/app/reducers/admin.js
```diff
@@ -0,0 +1,54 @@
...
+			};
+		case adminActions.RECEIVE_PREFERENCES:
+			return {
+				...state,
+				isFetching: false,
+				displayTitle: action.data.displayTitle,
+				defaultChartToRender: action.data.defaultChartToRender,
+				defaultBarStacking: action.data.defaultBarStacking
+			};
+		case adminActions.TOGGLE_IS_SUBMITTING_PREFERENCES:
```
jameeters: I think it's better to have two functions, one to set `submitted = true` the other to set it to `false`. Doing that makes it much harder to toggle `submitted` into the wrong state.

src/client/app/reducers/admin.js
johndiiorio: This is not called `submitted` because the flag gets flipped back after the request completes. Otherwise, it is not clear. The best example is the `isFetching` flag directly above it.

jameeters: The flag shouldn't be flipped back unless the request fails, until there are new changes to be submitted.

PR #241, opened 2017-11-29
Add more reading validation

jameeters:


src/server/services/readMamacData.js
```diff
@@ -11,6 +11,10 @@ const Reading = require('../models/Reading');
 const parseCsv = promisify(csv.parse);
 
 function parseTimestamp(raw) {
+	const timestampRegExp = /^\d{2}:\d{2}:\d{2} \d{1,2}\/\d{1,2}\/\d{1,2}/;
```
johndiiorio: Place end of string regex char

PR #271, opened 2018-01-30
Improve responsive support

### LINKED ISSUE #94: Mobile device support
We need first-class responsive mobile support, including placing the options for manipulating the graphs in a way that works on mobile. This is going to require some discussion about what we want the site to look like on a phone.

johndiiorio:
This PR updates Bootstrap to from version 3 to 4. As react-bootstrap only works for Bootstrap 3 elements, this replaces react-bootstrap with Reactstrap. Bootstrap 4 also removes icon support. Thus, we now have a dependency on font awesome (open for discussion).

Secondly, I move the navigation buttons in the header into the new MenuModalComponent (combination of UIOptions + HeaderButtons) when operating on smaller devices.

Overall, our webpack bundle size has decreased, our UI should look better, and responsiveness improved. Please thoroughly test to make sure everything works as intended. 

Closes #94


jameeters:
I think we could probably use the unicode left and right heavy angle brackets. ([Demo.](https://www.w3schools.com/charsets/tryit.asp?deci=10096)), but I also think the edit group UI needs a redesign and I'd be willing to put off the issue until then. (As long as that's not indefinitely.)

src/client/app/components/HeaderButtonsComponent.jsx
```diff
@@ -0,0 +1,92 @@
...
+		this.handleLogOut = this.handleLogOut.bind(this);
+	}
+
+	handleLogOut() {
+		localStorage.removeItem('token');
+		this.forceUpdate();
+	}
+
+	render() {
+		const urlArr = window.location.href.split('/');
```
simonbtomlinson: What's going on here? This is pretty complex logic that should be moved into a different function (or functions). And do we really want to use the current url as a source of state?

src/client/index.html
```diff
@@ -7,7 +7,7 @@
     <meta charset="utf-8">
     <title>Open Energy Dashboard</title>
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
-	<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
+	<link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet">
```
simonbtomlinson: Is there a reason why we're including fontawesome in this way when we include bootstrap via npm and webpack?

src/client/app/components/HeaderButtonsComponent.jsx
johndiiorio: This is not part this PR, but from a PR a couple months ago where the header component was redone to have it figure out which navigation buttons to display, instead of having the options passed down through props.

simonbtomlinson: Fair enough. I think it should be changed though.

PR #273, opened 2018-02-01
Validating JSONSchema for routes 

### LINKED ISSUE #246: Validate incoming JSON for all routes with jsonschema
Only some of the express routes are currently using jsonschema. We need all routes to be protected with jsonschema.

buip:
I hope I validated all the routes. If there is a route that I miss, please let me know!

This resolves #246


src/server/routes/preferences.js
```diff
@@ -28,11 +29,36 @@ router.use(authentication);
...
+		}
+	};
+	if (!validate(req.body, validParams).valid) {
+		res.sendStatus(400);
+	} else {
+		try {
+			const rows = await Preferences.update(req.body.preferences);
+			res.json(rows);
+		} catch (err) {
+			log.error(`Error while performing POST update preferences: ${err}`, err);
```
simonbtomlinson: You should probably send an error here.

simonbtomlinson:
Looks good to me now. @johndiiorio, can you take a look?

simonbtomlinson:
@johndiiorio I think those are either GET routes or not places where json is expected. Can you take another look?

johndiiorio:
@simonbtomlinson They still need to be validated as the user can send information through query parameters. Thus, `req.params` need to be validated instead of `req.body`.

buip:
There are two things I don't know if I should check:
req.query.barDuration
req.query.timeInterval

src/server/routes/authenticator.js
```diff
@@ -4,14 +4,20 @@
...
  * Middleware function to force a route to require authentication
  * Verifies the request's token against the server's secret token
  */
 module.exports = (req, res, next) => {
 	const token = req.headers.token || req.body.token || req.query.token;
-	if (token) {
+	const validParams = {
+		type: 'string',
+	};
+	if (!validate(validParams, token).valid) {
```
johndiiorio: You have validParams and token backwards. This causes all authenticated routes to fail.

buip:
I checked again to make sure the parameters are correct. Thank you for noticing it!

PR #279, opened 2018-02-02
Add meters properly in init

### LINKED ISSUE #278: Docker init script does not add meters

LeoTindall:
Closes #278

### DIFF
--- a/src/scripts/init.sh
+++ b/src/scripts/init.sh
@@ -22,6 +22,8 @@ if [ ! -e "$1" ] && [ "$1" != "NONE" ]; then
     exit 1
 fi
 
+IPS_FILE=$1
+
 while [ "$2" != "" ]; do
     case "$2" in
         -b) BUILD=yes;;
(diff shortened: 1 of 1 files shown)


src/scripts/init.sh
```diff
@@ -77,8 +79,8 @@ echo "Schema created or already exists."
 # Pull the meters into the DB and get data
 if [ $1 != "NONE"  ]; then
     set -e
-    echo "Trying to add meters."
-    npm run addMamacMeters $1 2> /dev/null
+    echo "Trying to add meters from '$IPS_FILE'."
+    npm run addMamacMeters $IPS_FILE 2> /dev/null
```
johndiiorio: Just curious, how does this fix the issue? Aren't you just aliasing $1 to IPS_FILE?

LeoTindall: Yes. The way I parse arguments causes arg 2 to be shifted to arg 1 and then parsed as an argument to `node` rather than to `addMamacMeters`.

PR #281, opened 2018-02-02
Scroll bar fix

sandeepacharya464:
- fixed the scroll bar issue. Now scroll bar does not display when you do not have any item. it displays as soon as you click an item.  @simonbtomlinson @johndiiorio


johndiiorio:
This PR slightly fixes the problem in Chrome, but makes the problem much worse in Firefox. I recommend closing this PR for now, and potentially re-open in the future with a better fix.

PR #282, opened 2018-02-03
Fix chart hotlinking, prevents admin preferences from taking precedence

johndiiorio:
This PR adds back the RouteContainer, which was accidentally removed a few PRs ago. This also adds a redux boolean in state.graph called hotlinked that tracks if a graph was hotlinked. This prevents admin preferences from overriding a linked graph.

### DIFF
--- a/src/client/app/actions/admin.js
+++ b/src/client/app/actions/admin.js
@@ -45,18 +45,20 @@ function markPreferencesNotSubmitted() {
 }
 
 function fetchPreferences() {
-	return dispatch => {
+	return (dispatch, getState) => {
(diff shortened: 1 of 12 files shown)


src/client/app/actions/admin.js
```diff
@@ -86,7 +86,7 @@ function submitPreferences() {
 }
 
 function shouldFetchPreferenceData(state) {
-	return !state.admin.isFetching;
+	return !state.admin.isFetching && !state.graph.hotlinked;
```
LeoTindall: This value should be used during _application_ of preferences, not loading.

PR #286, opened 2018-02-05
Allow admins to change a user's password

### LINKED ISSUE #285: Allow the create user script to change a user's password
The `npm run createUser` script does not allow changing a user's password. If an admin enters an email of a user that already exists, it should ask if the admin if the password should be changed.

The script should also be changed to something like `npm run usermod`.

johndiiorio:
Closes #285


INSTALL.md
```diff
@@ -37,7 +37,7 @@ OED_LOG_FILE=?                 // Path to the log file, defaults to ./log.txt
 8. Run ```npm run createdb``` to create the database schema.
 1. Run `npm run addMamacMeters` to load mamac meters from an `.csv` file.
 1. Run `npm run updateMamacMeters` to fetch new data for mamac meters in the database.
-1. Run `npm run createUser` and follow the directions to create a new admin user.
+1. Run `npm run usermod` and follow the directions to create a new admin user.
```
simonbtomlinson: You shouldn't shadow the name of a unix command.

johndiiorio:
That is the reason why I changed the name of the script to usermod. I don't think we need two separate scripts (one to create, one to update), we can just have one. I'll change the name, though.

src/server/services/user/utils.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// Asks the user for a password
+function askPassword(email) {
+	return new Promise(resolve => {
+		rl.question('Password: ', password => { resolve([email, password]); });
+	});
+}
+
+function terminateReadline(message) {
```
buip, line 38: This method is in user directory. Could we move it outside so that it can be used on other scripts as well?

simonbtomlinson: Honestly it would be great if we could abstract away the readline thing entirely.

buip: Maybe we should merge this in now and then refractor it with migrateDB?

PR #290, opened 2018-02-07
Use the '4 Weeks' label instead of 'Month'

### LINKED ISSUE #288: bar month label wrong
The button that says "Month" on bar charts is actually four weeks. We have decided to use four weeks since months vary in number of days. It needs to be relabeled so it is four weeks. Not sure what will easily fit but maybe "4 weeks" or "4 wks".

johndiiorio:
Closes #288


huss:
I assume this does the trick. I thought I should comment on one pull request :-)

PR #292, opened 2018-02-09
Fix file processing reading validation

johndiiorio:


src/server/routes/fileProcessing.js
```diff
@@ -24,7 +24,7 @@ router.post('/:meter_id', upload.single('csvFile'), async (req, res) => {
 		required: ['meter_id'],
 		properties: {
 			meter_id: {
-				type: 'number'
+				type: 'string'
```
simonbtomlinson: Can you add `pattern: '^\d+$'` (escaped whatever way is correct) to ensure it's an integer string?

sandeepacharya464:
I got it! Now, it looks good!

PR #293, opened 2018-02-09
Add error notification when user create a cyclic group

buip:


src/server/routes/groups.js
```diff
@@ -213,8 +213,11 @@ router.put('/edit', async (req, res) => {
 			});
 			res.sendStatus(200);
 		} catch (err) {
-			log.error(`Error while editing existing group: ${err}`, err);
```
johndiiorio: You should move this log to the else condition - we want to log when sending a 500 response.

Issue #244, opened 2017-12-04
Add button to hide chart options


johndiiorio:
Similar to how options are hidden in a modal when the screen size shrinks, we should add the ability for the user to be able to do this.

This admin should be able to set the default of this behavior in the admin panel.

LeoTindall:
I suggest that the admin also be able to decide that the modal is shown for non mobile users

huss:
Just to clarify from the conversations with users about this, we want the chart link to honor this when when the link is used. The idea is to allow website links that focus more attention and real estate on the graphic since that is a common use case. It should also help with fixed monitors (no mouse) and on smaller screens.

PR #295, opened 2018-02-10
Add admin preference to hide chart options

johndiiorio:
Implements #244 

This should not probably not be merged until after TypeScript goes through.


simonbtomlinson:
Let's make sure we merge typescript first. Merging this into the typescript branch would be really awful.

simonbtomlinson:
Don't we want a button to change this for a single page load, not a server-wide change? It seems like turning this on makes the site really unintuitive for the vast majority of users.

NoraCodes:
Yeah, I think this is honestly not a good way to implement this. For the "wall display" use case, we could just set a cookie or localStorage value and check it.

johndiiorio:
Superseded by #331

PR #303, opened 2018-02-21
Typescript - Final PR

NoraCodes:


package.json
```diff
@@ -7,8 +7,11 @@
     "start": "node ./src/bin/www",
     "dev": "webpack -d --watch --color --progress",
     "build": "cross-env NODE_ENV=production webpack -p",
+    "buildDev": "webpack -p",
```
johndiiorio: We should be using the `build` script instead of what `buildDev` does, so we should delete this line.

.travis.yml
```diff
@@ -36,7 +36,9 @@ before_script:
 script:
     # Validate legal requirements on source files
   - npm run checkHeader
-    # Check for eslint failures
+    # Validate that all areas which are intended to be typed, are typed
+  - npm run checkTypescript
```
johndiiorio, line 40: Should we also run `npm run typeCheck`?

src/client/app/actions/groups.ts
```diff
@@ -316,14 +272,17 @@ function submitGroupEdits(group) {
  * @return {function(*, *)}
  */
 export function submitGroupInEditingIfNeeded() {
-	return (dispatch, getState) => {
+	return (dispatch: Dispatch, getState: GetState) => {
 		if (shouldSubmitGroupInEditing(getState())) {
-			const rawGroup = getState().groups.groupInEditing;
+			const rawGroup = getState().groups.groupInEditing as t.GroupDefinition;
+			if (rawGroup === undefined) {
+				throw new Error('Unacceptable condition: state.groups.groupInEditing has no data.');
```
johndiiorio, line 279: I would like to remove all the `throw new Error` statements on the client, or at the very least display an error to the user. I don't think they serve much purpose except for debugging purposes, which should not be in production.

In this case, I would just do:
```
if (rawGroups) {
   ...
}
```

src/client/app/actions/groups.ts
```diff
@@ -343,11 +302,15 @@ export function submitGroupInEditingIfNeeded() {
  * Deletes the group in editing
  * @returns {function(*, *)}
  */
-export function deleteGroup() {
+export function deleteGroup(): Thunk {
 	return (dispatch, getState) => {
 		dispatch(markGroupInEditingDirty());
+		const groupInEditing = getState().groups.groupInEditing as t.GroupDefinition;
+		if (groupInEditing === undefined) {
```
johndiiorio, line 309: Again, another instance of the `throw new Error`

src/client/app/components/DashboardComponent.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { defaults } from 'chart.js';
+import UIOptionsContainer from '../containers/UIOptionsContainer';
+import LineChartContainer from '../containers/LineChartContainer';
+import BarChartContainer from '../containers/BarChartContainer';
+import MultiCompareChartContainer from '../containers/MultiCompareChartContainer';
+import { ChartTypes } from '../types/redux/graph';
+
+/* tslint:disable no-string-literal */
+// TODO TYPESCRIPT: There is an open PR for this: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21842
+defaults.global.plugins = {datalabels: {display: false}};
```
johndiiorio: This PR has been merged. Is this still an issue, and if so, we should comment it differently.

src/client/app/components/InitializationComponent.tsx
```diff
@@ -0,0 +1,48 @@
...
+	fetchPreferencesIfNeeded(): Promise<any>;
+	changeOptionsFromLink(options: LinkOptions): Promise<any>;
+}
+
+export default class InitializationComponent extends React.Component<InitializationProps, {}> {
+	private notificationSystem: NotificationSystem.System;
+
+	public componentWillMount() {
+		this.props.fetchMeterDetailsIfNeeded();
+		this.props.fetchGroupDetailsIfNeeded();
```
johndiiorio, line 25: This line is not part of development. Why do we fetch group details here?

src/client/app/components/InitializationComponent.tsx
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public componentWillReceiveProps(nextProps: InitializationProps) {
+		if (!_.isEmpty(nextProps.notification)) {
+			this.notificationSystem.addNotification(nextProps.notification);
+			this.props.clearNotifications();
+		}
+	}
+
+	public shouldComponentUpdate() {
```
johndiiorio: This function has been moved into RouteComponent (where it should go). Thus, it is safe to delete this function.

src/client/app/components/MultiSelectComponent.tsx
```diff
@@ -2,39 +2,45 @@
...
+}
+
+export default class MultiSelectComponent<I> extends React.Component<MultiSelectProps<I>, MultiSelectState> {
+	constructor(props: MultiSelectProps<I>) {
 		super(props);
 		this.onValuesChangeInternal = this.onValuesChangeInternal.bind(this);
 		// selectedOptions holds a list of the options that have been selected
 		this.state = {
-			selectedOptions: this.props.selectedOptions ? this.props.selectedOptions : null
+			selectedOptions: this.props.selectedOptions ? this.props.selectedOptions : undefined
```
johndiiorio, line 30: This should definitely be null.

src/client/app/components/RouteComponent.tsx
```diff
@@ -73,14 +78,15 @@ export default class RouteComponent extends React.Component {
 							options.groupIDs = info.split(',').map(s => parseInt(s));
 							break;
 						case 'chartType':
-							options.chartType = info;
+							options.chartType = info as ChartTypes;
+							options.chartType = info as ChartTypes;
 							break;
 						case 'barDuration':
-							options.barDuration = moment.duration(parseInt(info), 'days');
+							const maybeDuration = moment.duration(parseInt(info));
```
johndiiorio: I think there was a mistake here. This should definitely be assigned to `options.barDuration` as maybeDuration is never used. Should the `days` parameters still be removed?

src/client/app/components/RouteComponent.tsx
```diff
@@ -73,14 +78,15 @@ export default class RouteComponent extends React.Component {
...
 							break;
 						case 'barStacking':
 							if (this.props.barStacking.toString() !== info) {
-								options.changeBarStacking = true;
+								options.toggleBarStacking = true;
```
johndiiorio, line 88: GitHub doesn't let me select the line 4 down from here, but it is another case of the client-side erroring. Ideally this could be fixed, but not critical.

src/client/app/components/RouteComponent.tsx
```diff
@@ -73,14 +78,15 @@ export default class RouteComponent extends React.Component {
 							options.groupIDs = info.split(',').map(s => parseInt(s));
 							break;
 						case 'chartType':
-							options.chartType = info;
+							options.chartType = info as ChartTypes;
+							options.chartType = info as ChartTypes;
```
johndiiorio: Duplicated lines here

src/client/app/components/UIOptionsComponent.tsx
```diff
@@ -214,4 +178,71 @@ export default class UIOptionsComponent extends React.Component {
...
+	 * @param {Object[]} selection An array of {label: string, value: {type: string, id: int}} representing the current selection
+	 */
+	// private handleMeterSelect(selection: Array<{label: string, value: {type: string, id: number}}>) {
+	// 	this.props.selectMeters(selection.map(s => s.value));
+	// }
+
+	/**
+	 * Called when the user releases the slider, dispatch action on temporary state variable
+	 */
+	// TODO TYPESCRIPT: react-rangeslider doesn't have types. We should investigate what type this actually is
```
johndiiorio: Recommend switching out to use `rc-slider`, which is strictly better and should have good typings. The change should also be very fast.

src/client/app/components/UIOptionsComponent.tsx
```diff
@@ -214,4 +178,71 @@ export default class UIOptionsComponent extends React.Component {
...
+
+
+	/**
+	 * Toggles the bar stacking option
+	 */
+	private handleChangeBarStacking() {
+		this.props.changeBarStacking();
+	}
+
+	// TODO TYPESCRIPT this is an issue with typings for React.FormEvent<> and ChangeEvent<>
```
johndiiorio, line 226: This "might" (not sure) be fixed in React 16+. Recommend investigating and possibly upgrading to React 16.

src/common/TimeInterval.js
```diff
@@ -16,6 +16,9 @@ class TimeInterval {
 	toString() {
 		if (this.isBounded) {
 			// Using '_' as a separator character since it doesn't appear in ISO dates
+			if (this.startTimestamp === undefined || this.endTimestamp === undefined) {
+				throw Error('startTimestamp or endTimestamp was undefined in a bounded TimeInterval');
```
johndiiorio, line 20: Recommend using console.error here.

src/common/TimeInterval.js
```diff
@@ -16,6 +16,9 @@ class TimeInterval {
 	toString() {
 		if (this.isBounded) {
 			// Using '_' as a separator character since it doesn't appear in ISO dates
+			if (this.startTimestamp === undefined || this.endTimestamp === undefined) {
```
johndiiorio, line 19: Does it even make sense to throw an error here? Shouldn't we be able to handle half-open time intervals.

src/common/TimeInterval.js
```diff
@@ -69,4 +72,4 @@ class TimeInterval {
 	}
 }
 
-module.exports = TimeInterval;
+module.exports = { TimeInterval };
```
johndiiorio, line 75: We should be able to default export the TimeInterval class, e.g. `module.exports = TimeInterval`.

johndiiorio (changes requested):
Initial pass. Nothing too critical here.

src/client/app/actions/groups.ts
simonbtomlinson: I highly disagree with you. We shouldn't support our application when it's in weird bad states. Failing fast is always better as it's easier to debug (and at some point we will need to debug something in production), and as the user is just going to have the application subtly break on them later when the application accesses `undefined.startTimestamp` or something. Catching and doing nothing doesn't fix the problem, it just hides it, and it does a bad job of it.

jameeters: And for what it's worth, unless this has been fundamentally changed for typescript, the application will not recover from an empty group in editing marked dirty, and that's the one case that will lead to this error condition. The app will stop working properly whether or not we throw an error.

src/client/app/actions/groups.ts
jameeters: Again, this state is not recoverable. The application will not behave correctly once in this state whether or not an error is thrown. An error is the best way to guarantee that things fail as cleanly as possible.

src/common/TimeInterval.js
jameeters: It absolutely makes sense to handle half-bounded intervals, but they should be considered unbounded. So a bounded time interval must have two endpoints, and if either of them is missing we have an error condition.
I don't think we need half-bounded intervals now (if we did, we'd have them) so this seems correct to me.

src/common/TimeInterval.js
jameeters: As long as that's in addition to throwing the error.

src/client/app/components/UIOptionsComponent.tsx
NoraCodes: I've been working on this and unfortunately rc-slider is very broken; I can't get it to slide. If you could do this I would appreciate it a lot.

src/client/app/components/DashboardComponent.tsx
NoraCodes: I need to check that, but I think it is fixed and I'll remove the comment.

src/common/TimeInterval.js
NoraCodes: You already asked me about this in person, and already I told you that it doesn't work properly. I can't figure out a way to get it to work like that, but if you can I'd love to get this back to a default export.

src/client/app/components/InitializationComponent.tsx
NoraCodes: I experienced an issue with not having groups loaded and adding this line fixed that.

src/client/app/components/InitializationComponent.tsx
NoraCodes: Thanks! As with all things, feel free to make and commit that change; you can push to my fork, afaik

src/client/app/components/MultiSelectComponent.tsx
NoraCodes: Could you elaborate?

src/client/app/components/UIOptionsComponent.tsx
johndiiorio: I will look into this.

src/client/app/components/MultiSelectComponent.tsx
johndiiorio: There was a bug in MultiSelectComponent that @sandeepacharya464 and I had when `selectedOptions: undefined`, which was fixed by setting `selectedOptions: null`. I believe it had to do with the underlying Select component's PropTypes.

USAGE.md
```diff
@@ -57,7 +57,11 @@ OED_LOG_FILE=?                 // Path to the log file, defaults to ./log.txt
 1. Run ```systemctl enable oed.service``` to make the service start on server boot.
 1. Bring the app online with ```systemctl start oed.service```. Stop the app with ```systemctl stop oed.service```.
 
-### Administration ###
```
johndiiorio, line 62: You changed this file to USAGE.md but forgot to update the link in the README

src/client/app/components/MultiSelectComponent.tsx
NoraCodes: Thank you.

.travis.yml
NoraCodes: We couldn't while there were still unresolvable type errors; thanks to your PR we now can.

src/client/app/components/MultiSelectComponent.tsx
NoraCodes: Actually, after looking into this, that's not consistent with the type signature of `Select` (from `react-select`), and it doesn't cause a problem, so I don't see the need to change it.

src/client/app/components/RouteComponent.tsx
NoraCodes: Fixed; good catch, thank you.

src/client/app/components/UIOptionsComponent.tsx
NoraCodes: Agreed, I would like to do that.

src/common/TimeInterval.js
NoraCodes: Can I ask why? What does it get us?

USAGE.md
NoraCodes: Fixed

src/common/TimeInterval.js
jameeters: Are you asking me why I want the error or John why he wants the console log?

.travis.yml
johndiiorio: If you agree, I'd like to rename `checkTypescript` to something more descriptive. It sounds too similar to `typeCheck`.

src/common/TimeInterval.js
johndiiorio: Makes sense, I agree.

src/common/TimeInterval.js
NoraCodes: Why John wants the console error. Thrown errors get logged if unhandled.

johndiiorio:
Sorry about the merge conflicts. However, with the recent merge (which should help migrations), we should add documentation about the new `editUser` npm script. It doesn't need to be in this PR.

NoraCodes:
We're mergable!

Tested working:                                 
- Mobile layout                             
- Line chart & bar chart                    
- Creating Groups                           
- Editing groups                            
- Editing preferences                       
- Compare (all intervals)                   
- Import meter data                         
- CreateUser, UpdateMamacMeter, etc scripts 
- Chart Hotlinking

johndiiorio (approved):
Great job!

PR #304, opened 2018-02-23
Use Yarn instead of npm

johndiiorio:
Use yarn as the primary dependency manager instead of npm.

Pros:
- Faster than npm in most cases (including for OED)
- Better caching support
- Support for commands such as `yarn licenses` and `yarn why`
- Nicer logging
- Offline mode
- Workspaces
- Smaller lock-file
- Commands nearly identical to npm, but can be abbreviated (e.g. `npm install` -> `yarn` and `npm run dev` -> `yarn dev`).


jameeters (changes requested):
There will at the very least be a thorough discussion of this.

sandeepacharya464:
yarn and npm commands look very similar to me after viewing this pull request. So if there are obvious benefits of using  yarn, I would prefer yarn over npm.  But, really anything is fine for me.

PR #305, opened 2018-02-23
Add spinner component, show spinner while fetching chart readings

### LINKED ISSUE #296: "Work In Progress" Indicator
Given that fetching meter points can take some time, and that we already have actions for starting and finishing these downloads of data, it would be very easy to add a "working" or "WIP" indicator - perhaps a spinner or just a "downloading" label - that appears either over the chart or perhaps in the header.

johndiiorio:
Closes #296


jameeters:
~~Why not just load a small gif and avoid the dependency?~~
Nevermind, I misread `pagage-lock` as `package`.

NoraCodes (approved):
I like it. Let's merge this.

PR #306, opened 2018-02-26
Migrate Database

buip:
- Find the shortest path to migrate the database from the current version to new version
- `npm run migratedb` command can either update to the newest version or ask for the version user want to update to. It also logs a list of possible migrations.
- I will delete the list of folders in `src/server/migrations`. I just put it there for demonstrating purpose


src/server/migrations/0.1.0-0.3.0/migrate.js
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = {
+	fromVersion: '0.1.0',
+	toVersion: '0.3.0',
+	up: async db => {
+		// migration here
+		console.log('called 3');
```
johndiiorio: Remove console.log

src/server/migrations/0.5.0-0.6.0/migrate.js
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
johndiiorio: Recommend removing all of the migration dummy scripts and replacing them with one template script.

Then we will write migrations as needed.

src/server/migrations/registerMigration.js
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * register migration here. It allows for down migration however NOT RECOMMENDED
+ */
+const migrations = [
+	/* eslint-disable global-require */
+	require('./0.1.0-0.2.0/migrate'),
```
johndiiorio: See previous comment about removing dummy migrations.

src/server/models/Migration.js
```diff
@@ -0,0 +1,65 @@
...
+		this.fromVersion = fromVersion;
+		this.toVersion = toVersion;
+		this.updateTime = updateTime;
+	}
+
+	/**
+	 * Returns a promise to create the migration table.
+	 * @return {Promise.<>}
+	 */
+	static createTables() {
```
johndiiorio: Recommend calling it `createTable` as it only creates one table.

src/server/services/migrateDB.js
```diff
@@ -0,0 +1,68 @@
...
+
+function askToVersion() {
+	return new Promise(resolve => {
+		rl.question('To Version: ', toVersion => {
+			resolve(toVersion);
+		});
+	});
+}
+
+function terminateReadline(message) {
```
johndiiorio: This file, especially this method, is very similar to the createUser script. I have an open PR to refactor out some of the methods into multiple files. I recommend doing something similar (e.g. waiting for that PR to be merged and then refactoring, etc.)

src/server/sql/migration/get_all_migrations.sql
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- get the current as max of to_version column
```
johndiiorio: This comment does not make sense in respect to `SELECT * FROM migrations`. Can you elaborate?

johndiiorio (changes requested):
First pass, will look more in-depth later. Mostly minor changes requsted.

buip:
I abstracted `terminateReadline` and some `ask` methods into serviceUtils.js

src/server/migrations/migrateDatabase.js
```diff
@@ -0,0 +1,191 @@
...
+const { log } = require('../log');
+
+// file needed to run database transaction
+const requiredFile = [];
+
+/**
+ * @param migrationItems extracted from registerMigration.js
+ * @returns {Array} of migration pair
+ */
+function printMigrationList(migrationItems) {
```
simonbtomlinson: `print` doesn't really make sense as a name for this method, since it doesn't actually print anything. You could also change the body to `return migrationItems.map(item => `${item.fromVersion} -> ${item.toVersion});`

src/server/migrations/migrateDatabase.js
```diff
@@ -0,0 +1,191 @@
...
+	}
+	return list;
+}
+
+/**
+ * Create an array of unique versions;
+ * @param migrationItems extracted from registerMigration.js
+ * @returns {*[]} array of unique versions
+ */
+function getUniqueKeyOfMigrationList(migrationItems) {
```
simonbtomlinson: "unique key" isn't a great name for this either, since you aren't treating it like a key. Maybe something closer to `uniqueVersions`? But I'm not sold on that name either.

src/server/migrations/migrateDatabase.js
```diff
@@ -0,0 +1,191 @@
...
+	return migrationList;
+}
+
+/**
+ * If current version or version user wants to migrate is not in the migrationList, throw an Error
+ * @param curr current version of the database
+ * @param to version want to migrate to
+ * @param adjListArray adjacency list of version graph
+ */
+function checkIfFromAndToExist(curr, to, adjListArray) {
```
simonbtomlinson: I like using names more like `ensure` and `assert` for functions that throw an error if a condition isn't validated. For me, `check` implies that it's going to return a boolean to signal failure. You don't need to change this if you don't agree though.

src/server/migrations/migrateDatabase.js
```diff
@@ -0,0 +1,191 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const db = require('../models/database').db;
+const Migration = require('../models/Migration');
+const { compareSemanticVersion } = require('../util');
+const { log } = require('../log');
+
+// file needed to run database transaction
+const requiredFile = [];
```
simonbtomlinson: I don't like the idea of this being global state. It means we can only migrate once per run of the node server, and that's not good for testing.

src/server/migrations/migrateDatabase.js
```diff
@@ -0,0 +1,191 @@
...
+ * Open a database transaction and migrate the database by calling up() method.
+ * Insert row into migration folder
+ * @param neededFiles name of files needed to migrate
+ * @param allMigrationFiles is all of possible migration files
+ */
+async function migrateDatabaseTransaction(neededFiles, allMigrationFiles) {
+	try {
+		await db.tx(async t => {
+			neededFiles.forEach(neededFile => {
+				allMigrationFiles.forEach(async migrationFile => {
```
simonbtomlinson: This is firing off a bunch of promises without synchronizing on them in a meaningful way. You should run this in a normal loop instead and await on each part.

src/server/migrations/migrateDatabase.js
```diff
@@ -0,0 +1,191 @@
...
+					if (neededFile.fromVersion === migrationFile.fromVersion && neededFile.toVersion === migrationFile.toVersion) {
+						try {
+							await migrationFile.up(t);
+							const migration = new Migration(undefined, migrationFile.fromVersion, migrationFile.toVersion);
+							await migration.insert(t);
+						} catch (err) {
+							log.error('Error while migrating database', err);
```
simonbtomlinson: You should throw an error here. You want to abort the transaction if something goes wrong.

src/server/migrations/0.2.0-0.3.0/index.js
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = {
```
simonbtomlinson: You should replace this with a template file if you want to include anything. This file looks like it has a real, working migration at first glance.

src/server/migrations/registerMigration.js
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Register migration here
+ */
+const migrations = [
+	/* eslint-disable global-require */
+	require('./0.2.0-0.3.0')
```
simonbtomlinson: This should get changed along with the template thing I talk about further up.

simonbtomlinson (changes requested):
Partial review. I'll finish up with Migration.js and the other files I haven't commented on tomorrow.

src/server/migrations/0.2.0-0.3.0/index.js
buip: John and I were thinking of writing an actual migration on the next PR and then people will look at that as a sample instead of writing a template file. But I will change it for now until we write a migration

src/server/migrations/migrateDatabase.js
buip: Could you please help me with this? I don't understand this part

buip: I was trying to do this but it does not work
```
neededFiles.forEach(neededFile => {
			allMigrationFiles.forEach(migrationFile => {
				if (neededFile.fromVersion === migrationFile.fromVersion && neededFile.toVersion === migrationFile.toVersion) {
					db.tx(async t => {
						try {
							await migrationFile.up(t);
							const migration = new Migration(undefined, migrationFile.fromVersion, migrationFile.toVersion);
							await migration.insert(t);
						} catch (err) {
							throw new Error('Error while migrating database');
						}
					});
				}
			});
		});
```

src/server/services/servicesUtils.js
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
johndiiorio: I would just call this file `utils.js` to be consistent with other naming patterns.

buip: fixed

src/server/migrations/migrateDatabase.js
simonbtomlinson: That's because the fact that your `.forEach` calls hide the asyncness inside a function. Replace them with `for ... of ...` and it'll work.

src/server/services/migrateDB.js
```diff
@@ -0,0 +1,45 @@
...
+const migrationList = require('../migrations/registerMigration');
+
+function findMaxVersion(list) {
+	return findMaxSemanticVersion(getUniqueVersions(list));
+}
+
+(async () => {
+	let toVersion;
+	try {
+		const updateMax = await ask('Do you want to update to the max version? [yes/no]: ');
```
johndiiorio: My only wish is that this migratedb script could take in a version parameter argument (see createUser) that accepts a semantic version number or "latest" and does not go interactive. Thus, it could be run in the new `src/scripts/update.sh` script like `npm run migratedb --version latest`.

johndiiorio:
Outside my one comment above, this pull request looks good to go.

johndiiorio (approved):

simonbtomlinson (approved):

PR #308, opened 2018-03-02
Four weeks label on chart

### LINKED ISSUE #302: compare chart labels for 4 weeks
We changed OED so the button became 4 weeks instead of month. However, the actual charts still say month in the x-axis label and the title area (to indicate if more or less usage compared to previous time interval). While it is a little more clunky, I think we should probably change those too.

johndiiorio:
Closes #302.

To be merged after TypeScript.


johndiiorio:
Superseded by PR #314

PR #311, opened 2018-03-04
Collect all frontend api requests in one place

simonbtomlinson:
Pulls all api requests into a single `Api.ts` file that takes care of things like authentication.
I'd be happy to receive feedback about how `Api.ts` can be split up in a way that's easier to read while still easy to use.

This required modifying some server-side validation code in places that expected a token parameter. The token can be passed in a number of ways (see `src/server/routes/authenticator.js` for the relevant code), but lots of validation code assumed it would be passed in a specific way.


simonbtomlinson:
I've separated the parts of the `Api` class in commit 5b5b891.

src/client/app/actions/admin.ts
```diff
@@ -2,15 +2,15 @@
...
-import axios from 'axios';
 import { changeBarStacking, changeChartToRender } from './graph';
 import { showErrorNotification, showSuccessNotification } from '../utils/notifications';
-import { getToken } from '../utils/token';
 import { ChartTypes } from '../types/redux/graph';
 import { PreferenceRequestItem } from '../types/items';
 import * as t from '../types/redux/admin';
 import { ActionType, Dispatch, GetState, Thunk } from '../types/redux/actions';
 import { State } from '../types/redux/state';
+import {preferencesApi} from '../utils/api';
```
johndiiorio: We aren't using this spacing style in the project. Perhaps we need to add a tslint rule?

src/client/app/utils/api/ApiBackend.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import axios, {AxiosRequestConfig, AxiosResponse} from 'axios';
+import {getToken, hasToken} from '../token';
+
+/**
+ * Handles the actual sending and receiving of network requests.
+ *
+ * Should probably only be used by other api classes.
+ */
+class ApiBackend {
```
johndiiorio: I don't like the name ApiBackend, as it signifies something server-side. How about `ApiInternal.ts`, `Api.ts`, or `ApiDispatcher.ts`?

src/client/app/utils/api/GroupsApi.ts
```diff
@@ -0,0 +1,60 @@
...
+	public async create(groupData: GroupData): Promise<void> {
+		return await this.backend.doPostRequest<void>('api/groups/create', groupData);
+	}
+
+	public async edit(group: GroupData & GroupID): Promise<void> {
+		return await this.backend.doPutRequest<void>('api/groups/edit', group);
+	}
+
+	public async delete(groupID: number) {
+		return await this.backend.doPostRequest('api/groups/delete', {id: groupID});
```
johndiiorio, line 56: Why is this a POST instead of DELETE request? Recommend changing the server to correspond to be a DELETE request.

src/client/app/components/AdminComponent.tsx
```diff
@@ -184,16 +183,7 @@ export default class AdminComponent extends React.Component<AdminProps, {}> {
...
-			data.append('csvFile', file);
-			axios({
-				method: 'post',
-				url: `/api/fileProcessing/${this.props.selectedImportMeter.value}`,
-				data,
-				params: {
-					token: getToken()
-				}
-			})
+			metersApi.submitNewReadings(this.props.selectedImportMeter.value, file)
```
johndiiorio, line 186: Is there a reason directly below this line that you used a `.then` instead of `async`/`await`? Or would that get ugly fast because `await` needs to be in an async function?

src/client/app/components/LoginComponent.tsx
```diff
@@ -94,27 +94,24 @@ export default class LoginComponent extends React.Component<{}, LoginState> {
...
-			browserHistory.push('/');
-		})
-		.catch(err => {
-			if (err.response.status === 401) {
-				showErrorNotification('Invalid email/password combination');
-			} else {
-				// If there was a problem other than a lack of authorization, the user can't fix it.
-				// Log it to the console for developer use.
-				console.error(err); // tslint:disable-line no-console
+		(async () => {
```
johndiiorio, line 97: Would it be possible to mark `handleSubmit` as async instead of using an async IFFE? Or does that not work with React?

src/client/app/components/RouteComponent.tsx
```diff
@@ -46,16 +46,19 @@ export default class RouteComponent extends React.Component<RouteProps, {}> {
...
+		// Needs to be async because of the network request
+		(async () => {
+			try {
+				if (!(await verificationApi.checkTokenValid())) {
+					// Route to login page if the auth token is not valid
+					browserHistory.push('/login');
+				}
+			} catch (e) {
+				// In the case of a server error, the user can't fix the issue. Log it for developers.
+				console.error(e); // tslint:disable-line no-console
```
johndiiorio: We have been using `throw new Error()` instead of logging directly to the console (as of the TypeScript PR).

src/server/routes/groups.js
```diff
@@ -110,9 +110,8 @@ router.post('/create', async (req, res) => {
 	const validGroup = {
 		type: 'object',
 		maxProperties: 4,
```
johndiiorio: Don't forget to change maxProperties here and on all the other validation routes you changed.

src/client/app/components/LoginComponent.tsx
```diff
@@ -94,27 +94,24 @@ export default class LoginComponent extends React.Component<{}, LoginState> {
...
+				const token = await verificationApi.login(this.state.email, this.state.password);
+				localStorage.setItem('token', token);
+				browserHistory.push('/');
+			} catch (err) {
+				if (err.response.status === 401) {
+					showErrorNotification('Invalid email/password combination');
+				} else {
+					// If there was a problem other than a lack of authorization, the user can't fix it.
+					// Log it to the console for developer use.
+					console.error(err); // tslint:disable-line no-console
```
johndiiorio: Another use of console.eror -> throw new Error

src/client/app/utils/api/ApiBackend.ts
```diff
@@ -0,0 +1,70 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import axios, {AxiosRequestConfig, AxiosResponse} from 'axios';
+import {getToken, hasToken} from '../token';
+
+/**
+ * Handles the actual sending and receiving of network requests.
+ *
+ * Should probably only be used by other api classes.
```
johndiiorio: Recommend removing "probably", we definitely don't want it used outside other API classes.

src/client/app/utils/api/ApiBackend.ts
```diff
@@ -0,0 +1,70 @@
...
+		const response: AxiosResponse = await axios.post(url, body, this.buildConfig(params, headers, extraConfig));
+		return response.data as R;
+	}
+
+	private buildConfig(
+		params: { [key: string]: string},
+		headers: { [key: string]: string},
+		extraConfig: AxiosRequestConfig
+	): AxiosRequestConfig {
+		if (hasToken()) {
```
johndiiorio, line 53: Doesn't this always send a token regardless of the route? Is this best practice?

src/client/app/actions/admin.ts
```diff
@@ -46,45 +46,37 @@ function markPreferencesSubmitted(): t.MarkPreferencesSubmittedAction {
...
-				dispatch(receivePreferences(response.data));
-				if (!getState().graph.hotlinked) {
-					dispatch((dispatch2: Dispatch, getState2: GetState) => {
-						const state = getState();
-						dispatch2(changeChartToRender(state.admin.defaultChartToRender));
-						if (response.data.defaultBarStacking !== state.graph.barStacking) {
-							dispatch2(changeBarStacking());
-						}
-					});
+		const preferences = await preferencesApi.getPreferences();
```
johndiiorio, line 51: I know the old code did not have a catch condition, but should we be explicit and add a catch to every request?

johndiiorio (changes requested):

src/client/app/utils/api/GroupsApi.ts
simonbtomlinson: I didn't actually change the way we do any networking, as I didn't think it was in-scope for this pr. We can definitely change that in a different pr though.

src/client/app/actions/admin.ts
simonbtomlinson: I don't think there's a tslint rule for this. I do like `{ foo }` style better though, and a rule would be nice. See here: https://github.com/palantir/tslint/issues/1044. I'll fix it manually.

src/client/app/utils/api/ApiBackend.ts
simonbtomlinson: It does. I think this is a good way to do it. It makes things available to the server more often and there's less of a downside. We can just make the server send a 401 unauthorized if the user doesn't have the appropriate authorization level.

src/client/app/actions/admin.ts
simonbtomlinson: Maybe? I feel like we want to throw an error here, at least until we have more robust error handling (which isn't in scope for this pr, I don't think).

src/client/app/components/LoginComponent.tsx
simonbtomlinson: I can't call setState() inside an async function, so this has to stay an IFFE.

src/client/app/components/AdminComponent.tsx
simonbtomlinson: It's cleaner than wrapping it in an async IIFE in this case, and it was already like that, so I decided not to change it.

src/client/app/utils/api/GroupsApi.ts
NoraCodes: Yeah, I'm on board for changing this in a future PR.

buip (approved):
I don't know about the client side but the server side looks good to me!

src/server/routes/groups.js
```diff
@@ -157,9 +156,8 @@ router.put('/edit', async (req, res) => {
 	const validGroup = {
 		type: 'object',
 		maxProperties: 5,
```
buip: I think maxProperties should be 4 here

src/server/routes/groups.js
```diff
@@ -227,9 +225,8 @@ router.post('/delete', async (req, res) => {
 	const validParams = {
 		type: 'object',
 		maxProperties: 2,
```
buip: I think maxProperties should be 1 here

buip (changes requested):
Sorry for the last comment! I did not look carefully

buip (approved):

johndiiorio (approved):

PR #315, opened 2018-03-07
Added descriptive error for duplicate name.

### LINKED ISSUE #135: Warn users when creating groups or meters with the same name as an existing group or meter
What it says on the tin. If they want to, they can, but they shouldn't.

jamielepito:
Closes #135 
Changed the server to send 409 conflict error when a duplicate group name is detected and show a specific message on the client.


NoraCodes (approved):

src/server/routes/groups.js
```diff
@@ -148,7 +148,11 @@ router.post('/create', async (req, res) => {
 			res.sendStatus(200);
 		} catch (err) {
 			log.error(`Error while inserting new group ${err}`, err);
-			res.sendStatus(500);
+			if (err.toString() === 'error: duplicate key value violates unique constraint "groups_name_key"') {
+				res.sendStatus(409); //409 conflict
```
johndiiorio: I would not recommend using a 409 status code. Example from Wikipedia about 409 status codes: "conflict in the request, such as an edit conflict between multiple simultaneous updates."

I also recommend only using a limited number of status codes: 200, 400, 401, 403, 404, 429, and 500. Thus, I recommend that this should be a 400 - bad request.

src/client/app/actions/groups.ts
```diff
@@ -236,7 +236,11 @@ function submitNewGroup(group: t.GroupData): Thunk {
 			})
 			.catch(error => {
 				dispatch(markGroupInEditingNotSubmitted());
-				showErrorNotification('Failed to create group');
+				if (error.response !== undefined && error.response.status === 409) {
+					showErrorNotification(`Name '${group.name}' was already taken.`);
```
johndiiorio: For clarity, I would change this to "Group 'group name' is already in use."

johndiiorio (changes requested):

johndiiorio:
I think that this PR should also extend to warning a user when editing the name of an existing group to a name that already exists. Also, we should not log to a file whenever this event (new group or editing existing group) occurs. We should only log if we throw a 500 status code.

src/server/routes/groups.js
NoraCodes: > I also recommend only using a limited number of status codes

Why?

johndiiorio: It is considered good practice by many companies. Google uses 9, Dropbox uses 10, Facebook uses 8 (I believe). A couple good articles: https://blogs.dropbox.com/developers/2015/04/how-many-http-status-codes-should-your-api-use/ and http://blog.restcase.com/rest-api-error-codes-101/

sandeepacharya464:
@jamielepito  the code looks good. if you resolve the conflict, we can merge this in. :)

johndiiorio (approved):

PR #316, opened 2018-03-09
Use Webpack 4

johndiiorio:
Upgrade Webpack to version 4, which requires the newly separated `webpack-cli` package. Updated uglifyjs-webpack-plugin to be compatible with Webpack 4. Disabled performance hints, e.g. a warning whenever a production bundle size is greater than 250 kb.


simonbtomlinson:
Babel-loader 7.1.2 seems to want a peer dependency of webpack 2 or 3. Have you looked in to that?
I also get a warning when I try to build: "(node:4035) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead". Do you know what that's about? Can we get rid of it somehow?

johndiiorio:
@simonbtomlinson I cannot reproduce the warning with babel-loader 7.1.2. For the deprecation warning, I believe it is coming from one of our Webpack plugins trying to remain backwards compatible with the old plugin API. Should I investigate further, or is the warning ok?

NoraCodes:
My commit fixes all deprecation warnings, along with a few other dependency issues.

@simonbtomlinson

sandeepacharya464 (approved):
No comments. Looks fine.

sandeepacharya464:
Just make sure you resolve conflicts before you merge in. @johndiiorio

PR #317, opened 2018-03-19
Fixed wrong properties in validation

buip:


johndiiorio (approved):

src/server/routes/readings.js
```diff
@@ -37,7 +37,7 @@ router.get('/line/meters/:meter_ids', async (req, res) => {
 		maxProperties: 1,
 		required: ['meter_ids'],
 		properties: {
-			group_id: {
+			meter_ids: {
 				type: 'number'
```
simonbtomlinson, line 41: Shouldn't this be a string of comma-separated integers, not a number?

src/server/routes/readings.js
```diff
@@ -81,7 +81,7 @@ router.get('/line/groups/:group_ids', async (req, res) => {
 		maxProperties: 1,
 		required: ['group_ids'],
 		properties: {
-			group_id: {
+			group_ids: {
 				type: 'number'
```
simonbtomlinson, line 85: Same thing about number vs string here

src/server/routes/readings.js
```diff
@@ -126,7 +126,7 @@ router.get('/bar/meters/:meter_ids', async (req, res) => {
 		maxProperties: 1,
 		required: ['meter_ids'],
 		properties: {
-			group_id: {
+			meter_ids: {
 				type: 'number'
```
simonbtomlinson, line 130: Same thing about number vs string here

src/server/routes/readings.js
```diff
@@ -175,7 +175,7 @@ router.get('/bar/groups/:group_ids', async (req, res) => {
 		maxProperties: 1,
 		required: ['group_ids'],
 		properties: {
-			group_id: {
+			group_ids: {
 				type: 'number'
```
simonbtomlinson, line 179: Same thing about number vs string here

src/server/routes/users.js
```diff
@@ -31,7 +31,7 @@ router.get('/:user_id', async (req, res) => {
 		maxProperties: 1,
 		required: ['user_id'],
 		properties: {
-			group_id: {
+			user_id: {
 				type: 'id'
```
simonbtomlinson, line 35: Is "id" a type? I think this should be a string representing a number.

src/server/routes/meters.js
```diff
@@ -40,7 +40,7 @@ router.get('/:meter_id', async (req, res) => {
 		maxProperties: 1,
 		required: ['meter_id'],
 		properties: {
-			group_id: {
+			meter_id: {
 				type: 'number'
```
simonbtomlinson, line 44: I think this should be a string representation of an integer, not an integer, since it's from the url.

src/server/routes/groups.js
```diff
@@ -86,7 +86,7 @@ router.get('/deep/meters/:group_id', async (req, res) => {
 		maxProperties: 1,
 		required: ['group_id'],
 		properties: {
-			meter_id: {
+			group_id: {
 				type: 'number'
```
simonbtomlinson, line 90: I think this should be a string representing an integer, not an integer.

simonbtomlinson (commented):
I think we should revert this pr. I'm pretty sure it breaks routes.

PR #320, opened 2018-03-23
Fix validation properties, tested working

buip:
I changed the type and tested each route individually, and they worked. It would be great if you can test it one more time to be sure. Thank you!


simonbtomlinson (commented):
I like the idea of validating that the correct string format got passed in using a regex. You can use `^\\d+$` for a single number and `^\\d+(?:,\\d+)*$` for 1 or more numbers separated by commas. Those already have the extra backslashes you need when you use them as a js string.

buip:
I ran into some weird problem so I will close this PR and open a new one.

PR #321, opened 2018-03-23
Add meter

sandeepacharya464:
1) Refactored admin component into addMeter,addReadings, and preferences component.
2) Separated addMamacMeter script from methods associated with inserting meters.
3) Created a UI to add Mamac meters in the admin section. The file format is assumed to have 'ip' header.
4) Add reading UI and Add meter UI synchronized.


src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -0,0 +1,69 @@
...
+		this.handleMeterToImport = this.handleMeterToImport.bind(this);
+	}
+
+	public handleMeterToImport(files: File[]) {
+		const file = files[0];
+		let meters = [];
+		const reader = new FileReader();
+		let listValue = [];
+		reader.onload = () => {
+			const fileAsBinaryString = reader.result;
```
simonbtomlinson: I feel like all of this could be much more readable if it was replaced with a for loop that populated a new list. And `listValue` isn't a great variable name. I wouldn't know what this code was doing unless I already knew the format we want csv files to be in.

Also, this breaks on csv files that quote everything, which is a valid choice for csvs.

src/client/app/containers/admin/AddMetersContainer.ts
```diff
@@ -0,0 +1,16 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { connect } from 'react-redux';
+import AddMetersComponent from '../../components/admin/AddMetersComponent';
+import { fetchMetersDetailsIfNeeded } from '../../actions/meters';
+import { Dispatch } from '../../types/redux/actions';
+
+function mapDispatchToProps(dispatch: Dispatch) {
+	return {
+		fetchMeterDetailsIfNeeded: (alwaysFetch?: boolean) => dispatch(fetchMetersDetailsIfNeeded(alwaysFetch))
```
simonbtomlinson: You always pass true here (because you're always forcing it) so it might be cleaner to return `{ fetchMeterDetails: () => dispatch(fetchMeterDetailsIfNeeded(true)) }`

src/server/models/Meter.js
```diff
@@ -91,6 +91,13 @@ class Meter {
 		this.id = resp.id;
 	}
 
+	static insertAll(meters, conn = db) {
```
simonbtomlinson: You should write a test for this. sequence is pretty finicky and I had some difficulty getting it working for reading inserts.

simonbtomlinson (changes requested):

simonbtomlinson:
@johndiiorio can you look at the preferences part of this when you get a chance? It's not my area.

src/server/models/Meter.js
sandeepacharya464: This method is not used anywhere. I will just remove this.

src/client/app/containers/admin/AddMetersContainer.ts
sandeepacharya464: Yes, I will do that. That makes sense.

src/client/app/components/admin/AddMetersComponent.tsx
sandeepacharya464: changes made. now it accepts file with quotes. Also, it is much more readable.

src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+					const ips = items.replace(/\"/g, '');
+					if (items.length !== 0 && items !== 'ip') {
+						listOfIps.push(ips);
+					}
+				}
+				meters = listOfIps;
+				fileProcessingApi.submitNewMeters(meters)
+					.then(() => {
+						showSuccessNotification('Successfully uploaded meters');
+						this.props.fetchMeterDetailsIfNeeded(true);
```
johndiiorio: Duplicate lines here

src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+			fontWeight: 'bold',
+			margin: 0,
+			paddingBottom: '5px'
+		};
+
+		return (
+			<div>
+				<p style={titleStyle}> Add new meters: </p>
+				<Dropzone accept='text/csv, application/vnd.ms-excel,' onDrop={this.handleMeterToImport}>
+					<div>Upload meters list (CSV):</div>
```
johndiiorio: I would change this to "Upload meter data (CSV)".

Otherwise, a user may be confused and just upload a list of meter ips instead of our specified CSV format.

src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -0,0 +1,73 @@
...
+			if (dataLines[0] !==  'ip') {
+				showErrorNotification('Incorrect file format');
+			} else {
+				for (const items of dataLines) {
+					const ips = items.replace(/\"/g, '');
+					if (items.length !== 0 && items !== 'ip') {
+						listOfIps.push(ips);
+					}
+				}
+				meters = listOfIps;
```
johndiiorio: I would just get rid of the `meters` variable and it's declaration and just pass `listOfIps` to the backend.

src/server/routes/fileProcessing.js
```diff
@@ -8,16 +8,18 @@ const moment = require('moment');
 const streamBuffers = require('stream-buffers');
 const multer = require('multer');
 const streamToDB = require('../services/loadFromCsvStream');
+const { insertMeters } = require('../services/readMamacMeters');
 const authenticator = require('./authenticator');
 const validate = require('jsonschema').validate;
+const { log } = require('../log');
```
johndiiorio: I don't think this import is needed.

src/server/routes/fileProcessing.js
```diff
@@ -29,7 +31,7 @@ router.post('/:meter_id', upload.single('csvFile'), async (req, res) => {
 			}
 		}
 	};
-	if (!validate(req.params, validParams).valid || !req.file.buffer) {
+	if (!validate(req.params, validParams).valid) {
```
johndiiorio, line 33: Why was `!req.file.buffer` removed?

johndiiorio (changes requested):
Initial pass. I have not tested it out myself.

src/server/routes/fileProcessing.js
sandeepacharya464: This sometimes creates problem if users try to upload wrong file format (not csv). Since buffer is undefined at that point. So, this was removed.

sandeepacharya464:
@johndiiorio  and @simonbtomlinson, all requested changes have been made.

src/client/app/components/admin/AdminComponent.tsx
```diff
@@ -0,0 +1,37 @@
...
+	return (
+		<div>
+			<HeaderContainer />
+			<div className='container-fluid'>
+				<div className='row'>
+					<div className='col-12 col-lg-6'>
+						<PreferencesContainer />
+					</div>
+					<div className='col-12 col-lg-6'>
+						<AddReadingsContainer />
```
johndiiorio, line 30: You should add more separation between the AddReadings and the AddMeter Container.

src/client/app/components/admin/AdminComponent.tsx
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as React from 'react';
+import HeaderContainer from '../../containers/HeaderContainer';
+import FooterComponent from '../FooterComponent';
+import AddMeterContainer from '../../containers/admin/AddMetersContainer';
```
johndiiorio: This should be: `import AddMetersContainer from '../../containers/admin/AddMetersContainer';`

src/server/routes/fileProcessing.js
```diff
@@ -61,4 +62,31 @@ router.post('/:meter_id', upload.single('csvFile'), async (req, res) => {
...
+		}
+	};
+	if (!validate(req.body, validBody).valid) {
+		res.sendStatus(400);
+	} else {
+		try {
+			await insertMeters(req.body.meters.map(ip => ({ip})));
+			res.status(200).json({success: true});
+		} catch (err) {
+			res.status(403).json({success: false, message: 'Failed to upload meters.'});
```
johndiiorio: There is no reason to send the custom message. We are currently only sending a custom message if the client depends on a custom message to show a custom error notification.

src/client/app/components/admin/AddReadingsComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+			paddingBottom: '5px'
+		};
+		const smallMarginBottomStyle: React.CSSProperties = {
+			marginBottom: '5px'
+		};
+
+		return (
+			<div>
+				<p style={titleStyle}>Import readings for a selected meter:</p>
+				<SingleSelectComponent
```
johndiiorio, line 53: For a more aesthetically pleasing look, I would shrink the width of both this select and the title input in PreferencesComponent.

src/client/app/components/admin/AddReadingsComponent.tsx
```diff
@@ -0,0 +1,68 @@
...
+					style={smallMarginBottomStyle}
+					options={this.props.meters}
+					selectedOption={this.props.selectedImportMeter}
+					placeholder='Select meter to import data'
+					onValueChange={s => this.props.updateSelectedImportMeter(s.value)}
+				/>
+				{this.props.selectedImportMeter &&
+				<Dropzone accept='text/csv, application/vnd.ms-excel,' onDrop={this.handleFileToImport}>
```
johndiiorio, line 61: We should remove excel here.

src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -0,0 +1,70 @@
...
+		const titleStyle: React.CSSProperties = {
+			fontWeight: 'bold',
+			margin: 0,
+			paddingBottom: '5px'
+		};
+
+		return (
+			<div>
+				<p style={titleStyle}> Add new meters: </p>
+				<Dropzone accept='text/csv, application/vnd.ms-excel,' onDrop={this.handleMeterToImport}>
```
johndiiorio: We should remove excel here.

src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -0,0 +1,70 @@
...
+					<div>Upload meters data (CSV):</div>
```
johndiiorio: I'm worried that admin users may not know the file format to upload meters. I believe this warrants some further discussion.

johndiiorio (changes requested):

src/client/app/components/admin/AddMetersComponent.tsx
sandeepacharya464: I think this does more than just accepting csv. It was added long time back but I think it was added to prevent certain types of incorrect file formats.

src/client/app/components/admin/AddMetersComponent.tsx
sandeepacharya464: the instruction about file format is provided now.

sandeepacharya464:
@johndiiorio  all necessary changes have been made.

johndiiorio (approved):

simonbtomlinson (approved):

PR #324, opened 2018-03-26
i18n - TypeScript

johndiiorio:
Adds French translations


buip (approved):
The server side looks good too me

sandeepacharya464 (approved):
Great work John!  To me, it looks great.

.babelrc
```diff
@@ -3,5 +3,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 {
 	"presets": ["env", "react"],
-	"plugins": ["transform-object-rest-spread"]
+	"plugins": [
+		"transform-object-rest-spread", 
+		[
+			"react-intl", {
+				"messagesDir": "./src/client/app/translations/messages",
```
simonbtomlinson: If we remove this we shouldn't need to worry about generated files at all.

.babelrc
```diff
@@ -3,5 +3,13 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 {
 	"presets": ["env", "react"],
-	"plugins": ["transform-object-rest-spread"]
+	"plugins": [
+		"transform-object-rest-spread",
```
simonbtomlinson: Do we still need this line post-typescript? @LeoTindall do you know?

.gitignore
```diff
@@ -31,3 +31,6 @@ postgres-data
 # Text files
 *.txt
 !License.txt
+
+# Generated translations
```
simonbtomlinson: We shouldn't need this if we remove the line in the baberc file.

src/client/app/utils/translate.ts
```diff
@@ -0,0 +1,24 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import store from '../index';
+import { addLocaleData, IntlProvider, defineMessages } from 'react-intl';
+import * as en from 'react-intl/locale-data/en';
+import * as fr from 'react-intl/locale-data/fr';
+import * as localeData from '../translations/locales/data.json';
+
+export default function translate(messageID: string): string {
```
simonbtomlinson: This feels very heavyweight for something that runs each time a string gets translated. Can you cache most of this on the first run?

simonbtomlinson: You also might want to export it with a shorter name like `t` so it doesn't add a ton of line length when used. Django does something similar with its translations.

simonbtomlinson (changes requested):

src/client/app/utils/translate.ts
johndiiorio: `t` is frequently used throughout Redux to denote types

simonbtomlinson (approved):

PR #328, opened 2018-04-03
Refactor init.sh into install and upgrade scripts

johndiiorio:
This pull request does a number of things:

1. As the init.sh script was used for two different purposes, this PR breaks it up into install and update scripts.
2. The init.sh script had a number of flags. The new install script take only an optional flag (--production) and defaults to a development installation. The upgrade script does not take any flags. This change should simplify the installation process. The upgrade script will eventually incorporate the database migration script (when that PR is merged).
3. This renames the npm `dev` and `build` scripts to `webpack:dev` and `webpack:build` to specify exactly what is happening. This also adds a `npm run webpack` command for one off uses (such as in a development installation).
4. The install script does not accept meter CSV files, as users may want to upload many different CSV files. Developers can both use the `npm run add[meterType]Meters` and the web interface to upload CSVs.


src/scripts/updateOED.sh
```diff
@@ -0,0 +1,43 @@
...
+tries=0
+max_tries=10
+
+# Try to create the schema until it succeeds
+while [ $create_error == 0 ]; do
+    # Sleep to let PostgreSQL chill out
+    sleep 1
+    echo "Attempting to create database."
+    # Redirect stderr to a file
+    npm run createdb |& tee /tmp/oed.error > /dev/null
```
buip, line 26: I remembered when I tried to run createdb, it keeps saying that meter_type already exists. I don't know how to override that

NoraCodes: This is alright with that; it should just emit "Schema created or already exists" and move on.

webpack.config.js
```diff
@@ -9,7 +9,6 @@ const path = require('path');
 
 const BUILD_DIR = path.resolve(__dirname, 'src/client/public');
 const APP_DIR = path.resolve(__dirname, 'src/client/app');
-const COMMON_DIR = path.resolve(__dirname, 'src/common');
```
NoraCodes, line 12: Why? Don't we need stuff in src/common?

johndiiorio: We don't use COMMON_DIR anywhere in our webpack config (we don't bundle the stuff in that directory).

simonbtomlinson: We bundle in the `TimeInterval` class.

johndiiorio: We don't explicity point the webpack config at the src/common directory. We use it on the front-end, so webpack auto-bundles it.

NoraCodes: Makes sense!

NoraCodes (approved):

PR #329, opened 2018-04-05
Add robust meter diagnosis tools

NoraCodes:
These commits fix the updateMamacMeters error messages and add a script to check if meters are up, based on a CSV file of IPs (like ips.csv that we use for addMamacMeters).

### DIFF
--- a/src/scripts/checkMeters.sh
+++ b/src/scripts/checkMeters.sh
@@ -13,11 +13,11 @@ read line
 while read line
 do
 	echo "$line:"
-	curl -s -m 1 "http://$line" > /dev/null
+	curl -s -m 5 "http://$line" > /dev/null
 	success=$?
 	if [ "$success" == "0" ]; then
-		echo "	GOOD"
+		echo "	Request succeeds."
 	else
-		echo "	NO GOOD $success"
+		echo "	Request fails with code $success."
 	fi
 done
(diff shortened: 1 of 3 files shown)


src/scripts/checkMeters.sh
```diff
@@ -0,0 +1,23 @@
...
+
+while read line
+do
+	echo "$line:"
+	curl -s -m 1 "http://$line" > /dev/null
+	success=$?
+	if [ "$success" == "0" ]; then
+		echo "	GOOD"
+	else
+		echo "	NO GOOD $success"
```
johndiiorio: I think your message could be improved to `Meter failing with status $success`

src/scripts/checkMeters.sh
```diff
@@ -0,0 +1,23 @@
...
+
+USAGE="Usage: checkMeters.sh Takes a CSV file of meter IP addresses on standard in. Use as cat ips.csv | checkMeters.sh or checkMeters.sh < ips.csv"
+
+# Throw out header
+read line
+
+while read line
+do
+	echo "$line:"
+	curl -s -m 1 "http://$line" > /dev/null
```
johndiiorio: It appears that the max timeout is 1 second. I think this should be increased to 5-10 seconds.

johndiiorio (changes requested):
Your checkMeters script is not integrated into the rest of the system. Ideally, it would email admins that meters are down. However, I think having a basic implementation is good for this pull request.

jameeters (approved):
All changes look good to me, although I'm unsure about adding the checkMeters script to the main repository. It seems like we should have a better place for ancillary tools like this, although I don't know what that might be.

johndiiorio (approved):

PR #331, opened 2018-04-19
Toggle UI options

### LINKED ISSUE #244: Add button to hide chart options

johndiiorio:
Closes #244 and replaces PR #295.
Notes:
- Does not store the toggle option in the database
- Bumps React to version 16 (to fix `createPortal() is not a function`)


sandeepacharya464 (approved):
Looks good to me. Just a design thought. Do we really want people to go to menu just to show options when it is hidden. Or we have a small 'show button' link somewhere in the screen. That sounds more convenient to me. We can have a discussion about this in future. For now, this looks good to me. @johndiiorio  please talk to me if what I am saying is not clear.

PR #332, opened 2018-04-19
Groups view

sandeepacharya464:
- Make groups viewable to non-admin users.
- Create new routes for edit,create, and view groups


src/client/app/components/HeaderButtonsComponent.tsx
```diff
@@ -38,16 +38,13 @@ export default class HeaderButtonsComponent extends React.Component<HeaderButton
...
 				} else {
 					renderLoginButton = true;
 				}
 				break;
 			case 'groups':
 				renderAdminButton = true;
-				break;
-			case 'admin':
-				renderGroupsButton = true;
+				renderGroupsButton = false;
```
johndiiorio, line 46: Why do you hide the groups button on the admin panel?

sandeepacharya464: @johndiiorio the group button is still displayed on the admin panel. The case 'admin' was removed entirely. Because now instead of displaying in certain cases, logic is more about  not displaying in certain cases.  'renderGroupbutton' is set true initially now.

src/client/app/containers/groups/EditGroupsContainer.ts
```diff
@@ -8,25 +8,26 @@ import EditGroupsComponent from '../../components/groups/EditGroupsComponent';
...
- * @return {{meters: *, groups: *}}
- */
 function mapStateToProps(state: State) {
 	const allMeters = _.sortBy(_.values(state.meters.byMeterID).map(meter => ({ id: meter.id, name: meter.name.trim() })), 'name');
 	let allGroups = _.sortBy(_.values(state.groups.byGroupID).map(group => ({ id: group.id, name: group.name.trim() })), 'name');
 
 	const groupInEditing = state.groups.groupInEditing as GroupDefinition;
-	if (groupInEditing === undefined) {
-		throw new Error('Unacceptable condition: state.groups.groupInEditing has no data.');
+	if (groupInEditing === undefined || _.isEmpty(_.difference(Object.keys(groupInEditing), ['dirty']))) {
```
johndiiorio: I would add a comment here explaining what is going on

johndiiorio (approved):

Issue #312, opened 2018-03-05
bar values wrong


huss:
If you go to the demo site and select Chapin 2nd, 3rd & student then the sum of 2nd and 3rd should be student. This is what happens in the line graphic. However, this is not the case for bar representations (bar or compare). If you look at bar (say for one week) the sum is off and sometimes it is easy to see since student is less than 2nd and 3rd. If you go to the compare in a 4 week view it is somewhat off and in the week view it is way off. I tried to include the images but it will not work so I will send them around.

huss:
As noted, this did not fix the occasional bar values that are wrong. Anyone working on this should contact me since I have some examples of the problem. Since two different methods for getting the data give issues, the next step might be to see about the graphics package. Since we want to convert to Plotly anyway, maybe we should try that and see if the issue remains. If it is gone then we are done. If it remains then we need to carefully look at the data being used. Anything found should be added to the test suite to catch this type of problem. This is not a target for the 0.5 release but is a priority once that is out the door. I'm going to ask @CorinaSterbet to take the lead on this where helping someone else with it would be fine.

NoraCodes:
I think this is the same as #312 perhaps?

CorinaSterbet:
I looked into the issue and checked the sums of the meter data. For example, one issue stated that, for Maurer 1st, the sum of the day values was different from the week value (for the period Jan 3 - Jan 10, 2019). However, such a time interval does not exist at this moment for the Plotly bar graph. 

The closest time interval that the bar displays is the period Jan 1 - Jan 8. When I check the week interval section, the value for that week is 122 kW. In the day interval section, the sum of the values for each day of the period adds up to 122 kW. Therefore, I could not detect the issue.

huss:
I also checked that groups were the proper sum of their underlying meters. I did not see an issue either. Thus, I'm assuming this is fixed until someone finds otherwise.

PR #335, opened 2018-05-08
Fast pt compression

simonbtomlinson:
Holy crap, it's finally here!

I'm not quite done with this, but I figured people should start looking at it.

I have two things left to do:
1. Set up a cron job for running `npm run refreshReadingViews`. @LeoTindall, I'd appreciate it if you could help me with this; I'm not sure which bash script to edit.

2. Fix single-day width compare, which I broke. I just need to dig in to the compare code for this. It's on my list

I also need to get rid of all the references to the old way of doing compression, but I'm waiting for everything else to be done to do that.


simonbtomlinson:
The tests are getting stuck trying to create the `btree_gist` extension when not running under docker because of the way we have travis set up. @LeoTindall, do you know how I can give the non-docker postgres user admin permissions to fix that?

jameeters:
That's fixed in the baseline_refactoring branch.
Docker Version: https://github.com/OpenEnergyDashboard/OED/commit/fe5f115493fecef7066766b7a26faf86a06c1652
non-docker version: https://github.com/OpenEnergyDashboard/OED/commit/115f150b220f19c377b3c6cd3253d1ccaefcd685

shuheif:
Although it passes the Travis test, there's a problem that the database returns empty data for readings. I've spent many hours on this but couldn't figure out why. @simonbtomlinson, could you take a look at the code?

simonbtomlinson:
Sure, I'll try to get to it over Thanksgiving. Feel free to bug me about it
if I forget.

shuheif:
This PR is good to merge once #374 is approved to cover the comparison graph.

Although we expected this PR to solve #312, the issue still remains.

src/server/sql/reading/calculate_compare.sql
```diff
@@ -0,0 +1,106 @@
...
+		WHERE r.start_timestamp >= current_period_start AND r.end_timestamp <= current_period_end
+		GROUP BY m.id
+	),
+	prev_period AS (
+		SELECT
+			m.id AS meter_id,
+			SUM(r.reading) AS reading
+		FROM readings r
+			INNER JOIN unnest(meter_ids) m(id) ON r.meter_id = m.id
+		WHERE r.start_timestamp >= prev_period_start AND r.end_timestamp <= current_period_start -- prev_end == curr_start
```
kuchtact, line 52: prev_end = curr_start only when curr_period_duration = period_duration. I think that both the comment and logic should be replaced but it may occur so rarely that they are not equal that it won't make sense to implement. It is just one more addition though so...

src/server/sql/reading/calculate_compare.sql
```diff
@@ -0,0 +1,106 @@
...
+	prev_period_start TIMESTAMP;
+	curr_period_duration INTERVAL; -- How far into the current period we are
+	prev_period_partial_end TIMESTAMP;
+BEGIN
+	curr_period_duration := current_period_end - current_period_start;
+	ASSERT curr_period_duration <= period_duration, 'Current period is too long';
+	prev_period_start := current_period_start - period_duration;
+	prev_period_partial_end := prev_period_start + curr_period_duration;
+
+	RETURN QUERY
```
kuchtact, line 35: All of the following selects follow the same logic so I'll mention them all together. When selecting readings and checking that they are in the time range, would it not be better to just be looking for time range overlaps and just scale down the value of the reading if it is not fully within the range? This may be too many computations though so it is understandable not to have it but it does ignore some data and is an underestimate for both current and previous.

src/server/sql/reading/calculate_compare.sql
simonbtomlinson: That's not actually true. The start of the current period will always be the end of the one before it, but the end of the current one isn't always (and is usually not) `prev_end + period_duration` since the current period isn't usually over (instead it's `prev_end + curr_duration`).

But we can assume that the current one starts where the previous one ends.

I might have some variable names slightly wrong; I haven't looked at this in months.

src/server/sql/reading/calculate_compare.sql
simonbtomlinson: Yeah, that's likely a good idea. We should probably have an sql function that handles the scaling since it happens in so many places. It's definitely not too much of a performance hit since we should only be looking at small intervals here (as I mention in the comment at the top). I'll take a longer look at it.

src/server/sql/reading/calculate_compare.sql
kuchtact: I see it now. Thank you for the insight!

kuchtact:
In terms of doing all caps versus all lowercase for PgSQL functions, is there some format that should be followed? I noticed that it changes throughout the code but is usually consistent for the same type of function (like `sum` or `array_agg`).

simonbtomlinson:
> In terms of doing all caps versus all lowercase for PgSQL functions, is there some format that should be followed? I noticed that it changes throughout the code but is usually consistent for the same type of function (like `sum` or `array_agg`).

It should all be case-insensitive, and I should have been more consistent with it. I lean towards all-caps when something is defined by the sql standard since then I usually don't think of it as a postgres function, and towards lower-case when it's postgres-specific (which includes all the array stuff like `array_agg` and `unnest`).

I think postgres convention is all-lowercase for everything, but I first started writing mssql, where it was the opposite.

kuchtact:
> It should all be case-insensitive, and I should have been more consistent with it. I lean towards all-caps when something is defined by the sql standard since then I usually don't think of it as a postgres function, and towards lower-case when it's postgres-specific (which includes all the array stuff like `array_agg` and `unnest`).
> 
> I think postgres convention is all-lowercase for everything, but I first started writing mssql, where it was the opposite.

Maybe we should just capitalize all of the PgSQL functions as well because then it makes it very easy to see which are custom functions and which are not. [This](https://dba.stackexchange.com/questions/68264/official-postgresql-capitalization-conventions) has some opinions on code conventions so we could follow them.

src/server/sql/reading/calculate_compare.sql
kuchtact: This same kind of thing happens in `create_function_get_barchart_readings.sql` in the `SUM`. It is calculating the weight for the weighted average but this logic also works for creating these compares except we aren't averaging. Thus if we created a function like this, it would find use in multiple places.

simonbtomlinson:
> > It should all be case-insensitive, and I should have been more consistent with it. I lean towards all-caps when something is defined by the sql standard since then I usually don't think of it as a postgres function, and towards lower-case when it's postgres-specific (which includes all the array stuff like `array_agg` and `unnest`).
> > I think postgres convention is all-lowercase for everything, but I first started writing mssql, where it was the opposite.
> 
> Maybe we should just capitalize all of the PgSQL functions as well because then it makes it very easy to see which are custom functions and which are not. [This](https://dba.stackexchange.com/questions/68264/official-postgresql-capitalization-conventions) has some opinions on code conventions so we could follow them.

That's certainly an option, but you can also just click on them and intellij will tell you. I mostly was just capitalizing the ones that intellij bolded. Maybe we should open an issue to talk about this where other people can have input?

simonbtomlinson:
Yep, that's what I was thinking. I'll push a commit in a bit. It should
really improve readability in a lot of places.

src/server/sql/reading/calculate_compare.sql
simonbtomlinson: Yeah, this logic happens in at least 3 places.

huss:
> > > It should all be case-insensitive, and I should have been more consistent with it. I lean towards all-caps when something is defined by the sql standard since then I usually don't think of it as a postgres function, and towards lower-case when it's postgres-specific (which includes all the array stuff like `array_agg` and `unnest`).
> > > I think postgres convention is all-lowercase for everything, but I first started writing mssql, where it was the opposite.
> > 
> > 
> > Maybe we should just capitalize all of the PgSQL functions as well because then it makes it very easy to see which are custom functions and which are not. [This](https://dba.stackexchange.com/questions/68264/official-postgresql-capitalization-conventions) has some opinions on code conventions so we could follow them.
> 
> That's certainly an option, but you can also just click on them and intellij will tell you. I mostly was just capitalizing the ones that intellij bolded. Maybe we should open an issue to talk about this where other people can have input?

When we started OED I had the utopian dream that we would only use standard SQL. Then I thought we would isolate anything that was non-standard so it would be easy to convert to another system. At this point we are so wedded to PostgreSQL that we could not change without a huge effort. Thus, I don't think it matters too much but if it is easy to have a convention to know what should be standard SQL and what is PostgreSQL that might be nice. I agree with @simonbtomlinson that we should open a separate issue on this since it is somewhat separate from this pull request and we want to get this merged without any absolutely necessary or fast changes since others are waiting on this (along with release 0.5).

PR #337, opened 2018-05-11
Log mailer

buip:
Send email on ERROR level


src/server/logMailer.js
```diff
@@ -0,0 +1,59 @@
...
+		transporter = nodemailer.createTransport({
+			service: 'gmail',
+			auth: {
+				user: config.mailer.ident,
+				pass: config.mailer.credential
+			}
+		});
+	} else {
+		// tslint:disable-next-line no-console
+		console.error(`Unable to send e-mail due to unknown mailer method ${config.mailer.method}`, skipMail = true);
```
johndiiorio: Why do you print to the console instead of logging in this file?

src/server/logMailer.js
```diff
@@ -0,0 +1,59 @@
...
+const log = require('./log').log;
+
+/**
+ * Send an e-mail representing an error message.
+ * @param {string} level The level, as a string
+ * @param {string} message The string to e-mail
+ */
+function logMailer(level, message) {
+
+	var mailOptions = {
```
johndiiorio: Please do not use var.

src/server/logMailer.js
```diff
@@ -0,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const nodemailer = require('nodemailer');
+const mg = require('nodemailer-mailgun-transport');
+const config = require('./config');
+const log = require('./log').log;
```
johndiiorio: Log is not used but should be used.

USAGE.md
```diff
@@ -27,14 +27,23 @@ Killing the running process (ctrl+C) will stop the app. You can get rid of the D
...
 OED_DB_HOST=?                  // The host for your postgres db, likely localhost
 OED_DB_PORT=?                  // The port for your postgres db, likely 5432
-OED_TOKEN_SECRET=?             // Token for authentication. Generate something secure and random
+
 OED_LOG_FILE=?                 // Path to the log file, defaults to ./log.txt
+
+OED_MAIL_METHOD=none		   // Method of sending mail. Supports "gmail", "mailgun", "none". Case insensitive.
+OED_MAIL_IDENT=user@gmail.com  // Identifier; username for gmail, domain for mailgun.
+OED_MAIL_CREDENTIAL=?		   // Credential; password for gmail, API key for mailgun.
+OED_MAIL_FROM=user@gmail.com   // From address for email
```
johndiiorio: I would change all instances of `@gmail.com` to `@example.com`

docker-compose.yml
```diff
@@ -28,6 +28,12 @@ services:
             - OED_DB_PORT=5432
             - OED_TOKEN_SECRET=?
             - OED_LOG_FILE=log.txt
+            - OED_MAIL_METHOD=gmail
+            - OED_MAIL_IDENT=openenergydashboard@gmail.com
```
johndiiorio: I would change all of these to be dummy variables, such as `mydomain@example.com`

USAGE.md
```diff
@@ -27,14 +27,23 @@ Killing the running process (ctrl+C) will stop the app. You can get rid of the D
...
 OED_DB_TEST_DATABASE=?         // The test database you just created, so likely oed_testing
 OED_DB_PASSWORD=?              // The password for your postgres user
 OED_DB_HOST=?                  // The host for your postgres db, likely localhost
 OED_DB_PORT=?                  // The port for your postgres db, likely 5432
-OED_TOKEN_SECRET=?             // Token for authentication. Generate something secure and random
+
 OED_LOG_FILE=?                 // Path to the log file, defaults to ./log.txt
+
+OED_MAIL_METHOD=none		   // Method of sending mail. Supports "gmail", "mailgun", "none". Case insensitive.
+OED_MAIL_IDENT=user@gmail.com  // Identifier; username for gmail, domain for mailgun.
```
johndiiorio: How does IDENT differ from MAIL_FROM?

USAGE.md
```diff
@@ -27,14 +27,23 @@ Killing the running process (ctrl+C) will stop the app. You can get rid of the D
...
+OED_MAIL_CREDENTIAL=?		   // Credential; password for gmail, API key for mailgun.
+OED_MAIL_FROM=user@gmail.com   // From address for email
+OED_MAIL_TO=admin@example.com  // Who gets the e-mail
+OED_MAIL_ORG=Development	   // Organization Name
```
johndiiorio: Would change this to a dummy name, such as `My Organization Name`

src/server/config.js
```diff
@@ -28,6 +28,20 @@ config.database = {
 	port: process.env.OED_DB_PORT
 };
 
+// Mail configuration
+let method = process.env.OED_MAIL_METHOD;
+if (method !== null & method !== undefined) {
```
johndiiorio: I think you want a `&&` instead of `&`.

src/server/log.js
```diff
@@ -31,6 +32,7 @@ class Logger {
 
 	constructor(logFilePath) {
 		this.level = LogLevel.INFO;
+		this.emailLevel = LogLevel.ERROR;
 		this.logToFile = true;
 		this.logToConsole = true;
 		this.logFile = logFilePath;
```
johndiiorio, line 38: Not used

src/server/log.js
```diff
@@ -43,80 +45,99 @@ class Logger {
...
+				});
 			}
 		}
-		if (this.logToFile) {
-			fs.appendFile(logFile, messageToLog, err => {
-				if (err) {
-					console.error(`Failed to write to log file: ${err}`); // tslint:disable-line no-console
+
+		// Only send an e-mail if given a high enough priority level.
+		if (level.priority <= this.emailLevel.priority) {
```
johndiiorio: Don't you want `level.priority >= this.emailLevel.priority`?

src/server/log.js
```diff
@@ -43,80 +45,99 @@ class Logger {
...
+			let messageToMail = `At ${new Date(Date.now()).toISOString()}, an ${level.name} event occurred.\n`;
+			messageToMail += `${message}\n`;
+			// Add a stacktrace to the message if one was provided
+			if (error !== null) {
+				if (error.stack) {
+					messageToMail += `Stacktrace: \n${error.stack}\n`;
+				} else {
+					messageToMail += `An error was included, but it was not an Error object:\n${error}`;
```
johndiiorio: Is this text really necessary to email?

src/server/logMailer.js
```diff
@@ -0,0 +1,59 @@
...
+		to: config.mailer.to,
+		subject: `[OED ${config.mailer.org}] Open Energy Dashboard ${level}`,
+		text: message
+	};
+
+	let transporter;
+	if (config.mailer.method === 'none') {
+		return;
+	} else if (config.mailer.method === 'mailgun') {
+		transporter = nodemailer.createTransport(mg({
```
johndiiorio: I think it would be nicer if you moved the declarations of the transporters outside the function call.

USAGE.md
```diff
@@ -27,14 +27,23 @@ Killing the running process (ctrl+C) will stop the app. You can get rid of the D
 1. Create a .env file in the root directory of the project with the following, replacing (?) with the desired information: <br>
 ```
 OED_SERVER_PORT=?              // The port that the server should run on. 3000 is a good default choice
+OED_TOKEN_SECRET=?             // Token for authentication. Generate something secure and random
```
johndiiorio: I know this isn't part of your pull request, but can you change the comment types of all the env variables from `//` to `#`?

johndiiorio (changes requested):
It looks pretty good. If possible, could you set it up so that only one email is sent per time period? Otherwise the admin will get spammed with lots of emails.

huss:
I agree that limiting the number of emails by time and also not repeating the same issue multiple times would be good. However, I want the basic functionality working now and the emails will only be going to the project for monitoring. Before we release so others get the emails we will need to fix this up. If some changes are quick then they can be done but I don't want to delay this past the upcoming release unless absolutely necessary.

src/server/log.js
buip: I think the way it is setup is the low means high level and we only send email when there is an error

src/server/logMailer.js
buip: Used

src/server/log.js
buip: I think we used it in this line `if (level.priority <= this.emailLevel.priority)`

buip:
I add a feature that sends an email every hour. When there is no error, it will not send. If there are multiple errors, it will combine it into one email and send. After the email is sent, it will clear the message stack.

src/server/logMailer.js
```diff
@@ -0,0 +1,88 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const nodemailer = require('nodemailer');
+const mg = require('nodemailer-mailgun-transport');
+const config = require('./config');
+const schedule = require('node-schedule');
```
johndiiorio: We have been using Linux's built-in cron scheduler to perform scheduled tasks. This raises the question, should we continue using cron or rely on NodeJS? @simonbtomlinson, @jameeters, and @LeoTindall, what do you think?

buip: I keep seeing this in Travis: `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`. I don't know if node-schedule caused this

johndiiorio: It might be causing it, especially if you weren't getting it before adding it. Try using a normal bash script located in src/scripts to call your mail script.

buip: I removed it and Travis ran again. I will add the Cron script. I think I will create a JS script in service to do that. Can I do sth like this in the bash script: `docker-compose run --rm web npm run --silent sendLogEmail &>> /dev/null &`

NoraCodes: Yeah, I think it would work better to do it that way. Thanks @buip.

src/server/log.js
buip: I am not sure. Do you want to include this in the email @huss ?

buip:
Thank you for reviewing my PR @johndiiorio and @LeoTindall. I think I made all the changes. Could you please take a look again? Thank you

NoraCodes (approved):
I like your approach and the code is good. I say merge it.

huss (approved):
I think this is now all okay unless @johndiiorio thinks their comment still needs addressing. I'm approving so this can be merged and pushed to development server.

PR #339, opened 2018-06-03
Log mailer redesign

buip:
My last implementation had some data persistence issue so I switch to a new design. I added the messages to the database, send it at the end of the day, and remove everything from that table.


huss (approved):
This generally looks okay so I am going to approve so we can move forward on this since it currently only sends messages to the OED team. It would be good for someone with more knowledge of this code to check it out.

PR #340, opened 2018-06-03
Migration script for 0.4.0

buip:
We have to execute the create migration table and insert default migration first.

The specific steps are: 
1. Run `docker-compose exec database psql -U oed`

2. Then run
`
CREATE TABLE IF NOT EXISTS migrations (
	id SERIAL PRIMARY KEY,
	from_version VARCHAR(20),
	to_version VARCHAR(20) NOT NULL,
	update_timestamp TIMESTAMP DEFAULT NOW() NOT NULL
);
`

3. Then run
`
INSERT INTO migrations (from_version, to_version, update_timestamp)
VALUES (NULL, '0.3.0', NOW())
RETURNING id;
`

Migration script from 0.3.0 to 0.4.0


huss (approved):
We want to migrate so I'm approving this. As with the other change (#339), it would be good for someone else to review.

PR #342, opened 2018-09-03
Fix long timeouts for adding new meters

NoraCodes:
As well as general improvements to error handling for adding new meters,
like not adding meters that already exist.


shuheif (approved):
It looks good to me. I think it would be nicer to show an error message when a timeout happens or when meters already exist in the future

PR #349, opened 2018-09-26
Add Nodemon; now `start:dev` watches and restarts

NoraCodes:
Also changed in devstart.sh so the guide uses the best version for
development.


buip (approved):
Have been waiting for this for a long time 💯

PR #350, opened 2018-09-30
Allow meters to be made invisible to unauthenticated users

NoraCodes:
This is a WIP pull request. I am interested in feedback but it is not ready to be merged.

This PR adds the ability to turn off the display of meters, and prevents unauthenticated users from seeing non-displayable meters. It also adds a meter management UI.


src/client/app/components/HeaderButtonsComponent.tsx
```diff
@@ -24,33 +24,14 @@ export default class HeaderButtonsComponent extends React.Component<HeaderButton
...
-		let showOptions = false;
-		let renderLoginButton = false;
-		let renderHomeButton = true;
-		let renderAdminButton = false;
-		let renderGroupsButton = true;
+		const showOptions = false;
+		const renderLoginButton = !hasToken();
+		const renderHomeButton = getPage() !== '';
+		const renderAdminButton = hasToken() && getPage() !== 'admin';
+		const renderGroupsButton = hasToken() && getPage() !== 'groups';
```
shuheif: I think we don't need the hasToken() check here as the current development branch allows rendering the button without logging in.

shuheif:
This new feature is cool! I think the structure of both the front-end and the back-end makes sense and looks good to me. As long as I see, there are 
- 'column "displayable" does not exist' error when adding new meters, and 
- the implementation for actual enabled/disabled switch button

 to be done. Do you have any concerns or anything that you want to discuss about at this point @LeoTindall?

src/client/app/components/HeaderButtonsComponent.tsx
NoraCodes: Like this?

```suggestion
		const renderGroupsButton = getPage() !== 'groups';
```

NoraCodes:
> 'column "displayable" does not exist' error when adding new meters, and

I don't get this error; did you run the database migration?

> the implementation for actual enabled/disabled switch button

This is on hold until Obvius stuff is done; once I get time to work on this, I would definitely appreciate your input.

src/client/app/components/HeaderButtonsComponent.tsx
shuheif: Yes

NoraCodes:
This pull request is no longer WIP and requires review

shuheif:
The code looks good in general. Although the Travis test didn't work, the code passes all the tests on my laptop. I noticed a few things:

- I think the Group button on the header should be visible to the normal users. (As mentioned before)

- The text of "Save meter edits" button is cut off in the middle.

- I think it would be nicer to hide the save button to non-admin users if possible.

- What happens if a group contains disabled/hidden meters? I couldn't check this because the add-group feature is currently broken, also I couldn't find the code to support groups. I'm sorry if this is out of scope for this PR.

NoraCodes:
> The code looks good in general. Although the Travis test didn't work, the code passes all the tests on my laptop. I noticed a few things:
> 
>     * I think the Group button on the header should be visible to the normal users. (As mentioned before)

Definitely, I forgot that I overwrote that. I'll fix it.

>     * The text of "Save meter edits" button is cut off in the middle.

Can you screenshot this? I can't repro.

>     * I think it would be nicer to hide the save button to non-admin users if possible.

Definitely, didn't even think of that but you're right.

>     * What happens if a group contains disabled/hidden meters? I couldn't check this because the add-group feature is currently broken, also I couldn't find the code to support groups. I'm sorry if this is out of scope for this PR.

The data is still included in the group. This is useful to admins so they can group meters together which are nonsensical by themselves.

shuheif:
Here're the screenshots. Apparently, the save button shrinks as the screen gets smaller. I checked it on Google Chrome and Safari.
screen shot 2018-11-15 at 12 23 33 am
screen shot 2018-11-15 at 12 23 58 am

PR #351, opened 2018-09-30
Fix TSC/ATLoader not giving errors in some cases

NoraCodes:
ATLoader needs to have a plugin enabled to give good error messages and we hadn't done that, so here it is.


shuheif:
The change looks good to me.

shuheif (approved):

Issue #299, opened 2018-02-14
include CSV line number when logging issues on meter reads


huss:
If the CSV meter data has an issue, we log it. It would be helpful if it included the line number within the CSV file to help in locating the problem.

PR #353, opened 2018-10-16
CSV line numbers for errors

CorinaSterbet:
Solves issue #299

### DIFF
--- a/src/server/services/readMamacData.js
+++ b/src/server/services/readMamacData.js
@@ -14,11 +14,11 @@ function parseTimestamp(raw, line) {
 	raw = raw.trim();
 	const timestampRegExp = /^\d{2}:\d{2}:\d{2} \d{1,2}\/\d{1,2}\/\d{1,2}$/;
 	if (!timestampRegExp.test(raw)) {
-		throw new Error(`CSV line ` + line + `: Raw timestamp ${raw} does not pass regex validation`);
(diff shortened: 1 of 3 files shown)


src/server/services/readMamacData.js
```diff
@@ -10,15 +10,15 @@ const Reading = require('../models/Reading');
 
 const parseCsv = promisify(csv.parse);
 
-function parseTimestamp(raw) {
+function parseTimestamp(raw, line) {
 	raw = raw.trim();
 	const timestampRegExp = /^\d{2}:\d{2}:\d{2} \d{1,2}\/\d{1,2}\/\d{1,2}$/;
 	if (!timestampRegExp.test(raw)) {
-		throw new Error(`Raw timestamp ${raw} does not pass regex validation`);
+		throw new Error('CSV line ' + line + `: Raw timestamp ${raw} does not pass regex validation`);
```
buip: ``throw new Error(`CSV line ${line}: Raw timestamp ${raw} does not pass regex validation`);``
This will be more consistent since you don't have to concat the string.

buip (changes requested):

src/server/services/readMamacData.js
CorinaSterbet: Makes sense! Fixed.

NoraCodes (approved):

buip (approved):

PR #355, opened 2018-10-25
Keep any data already in the database when inserting readings

shuheif:
Changed to ignore a new reading when the database already has data for the same timestamp.


NoraCodes (approved):
Looks good to me

Issue #100, opened 2017-09-08
https


huss:
We should figure out how to support https, esp. as we develop login/admin

LeoTindall:
Perhaps we can use a self-signed cert for dev and then get a LetsEncrypt
cert at some point

huss:
We have a certificate from Beloit College that should be working with
Apache. I think we need to figure out how to make it work with our web
server software.

huss:
Now that we have users and an admin login, the issues of https is becoming more important. I suspect this has to do with the port we are using as you can do a secure connection to the Apache web server on the same machine. While this relates to our development server, I think it will likely be similar on other Linux systems. Anyone with spare time interested in looking into this?

LeoTindall:
I'd be happy to; I don't have access to the server, but if someone could send me the Apache config I'll look at it

huss:
I want to up the ante on this one. I have now been in two circumstances where people at other places cannot access our OED site. I suspect it is related to not having a good certificate. In one case it was very strange because they could not seem to go to explicit ports but could go to the named site (oed.beloit.edu/demo). I cannot explain why.

NoraCodes:
This is not a problem with the OED software and I think it should be closed here - OED supports HTTPS via reverse proxy, it's just that the main installation's reverse proxy isn't working.

huss:
This is an operations issue on our own server and may become an issue for others. Until we have a way to track OED related issues that are not directly the code we can leave this here.

NoraCodes:
There exists a reverse proxy bug which I am working on. This comment is a note that, when merging Obvius functionality, we need to ensure that we support X-Real-IP for logging, or give up on logging IP addresses.

huss:
This was resolved about a year ago but not closed so here I go.

PR #359, opened 2018-11-22
Support for reverse proxy through Apache or Nginx.

NoraCodes:
Move database container spec to the new directory containers/
Add a new environment variable, OED_SUBDIR
String replace in index.html request

Addresses #100


FaiqHussain:
The file src/client/index.html I have some questions
1) Line 9 and 12 has the html command of Link there. It starts with <Link and the things needed but shouldn't it be closed with </Link> to make the command workable?
2) Line 18 has </script> on it. I don't see the <script> that goes with it?

NoraCodes:
> Line 9 and 12 has the html command of Link there. It starts with <Link and the things needed but shouldn't it be closed with  to make the command workable?

The `link` tag (not command; HTML is a declarative language) does not need to be closed, per MDN documentation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link).

> Line 18 has </script> on it. I don't see the <script> that goes with it?

It's on line 16.

MatthewDNielsen:
I've read through the code and everything seems fine to me.  I don't know if there are any other reasons we're holding up this pull request.  I would like to note however that from my perusal of things, and due to how old this pull request is now, it looks like four or so of the files modified by this pull request have since been modified in other commits to development.  Those files are USAGE.md, database/init.sql and database/nuke.sql (both of which are to be moved by this pull request), and src/server/app.js.  I don't know if anything particular needs to be done about this matter, but it felt like something I should flag.

NoraCodes:
For the benefit of any reviewers and @huss, this patch only adds the _capability for OED to be hosted behind a reverse proxy at a subdirectory path_.

Our current hosting situation will not change when this patch is merged.

Prior this patch, OED would either be in a situation where internal links and API calls would go to the root instead of the /path, or would be unable to recognize user calls and API calls because they were not matching on the /path/api/... but rather looking for /api/...

Our hosting situation will have to change after this patch is merged.

We will need to configure Apache to pass through requests as a reverse proxy and terminate HTTPS for us.

NoraCodes:
This failed CI because ReactBootstrap updated just a few minutes before I pushed my stuff, so I've added the fixes related to their type changes.

NoraCodes:
Sorry about that, all fixed.

USAGE.md
```diff
@@ -32,19 +50,23 @@ terminal.
...
+Windows is not a supported platform. You can probably run OED there, but you will
+experience problems.
 
-Docker is not supported on Windows. Windows is incapable of supporting LXE containers,
-the technology on which Docker is based, and Microsoft
-[artifically limits](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/about/)
-the use of Docker for Windows, a virtualized technology.
+Docker is a Linux technology. It is based on Linux kernel features including namespaces
+and cgroups. On Mac OS, you can use [Docker for Mac](https://docs.docker.com/docker-for-mac/)
+to run OED in Docker. Microsoft [artifically limits](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/about/)
```
huss: The term "artifically limits" (correct spelling of artificially) does not seem a neutral term. From what I understand, it does not run on the home version of Windows but does run on others. Is that correct for OED? If so, should we state that or link to MS and say they specify which versions of the OS are supported?

USAGE.md
```diff
@@ -32,19 +50,23 @@ terminal.
...
+the use of Docker for Windows, so you may experience problems.
 
-Docker is supported on MacOS via the [Docker for Mac](https://docs.docker.com/docker-for-mac/)
-software, a virtualized technology. You may experience issues with the `devstart` script
-which uses features that have problems on Docker for Mac to watch files for edits.
+Open Energy Dashboard universally uses UNIX filenames, which are seperated with forward
+slashes. Linux uses UNIX filenames. Mac OS uses UNIX filenames. BSD uses UNIX filenames.
+Solaris uses UNIX filenames. SmartOS uses UNIX filenames.
+
+For reasons known only to Bill Gates and God, Windows uses backwards slashes instead.
```
huss: Again, this has a negative tone. I think we should say that the forward slash is used in our scripts and this is not compatible with Windows OS.

huss (changes requested):
Since others have checked and found the changes okay, I'm prepared to approve once the two comments I have made are addressed.

USAGE.md
NoraCodes: Understood, I'll reword it.

USAGE.md
NoraCodes: > The term "artifically limits" (correct spelling of artificially) does not seem a neutral term. From what I understand, it does not run on the home version of Windows but does run on others. Is that correct for OED?

This is precisely what I mean by an "artificial" limit. They have the capacity to support it on all versions of Windows and choose not to do so. That link explains all the various complexity inherent therein. I can change it to make that more clear.

huss: I think we should try to stay neutral about what we say about OSes. I might prefer something similar to: "Docker is not supported on all versions of Windows. You should see [link] for details from Microsoft. You will need to manually install OED ([link to directions?] if this is the case for you." How does this sound?

NoraCodes: I'm somewhat reluctant to do so given that this _is in fact_ a bad faith action by Microsoft Corporation. They previously allowed users to do something and then just _stopped_ letting them, almost totally invalidating an entire semester's worth of work in the process. I don't feel like there's a particularly good reason for us to be neutral about that.

I'll to take the language out, but given that I've spent a significant portion of my time this year trying (and often failing) to get other students set up with a working development environment on Windows, I think it might be a good idea to have a conversation, as a team, about whether or not we actually want to make an effort to support that platform.

As much as we may want to be neutral, the fact remains that people on Linux and Mac OS X are able to get set up with the project in a few minutes, while it takes hours at best on Windows. The whole point of Dockerization was to solve that, and it did, until Microsoft decided to hold that feature for ransom.

huss (approved):
I'm approving this but I think we might want to point to the Microsoft website and not state what is or is not supported. As you noted, they have chosen to make changes that impacted Docker and this project. Thus, I am a bit weary of stating what is the situation given it might change and rather leave it to them to post the situation.

You also noted that supporting certain Microsoft OS versions is much more difficult given decisions made on how to deploy OED. I agree that we should have a project discussion about what environments we readily support and then can post that on GitHub. If we choose not to support some versions (or all) of Microsoft OS then we will be making a statement but it will be indirect. We would most likely tell why this is the case for technical reasons rather than the motive behind another entity's action. However, before we do this I think we should discuss the options and make a decision as a team, just like we do for other significant choices.

USAGE.md
```diff
@@ -32,19 +50,18 @@ terminal.
...
-FreeBSD, OpenBSD, and Plan9.
+Windows is not a supported platform. You can probably run OED there, but you will
+experience problems for the following reasons.
 
-Docker is not supported on Windows. Windows is incapable of supporting LXE containers,
-the technology on which Docker is based, and Microsoft
-[artifically limits](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/about/)
-the use of Docker for Windows, a virtualized technology.
+First, Docker is a Linux technology. It is based on Linux kernel features including namespaces
+and cgroups. On Mac OS, you can use [Docker for Mac](https://docs.docker.com/docker-for-mac/)
```
jameeters, line 59: If the section header is going to say "Non-Unix", then Mac OS should go somewhere else. Despite the fact that it's a different world from the other unices as far as docker is concerned, nobody is going to be looking for Mac instructions under the heading "Non-Unix".

NoraCodes: That's a good point, thanks! This language was here before this PR so I will move it in another PR.

Issue #347, opened 2018-09-21
"conn is not a function" error when creating a new group


shuheif:
I'm getting "conn is not a function" error when creating a new group. This is because getDB func should not have () when it is used as conn variable inside methods. 
Related to #346 

`[ERROR@2018-09-21T19:00:34.895Z] Error while inserting new group BatchError 
    stat: { total: 5, succeeded: 0, failed: 5, duration: 2 }
    errors: [
        0: TypeError: conn is not a function
            at Function.getByID (/usr/src/app/src/server/models/Meter.js:76:21)
`

NoraCodes:
It should be passed in as a function and called inside model methods. E.g.:

```
function whatever(arg1, arg2, conn = getDB) {
    conn().whatever(arg1, arg2)...
}
```

That is, the function should be called _at the time the model method is called_, rather than _at the time the model method is created_ which is what would happen if one wrote:

```
function whatever(arg1, arg2, conn = getDB()) {
    conn.whatever()
}
```

shuheif:
I agree that it should be called inside model methods as your first example. I guess the problem is not the parenthesis but the other part of the code. Below is the error message I got when I tried to create a new group.

```
[ERROR@2018-09-24T20:24:26.484Z] Error while inserting new group BatchError {
    stat: { total: 4, succeeded: 0, failed: 4, duration: 0 }
    errors: [
        0: TypeError: conn is not a function
            at Function.getByID (/usr/src/app/src/server/models/Meter.js:76:21)
            at Group.adoptMeter (/usr/src/app/src/server/models/Group.js:127:29)
            at req.body.childMeters.map.mid (/usr/src/app/src/server/routes/groups.js:146:71)
            at Array.map (<anonymous>)
            at Task.getDB.tx (/usr/src/app/src/server/routes/groups.js:146:51)
            at <anonymous>
            at process._tickCallback (internal/process/next_tick.js:118:7)
        1: TypeError: conn is not a function
            at Function.getByID (/usr/src/app/src/server/models/Meter.js:76:21)
            at Group.adoptMeter (/usr/src/app/src/server/models/Group.js:127:29)
            at req.body.childMeters.map.mid (/usr/src/app/src/server/routes/groups.js:146:71)
            at Array.map (<anonymous>)
            at Task.getDB.tx (/usr/src/app/src/server/routes/groups.js:146:51)
            at <anonymous>
            at process._tickCallback (internal/process/next_tick.js:118:7)
[...]

PR #363, opened 2018-11-28
Fix database errors for creating and editing groups

shuheif:
This fixes the "conn is not a function" errors when creating and editing groups. #347 

The Travis test failed because of the flatmap-stream package, which is removed in the fix-group-features branch.


sandeepacharya464 (commented):
Looks good to me Shuhei. I am glad that the "conn error" is fixed.

src/server/routes/groups.js
```diff
@@ -139,13 +138,10 @@ router.post('/create', async (req, res) => {
...
-				const newGroup = new Group(undefined, req.body.name);
-				await newGroup.insert(() => t);
-				const adoptGroupsQuery = req.body.childGroups.map(gid => newGroup.adoptGroup(gid, t));
-				const adoptMetersQuery = req.body.childMeters.map(mid => newGroup.adoptMeter(mid, t));
-				return t.batch(_.flatten([adoptGroupsQuery, adoptMetersQuery]));
-			});
+			const newGroup = new Group(undefined, req.body.name);
+			await newGroup.insert();
+			await req.body.childGroups.map(gid => newGroup.adoptGroup(gid));
+			await req.body.childMeters.map(mid => newGroup.adoptMeter(mid));
```
NoraCodes, line 144: If these don't happen in a transaction, is it possible for the adoption of groups to succeed and the adoption of meters to fail, leaving the database in a bad state?

NoraCodes (changes requested):

src/server/routes/groups.js
shuheif: You're right, the database could be in a bad state. They should be called as a transaction or a task.

PR #368, opened 2018-12-05
Update USAGE.md for modern Windows quirks

NoraCodes:


NoraCodes:
Anyone know why this fails in CI? Very weird as I only changed USAGE.md

huss:
I have not looked at the details but from my quick look I wonder if someone added a migration test but that is not yet in this branch. Just a somewhat wild guess.

buip:
You can try to merge development in again

USAGE.md
```diff
@@ -75,36 +75,62 @@ Killing the running process (ctrl+C) will stop the app. You can get rid of the D
 1. Install Node.js. Node.js is the runtime used to execute our serverside JavaScript application.
 1. Install NPM, the Node Package Manager. NPM is used to manage our JavaScript dependencies.
 1. Clone this repository.
+1. Create a CSV file with a single column called "ip" with your meter IP addresses and copy it into the directory in which the project resides.
```
huss: My recollection is it must be "ip" and "IP" does not work. It might be good to make this very clear. Something like: "it must be lowercase for the "ip" column heading".

huss (approved):
This is not due to anything you did but I think it would be good to make it clear that the loading of a CSV for meters and the fetch data (already a comment) is for MAMAC meter data acquisition.  As we prepare for other systems this seems a good idea. Maybe even something at the high level so people are well aware.

USAGE.md
buip: why don't we have a ip csv file already sitting in our repo with some default ip. It'just my thought. I don't know if it's more convenient or not.

NoraCodes: @huss this was originally your decision; we could certainly do this, but would need to decide on which IPs we could use. Certainly not the Beloit IPs, as they will not be accessible to non-Beloit users.

huss: I think the reworded USAGE file is clear and for format of this CSV is sufficiently simple that a developer/site can do this. Thus, lets not include a default CSV since that can have issues of people reading those meters. We should probably fix up the ip vs IP issue as we made reading of multiple types of CVS files a part of OED and have a common package to do this. Thus, lets defer acting on that.

huss:
I think this is okay to merge once we understand why it is failing the Travis check.

NoraCodes:
I rebased on current development again. If it passes CI I'm disposed to merge.

PR #369, opened 2018-12-05
Change reading type from int to real.

wendelap:


src/server/migrations/0.3.0-0.4.0/index.js
```diff
@@ -9,14 +9,11 @@ module.exports = {
...
-			await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/baseline/create_function_get_average_reading.sql'));
-		} catch (err) {
-			throw new Error('Error while migrating each sql file');
-		}
+		await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/preferences/create_language_types_enum.sql'));
+		await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/preferences/add_language_column.sql'));
+		await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/logemail/create_log_table.sql'));
+		await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/baseline/create_baseline_table.sql'));
+		await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/baseline/create_function_get_average_reading.sql'));
+		await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/readings/set_reading_type_to_real.sql'));
```
buip, line 17: Why do you take it out of the try-catch?

buip (changes requested):

src/server/migrations/0.3.0-0.4.0/index.js
wendelap: It erases the details of errors that occur during the migration. During debugging, we want those to propagate up.

src/server/migrations/0.3.0-0.4.0/sql/baseline/create_baseline_table.sql
```diff
@@ -5,7 +5,7 @@
 -- Allows us to use gist in this database
 CREATE EXTENSION IF NOT EXISTS btree_gist;
 -- create baseline table
-CREATE TABLE baseline (
+CREATE TABLE IF NOT EXISTS basline (
```
shuheif: could you fix this typo?

wendelap: Done.

src/server/migrations/0.3.0-0.4.0/index.js
wendelap: @buip I answered your question a week ago. If you could please either respond to this or resolve the request, it would be greatly appreciated. Thanks.

buip (approved):

Issue #371, opened 2018-12-11
make compare graphs compatible with fast pt


huss:
Fast PT does not support short time interval comparison graphs. To fix this we need to

- [ ] Augment back-end to get sub-daily data needed. The first step will be to use the raw data since the number of points will not be great with the current comparison graphs.

- [ ] The current comparison graphs are tied to bar charts. These will be decoupled and comparison will now use the new back-end.

- [ ] The back-end will be enhanced so it used Fast PT data for daily segments and then add in the extra sub-day data from the raw data.

huss:
This is done completely. It was part of fast-pt and converting to plotly.

PR #374, opened 2019-01-21
Compare with shift

shuheif:
Checking out from the fast-pt-compression branch #335, this PR improves the comparison graph by implementing the items addressed in #371 except the last one.

Here are the changes:

1. Created a new API that supports arbitrary time interval and shifts to compare with. The API uses the raw data to calculate usages.

1. Modified the front-end side to support the new API. The comparison data is decoupled from the bar chart and has its own Redux state.

1. Removed the total usage bars from the comparison chart in order to avoid confusion. Now, the chart shows only two bars: current usage and previous usage.

There is a minor error in this PR that I would like to get help with. The compare component is not re-rendering when receiving new comparison data. I checked that the new state contains the accurate data. It correctly shows the graph after switching the compare period with the buttons. [This](https://redux.js.org/faq/react-redux#why-isn-t-my-component-re-rendering-or-my-mapstatetoprops-running) and
[this page](https://redux.js.org/troubleshooting#nothing-happens-when-i-dispatch-an-action) seem to be pointing the problem, but I couldn't find the error in the code.


src/client/app/actions/barReadings.ts
```diff
@@ -91,7 +90,7 @@ function fetchMeterBarReadings(meterIDs: number[], timeInterval: TimeInterval):
 	return async (dispatch: Dispatch, getState: GetState) => {
 		const barDuration = getState().graph.barDuration;
 		dispatch(requestMeterBarReadings(meterIDs, timeInterval, barDuration));
-		const readings = await metersApi.barReadings(meterIDs, timeInterval, barDuration);
+		const readings = await compressedReadingsApi.meterBarReadings(meterIDs, timeInterval, Math.round(barDuration.asDays()));
```
NoraCodes, line 93: Since we have the API object model, we could probably do the toDays and rounding within the meterBarReadings if it gets a Duration.

src/client/app/containers/ExportContainer.ts
```diff
@@ -8,6 +8,16 @@ import ExportComponent from '../components/ExportComponent';
 import { ChartTypes } from '../types/redux/graph';
 import { ExportDataSet } from '../types/readings';
 import { State } from '../types/redux/state';
+import {CompressedBarReading, CompressedBarReadings, CompressedLineReading} from '../types/compressed-readings';
+
+
+function transformLineReadingToLegacy(reading: CompressedLineReading): [number, number] {
```
NoraCodes, line 14: I think `decompressLineReading` is a more semantically accurate name here.

src/client/app/containers/ExportContainer.ts
```diff
@@ -8,6 +8,16 @@ import ExportComponent from '../components/ExportComponent';
...
+	return [reading.startTimestamp, reading.reading];
+}
+
+function transformBarReadingToLegacy(reading: CompressedBarReading): [number, number] {
```
NoraCodes, line 18: I think `decompressBarReading` is a more semantically accurate name here.

src/common/TimeInterval.d.ts
```diff
@@ -16,4 +16,6 @@ export class TimeInterval {
 	public contains(other: TimeInterval): boolean;
 	public valueOf(): string;
 	public duration(specifier: string): number;
+	public getStartTimestamp(): moment.Moment;
```
NoraCodes, line 19: Not really a defect with this PR, but rather with TimeInterval; technically I think these are times, not timestamps, because their internal representation is not specified (it's just a Moment). Perhaps we should change the name?

src/server/migrations/0.3.0-0.4.0/index.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
 			await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/logemail/create_log_table.sql'));
 			await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/baseline/create_baseline_table.sql'));
 			await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/baseline/create_function_get_average_reading.sql'));
+			await db.none(sqlFile('../migrations/0.3.0-0.4.0/sql/compareReadings/create_function_get_compare_readings.sql'));
```
NoraCodes, line 18: Is this sufficient to perform all needed database changes from a 0.3 database?

NoraCodes (changes requested):
Basically looks good with some minor gripes, awesome work. Thank you.

One other thing: it would be very convenient to have a single script which can be run to both update meters and update the materialized views for this, for developer installs which are not always running.

src/client/app/containers/ExportContainer.ts
simonbtomlinson: I kind of like legacy cause it makes it obvious that it's not a way that we like doing this anymore.

src/client/app/actions/barReadings.ts
simonbtomlinson: Yeah, that makes sense to me.

src/common/TimeInterval.d.ts
simonbtomlinson: To me a time explicitly doesn't include a date (and it means that in python, postgres, ruby-on-rails, and other places). I like timestamp here, or datetime, because it conveys that we're tracking a moment in time as opposed to a moment in a day, and we have reasonably close to unix timestamp precision.

src/client/app/containers/ExportContainer.ts
simonbtomlinson: I kind of like legacy since it really tells you that we shouldn't be using naked `[number, number]` pairs all over the place. But I'm flexible.

It's also still a compressed reading. To decompress it you'd create more than one point.

src/client/app/actions/barReadings.ts
simonbtomlinson: My only issue with that is that it could give the impression that any precision is honored, not just an integer number of days. But with reasonable commenting I can definitely see changing that argument to `int | moment.Duration`, especially since this is typescript.

simonbtomlinson:
@NoraCodes the only issue with having the thing update when pulling new readings is that we explicitly only want to update the views every 24 hours, even though we want to pull every hour or so. We could have it update if there are no rows in the views, maybe? I think that would cover every use case pretty well.

huss:
> @NoraCodes the only issue with having the thing update when pulling new readings is that we explicitly only want to update the views every 24 hours, even though we want to pull every hour or so. We could have it update if there are no rows in the views, maybe? I think that would cover every use case pretty well.

I just want to note we will soon have sites that get us data every 15 minutes and about once a day with lots of points for the previous day. I'm fine with the solution since I think it covers these cases but wanted to ask since I am not certain. This data is not coming via the CSV read of MAMAC meters.

simonbtomlinson:
> > @NoraCodes the only issue with having the thing update when pulling new readings is that we explicitly only want to update the views every 24 hours, even though we want to pull every hour or so. We could have it update if there are no rows in the views, maybe? I think that would cover every use case pretty well.
> 
> I just want to note we will soon have sites that get us data every 15 minutes and about once a day with lots of points for the previous day. I'm fine with the solution since I think it covers these cases but wanted to ask since I am not certain. This data is not coming via the CSV read of MAMAC meters.

@huss 
We should really run a separate daily cron script for the view refresh, but other than that it shouldn't cause any new issues. If the refresh runs a few minutes after the import to give it time to finish that would be ideal.

src/common/TimeInterval.d.ts
NoraCodes: Ah, that makes sense. Perhaps we should call it startMoment and endMoment, or just start and end? timeinterval.start and timeinterval.end reads pretty well, semantically.

simonbtomlinson: Start and end seem fine to me unless end is somehow reserved. I've been doing ruby for so long it seems a little weird.

NoraCodes:
>  @NoraCodes the only issue with having the thing update when pulling new readings is that we explicitly only want to update the views every 24 hours, even though we want to pull every hour or so. We could have it update if there are no rows in the views, maybe? I think that would cover every use case pretty well.

That makes sense, definitely. That check should probably be in place anyway, in case an over-eager sysadmin runs the update script too often.

src/server/models/Reading.js
```diff
@@ -54,6 +55,34 @@ class Reading {
 		return getDB().none(sqlFile('reading/create_function_get_barchart_readings.sql'));
 	}
 
+	/**
+	 * Returns a promise to create the function and materialized views that aggregate
+	 * readings by various time intervals.
+	 * @return {Promise<void>}
+	 */
+	static createCompressedReadingsMaterializedViews(conn = getDB) {
```
simonbtomlinson, line 63: This still isn't going to work, as I've mentioned in other reviews and possibly an issue a while back. The argument here needs to be compatible with the argument you get in the callback from pg-promise's `db.tx`, otherwise there's no way to wrap the call in a transaction (unless you wrap the transaction object, but that seems like a bad solution). If you want to inject the db in this way you might be able to get away with wrapping the whole class declaration in a function that gets passed db and then calling it with (getDB()) and exporting the result. But I feel like you should explore mocking it as it gets exported instead.

simonbtomlinson (changes requested):
I only took a cursory glance at this, and I'll look more over the weekend when I have more time, but the only real issue I see so far is the `getDB` thing.

src/server/models/Reading.js
NoraCodes: So, the basic issue here is that all these functions want a default connection. Perhaps simply striking that out and forcing any usage of anything that touches the database to pass in a connection would work better?

This actually has a few advantages. First of all, it would definitely remove the use of getDB. Second, it would solve the problem of race conditions for database connection initialization, which was why we moved to getDB in the first place. Third, it would make explicit any use of the database, which is almost certainly a good idea anyway.

simonbtomlinson:
That's actually a great idea.

PR #377, opened 2019-02-02
Test determinism improvements

NoraCodes:
For some reason, tests fail in CI a lot even when they don't fail on dev machines. This PR is an investigation of that phenomenon.


NoraCodes:
OK, this is good to go and I think it will solve our random failure woes.

huss (approved):
I have not tested but this seem okay. I don't know if some version info needs changing based on the meeting today about the next release number. Also, I suspect this is going to be impacted by the DB state change also discussed but I assume that is a known change.

PR #393, opened 2019-02-25
Add inline documentation for groups frontend.

Zeewild:
Added them comments.


src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -6,10 +6,16 @@ import * as React from 'react';
...
 
+/*
+ * Establishes text for chart links.
+ */
 interface ChartLinkProps {
 	linkText: string;
 }
 
+/*
+ * Is the link visible?
```
NoraCodes: I think it makes more sense to put this comment on the actual line with the `showLink` field definition, or to annotate it, like:
```
/**
 * The type used to represent the state of a ChartLinkComponent.
 * showLink: whether the link is visible or not.
 */

File: src/client/app/components/ChartSelectComponent.tsx [hunk 2]
```diff
@@ -11,6 +11,9 @@ import ButtonGroup from 'reactstrap/lib/ButtonGroup';
 import { FormEvent } from 'react';
 import { FormattedMessage } from 'react-intl';
 
+/*
+ * Calls method to change the type of chart when a chart is selected.
```
NoraCodes, line 15: This is a type definition; it doesn't call anything.

src/client/app/components/LoginComponent.tsx
```diff
@@ -12,6 +12,9 @@ import { showErrorNotification } from '../utils/notifications';
 import { verificationApi } from '../utils/api';
 import translate from '../utils/translate';
 
+/**
+ * Creates components used by the LoginState
```
NoraCodes, line 16: This is a type definition. It doesn't actually execute; rather, it's used to make sure we didn't access variables that don't exist, or use the variables that _do_ exist in the wrong way.

Check out the comments on the type definitions on src/client/app/types to see how comments for these might look.

src/client/app/components/admin/AddMetersComponent.tsx
```diff
@@ -21,7 +21,11 @@ class AddMetersComponent extends React.Component<AddMetersPropsWithIntl, {}> {
 		super(props);
 		this.handleMeterToImport = this.handleMeterToImport.bind(this);
 	}
-
+	
+	/**
+	 * Takes an array of type "File" as "files" and does something to handle it.
```
NoraCodes: The "something" here is:
1. checks that the string is really there
2. splits it on line endings
3. grabs the IPs of all the meters in the file
4. submits those IPs as new meters

src/client/app/styles/spinner.css
```diff
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* Creates general spinner visuals, used and modified by spinner-item-1 and spinner-item-2 */
```
NoraCodes: As a declarative element, I would swap "creates" for "specifies", or simply remove it: "General spinner visuals..."

NoraCodes (changes requested):
This basically looks OK to me, barring the requested changes, but I might suggest pulling out the comments on the `render()` functions into some external document that describes how we use React for document structure, since they are repetitious.

Also, these commit messages are somewhat uninformative. There's a good guide on how to write useful commit messages [here](https://chris.beams.io/posts/git-commit/).

src/client/app/components/ExportComponent.tsx
```diff
@@ -50,6 +53,10 @@ export default function ExportComponent(props: ExportProps) {
 		const name = `oedExport_${chartName}_${startTimeString}_to_${endTimeString}.csv`;
 		graphExport(compressedData,	name);
 	};
+	/*
+	 * TODO: I am noticing that some files have specific "render" methods for the JSX code, 
+	 * while others just return it. Is there a reason for this inconsistency I am missing?
+	 */
```
jameeters: The difference is between functional and class-based components. https://reactjs.org/docs/components-and-props.html

jameeters:
I think the usage of `render()` is pretty well covered in the react docs, and our wiki already points to those docs, so I don't think we should have our own homebuilt react documentation. I think the comments on `render` can be safely dropped.

huss:
These comments were not merged at the time (my fault) and are now out of date so some no longer apply (we changed graphics package). Also, there remains a moderate number of TODO items and the requested changes are not addressed. I appreciate the effort by @Zeewild but given all of this I'm going to close this PR without merging.

PR #424, opened 2019-04-29
Fix issue where groups editing and creating mixed.

### LINKED ISSUE #415: Groups editing shows wrong name after cancellation
When you modify the name of a group, hit cancel, and then go to edit another group, it shows the previously entered name in place of meter's actual name. (previously #411)

NoraCodes:
This closes #415. Specifically, when transitioning between states in the
groups editing UI, state was not being kept track of. When a user
pressed "edit", a group was placed into editing, but it was not removed
when that action was cancelled, so that group remained "in editing"
if/when a new group was edited or created.

This meant that edits would go to the wrong group, and creating a new
group would actually over-write the old one.

This was corrected by calling the right state transitions in the right
places and resetting the group in editing state during a cancellation
transition.


huss (approved):
This seems okay but someone with more background in this area should check.

NoraCodes:
Merging in the absence of any complaints.

Issue #403, opened 2019-04-15
Compare doesn't update with selected meters until a time-frame is selected


aaronshirst:
Not sure why this one is happening; there's a chance that it is an error on my part or on the part of my machine-- however, after being able to load the Compare graphs normally, my OED build became unable to load Compare graphs, loading an entirely blank screen upon hitting the [Compare] button on the main page when comparing 2-3 meters.

NoraCodes:
I get a similar issue, which I think is the same thing. When you select groups/meters for compare, the data is loaded but not displayed until a new time-frame is selected. Like #407 .

NoraCodes:
Spent a while looking into this. It looks like an async issue in how compare handles getting new readings data. First of all, in `DashboardContainer`, there is the problem that `compareLoading` is based on `state.readings.bar.isFetching` when it should be `state.readings.compare.isFetching`.

That aside, what happens _before_ you select a time interval is that the readings returned by `metersApi.compareReadings` in `fetchMeterCompareReadings` are null; that is, `{ prev_use: null, curr_use: null }`. _After_ you select a time interval, you get back useful results.

It appears that the server does infact give `null` for both fields when asked. Specifically, for instance, the request:
```
GET /api/compareReadings/meters/11?curr_start=2019-04-28T05:00:00.000Z&curr_end=2019-04-29T05:01:25.797Z&shift=P7D
```

gives back nulls while 
```
GET /api/compareReadings/meters/11?curr_start=2019-04-07T05:00:00.000Z&curr_end=2019-04-29T05:01:40.967Z&shift=P28D
```

does not. Perhaps this is simply an issue with missing data and how we handle it?

PR #428, opened 2019-05-03
Fix chart links setting bar duration in seconds.

### LINKED ISSUE #418: Hotlinks do not actually bring the user to the linked page
The link from "toggle chart link" does not work.
When "toggle chart link" pasted into new tab no actions (other than default ones) are triggered.
Simply opens fresh OED page.

NoraCodes:
This lead to "week" setting 7 seconds, not 7 days, and similar for other
durations. This closes #418, while technically not fixing all of it,
the other portion is related to #403 which is out of scope.


huss (approved):
Short and sweet.

Issue #392, opened 2019-02-24
The reading values and time labels are mismatched at the bar graph display


CorinaSterbet:
The meter readings are added to the "datasets" variable in the order received while the dates are being sorted based on the first date of the time interval and only then added to "labels". As a result, the two variables are incorrectly associated when passed as the chart data for display.

huss:
There is already a fix for this that will be coming with the new Plotly graphics.

huss:
As expected, this was addressed when OED switched to plotly  but it was not tagged with that PR. Therefore, this issue has been addressed.

PR #436, opened 2019-05-28
Integrate Plotly into OED

CorinaSterbet:
This pull request replaces the old graphics package Chart.js with the new Plotly.js. 

The code for all graph types (line, bar, compare) has been remodeled/rewritten to correspond to the requirements of the new graphics package. Some notes:
- the line graph now features a new option to zoom in and out on specific time intervals chosen by the user;
- this pull request solves issue #392 which caused, for the bar graph, the values and meter names to be mismatched at the graph display;
- I have looked into issue #312 -- however, I could not detect the errors mentioned there. When I checked the week value and compared it against the sum of the day values, they came up the same (which is what we want).


src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
...
 			}
 		}
 	}
 
-	const options = {
-		animation: {
-			duration: 0
+	// Customize the layout of the plot
+	const layout: any = {
+		autozise: true,
```
huss:
```suggestion
		autosize: true,
```
Wrong name/typo. autosize as per <https://plot.ly/javascript/reference/#layout-autosize>

src/client/app/types/react-plotlyjs-ts/index.d.ts
```diff
@@ -0,0 +1,30 @@
+/// <reference types="react" />
```
huss:
```suggestion
/// <reference types="react" />
// TODO <https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html> gives limited 
// circumstances where reference should be used. We need to verify this is a good use.
```
<https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html> gives limited circumstances where reference should be used. Wanted to check that this is the best way to do this.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
 					throw new Error('Unacceptable condition: readingsData.readings is undefined.');
 				}
 
-				const dataPoints: ChartPoint[] = _.values(readingsData.readings).map(compressedReadingToChartPoint);
+				// Create two arrays for the x and y values. Fill the array with the data from the compressed readings
+				const xData: string[] = [];
+				const yData: number[] = [];
+				const hoverText: string[] = [];
+				_.values(readingsData.readings).forEach(reading => {
```
huss:
```suggestion
				_.values(readingsData.readings).forEach(reading => {
// TODO The meter code explicitly creates the array but this does not. This seems cleaner (a little harder to 
// see data on debug). Whatever is chosen, it should probably be the same way.
```

src/client/app/containers/LineChartContainer.ts
```diff
@@ -31,19 +24,50 @@ function mapStateToProps(state: State) {
...
+					x: xData,
+					y: yData,
+					text: hoverText,
+					hoverinfo: 'text',
+					type: 'scatter',
+					mode: 'lines',
+					line: {
+						shape: 'spline',
+						width: 3
+						// smoothing: 1.3
```
huss:
```suggestion
```
Seems left over item so remove at this point.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
...
 		}
 	}
 
-	const options = {
-		animation: {
-			duration: 0
+	// Customize the layout of the plot
+	const layout: any = {
+		autozise: true,
+		// width: 700,
```
huss: Should these 3 lines be removed?

src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
...
+		// height: 700,
+		// title: 'First Test',
+		showlegend: true,
```
huss, line 116:
```suggestion
		showlegend: true,
// Just for the record, showlegend is listed as true by default in the Plotly documentation but it turns out 
// it does not show the first one unless you set it this way so this is needed.
```

src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
...
+			y: 1.1,
+			orientation: 'h',
 		},
-		elements: {
-			point: {
-				radius: 0
-			}
+		yaxis: {
+			title: 'kW',
+			showgrid: true,
```
huss:
```suggestion

```
It is okay to set showgrip: true but that is the default. Testing shows grids are present without this line (or in xaxis). Given we don't include all the defaults, it is suggest to remove this.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
...
 		}
 	};
 
-	const data: ChartData = { datasets };
-
-	const props: ChartComponentProps = {
-		data,
-		options,
-		redraw: true
+	// Assign all the paramaters required to create the Plotly object (data, layout, config) to the variable props, returned by mapStateToProps
```
huss:
```suggestion
	// Assign all the parameters required to create the Plotly object (data, layout, config) to the variable props, returned by mapStateToProps
```
paramaters -> paramaters

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
 
+import * as moment from 'moment';
+import { TimeInterval } from '../../../common/TimeInterval';
+
+import Button from 'reactstrap/lib/Button';
+import { FormEvent } from 'react';
+import { FormattedMessage } from 'react-intl';
+import { Dispatch, Thunk, ActionType } from '../types/redux/actions';
+
+import * as Plotly from 'plotly.js';
```
huss, line 21:
```suggestion
import * as Plotly from 'plotly.js';
// TODO lowercase plotly for the import in index.d.ts but uppercase here.
```

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
-		ChartToRender = MultiCompareChartContainer;
+export default class DashboardComponent extends React.Component<DashboardProps, {}> {
+	constructor(props: DashboardProps) {
+		super(props);
+		this.handleTimeIntervalChange = this.handleTimeIntervalChange.bind(this);
 	}
 
-	const optionsClassName = props.optionsVisibility ? 'col-2 d-none d-lg-block' : 'd-none';
-	const chartClassName = props.optionsVisibility ? 'col-12 col-lg-10' : 'col-12';
+	public render(){
```
huss:
```suggestion
	public render() {
```
space before {

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
 				</div>
 			</div>
-		</div>
-	);
+		);		
+	}
+
+	private handleTimeIntervalChange(mode: string) {
+		if (mode == "all"){
+			this.props.changeTimeInterval(TimeInterval.unbounded());
```
huss:
```suggestion
			this.props.changeTimeInterval(TimeInterval.unbounded());
```
3 tabs on end of line

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
 			</div>
-		</div>
-	);
+		);		
+	}
+
+	private handleTimeIntervalChange(mode: string) {
+		if (mode == "all"){
+			this.props.changeTimeInterval(TimeInterval.unbounded());			
+		}else{
```
huss:
```suggestion
		} else {
```
spacing on } and {

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
+			let sliderContainer: any = document.querySelector(".rangeslider-bg");
+			let sliderBox: any = document.querySelector(".rangeslider-slidebox");
+			let root: any = document.getElementById("root");	
+
+			if (sliderContainer && sliderBox && root){
```
huss:
```suggestion
			if (sliderContainer && sliderBox && root) {
```
space before {

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
+				let deltaSeconds: number = maxTimeStamp - minTimeStamp;
+				let secondsPerPixel: number = deltaSeconds / fullWidth;
+
+				// Get the new min and max times, in seconds, from the slider box
+				let newMinXTimestamp = Math.floor(minTimeStamp + (secondsPerPixel * sliderMinX));
+				let newMaxXTimestamp = Math.floor(minTimeStamp + (secondsPerPixel * sliderMaxX));
+
+				let timeInterval = new TimeInterval(moment(newMinXTimestamp), moment(newMaxXTimestamp));
+				this.props.changeTimeInterval(timeInterval);
+			}
```
huss:
```suggestion
			}
```
3 tabs at end of line

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
-		</div>
-	);
+		);		
+	}
+
+	private handleTimeIntervalChange(mode: string) {
+		if (mode == "all"){
+			this.props.changeTimeInterval(TimeInterval.unbounded());			
+		}else{
+			let sliderContainer: any = document.querySelector(".rangeslider-bg");
```
huss: This series of calculations seems to work fine and makes some sense. What I have been unable to find/figure out is where the rangeslider-bg & rangeslider-slidebox are set/documented. I did find <https://plot.ly/javascript/reference/#layout-xaxis-rangeslider> on Plotly but that does not match this. I suspect I'm missing something basic but wanted to be sure this is absolutely correct.

src/client/app/components/DashboardComponent.tsx
```diff
@@ -11,60 +11,132 @@ import MultiCompareChartContainer from '../containers/MultiCompareChartContainer
...
+						)}
+					</div>
 				</div>
 			</div>
-		</div>
-	);
+		);		
+	}
+
+	private handleTimeIntervalChange(mode: string) {
```
huss, line 107:
```suggestion
	private handleTimeIntervalChange(mode: string) {
// TODO Should mode be an enum since it has clearly defined values. Since the if/else uses it below 
// it could use a boolean but that implies there will only be two possible values forever 
// which seems likely to be the case.
```

src/client/app/containers/BarChartContainer.ts
```diff
@@ -23,18 +21,26 @@ function mapStateToProps(state: State) {
 			const readingsData = byMeterID[timeInterval.toString()][barDuration.toISOString()];
 			if (readingsData !== undefined && readingsData.readings !== undefined && !readingsData.isFetching) {
```
huss: This file does && readingsData.readings !== undefined whereas LineChartContainers.ts has a separate if with a thrown Error. Would be good to be consistent unless there is a reason not to be. Same for groups below.

src/client/app/containers/BarChartContainer.ts
```diff
@@ -23,18 +21,26 @@ function mapStateToProps(state: State) {
...
 				const label = state.meters.byMeterID[meterID].name;
-				const color = getGraphColor(label);
+				const xData: string[] = [];
```
huss, line 29: A number of comments were added to LineChartContainer.ts and it might be nice to have the mirror ones here.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -31,19 +24,50 @@ function mapStateToProps(state: State) {
...
 				}
 
-				const dataPoints: ChartPoint[] = _.values(readingsData.readings).map(compressedReadingToChartPoint);
+				// Create two arrays for the x and y values. Fill the array with the data from the compressed readings
+				const xData: string[] = [];
+				const yData: number[] = [];
+				const hoverText: string[] = [];
+				const readings = _.values(readingsData.readings);
+				readings.forEach(reading => {
+					const timeReading = moment(reading.startTimestamp);
```
huss, line 33: The code here differs slighly from BarCharContainer.ts. They do the same basic work so it would be nice if they did it in a similar way.

src/client/app/containers/BarChartContainer.ts
```diff
@@ -3,18 +3,16 @@
...
-
+import PlotlyChart, { IPlotlyChartProps } from 'react-plotlyjs-ts';
 
 function mapStateToProps(state: State) {
 	const timeInterval = state.graph.timeInterval;
 	const barDuration = state.graph.barDuration;
-	const datasets: ChartDataSets[] = [];
+	const datasets: any[] = [];
 
 	const labelsSet = new Set();
```
huss:
```suggestion
```
This appears left over from the old previous version so remove.

src/client/app/containers/BarChartContainer.ts
```diff
@@ -45,87 +51,77 @@ function mapStateToProps(state: State) {
...
 
-	const options = {
-		animation: {
-			duration: 0
+	// Customize the layout of the plot
+	const layout: any = {
+		barmode: 'group',
+		bargap: 0.2,
+		bargroupgap: 0.1,
+		autozise: true,
```
huss: autosize not autozise.

src/client/app/containers/BarChartContainer.ts
```diff
@@ -45,87 +51,77 @@ function mapStateToProps(state: State) {
...
 		}
 	};
 
-	const data: ChartData = { datasets, labels };
-
-	const props: ChartComponentProps =  {
-		data,
-		options,
-		redraw: true
+	// Assign all the paramaters required to create the Plotly object (data, layout, config) to the variable props, returned by mapStateToProps
```
huss: paramaters -> parameters

src/client/app/containers/BarChartContainer.ts
```diff
@@ -45,87 +51,77 @@ function mapStateToProps(state: State) {
...
-			}]
+		xaxis: {
+			showgrid: true,
+			gridcolor: '#ddd',
+			tickfont: {
+				size: 10
+			},
+			tickangle: -45,
+			autotick: true,
+			nticks: 10,
```
huss, line 118: <https://plot.ly/javascript/reference/#scattergl-marker-colorbar-tickmode> does not seem to make it clear if tickmode is auto (at least not from my very modest checking). nticks only has impact in this case. Should we set tickmode?

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -41,107 +28,76 @@ function mapStateToProps(state: State, ownProps: CompareChartContainerProps): Li
...
-				ticks
-			}]
+	const layout: any = {
+		title: `<b>${changeSummary}</b>`,
+		titlefont: {
+			size: 10,
+			color: colorize(entity.change)
 		},
+		hovermode:'closest',
+		autozise: true,
```
huss:
```suggestion
		autosize: true,
```
autosize

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -41,107 +28,76 @@ function mapStateToProps(state: State, ownProps: CompareChartContainerProps): Li
...
-					weight: 'bold'
-				},
-				display: true,
-				formatter: (value: number) => `${value} kW`
-			}
+		margin: {
+			t: 20,
+			b: 120,
+			l: 60,
+			r: 20,
```
huss: In the Plotly reference at <https://plot.ly/javascript/reference/#layout-margin>, I cannot find the r attribute for margin. Want to verify this is correct usage.

nicolebryant: I noted a similar issue but found that the attribute is seen in the search function of plotly's documentation although not found in the documentation itself. I have put in an issue on Plotly's Documentation GitHub regarding this.

src/client/app/containers/BarChartContainer.ts
huss:
```suggestion
		autosize: true,
```
Wrong name/typo.

src/client/app/containers/BarChartContainer.ts
huss:
```suggestion
	// Assign all the parameters required to create the Plotly object (data, layout, config) to the variable props returned by mapStateToProps
```

src/client/app/containers/LineChartContainer.ts
```diff
@@ -54,73 +78,75 @@ function mapStateToProps(state: State) {
...
 
-	const options = {
-		animation: {
-			duration: 0
+	// Customize the layout of the plot
+	const layout: any = {
+		autozise: true,
+		// width: 700,
+		// height: 700,
+		// title: 'First Test',
```
huss:
```suggestion
```
Seems left over from development so remove.

src/client/app/containers/CompareChartContainer.ts
huss: Let's assume it is okay to use, esp. since not causing any issues.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -31,19 +24,49 @@ function mapStateToProps(state: State) {
...
+					mode: 'lines',
+					line: {
+						shape: 'spline',
+						width: 3
+					},
+					marker: {color: getGraphColor(label)}
 				});
 			}
 		}
 	}
```
huss, line 68:
```suggestion

// TODO The meters and groups code is very similar and maybe it should be refactored out to create a function to do 
// both. This would mean future changes would automatically happen to both.
```

src/client/app/components/DashboardComponent.tsx
huss: I've decided to let this go and close this out.

src/client/app/containers/BarChartContainer.ts
huss: Code modified to do this.

src/client/app/containers/BarChartContainer.ts
huss: Upon rereading, it states for the other two cases what is the default so it seems it is likely okay in this case. Since working, I am going to close this.

src/client/app/containers/BarChartContainer.ts
huss: Addressed in commit 62e075f

src/client/app/containers/LineChartContainer.ts
huss: Addressed in commit 62e075f

huss:
The code in this PR does what it was designed to do. However, there are some other features we want before we deploy the new Plotly graphics system for general use. See issues #456, #458, #459 & #460. This is just to document this fact.

huss:
This is to document that I verified all graphic data (line with day and hourly, all bar & all compare) against the raw meter data. They were in agreement but some edge issues were found that will be new issues and outside the work of this PR. This should be a strong test that fast-pt and the new graphics in Plotly are doing what they are supposed to do.

huss:
The remaining TS lint error:
TypeError: ts.unescapeIdentifier is not a function
seems likely to be related to the versions of tslint/tsutil we are using per <https://github.com/palantir/tslint/issues/4266>. We want to migrate to ESLint (#463) so hope it goes away then. This is not causing an error in the Travis build so not resolving now.

nicolebryant:
This PR looks good to merge based on the discussions of code reviews and testing with @huss as well as the updates made to the code. There are still some minor changes that need to be made later that aren't breaking or causing the Travis build to fail, all listed in the comments on this PR.

huss (approved):
This PR looks good to merge based on the discussions of code reviews and testing with @huss as well as the updates made to the code. There are still some minor changes that need to be made later that aren't breaking or causing the Travis build to fail, all listed in the comments on this PR

PR #443, opened 2020-01-10
updated colors to make them more readable

mm413:
I changed the table in getGraphColor.ts in order to have the colors stand out more from one another.


huss:
Thanks to @mm413 for working on this. I have a few thoughts:

- The branch this is based on is older Plotly work. The current version that should get merged into OED is CorinaSterbet:plotly-and-upstream-dev which is in PR #436. It would be better for the project and easier to integrate if the code was developed from that repo. I can also more readily check the colors graphically once that happens.
- There are some Travis issues involving correct formatting. If you feel comfortable then please address them.
- Any changes to the proposed Plotly code will be held until that branch is merged into development. At that time a lot of changes are expected to be made and extensive testing done. I just want people to know why there will be a delay in approving/merging.
- I think changing the colors is what was asked. However, the larger issue is having lines with similar/same colors. The hash-based method used (not part of this PR) seems good but other techniques might work better where the color of a meter/group is kept the same across graphs but colors are not repeated until all colors are used. There is also the possibilities of putting meters and groups in different parts of the color spectrum so it is easier to tell which is being graphed. I'm happy to work with someone on these ideas.

huss:
This will address issue #495.

huss:
This work and the better selection of color choices are being worked. The PR from that work should supersede this PR.

Issue #444, opened 2020-01-15
check on null/undefined typescript warning


huss:
RouteComponent.tsx had the line:
const info: string = infoObj.toString();
It was changed to:
const info: string = infoObj!.toString();
because travis was reporting:
         ERROR in [at-loader] ./src/client/app/components/RouteComponent.tsx:108:27 
    		TS2533: Object is possibly 'null' or 'undefined'.
This eliminated the error so the builds would pass. However, this basically removes the complaint but does not directly address a possible underlying issue. The code should be check and changed as appropriate.

This fix was pushed in quickly to get rid of the travis errors so all pull commits will not fail but may not be the best fix.

huss:
This was fixed and a TODO left to see if something else needs to be done. The main issue is fixed.

PR #445, opened 2020-01-15
fix travis issue for null/undefined

huss:
This change will silence the error so travis builds pass. The original merges did not have any errors so appears to be some sort of new issue that catches it. I have opened issue #444 to see if a better fix is needed but this is being done quickley this way so others will not get failures.


huss:
This is a simple change to fix the build issue. As such, I am approving without review so others don't have issues.

PR #447, opened 2020-01-18
Add validateData & validateDataTests

truonghh99:
Use validateData to detect out-of-bound values in updateMeters.

### DIFF
--- a/src/server/services/updateMeters.js
+++ b/src/server/services/updateMeters.js
@@ -4,7 +4,7 @@
 
 const _ = require('lodash');
 const Reading = require('../models/Reading');
-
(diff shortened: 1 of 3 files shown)


huss:
A standard pipeline for accepting and checking all input data is being created by @truonghh99 at this time. It will either be an update to the PR or a new one and include these features. Thus, this PR is on hold pending those changes.

huss:
I think PR #482 includes this work on replaces it. Should this PR be closed (question for @truonghh99)?

truonghh99:
Agree. This work has been replaced by validateReadings.js in PR #482

PR #452, opened 2020-02-25
Add checkOnlineStatus scripts

Jerry2001:


src/server/services/checkOnlineStatus.js
```diff
@@ -0,0 +1,23 @@
...
+//This script may encounter redirecting problem due to the web policy
+
+const { log } = require('../log');
+var script = document.createElement('script');
+script.src = 'https://ajax.googleapis.com/ajax/libs/jquery/3.4.1/jquery.min.js';
+document.body.appendChild(script);	
+
+$.ajax({
+	type: "HEAD",
+	url: 'YOUR SITE URL',
```
huss, line 19: I think there should be a comment here to clearly indicate that this line needs to be edited and how. Also, put a comment with the top ones to say to look for where the edit needs to be made. Same change for the PHP version.

huss (changes requested):
I'm not sure why there is a Travis build failure. I'm not sure it relates to these changes but it should be cleared up before this is merged.
Also, you can update a PR so it generally is not necessary to create a new one each time you want to make changes.

src/server/services/checkOnlineStatus.js
Jerry2001: I added comments

PR #453, opened 2020-03-10
Messages sent to log.txt, not console

corralem:
Wanted to log messages to a permanent file instead of console.


huss (approved):
Seems fine at this point. Basic change and tested so approving.

PR #482, opened 2020-05-02
Add pipeline to insert meter data to database

### LINKED ISSUE #196: sanity checking meter data

### LINKED ISSUE #467: standardize API and pipeline for acquiring meter data
OED is in the process of allowing more ways to acquire data. To allow for standardized error checking, to clean up the AP and code, and to make future additions much easier, we need to have a single pipeline for accepting data. This will impact #370 and #163 as they will  use this new interface. We plan to have this done before milestone 1.0 but it is assigned to that until we are sure when it will be ready.

truonghh99:
Add a pipeline to read, convert, validate, and insert data from Obvius, Metasys, and Mamac meters to database. 

Attempt to solve issues and fixes #467 and fixes #196.

Note: this pull request also contains some modifications in the database settings to accept Obvius data. These changes can be removed if NoraCodes:obvius-old is successfully merged to development.


src/server/models/Configfile.js
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Returns a promise to remove all entries in the table.
+	 * @return {Promise.<>}
+	 */
+	static purgeAll(conn = getDB) {
+		return conn().none(sqlFile('obvius/purge_configs.sql'));
+	}
+
+	static mapRow(row) {
```
Jerry2001, line 47: This file doesn't have comments in several functions

src/server/models/Configfile.js
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	/**
+	 * Returns a promise to remove all entries in the table.
+	 * @return {Promise.<>}
+	 */
+	static purgeAll(conn = getDB) {
+		return conn().none(sqlFile('obvius/purge_configs.sql'));
+	}
+
```
Jerry2001: This file doesn't have comments in several functions

src/server/sql/obvius/insert_new_config.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
Jerry2001, line 4: There should be a comment describing what this file does.

src/server/sql/obvius/create_config_table.sql
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
Jerry2001, line 4: There should be a comment describing what this file does.

src/server/services/pipeline-in-progress/insertObviusData.js
```diff
@@ -0,0 +1,41 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const moment = require('moment');
+const Meter = require('../../models/Meter');
+const Reading = require('../../models/Reading');
+const { log } = require('../../log');
+const demuxCsvWithSingleColumnTimestamps = require('./csvDemux');
+const loadArrayInput = require('./loadArrayInput');
+
```
Jerry2001, line 11: Add comment to this function and file

truonghh99: Thanks for your comments! However, all the obvius-related files are merged directly from Nora's old-obvius branch for testing purpose only and will be removed soon. The complete version of them can be found in obvius2020.

Jerry2001: Thank you for the information

truonghh99:
This pull request is now in sync with obvius2020. TODO:
- Add tests for Obvius data
- Generalize handleCummulative

src/server/services/pipeline-in-progress/insertMamacData.js
```diff
@@ -0,0 +1,38 @@
...
+							return [readRate, startTimestamp, endTimestamp];
+						},
+						readAsStream = true,
+						isCummulative = false,
+						readingRepetition = 1,
+						conditionSet = undefined,
+						conn = conn);
+}
+
+module.exports = insertMamacData;
```
ak476519, line 39: I think this file is going to eventually replace the loadMamacReadingsFromCsvFile.js in the src/server/services/ directory after this PR is merged since they serve the same function. 

And I assume other files in the pipeline-in-progress folder works in the same way. If that's the case, I believe it's important to remove the test files dependent on the previous version as well(when this PR gets accepted).

src/server/services/pipeline-in-progress/handleCummulative.js
```diff
@@ -0,0 +1,52 @@
...
+ *  row 1: reading #0 + #1
+ * 	row 2: reading #0
+ * => reading #1 = row 0 - row 1
+ *    reading #2 = row 1 - row 2
+ *    reading #0 is cummulative value from unknown readings that are may or may not inserted before
+ * @param {2d array} rows
+ * @param readingRepetition value is 1 if reading is not duplicated. 2 if repeated twice and so on (E-mon D-mon meters)
+ */
+
+function handleCummulative(rows, readingRepetition) {
```
ak476519: small typo

ak476519 (approved):
Looks great! With a few fixes mentioned in the comments, I believe this is good to be merged XD
*And I believe we'll need to expand the usage instructions for commands introduced by this PR.

src/server/services/pipeline-in-progress/convertToReadings.js
```diff
@@ -0,0 +1,29 @@
...
+const { validateReadings } = require('./validateReadings');
+const moment = require('moment');
+
+/**
+ * Convert an matrix of number values to an array of Readings
+ * @param {2d array} toConvert a number array in which each row represents a Reading values (reading, startTime, endTime)
+ * @param {int} meterID ID of the meter which all the Reading values belong to
+ * @param {dict} conditionSet used to validate readings (minVal, maxVal, minDate, maxDate, interval, maxError)
+ */
+
```
huss: I believe other OED code does not leave a space between the comment and the function code.

src/server/services/pipeline-in-progress/convertToReadings.js
```diff
@@ -0,0 +1,29 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const Reading = require('../../models/Reading');
+const { log } = require('../../log');
+const { validateReadings } = require('./validateReadings');
+const moment = require('moment');
+
+/**
+ * Convert an matrix of number values to an array of Readings
```
huss, line 13: I think this comment should state that it returns null if any reading is not validated.

src/server/services/pipeline-in-progress/convertToReadings.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {2d array} toConvert a number array in which each row represents a Reading values (reading, startTime, endTime)
+ * @param {int} meterID ID of the meter which all the Reading values belong to
+ * @param {dict} conditionSet used to validate readings (minVal, maxVal, minDate, maxDate, interval, maxError)
+ */
+
+function convertToReadings(toConvert, meterID, conditionSet) {
+	readings = toConvert.map(row => new Reading(meterID, row[0], row[1], row[2]));
+	if (conditionSet !== undefined && !validateReadings(readings, conditionSet)) {
+		log.warn(`ERROR WHEN VALIDATING READINGS FROM METER ${ipAddress}`);
```
huss: Would it make sense to add to the warning message in the log that all the readings were not added to OED? Also, is this an error since values are dropped?

src/server/services/pipeline-in-progress/handleCummulative.js
```diff
@@ -0,0 +1,52 @@
...
+ * 	row 0: reading #0 + #1 + #2
+ *  row 1: reading #0 + #1
+ * 	row 2: reading #0
+ * => reading #1 = row 0 - row 1
+ *    reading #2 = row 1 - row 2
+ *    reading #0 is cummulative value from unknown readings that are may or may not inserted before
+ * @param {2d array} rows
+ * @param readingRepetition value is 1 if reading is not duplicated. 2 if repeated twice and so on (E-mon D-mon meters)
+ */
+
```
huss: Extra blank line.

src/server/services/pipeline-in-progress/handleCumulative.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Handle cumulative data, assume that last row is the first reading (skip this row).
+ * @Example
+ * 	row 0: reading #0 + #1 + #2
+ *  row 1: reading #0 + #1
+ * 	row 2: reading #0
+ * => reading #1 = row 0 - row 1
+ *    reading #2 = row 1 - row 2
+ *    reading #0 is cummulative value from unknown readings that are may or may not inserted before
```
huss: cummulative->cumulative & that are may not inserted before->that may or may not have been inserted before (is this what you meant?)

src/server/services/pipeline-in-progress/insertMamacData.js
```diff
@@ -0,0 +1,38 @@
...
+						meterID = meter.id,
+						mapRowToModel = row => {
+							const readRate = row[0];
+							// Mamac timestamps look like 11:00:00 7/31/16
+							const endTimestamp = moment(row[1], 'HH:mm:ss MM/DD/YYYY');
+							const startTimestamp = moment(endTimestamp).subtract(readingDuration);
+							return [readRate, startTimestamp, endTimestamp];
+						},
+						readAsStream = true,
+						isCummulative = false,
```
huss: cummulative->cumulative

src/server/services/pipeline-in-progress/insertMetasysData.js
```diff
@@ -0,0 +1,42 @@
...
+						mapRowToModel = row => {
+							readRate = row[3].replace(' kW', '');
+							readRate = Math.round(parseFloat(readRate));
+							// Metasys timestamps look like 11:00:00 7/31/16
+							endTimestamp = moment(row[0], 'MM/DD/YY HH:mm');
+							startTimestamp = moment(endTimestamp).subtract(readingInterval, 'minutes');
+							return [readRate, startTimestamp, endTimestamp];
+						},
+						readAsStream = false,
+						isCummulative = cumulativeIndicator,
```
huss: cummulative->cumulative

src/server/services/pipeline-in-progress/insertObviusData.js
```diff
@@ -0,0 +1,41 @@
...
+		return loadArrayInput(dataRows = data[i],
+							meterID = meter.id,
+							mapRowToModel = rawReading => {
+								const reading = rawReading[1];
+								const startTimestamp = moment(rawReading[0], 'YYYY-MM-DD HH:mm:ss');
+								const endTimestamp = startTimestamp.clone();
+								endTimestamp.add(moment.duration(60, 'minutes'));
+								return [reading, startTimestamp, endTimestamp];
+							},
+							isCummulative = false,
```
huss: cummulative->cumulative

src/server/services/pipeline-in-progress/loadCsvInput.js
```diff
@@ -0,0 +1,39 @@
...
+const loadCsvStream = require('./loadCsvStream');
+const { log } = require('../../log');
+
+/**
+ * Read a CSV file and select needed column to return an array of reading value and reading time
+ * @param {string} filePath
+ * @param {string} meterID
+ * @param {function} mapRowToModel a customized function that map needed values from each row to the Reading model
+ * @param {boolean} readAsStream true if prefer to read file as CSV stream
+ * @param {boolean} isCummulative true if the given data is cummulative
```
huss: cummulative->cumulative & other 3 places used.

src/server/services/pipeline-in-progress/handleCumulative.js
```diff
@@ -0,0 +1,52 @@
...
+	// Initialize timestamps and other variables
+	let startTimestamp = moment(0);
+	let endTimestamp = moment(0);
+	let meterReading = 0;
+	let meterReading1 = 0;
+	let meterReading2 = 0;
+
+	for (let index = 0; index < rows.length; ++index) {
+		// To read data where same reading is repeated. Like E-mon D-mon meters
+		if (index !== 0 && Math.abs((index - readingRepetition) % readingRepetition) === 0) {
```
huss: While this logic works and is similar to code already in OED, I wonder if it could be simpler. First, I'm unclear on why the Math.abs is needed if you are checking for equality with 0. More generally, if the loop above was for (let index = readingRepetition, ....) then it would automatically skip 0 (so that check can be removed) and all values until the first row that meets the % requirement and you would never be taking the mod of the negative number so the Math.abs is certianly not needed. Would these changes make sense?

src/server/services/pipeline-in-progress/loadArrayInput.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {string} meterID
+ * @param {function} mapRowToModel a customized function that map needed values from each row to the Reading model
+ * @param {boolean} isCumulative true if the given data is Cumulative
+ * @param {array} conditionSet used to validate readings (minVal, maxVal, minDate, maxDate, interval, maxError)
+ */
+
+async function loadArrayInput(dataRows, meterID, mapRowToModel, isCumulative, readingRepetition, conditionSet, conn) {
+	readingsArray = dataRows.map(mapRowToModel);
+	if (isCumulative) {
+		readingsArray = handleCumulative(readingsArray, readingRepetition);
```
huss: I'm concerned about bad data and [handleCumulative](https://github.com/truonghh99/OED/blob/pipelinedInsertionToDatabase/src/server/services/pipeline-in-progress/handleCumulative.js). In handleCumulative, if a value is negative then it is assumed that there was a wrap in meter values (such as occurs at midnight). However, if a the conditionSet check below does not allow negative values but one is present then handleCumulative is going to assume a wrap has occurred. The fix may not be trivial. You cannot do the checks first since the values are not yet correct. Thus, it seems to be an issue that may need to be addressed in handleCumulative (see comment there). However, that is mixing the functionality that is now separate so I'm not yet thrilled with that. Other ideas?

src/server/services/pipeline-in-progress/handleCumulative.js
```diff
@@ -0,0 +1,52 @@
...
+			// set start_timestamp and end_timestamp
+			startTimestamp = moment(rows[index][1], 'MM/DD/YY HH:mm');
+			endTimestamp = moment(rows[index - readingRepetition][2], 'MM/DD/YY HH:mm');
+			// meterReading
+			meterReading1 = rows[index - readingRepetition][0];
+			meterReading2 = rows[index][0];
+			meterReading = meterReading1 - meterReading2;
+			// To handle cumulative readings that resets at midnight
+			if (meterReading < 0) {
```
huss: In [loadArrayInput](https://github.com/truonghh99/OED/blob/pipelinedInsertionToDatabase/src/server/services/pipeline-in-progress/loadArrayInput.js), this is called before the data is checked for issues (see comment there). This means that a bad value that is negative will be assumed to be a reset at midnight. It also means that any negative data that is valid will cause an issue. One possible fix would be to check if the reading(s) are negative and then not do this process. This seems relatively safe since we don't normally accept negative readings, esp. if cumulative. This could also be an option via a parameter. We could also enhance this by checking if the reading value is close to midnight. Another option (maybe with the previous one(s)) is to add a parameter to allow/skip the reset as we already have something for readingsRepetition so the user needs to know about their meter data. I'm open to ideas here.

Another potential issue is that a weird value could be missed/located at the wrong row due to subtracting values but I don't think this is likely and it is likely to show up for some row(s) in the file. I think we can forget about this.

src/server/services/pipeline-in-progress/loadCsvInput.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {array} conditionSet used to validate readings (minVal, maxVal, minDate, maxDate, threshold, maxError)
+ * @param {array} conn connection to database
+ */
+
+async function loadCsvInput(filePath, meterID, mapRowToModel, readAsStream, isCummulative, readingRepetition, conditionSet, conn) {
+	try {
+		if (readAsStream) {
+			const stream = fs.createReadStream(filePath);
+			return loadCsvStream(stream, meterID, mapRowToModel, conditionSet, conn);
```
huss, line 29: This is a comment for the record and not necessarily expected for the PR. The stream version cannot deal with cumulative or repeated data as the other input can. I wonder if we want to make them consistent?

src/server/services/pipeline-in-progress/loadCsvInput.js
```diff
@@ -0,0 +1,39 @@
...
+		} else {
+			const dataRows = await readCSV(filePath);
+			return loadArrayInput(dataRows, meterID, mapRowToModel, isCummulative, readingRepetition, conditionSet, conn);
+		}
+	} catch (err) {
+		log.error(`Error updating meter ${meterID}: ${err}`, err);
```
huss: Is there any way to give more information in the message so it is easier to trace down? I don't know if the filePath is useful or not. Maybe some new value to provide?

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+
+const Reading = require('../../models/Reading');
+const { log } = require('../../log');
+
+/**
+ * Validate an array of Readings value according to certain criteria
+ * @param {array of Readings} arrayToValidate
+ * @param {dict} conditionSet used to validate readings (minVal, maxVal, minDate, maxDate, interval, maxError)
+ */
+
```
huss: Extra blank line.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {number} maxError maximum number of errors to be reported, ignore the rest
+ */
+function checkValue(arrayToValidate, minVal, maxVal, maxError) {
+	validValues = true;
+	for (reading of arrayToValidate) {
+		if (maxError <= 0) {
+			break;
+		}
+		if (reading.reading < minVal) {
+			log.warn(`ERROR WHEN CHECKING READING VALUE FROM METER ${reading.meterID}: ${reading.reading} is smaller than lower bound ${minVal}`);
```
huss: The message in checkDate does meterID: TIME timestamp. Should this one be meterID: Value value to be parallel and make the value clear? Same below for maxVal.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+			validDates = false;
+		}
+	}
+	return validDates;
+}
+
+/**
+ * Check and report any out-of-bound reading value. Can be ignored by passing MIN_VALUE & MAX_VALUE
+ * @param {array of Readings} arrayToValidate
+ * @param {number} minVal maximum acceptable reading value
```
huss: Might want to add the minVal and maxVal are exclusive (so not included in errors) just to be very clear. Same would apply to date.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}
+	return validValues;
+}
+
+/**
+ * Check and report unequal intervals. Can be ignore by passing null interval
+ * @param {array of Readings} arrayToValidate
+ * @param {number} interval expected gap between 2 consecutive reading times in seconds
+ * @param {number} maxError maximum number of errors to be reported. Ignore the rest
```
huss: maxError is not actually a parameter. It stops after the first detected issue.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+			validValues = false;
+		}
+	}
+	return validValues;
+}
+
+/**
+ * Check and report unequal intervals. Can be ignore by passing null interval
+ * @param {array of Readings} arrayToValidate
+ * @param {number} interval expected gap between 2 consecutive reading times in seconds
```
huss: Should interval be threshold? Is description correct?

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {number} maxError maximum number of errors to be reported. Ignore the rest
+ */
+function checkIntervals(arrayToValidate, threshold) {
+	if (threshold === null) {
+		return true;
+	}
+	interval = arrayToValidate[1].startTimestamp - arrayToValidate[0].endTimestamp;
```
huss: This is probably fine but should you use .diff as you did in the loop?

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+	lastTime = arrayToValidate[1].endTimestamp;
+	for (reading of arrayToValidate) {
+		if (reading === arrayToValidate[0]) {
```
huss, line 92: I think a few comments on blocks of code would help.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+			continue;
+		}
+		currGap = reading.startTimestamp.diff(lastTime, 'seconds');
+		if (currGap < interval - threshold || currGap > interval + threshold) {
+			log.warn(`UNEQUAL INTERVAL IS DETECTED FROM METER ${reading.meterID}`);
```
huss: I think adding time stamps and, maybe, the reading value would be useful in this message.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+		return true;
+	}
+	interval = arrayToValidate[1].startTimestamp - arrayToValidate[0].endTimestamp;
+	lastTime = arrayToValidate[1].endTimestamp;
+	for (reading of arrayToValidate) {
+		if (reading === arrayToValidate[0]) {
+			continue;
+		}
+		currGap = reading.startTimestamp.diff(lastTime, 'seconds');
+		if (currGap < interval - threshold || currGap > interval + threshold) {
```
huss: This seems okay but I think it might be clearer with Math.abs(currGap - interval) > threshold.

src/server/services/pipeline-in-progress/loadArrayInput.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+const Reading = require('../../models/Reading');
+const convertToReadings = require('./convertToReadings');
+const { log } = require('../../log');
+const handleCumulative = require('./handleCumulative');
+
+/**
+ * Select and process needed values from a matrix and return an array of reading value and reading time
+ * @param {2d array} dataRows
```
huss: It seems you must have a valid type inside the {} and starting with a number causes issues (look at the popup hover documentation in VSC to see). I think this probably should be:
@param {object[[]]} dataRows where each row is defined by mapRowToModel function
where I added a comment on what is in the 2D array. You don't see the type in JS popup documentation but it is good to have it there.

src/server/services/pipeline-in-progress/convertToReadings.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+const Reading = require('../../models/Reading');
+const { log } = require('../../log');
+const { validateReadings } = require('./validateReadings');
+const moment = require('moment');
+
+/**
+ * Convert an matrix of number values to an array of Readings
+ * @param {2d array} toConvert a number array in which each row represents a Reading values (reading, startTime, endTime)
```
huss: see loadArrayInput.js for issue with {2d array}

src/server/services/pipeline-in-progress/handleCumulative.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Handle cumulative data, assume that last row is the first reading (skip this row).
+ * @Example
+ * 	row 0: reading #0 + #1 + #2
+ *  row 1: reading #0 + #1
+ * 	row 2: reading #0
+ * => reading #1 = row 0 - row 1
+ *    reading #2 = row 1 - row 2
+ *    reading #0 is cummulative value from unknown readings that are may or may not inserted before
+ * @param {2d array} rows
```
huss: see loadArrayInput.js for issue with {2d array}

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const Reading = require('../../models/Reading');
+const { log } = require('../../log');
+
+/**
+ * Validate an array of Readings value according to certain criteria
+ * @param {array of Readings} arrayToValidate
```
huss: see loadArrayInput.js for issue with {2d array} and this similar.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+function validateReadings(arrayToValidate, conditionSet) {
+	validDates = checkDate(arrayToValidate, conditionSet['minDate'], conditionSet['maxDate'], conditionSet['maxError'] / 2);
+	validValues = checkValue(arrayToValidate, conditionSet['minVal'], conditionSet['maxVal'], conditionSet['maxError'] / 2);
+	validIntervals = checkIntervals(arrayToValidate, conditionSet['threshold']);
+	return validDates && validValues && validIntervals;
+}
+
+/**
+ * Check and report any out-of-bound date. Can be ignored by passing null minDate and maxDate
+ * @param {array of Readings} arrayToValidate
```
huss: see loadArrayInput.js for issue with {2d array} and this similar.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -0,0 +1,110 @@
...
+			--maxError;
+			validDates = false;
+		}
+	}
+	return validDates;
+}
+
+/**
+ * Check and report any out-of-bound reading value. Can be ignored by passing MIN_VALUE & MAX_VALUE
+ * @param {array of Readings} arrayToValidate
```
huss: see loadArrayInput.js for issue with {2d array} and this similar.

src/server/test/db/validateReadingsTests.js
```diff
@@ -0,0 +1,67 @@
...
+			} else {
+				expect(results[i]).to.equal(false);
+			}
+		}
+	});
+	mocha.it('detects out-of-bound data', async () => {
+		const minVal = 10;
+		const maxVal = 20;
+		// testing sequence: lower, in-bound, higher, in-bound, lower, in-bound, higher
+		let readingVals = [9, 10, 21, 20, -Number.MAX_VALUE, 15, Number.MAX_VALUE];
```
huss: Maybe use Number.MIN_VALUE instead of -Number.MAX_VALUE. It is clearer and safer (if a language is not symetric in values).

src/server/test/db/validateReadingsTests.js
```diff
@@ -0,0 +1,67 @@
...
+		)).map(reading => checkValue([reading], minVal, maxVal, Number.MAX_VALUE));
+		for (let i = 0; i < 7; ++i) {
+			if (i % 2 === 0) {
+				expect(results[i]).to.equal(false);
+			} else {
+				expect(results[i]).to.equal(true);
+			}
+		}
+	});
+	mocha.it('detects unequal intervals', async () => {
```
huss, line 53: These tests work fine but I'm wondering about changing the interval to make it a tighter test. In the unequal case it could be set to a non-zero value such as 120 which is exactly equal to the change in gap (59 seconds for the first two and 2:59/179 seconds for the second two). In the equal case, the first two have a gap of 0 while the second two are 29 seconds so the interval could be 29 seconds. I tested these changes work. I even tested setting the second to 28 seconds and it then failed as expected.

src/server/test/db/validateReadingsTests.js
```diff
@@ -0,0 +1,67 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+const { mocha, expect } = require('../common');
+const { checkDate, checkValue, checkIntervals} = require('../../services/pipeline-in-progress/validateReadings');
+const Reading = require('../../models/Reading');
+const moment = require('moment');
+
+mocha.describe('PIPELINE: Validate Readings', () => {
```
huss, line 11: Should there be one test of the validateReadings function to make sure the overall one returns the correct value? (Seems quite likely it works fine for such a simple function but testing is good.)

huss (changes requested):
I want to note that the delay in getting this PR approved/merged relates to project considerations and nothing about the work done or the efforts of @truonghh99.
This PR is very good work and does not have any issues while running. I note the following that are global thoughts or ones that might be addressed outside this PR:
- We need to merge in the huss branch (already has PR) to fix a number of issues. I want to record that I already did this to do more testing. There were 4 files with merge conflicts and I have a record of what to do about them. I then ran all the tests and everything passed.
- I tested updating and viewing real Mamac meters. It works as expected.
- There are old files in src/server/services/ that relate to new files in pipeline. Some are: addMamacMeters.js, loadFromCsvStream.js, loadMamacReadingsFromCsvFile.js, readCSV.js, readMamacData.js, readMamacMeters.js, readMetasysData.js, updateMeters.js & also consider testing files (insertMamacReadingsFromCSVTests, metasysTests, etc.). We need to decide what should be done with these seemingly related versions. This is an expansion of comment by @ak476519. Issue #510 addresses this.
- The conditionSet is undefined in insertMamacData, insertMetasysData & insertObviusData. We need to decide what to put in for these so we actually do testing on inputted data. This may require new admin input to decide.  Issue #511 addresses this.
- It seems insertMetasysData is only used in testing now. Should it be used in the actual OED code? Issue #512 addresses this.
- I want to amplify @truonghh99 comment that handleCumulative drops the logical first value. We need to think about how we deal with sorted, reverse sorted & cumulative to get what we want. Dropping a value (esp. without warning) seems undesirable. Issue #513 addresses this.
- We can improve the tests in insertMamacDataTests by verifying the values. See Metasys test on doing one value. A better alternative would be to test a middle value or all of them by making the values regular. The same applies to insertMetasysDataTests. Issue #514 addresses this.
- insertMetasysDataTests has issues of wrong gap on 9/26/16 at 2:00 and bad date right before this. Should this be two tests? This one is being let go.
- There are likely changes coming to the obvius version of insertObviusData.js by @lindavin. These include logging (maybe changing) how meter creation is done & not assuming readings are 60 minutes. We need to coordinate the obvius changes with this code.
- What happens if the meter does not exist when OED tries to create it in insertMetasysData.js? I'm concerned it will cause an issue where the code encounters but does not deal with an error. I know there are places we don't check for this, however, here this is typically data coming from another source where we are using the filename to get the meter name. Unlike the internal uses where we control the name, this seems to be a place where a mistake can really happen. I do know the original code did not check for this but I'm seeing it now. Issue #515 addresses this.
- The stream version (see loadCsvInput.js) cannot deal with cumulative or repeated data as the other input can. I wonder if we want to make them consistent? Issues #516 addresses this.

src/server/services/pipeline-in-progress/handleCumulative.js
truonghh99: This is a really good point. I'm just not sure I fully understand your solution. Can you please clarify what you mean by "check if the reading(s) are negative and then not do this process"? Are you referring to the actual reading values (meterReading1, meterReading2)? I may be missing something since I don't really see how this would solve the problem (meterReading1 and meterReading2 can both be non-negative/valid to generate a negative/invalid difference).

I do like the idea of having an "allow reset" parameter though, I think we can also let the user define the time period to accept "reset" reading value (e.g. within 1 hour after midnight). If we have that information, we can simply check for the timestamp every time we get a negative value and immediately accept or reject it accordingly. My only concern is whether we're asking too much from the user, but I think having a nice "default" setting may help

huss: I was referring to meterReadings1 or 2. My thinking was that if either is negative then we don't do the wrap (the if that resets the value) since that can take a negative reading value and reset it to positive. The only negative reading I know of is for temperature and that is not done cumulatively so it should not be in this code. Thus, something should be wrong with a negative value so trying to take differences is problematic. The concern is that a negative reading might be covered up since the reset could make it positive. I agree with you that two positive values can lead to a negative difference but that is actually an error in cumulative data so it should be left and detected later. Given this, I propose that the check be changed to
if (meterReading1 < 0 || meterReading < 0) {
This means a negative reading is passed along and not reset. I don't think we need to check 
meterReading2 (assuming my limited analysis is right) since it will just make it more positive (the difference) and it would be detected on the next loop cycle when it becomes meterReading1.

I'm not sure the perfect default value for how close to midnight exists. I think using one hour seems fine. This means we are much less likely to reset bad values since they will not be near midnight. I agree this asks something of the user but the default means it will likely work without them doing anything. I think this would be a good change.

huss:
All the updated changes seem good. I have locally tested the code and it passes. The CodeQL failures relate to Obvius and are not part of this PR. They are being addressed separately. Once the final item that is under discussion is resolved this PR is ready to be merged. I have opened issues for remaining items listed that are outside this PR.

src/server/services/pipeline-in-progress/handleCumulative.js
truonghh99: I think I got your idea now, thank you for the clarification! Just to confirm, I'm understanding that there are 4 different cases that need to be handled differently in handleCumulative:

1. Positive difference between 2 positive readings: accept
2. Positive difference between 2 negative readings: reject immediately
3. Negative difference between 2 positive readings, happens at midnight: convert difference to meterReading1, accept as a reset
4. Negative difference between 2 positive readings, happens anytime else: reject eventually in validateReadings

For case 2, my solution is to reject the whole set of readings (since an invalid cumulative value would shift all the others, accepting partially may be messy). The condition is meterReading1 < 0

For case 3, I think we can check for the date difference between the start and endTimeStamp to see if the value has been read overnight. This way, there's no need for user input nor a default "midnight threshold." The condition is meterReading < 0 && endTimestamp.isAfter(startTimestamp, 'date')

Am I missing anything? I've added code to implement these new catches and a little addition to handleCumulativeTests, please let me know if you notice any problem.

Thank you!

src/server/services/pipeline-in-progress/insertMamacData.js
huss: Yes & consistent with my comments. Issue #510 lists this issue to address outside this PR.

huss (approved):

huss:
The Travis failure is an ongoing issue but local testing showed all is fine. This is good work that is ready for deployment.

Issue #458, opened 2020-03-26
Bar stacking not yet implemented with plotly


huss:
The change to plotly means this feature needs to be fixed so it works.

PR #485, opened 2020-05-12
Fix bar stacking

nicolebryant:
Bar chart page will now implement bar stacking option when toggled.
Fixes issue #458 

A helpful page for adding different features for bar charts from plotly can be found [here](https://plotly.com/javascript/bar-charts/)


huss (approved):
Tested locally and runs fine. Simple code change and seems fine.

Issue #481, opened 2020-04-21
when to run npm install


huss:
About a year ago, installOED.sh was modified so it does no do an npm install if node_modules exist. The idea was to make it idempotent/reproducible. I see an advantage of not allowing package changes unless wanted. However, if package-lock.json changes then the install script does not change the node modules. This can cause developers or anywhere an upgrade occurs to not align the node modules with the updated package-lock.json that might be needed for new packages or updates (for security, etc.). I could see some possible changes to address this an am seeking input on what to do, including other ideas. First, npm install could be run if node_modules is older than package-lock.json and maybe package.json so the update occurs if changes were made. Second, we could add flags to force or stop the npm install so you could make it happen or you could stop it from happening even if packages were changed. Third, we could not change the script but make people remove or touch node_modules to cause an npm install to occur (or via flags on the script). Fourth, other ideas? I don't really like number 3 since it is easy to miss important upgrades. I think doing both number 1 and 2 should not take much work and allows people to override the default behavior. If we do this then we need to decide on the default behavior. I'm leaning toward having npm install happen if node_modules is out of date for package-lock.json or package.json (but am a little less inclined for package.json). This means a developer/site needs to do something to stop an npm install on changes or to force an npm install if there has been no changes. So, please weigh in so we can decide what to do. Thanks.

PR #486, opened 2020-05-12
fix security, typescript & modify install

huss:
1. This address GitHub reported security issue.
1. Typescript was updated but not to the most recent version to remove issues during checking. At some point we need to go to latest but that introduced some new issues. One fix addressed #444 but there is a TODO in the code to see if a better solution can be found.
1. The install script was modified (see commit) so it now recreates node_modules/ each time unless the --keep_node_modules argument is supplied. When rebuilt, the install is slower but it does make sure the updates to the packages are installed (unless asked not to do). It also avoids changing package-lock.json but does error out if it is out of date relative to package.json. This addresses issue #481.

One note to developers. If you do "npm i" then it can change the package-lock.json to make it match the information on npm for a package. I changed minimist version for geojason-rewind manually (add ^) to allow for it to use a slightly newer version to fix the security issue. If this package is not upgraded (has not been) and anyone does an npm install then they may need to manually apply this patch again.


huss:
The latest changes fix a typo and update reference to Beloit.

ak476519 (approved):
I agree that this is a good temporary solution to the typescript problem. The difficulties in completely solving this typing issue and the trouble it is causing in the development of this project on a larger scale makes it a better option to move on for now. 
Adding an 'npm ci' option to the installation script helps to avoid many problems caused by incongruency in the modules when the package versions are updated.

huss:
This has been locally tested and passed so the ongoing Travis failure is being ignored. @ak476519 review & approval is being considered sufficient as very few developers now have write access.

PR #491, opened 2020-05-13
Coverity

pathik10:
Applying fixes for CID 280473, 280469, 280445 and 280444.


huss (approved):
The changes have been reviewed and are okay. I also fixed two other issues reported by Coverity Scan that were not an issue but needed to have a variable initialized to avoid (I hope) in the future. The testing was also fixed that was not running before and needed revision.

It is good to note that Coverity Scan only found two issues with the OED code in testing (issues found inside node packages were not addressed). No issue (including node packages) is considered serious. OED now should pass all the Coverity Scan tests for code written by the OED project.

PR #501, opened 2020-08-11
Resolve bugs in Compare graphs

### LINKED ISSUE #403: Compare doesn't update with selected meters until a time-frame is selected

### LINKED ISSUE #461: Bar graphs not working with groups or compare
Groups do not seem to appear in bar or compare graphs. This appears to be present in version 0.5 (at least for bar graphs) so it is unclear when this happened. It is also unclear if the issues are related for the two types of graphs.

ak476519:
This PR is a fix for the following issues: 

Fixes #461 where compare graph does not fetch data for groups, fixed the post request in compressed Readings API that's causing this problem.
Fixes #403 turns out to be a bug during Redux state change. When the selected groups/meters change, there's an incongruence between the time interval in Redux state(which is used to check if the desired data is available when graphing compare) and the time interval of data fetched from database by this request. Added code so that changes in meter/group selection affect the Redux state the same way changes in compare periods affect Redux states.


src/client/app/actions/compareReadings.ts
```diff
@@ -80,13 +81,12 @@ function receiveGroupCompareReadings(groupIDs: number[], timeInterval: TimeInter
 /**
  * Fetch the data for the given meters over the given interval. Fully manages the Redux lifecycle.
  * @param {[number]} meterIDs The IDs of the meters whose data should be fetched
- * @param {TimeInterval} timeInterval The time interval over which data should be fetched
```
huss, line 83: These two params seem to be replaced by comparePeriod and probably the same as param in next function.

src/client/app/containers/ChartLinkContainer.ts
```diff
@@ -20,6 +20,12 @@ function mapStateToProps(state: State) {
 			linkText += `&barDuration=${state.graph.barDuration.asDays()}`;
 			linkText += `&barStacking=${state.graph.barStacking}`;
 			break;
+		case 'line':
+			linkText += `&serverRange=${state.graph.timeInterval.toString()}`;
```
huss, line 28: Should these commented out lines stay?

huss (changes requested):
The PR addresses the issues listed. It passes local testing. I have merged development and fixed linting issues. There are only a couple of comments about commenting that I hope can be addressed before the PR is merged.

huss:
To move version 0.7 forward, the changes to comments are being delayed.

Issue #141, opened 2017-10-06
test data


huss:
We should generate test data that looks like actual data and be fed into OED. This will be used for testing and also for demos instead of using the more limited real data. For testing the scaling/performance of the system, we should generate 10 years of data for 1000 meters on a minute interval (for now). I estimate this will be in the 2GB range of data. The meter names can be something like M0001, M0002, etc. and the data can be random in some range (such as 0-20 KWh).

LeoTindall:
I'd like to generate some meters which have data whose waveforms are:

- Sinusoidal with high frequency Perlin component (this is how most campus meters currently look)
- Constant with high frequency Perlin component (Mauer looks like this)
- Square with ultra-low frequency (seasonal) sinusoidal component and high frequency Perlin component
- Perlin noise with ultra-low frequency (seasonal) sinusoidal component

I think these will provide both a pleasing visual demonstration and a good exercise of the compression system.

huss:
There is now some standard development data that is documented on the wiki page. It is expected to expand the set of data in the future. Also, when we allow CSV via a URL, the adding of test data will be even easier (#370). We also want to start using some of this test data for Travis testing.

huss:
We can now generate sine/cosine data. We are in the process of using those functions to generate different types of data for developer usage and testing.

huss:
The test data generation was completed with PR #614. We still will probably do more types of data in the future. We still need to integrate this data into the testings code.

huss:
At this point in time, the one new type of test data that is desired is something that looks like real data from a meter. This would allow for automated data generation for the website and developers. Some ideas are above. We want it to vary with time (maybe season of the year) and have some quasi-random looking component where we can easily generate the same data over and over. See src/server/data/ for the current data generated and the generate... scripts listed in package.json.

huss:
Progress has been made on this and will continue after v1.0.

kramosss:
Hi, I would like to work on this issue.

huss:
OED now loads some real (or real-like) data for the website via CSV. Given real data has not been important for testing or development and it can be generated outside OED and added, this issue is no longer of great importance. If anyone feels otherwise or has thoughts then let OED know. I'm closing this issue based on this comment.

PR #505, opened 2020-10-14
Lindavin/testing data

lindavin:
This pull request adds a way to generate testing data so that developers can have a visual test on OED. Currently it only generates a sine function over time.


src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * generateTestData.js exports four functions: 
+ * generateDates,
+ * generateSine, 
+ * write_to_csv,
+ * generateSine 
+ *
```
huss: I think we prefer not to have final, empty lines in comments.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ */
+
+// Imports
+const fs = require('fs');
+const stringify = require('csv-stringify');
+const _ = require('lodash');
+const moment = require('moment');
+
+/* Our main export is the generateSine function. We break this into several parts:
+ * 1. Generate the moments in time within a specificed range and at a specified time step from a given range.
```
huss: specificed (sp).

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * 2. For each moment determine how much time as elapsed (as a decimal) within its respective period.
+ * 3. Calculate the value of sine at that moment in time.
+ * 		Conceptually this is sin(decimal_percentage * 2*PI) and it works because the sine function we will
```
huss: Maybe consistent spaces around * in formula?

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * this is { minute: 20 } or 20 minutes and to at be at least 1 second. 
+ * @returns {String[]} An array of timestamps between startDate and endDate, at a given timestep
+ * (default 20 minutes). The first element of the output will be the startDate, but the last element
+ * may not necessarily be the endDate.
+ */
+function generateDates(startDate, endDate, timeStep = { minute: 20 }) {
+	// Check timeStep is at least 1 second, if not throw an error.
+	const temp = moment();
+	if (temp.clone().add(timeStep).isBefore(temp.clone().add({ second: 1 }))) {
+		throw Error('The timestep needs to be at least 1 second.')
```
huss: I think it is always a good idea to provide as much information as possible in an error message. Here I would add the timeStep provided.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+	};
+	const array_of_moments = [];
+	const startMoment = moment(startDate);
+	const endMoment = moment(endDate);
+	while (!startMoment.isAfter(endMoment)) {
```
huss: Would this be clearer and simpler as
endMoment.isAfter(startMoment)

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * @source: https://stackoverflow.com/questions/18960327/javascript-moment-js-calculate-percentage-between-two-dates
+ */
+function _momentPercentage(startTime, endTime, currentMoment) {
+	// Check pre-conditions
+	if (endTime.isBefore(startTime)) { throw RangeError('The endTime must be after or equal to the startTime.') };
+	if (currentMoment.isBefore(startTime)) { throw RangeError('The currentMoment must be after or equal to the starTime.') };
+	if (currentMoment.isAfter(endTime)) { throw RangeError('The currentMoment must be before or equal to the endTime.') };
+	if (startTime.isAfter(endTime)) { throw RangeError('The startTime must be before or equal to the endTime.') };
+
+	if (endTime - startTime <= 0) { return 1 };
```
huss: This test seems to overlap the above checks. I don't see how endTime - startTime could be negative given you check endTime before startTime and startTime after endTime. I do see that the case of them being equal is not caught above and needs to be done (there is the isSame() function). I think it needs to be decided if === is an error or should return 1 given currentMoment must equal the other two in this case.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+	return (currentMoment - startTime) / (endTime - startTime);
+} // _momentPercentage
+
+/**
+ * Takes each moment and converts them into the percentage of time elapsed in 
+ * its specific period as a decimal from 0 to 1.
+ * @param {moment[]} array_of_moments Array of moment objects 
+ * @param {Object} period_length Object whose keys describe the length of the
```
huss: describe the length of the period (?)

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+} // _momentPercentage
+
+/**
+ * Takes each moment and converts them into the percentage of time elapsed in 
+ * its specific period as a decimal from 0 to 1.
+ * @param {moment[]} array_of_moments Array of moment objects 
+ * @param {Object} period_length Object whose keys describe the length of the 
+ * length of the period, which should be greater than the timestep between 
+ * consecutive moments. 
+ * @returns {Number[]} an array where each element corresponds to the percentage of time elasped at the
```
huss: elasped (sp)

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * @source: https://www.quora.com/In-JavaScript-how-do-I-test-if-a-number-is-close-to-zero
+ */
+function isEpsilon(number, epsilon = 1e-10) {
+	return Math.abs(number) < epsilon;
+} // isEpisilon
+
+/**
+ * Generates sine data over a period of time. By default the timeStep is 20 minutes. 
+ * and the period_length is one day.
+ * @param {String} startTimeStamp, the time's format is 'YYYY-MM-DD HH:MM:SS'
```
huss: I think these comments could better describe how the TimeStamps are used and what values are valid for options along with the defaults (amplitude not given above).

Probably should remove the , after the variable names.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * @param {String} endTimeStamp, the time's format is 'YYYY-MM-DD HH:MM:SS' 
+ * @param {Object} options controls the timeStep and the period_length, the timeStep needs to be at least
+ * 1 second. 
+ * @returns {String[][]} Matrix of rows representing each csv row of the form timeStamp, value
+ */
+function _generateSineData(startTimeStamp, endTimeStamp, options = { timeStep: { minute: 20 }, period_length: { day: 1 }, maxAmplitude: 2 }) {
+	const defaultOptions = {
```
huss: I  may be missing something but it seems the default options are set in both the arguments and here. If so, would it be safer to only set in one place so it only has to be changed in one place? See how done another way in generateSine.

Also, is defaultOptions a good name if it might be the default ones or the ones passed? Is it the usedOptions?

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+	}
+	const dates = generateDates(startTimeStamp, endTimeStamp, defaultOptions.timeStep);
+	const dates_as_moments = dates.map(date => moment(date));
+	const halfMaxAmplitude = defaultOptions.maxAmplitude / 2;
+	// We take our array of moment percentages and scale it with the half max amplitude
+	// and shift it up by that amount. 
+	const sineValues = momenting(dates_as_moments, defaultOptions.period_length)
+		.map(x => {
+			const result = Math.sin(Math.PI * 2 * x);
+			const scaledResult = halfMaxAmplitude * (isEpsilon(result) ? 0 : result) + halfMaxAmplitude;
```
huss: I'm curious why you make small values 0. I can see that you get roundoff/precision errors with sin but they should be very small. Also, they will occur at all values including ones close to 1. Given this, should we just accept the sin result? Maybe there is an idea here I am missing.

I did have the thought that you want this to go to 0 for testing to compare to an integer. If that is the case then maybe changing the test to do a close to equal (such as .closeTo) might be better.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+ * ]
+ * 
+ * Sources: 
+ * https://csv.js.org/stringify/api/
+ * https://stackoverflow.com/questions/2496710/writing-files-in-node-js
+ */
+function write_to_csv(data, filename = 'test.csv') {
+	stringify(data, (err, output) => {
+		if (err) {
+			return console.log(err);
```
huss: I wonder if this should cause an exception so any code calling this code knows an error occurred. generateDates does that.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+			console.log("The file was saved!");
+		});
+	});
+} // write_to_csv
+
+/**
+ * Creates a csv with sine data 
+ * @param {String} startTimeStamp, the time's format is 'YYYY-MM-DD HH:MM:SS'
+ * @param {String} endTimeStamp, the time's format is 'YYYY-MM-DD HH:MM:SS'
+ * @param {Object} options, options is an object that will be parsed by
```
huss: Are all the options parsed by moment or just the ones that represent a time stamp? I think here and/or somewhere else that is referenced (if document some elsewhere as suggested), the meaning of each option should be given.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,189 @@
...
+	const chosen_data_options = { timeStep: options.timeStep, period_length: options.period_length, maxAmplitude: options.maxAmplitude };
+	write_to_csv(_generateSineData(startTimeStamp, endTimeStamp, chosen_data_options), options.filename);
+} // generateSine
+
+module.exports = {
+	generateDates,
+	generateSine,
+	write_to_csv,
+	momenting,
+	_generateSineData
```
huss: I see you use _generateSineData in testing but it seems a little funny to export a function with _ that normally means it is for internal use/private. I don't know if there is a better way to do this.

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+	});
+	mocha.it('should be able to generate dates at different time steps.', () => {
+		const startTimeStamp = '2020-10-01 21:00:00';
+		const endTimeStamp = '2020-10-02 21:00:00';
+		const timeStep = { days: 1 };
+		expect(generateDates(startTimeStamp, endTimeStamp, timeStep)).to.deep.equal(
+			[startTimeStamp, endTimeStamp]
+		);
+	});
+	mocha.it('should be able to generate dates at a time step 15 seconds', () => {
```
huss: I like these tests but both hit the endTimeStamp exactly. Should one have it so it is not the case that endTimeStamp = startTimeStamp + n * timeStep, where n is an integer?

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+			result
+		);
+	});
+});
+
+mocha.describe('momenting', () => {
+	mocha.it('should cover the simple singleton case', () => {
+		const date = '2019-09-10 00:00:15';
+		const periodOfMoments = [moment(date)];
+		expect([1]).to.deep.equals(momenting(periodOfMoments));
```
huss: It seems you are calling momenting with one argument but it should take two. Is this correct?

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+	mocha.it('should be able to partition an array with just two elements', () => {
+		const startMoment = moment('2019-09-10 00:00:15');
+		const timeStepMs = 15000;
+		expect([0, 1]).to.deep.equals(momenting([startMoment, startMoment.clone().add(timeStepMs)], timeStepMs));
+	});
+	mocha.it('should cover the a skipped datetime', () => {
+		const test = ['2019-09-10T00:00:15',
+			'2019-09-10T00:00:30', '2019-09-10T00:01:00'];
+		const timeStepMs = 15000;
+		expect(momenting(test.map(timeStamp => moment(timeStamp)), timeStepMs)).to.deep.equals([0, 1, 1]);
```
huss, line 68: Small point but in the first two tests you put the expected array first and in the second two you put it last.

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+	});
+	mocha.it('should work on irregular timesteps', () => {
+		const test = ['2019-09-10T00:00:15',
+			'2019-09-10T00:00:30', '2019-09-10T00:00:50'];
+		const timeStepMs = 15000;
+		expect(momenting(test.map(timeStamp => moment(timeStamp)), timeStepMs)).to.deep.equals([0, 1, 5 / 15]);
+	});
+});
+
+mocha.describe('Generate sinewave helper', () => {
```
huss: sinewave -> sine wave, below is similar in next test.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,209 @@
...
+	// Check timeStep is at least 1 second, if not throw an error.
+	const temp = moment();
+	if (temp.clone().add(timeStep).isBefore(temp.clone().add({ second: 1 }))) {
+		throw Error('The time step needs to be at least 1 second.');
+	}
+	const arrayOfMoments = [];
+	const startMoment = moment(startDate);
+	const endMoment = moment(endDate);
+	while (!startMoment.isAfter(endMoment)) {
+		array_of_moments.push(startMoment.format('YYYY-MM-DD HH:mm:ss'));
```
huss: Should be arrayOfMoments here and below. Editing oversight when linting issues were resolved in recent commit.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,209 @@
...
+ * [...row],
+ * [...row],
+ * ...
+ * ]
+ *
+ * Sources:
+ * https://csv.js.org/stringify/api/
+ * https://stackoverflow.com/questions/2496710/writing-files-in-node-js
+ */
+function write_to_csv(data, filename = 'test.csv') {
```
huss: I think this function name should be changed to remove _ and use camelcase to be consistent with variable name changes from linting rules.

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+	mocha.it('should be able to generate data for half a day', () => {
+		const startTimeStamp = '2019-09-10 00:00:00';
+		const endTimeStamp = '2019-09-11 00:00:00';
+		expect(_generateSineData(startTimeStamp, endTimeStamp, { timeStep: { hour: 12 } }))
+		.to.deep.equals([[startTimeStamp, '1'], ['2019-09-10 12:00:00', '1'], [endTimeStamp, '1']]);
+	});
+});
+
+mocha.describe('Generate Sinewave', () => {
+	mocha.it('should properly write to file', async () => {
```
huss, line 148: Took me a little while to figure out this mistake. period_length -> periodLength.

src/server/data/generateTestData.js
```diff
@@ -0,0 +1,209 @@
...
+	stringify(data, (stringifyErr, output) => {
+		if (stringifyErr) {
+			log.error(stringifyErr);
+		}
+		fs.writeFile(filename, output, err => {
+			if (err) {
+				return log.error(err);
+			}
+			log.info('The file was saved!');
```
huss: I think this message (and the error ones above) should have more information. It should say the file name and that it was doing test data generation. This is needed because the log file has lots of output and it is not easy to tell where a message came from without context.

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+		const startTimeStamp = '2019-09-10 00:00:00';
+		const endTimeStamp = '2019-09-11 00:00:00';
+		const filename = 'test1.csv';
+		const timeOptions = { timeStep: { minute: 20 }, period_length: { day: 1 } };
+		const maxAmplitude = 2;
+		const data = _generateSineData(startTimeStamp, endTimeStamp, { ...timeOptions, maxAmplitude: maxAmplitude });
+		await generateSine(startTimeStamp, endTimeStamp, { ...timeOptions, filename: filename, maxAmplitude: maxAmplitude });
```
huss: filename: filename -> filename: `${__dirname}/${filename}`
Below the code reads from this location. Docker may write a file (by default) in the main directory of the application where it is started. This often is not the src/server/test directory. Since we don't want the test data file in another directory, I think we should do this change. I tested it on my machine and it worked properly. See below for another comment.

src/server/test/generateTestDataTests.js
```diff
@@ -0,0 +1,115 @@
...
+		// https://stackabuse.com/reading-and-writing-csv-files-in-nodejs-with-node-csv/
+		const dataFromFile = await fs.readFile(`${__dirname}/${filename}`);
+		const records = await parseCsv(dataFromFile);
+
```
huss, line 159: I wonder if we should delete the test output file after it is read for several reasons. First, if a file is left over from previous testing then it may be reused if this test fails before the file write and it could pass when it should not. Second, the file shows up as a change in git that might lead to it being put onto GitHub. Third, leaving around files may not be a good idea (it is very nice during debugging to see the output but not in general). Another option would be to use the date/time as part of the file name but I think it is best to delete it.

huss (changes requested):
This code works well. I have asked for a number of changes or made suggestions in my review. Note that some are marked Outdated but that seems to be from @lindavin doing a commit after I make the comment. As such, they should still be looked at. I did the testing on my machine and it runs correctly (with suggested changes). I also loaded the test data in OED. It looks fine but there are some issues that will be addressed outside this PR. This code may be altered in the future to generate more types of sample data as we do more testing on OED. It is also intended to be used to all for more systematic testing of OED in the automated tests rather than just using a few simple points.

huss:
I have been looking at and analyzing the graphs more. The line graph you get (see issue #519) has values from 0-6 even thought the generated values are 0-2. The reason is that OED takes the reading which is kW and converts it to kWh. Since each reading is 20 minutes, it scales the reading by dividing by 20/60 (fraction of an hour that each point represents) so each reading is graphed as reading/(20/60) or reading * 3. Since the intent was the test code would generate data with a given amplitude, I think we should normalize it by the fraction of an hour represented (the inverse of what OED is doing) so we see the expected amplitude on the graph and the actual readings will not be these values.

huss:
I used the test data to actually test OED. It was useful to create sin^2 and cos^2 so I could create a group that is the sum that should then be the sum of these amplitudes since sin^2 + cos^2 is 1. While it is not hard to get cos from sin, I think it would be good to have a way to get the cos as we do sin in the test code. This could either be a new function or one adding a parameter to choose which the user wants (with renaming of the functions).  The user can easily do the square of each point if that is what they want.

huss:
This is a major step in doing #141. We will need to integrate this data fully into OED to finish that process.

lindavin:
Now testing data is rewritten in typescript, the cosine function is added, and a new parameter to normalize data for OED has been added.

src/server/data/generateTestingData.ts
```diff
@@ -0,0 +1,251 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * generateTestData.js exports four functions:
+ * generateDates,
+ * generateSine,
```
huss: You added generateCosine in last revisions.

src/server/data/generateTestingData.ts
```diff
@@ -0,0 +1,251 @@
...
+ * @param {object} options the parameters for generating a data file for OED
+ */
+async function generateCosine(startTimeStamp: string, endTimeStamp: string, options: GenerateSinusoidalDataFileOptions) {
+	const chosenOptions: GenerateSinusoidalDataFileOptions = {
+		timeStep: options.timeStep || { minute: 20 },
+		periodLength: options.periodLength || { day: 1 },
+		maxAmplitude: options.maxAmplitude || 2,
+		filename: options.filename || 'test.csv',
+		normalizeByHour: options.normalizeByHour || false,
+		phaseShift: (options.phaseShift || 0) + (Math.PI / 2)
```
huss: I might put a comment here that this shifts sine to cosine by adding pi/2.

src/server/data/generateTestingData.ts
```diff
@@ -0,0 +1,251 @@
...
+import * as moment from 'moment';
+import * as promisify from 'es6-promisify';
+import { log } from '../log';
+
+const stringifyCSV = promisify(stringify); // this is a strange error; when compiled it is okay
+/* Our main export is the generateSine function. We break this into several parts:
+ * 1. Generate the moments in time within a specified range and at a specified time step from a given range.
+ * 2. For each moment determine how much time as elapsed (as a decimal) within its respective period.
+ * 3. Calculate the value of sine at that moment in time.
+ * 		Conceptually this is sin(decimal_percentage * 2* PI) and it works because the sine function we will
```
huss: space before *

src/server/data/generateTestingData.ts
```diff
@@ -0,0 +1,251 @@
...
+ * this is { minute: 20 } or 20 minutes and needs to be at at least 1 second.
+ * @returns {string[]} An array of timestamps between startDate and endDate, at a given time step
+ * (default 20 minutes). The first element of the output will be the startDate, but the last element
+ * may not necessarily be the endDate.
+ */
+function generateDates(startDate: string, endDate: string, timeStep: moment.MomentInputObject = { minute: 20 }): string[] {
+	// Check timeStep is at least 1 second, if not throw an error.
+	const temp = moment();
+	if (temp.clone().add(timeStep).isBefore(temp.clone().add({ second: 1 }))) {
+		throw Error(`The time step provided is ${JSON.stringify(timeStep)} needs to be at least 1 second.`);
```
huss: throw new Error. Several others below.

src/server/test/generateTestDataTests.js
lindavin: I thought that this would be okay since I just wanted to test the singleton case which would have no period length.

src/server/data/generateTestData.js
lindavin: There is no deep.closeTo implemented for chai so instead I used a loop with closeTo.

src/server/test/generateTestDataTests.js
lindavin: Okay, I changed the time step to 13 minutes which should work since there is an integer amount of hours between the end and start timestamps.

src/server/test/generateTestDataTests.js
lindavin: I added `undefined` as the second parameter for this reason.

huss (approved):
I merged with development and it runs fine. All tests are passing locally. Since these changes are independent of other PRs ahead of it, it is okay to merge now. I thank @lindavin for working on this and adding features added after initial development.

PR #506, opened 2020-10-27
Tooltips

### LINKED ISSUE #259: Inline documentation system
We should add small circled-questionmark buttons which, when clicked, bring up inline help for the nearby elements.

lindavin:
This pull request implements the tooltips on the home page. For some tooltips we use the TooltipHelpComponent. In other positions on the home page, it was easier to use an alternative implementation of tooltips: TooltipHelpComponentAlternative. 

This fixes #259. We will need to fully integrate this change by making tooltips for all pages in OED.

Note merging this changes the node modules.


sasankthapa:
I think everything is working as intended but there are some minor issues about styling like:

image

The buttons on the graph are messing with the texts and most of the links are wrapped outside the box.

huss:
While it would be nice to fix the formatting to avoid going outside the tooltip box, I suspect the box size is set and the link URL is too long. I think an easy fix that also makes sense is to use a name/words that link to the help page. The URL was put there to help plan the help page location but I don't see a lot of value in having it in the actual tooltip. Thus, I will edit the tooltip text to fix this and add it as needed.

As for the items on the page showing behind the tooltip and making it harder to read, I suspect this is designed to allow people to see the page while reading the tooltip. We could change (if that is an option) so the tooltip background is solid but I don't know if we want that all the time. I'm open to suggestions here.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -5,6 +5,8 @@
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { Button } from 'reactstrap';
+import TooltipHelpComponent from './TooltipHelpComponent';
```
huss: This import does not seem to be used.

src/client/app/components/DashboardComponent.tsx
```diff
@@ -19,6 +19,8 @@ import { FormattedMessage } from 'react-intl';
 import { Dispatch, Thunk, ActionType } from '../types/redux/actions';
 
 import * as Plotly from 'plotly.js';
+import TooltipHelpComponent from './TooltipHelpComponent';
```
huss: import does not seem to be used.

src/client/app/components/HeaderButtonsComponent.tsx
```diff
@@ -9,6 +9,9 @@ import { FormattedMessage } from 'react-intl';
 import MenuModalComponent from './MenuModalComponent';
 import { hasToken } from '../utils/token';
 import getPage from '../utils/getPage';
+import TooltipHelpComponent from './TooltipHelpComponent';
```
huss: This and next imports do not seem to be used.

huss (changes requested):
The review chages are not showing up on GitHub so maybe I forgot to finish review. Doing it now.

huss:
This PR is now complete. It will likely be merged after the PRs that came before it (or sooner) and slated for version 0.7. I will do the final checks to be sure merging in the current development does not cause any issues.

I want to thank @lindavin for doing all of this and making changes as new ideas were thought of.

huss:
I have reviewed the recent changes (that I asked for). All still seems fine. Note I may be changing the links in the tooltips as I create the help pages and desire changes.

huss (approved):
I have looked these over and all is fine. Changes requested plus more done.

PR #507, opened 2020-11-10
resolve install issues

### LINKED ISSUE #481: when to run npm install

huss:
This PR primarily addesses install errors. It should fix up errors that new developers were seeing. It made the following changes:
- The version of pg-promise was changed to a newer version. This fixed database connection issues. The version of Postgres was also increased to allow some new features (such as printing version for debugging). The change to pg-promise broke some tests and they were fixed for the new issue.
- A Postgres password for the database is now set as this is needed in newer versions of Postgres.
- Fixed a bug when swapping database connections that caused failures. This error was clearer to see after upgrade of pg-promise.
- Many of the database initalization steps did not protect against running when the item already existed. This caused an error that stopped later initialization steps. Developers with a properly installed database did not see the problem since the later steps had been done previously.
- Also worked on migrations to address same database issue.
- The install script was changed so it only catches known errors and stops otherwise. This applies to the database initialization and the user creation. With database initialization the script retries multiple time to overcome temporary errors, eg., database not yet ready.
- A new parameter (--continue_on_db_error) was added to the install script that allows it to continue even if an error is detected. Default is to stop on error. This can be used for debugging and to get the web container running for debug/work if you don't need the database properly installed.
- The install script now accepts install_args so you can pass parameters via docker-compose up. For example, to keep the node modules, you can do:  install_args="--keep_node_modules" docker-compose up.
- Fixed bug where createDB passed incorrect parameters to logging.
- Fixed bug where the install script stopped even if the last database creation attempt worked.

The code passes most tests locally (Travis still is not working). There are still 6 issues that seem related to moment that must be addressed. It is being committed now before all known issues are addressed since it fixed ongoing issues that developers are seeing. For example, we still get a warning for Postgres about trust that will be addressed later.

This fixes #481


src/server/sql/group/create_groups_tables.sql
```diff
@@ -154,8 +154,24 @@ CREATE FUNCTION check_cyclic_groups()
 	$$
 	LANGUAGE 'plpgsql';
 
-CREATE TRIGGER check_cyclic_groups_trigger
+-- This will avoid trying to create the trigger if it exists as that causes an error. This is an issue since
+-- the OED install stops the creation of database items after this.
+DO
+$$
+BEGIN
+IF NOT EXISTS(SELECT *
```
wadoodalam, line 162: Usually IF NOT EXISTS gives out a problem in these types of queries. Exception handling might be another way of doing it.

src/server/test/db/barchartReadingsTests.js
```diff
@@ -32,7 +32,7 @@ mocha.describe('Barchart Readings', () => {
 		// TODO: This is awkward. Is there a better way to have conn as a non-optional
 		// without requiring changing EVERY call to getBarchartReadings?
 		// If not, well, we need to change every call.
-		const barchartReadings = await Reading.getBarchartReadings([meter.id], moment.duration(2, 'h'), null, null, conn);
+		const barchartReadings = await Reading.getBarchartReadings([meter.id], moment.duration(2, 'h').toISOString(), null, null, conn);
```
wadoodalam, line 35: Not sure if need the value in string format or not.

huss: This was added to resolve issues after the upgrade of pg-promise. I don't know if this is the best fix but it works for now.

src/server/sql/group/create_groups_tables.sql
huss: I'm unclear on what problem this might cause. Can you elaborate on what the concern is? Also, while exception handling can be used in some circumstances, it can be heavier weight and, when possible, it seems better to deal with directly rather than as an error.

ak476519 (approved):
The changes so far seems adequate, I believe we are good to go. 
Although the production code works fine, we do have test cases that become problematic after this installation update. Adding workable test cases to replace momentPatchTest.js should be addressed as a new issue.

PR #508, opened 2020-11-24
Fix timeshift issue

### LINKED ISSUE #476: Plotly graphs and timezone shifts
This may relate to #280 on meter timezones. The seeming shift between the meter data time in UTC and local time is causing issues in a number of our graphics:

- Line graphs have the data shifted by the difference between UTC and the local timezone. For example, if the graph is zoomed in so you can see each time point from the meter and a point has the time 1:00 then the corresponding value in the database is 6:00 (if you are -5 hours shifted from UTC as is the case in the summertime in Central U.S.A time). The data coming from the meters for graphing is showing UTC and the shift to local time. If you look at graphs where each day is shown, they are labeled the previous day shifted from 0:00 by UTC. So, for example, if you have 10/12/19 at 7:00 then it was the sum of 10/13/20 0:00-23:00 (start time) from the database (if -5 hours from UTC). All of this also has the consequence that locales that shift clocks by season (daylight saving in U.S.A.) have the times shift by that amount as listed on the graph and it varies by time of year (see issue #474). This also probably has a relationship to #280. The data should not be shifting like this and a day should start at 0:00 (but see #477 for desired labeling)
- Bar graphs have an issue but it may be different from this one. For example, a bar labeled 10/11/2019-10/12/2019 is the sum of the meter data for 10/12/2019. Once the issues with line graph shifts and labels are resolved, the correspondence between summing the line values to get a bar value will work and be clear. Relates to #477.
- Compare also has the issue but it works differently than line. The end time for the current bar is the current time with any fractional hour dropped. Thus, if it is currently 13:33 the end time is 13:00. This means the start time of the last point is one hour less (if have hour points). The start time is this time shifted back to the start of the time frame being considered (0:00 of that day). For day it is start of the same day, for week it is the Sunday of the same week and for 4 weeks is the 4 Sundays earlier (3 Sundays before the week start). Now for the time shift issue. All these times are now shifted from local to UTC time. If that is -5 hours (as above) it adds that time onto each time. Note that it makes the time later here unlike the description for line above because the shift is happening the other way. So, if the current time is 13:33 on Wednesday, 4/8/20 then the end time for the sum will be 17:00 (13:33 becomes 13:00 end time so start is one hour earlier with hour meter points to 12:00 and then add 5 hours). The start time for day will be Wednesday, 4/8 at 5:00 since it should be 0:00 but is shifted 5 hours. This means you sum the points with start time of 4/8 5:00-4/8 17:00 in the database meter values. For week and 4 week the end time is the same but the start time is Sunday, 4/5/20 at 5:00 and Sunday, 3/15 at 17:00. While the number of points summed is correct, you are summing past the current time so these values are effectively 0 and the value is generally smaller than it should be. It should also be obvious that this shift, esp. compared to line, is very confusing for the user. We need to get rid of the local to UTC shift. Finally, the previous bar (labeled last on graph) bar is similar except you shift the date/time from the current (labeled this on graph) value back by the number of days that is chosen. So, for day you go back one day so you sum 4/7/20 5:00-4/7/20 17:00, for week you go 7 days so you sum 3/29/20 5:00-4/1/20 17:00, and 4 weeks is the sum of 2/16/20 5:00-3/11/20 17:00. The same shift issue that exists for the previous bars may naturally clear up after the current time is fixed.

This may seem complex/hard to understand so feel free to contact @huss who worked this out.

ak476519:
Documented time shift issue fixed #476. 

The issue was caused by timestamps without timezones and a localization feature of moment.js.
After evaluating different solutions, a easier fix was selected because of the following reasons:
1. The package that is causing the problem, moment.js, has already stopped updating and its functionality in this project would be replaced with other packages such as JavaScript's updated date library in the following years. It's more feasible to use this time in other goals that are more relevant to long term.
2. This easier (and the cleaner) fix, has minimal changes to the codebase. This would make it easier to convert to other packages in the future. 

Note merging this changes the node modules.


src/client/app/containers/LineChartContainer.ts
```diff
@@ -30,8 +31,9 @@ function mapStateToProps(state: State) {
 				const hoverText: string[] = [];
 				const readings = _.values(readingsData.readings);
 				readings.forEach(reading => {
+					// const timeReading = recreateTimezoneOffset(reading.startTimestamp, 'America/Chicago').tz('America/Chicago');
```
huss: Should this commented out line and the import be removed?

src/client/app/utils/timezoneHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+export function recreateTimezoneOffset(origin: number, targetZone: string): moment.Moment {
+	const localDate = moment(origin).utc().format().substring(0, 19);
+	return moment.tz(localDate, targetZone);
+}
+
+/**
+ * Checks if moment-timezone can recognize the provided timezone name
+ * @param timezone
+ */
+function isValidTimezone(timezone: string): boolean {
```
huss, line 24: I don't see where this is used.

src/client/app/utils/timezoneHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as moment from 'moment-timezone';
+import {State} from '../types/redux/state';
+
+/**
+ * recreate a timestamp created in default timezone(UTC) with its true timezone
+ * this is invoked only when the user wishes to export data in their absolute time in UNIX timestamps
+ */
+export function recreateTimezoneOffset(origin: number, targetZone: string): moment.Moment {
```
huss, line 15: This only seems used by commented out code in LineChartContainer.ts (see comment there). Not sure if this is needed in the end.

huss (changes requested):
For some reason I cannot add any more comments to some parts of the code so I'm leaving them here. I hope I have the descriptions correct and I'm correct about what I saw.

- meters work fine with lines. With bar the values are all fine but the dates may be shifting as described for groups below.
- groups with lines have the correct value but the time is earlier by the shift from UTC. I think the same change for meters needs to be applied to groups.
- groups with bar chart has the wrong value. My testing indicates that they are actually summing the right number of points but they are all shifted earlier in a similar way to lines. So, you get 24 points for a day bar but it is the wrong range of points.
- groups with bar chart also can have the wrong days. This screenshot:
Screen Shot 2021-01-19 at 12 06 45 PM
shows that the groups bar (it was the sum of the two meters displayed) has different days than the meters. I suspect that the calculation of the days involved has a timezone issue for the groups (and for meters as noted above).
- compare graphs have a timezone issue. The range of values summed seem to shift forward by the UTC timezone difference. This is the opposite direction than groups with bar. This also applies to meters here.
- exporting is working fine.
- There are some linting errors but I think they will go away when the newer development is merged (there will be conflicts).
- testing is passing.

I'm asking for @ak476519 to look at these issues. Once complete I will recheck all values and can then approve this PR. If anything is not clear or you cannot see what I say then just let me know.

src/client/app/containers/LineChartContainer.ts
huss: done.

src/client/app/utils/timezoneHandler.ts
huss: Added comment but left in case needed for export PR.

src/client/app/utils/timezoneHandler.ts
huss: See other comment on leaving for now.

huss:
I've merged this with development. I addressed the issues stated in the PR. I also fixed for similar issue with maps. Local check/testing is fine. This should be ready to merge.

huss (approved):

PR #509, opened 2020-11-30
DB changes and Interface changes for timezone

### LINKED ISSUE #280: Store timezones for each meter

### LINKED ISSUE #465: docked/hidden menu does not show graphics options
This seems to have started in version 0.5 but not positive. When the menu gets docked it only shows buttons for the main pages (admin, meters, etc.) and not anything involved in the graphs. It did show them before and it should.

sasankthapa:
This is a PR for @wadoodalam DB changes as well. There were 2 linting errors which you could see by running "npm run check".

A big issue for this PR is that the time zone select doesn't directly appear since we are getting all the timezones from the backend. One way to fix this issue would be to add timezones to Redux but it seems to be a big change to fix a small issue. Instead until the timezones are loaded, "Please Reload" shows up in place of the timezone select. Reloading localhost:3000 wont solve this, instead to see the options for timezone, go to another page (OED main page) then go back to the admin page then it should show up.

A minor issue that I saw was the "x" in the timezone select options doesn't do anything.

This fixes #280 and fixes #465.

Note merging this changes the database configuration. Note merging this **may** change the node modules. The PR does not list package.json as changed but it was a changed file so @huss thinks it came from merging in other work from development.


src/server/test/db/meterTests.js
```diff
@@ -15,7 +15,7 @@ function expectMetersToBeEquivalent(expected, actual) {
 mocha.describe('Meters', () => {
```
huss, line 15: I think for this test and the next one it would be good to use some value for timezone to check the code more.

src/server/models/Timezone.js
```diff
@@ -0,0 +1,35 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const database = require('./database');
+
+const sqlFile = database.sqlFile;
+
+class Timezone {
+	/**
+	 * Creates a new reading
```
huss: This does not seem to describe what this does.

src/server/models/Meter.js
```diff
@@ -15,14 +15,16 @@ class Meter {
 	 * @param enabled This meter is being actively read from
 	 * @param displayable This meters is available to users for charting
 	 * @param type What kind of meter this is
+	 * @param meterTimezonep Default timezone for meter
```
huss: imezonep typo

huss (changes requested):
Overall, this PR does what it should and works correctly. There are some check and tests that are failing but I think many/all might be cleared up when development is merged. I will be doing that next and pushing out thoses changes to this PR but wanted to post the review first.

There is an issue we have discussed that the timezone does not always show up when you load the admin page or meter page. It now tells you to try refreshing the page. I tried that but it does not seem to work for me. The alternative we talked about of going to another page and then coming back does work. If that works for you too then maybe we can change what the page tells the user to something like: "Please go to another tab/page and come back to this one to see the timezone information".

src/server/models/Meter.js
huss: I fixed this while merging in development. It will be seen once I finish and push out my changes.

huss:
The issue with the timezone not showing up on pages seems gone with PR #517. I don't know if others are okay or if we still need to address.

src/server/models/Timezone.js
huss: done by @sasankthapa

huss (approved):

huss:
Merged in latest development and tested. All looks good.

PR #520, opened 2020-12-02
Obvius2020

### LINKED ISSUE #163: Read non-Mamac meter data via Obvius

lindavin:
This PR brings in the previously done Obvius2020 pipeline, with the added work of properly creating meters from a .ini configuration file.

Fixes #163.

Note merging this changes the database configuration. Note merging this changes the node modules.


src/server/models/obvius/processConfigFile.js
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const ini = require('ini');
+const Meter = require('../../models/Meter');
+const ConfigFile = require('../../models/obvius/Configfile');
+
+/**
+ * Creates meters from a config file
```
huss: The code actually seems to create an array of meters as the return indicated and not create the meters.

src/server/models/obvius/processConfigFile.js
```diff
@@ -0,0 +1,43 @@
...
+function processConfigFile(configFile) {
+	const config = ini.parse(configFile.contents);
+	const regularExpression = /([0-9][0-9])/;
+	// For the metersHash we assume each key corresponds
+	// to a hash of structure { NAME: <alternative name>, UNITS: <units>, LOW:<>, HIGHT:<>, CONSOLE:<> }
+	const metersHash = {};
+	// Array of Meter (from models) objects
+	const metersArray = [];
+	for (key of Object.keys(config)) {
+		const [, meterNumber, characteristic] = key.split(regularExpression);
```
huss, line 23: I think my knowledge of JS is too limited but I wanted to check that [, is correct. The code does seem to work fine.

src/server/routes/obvius.js
```diff
@@ -0,0 +1,199 @@
...
+		}
+		const conn = getConnection();
+		for (const fx of req.files) {
+			log.info(`Received ${fx.fieldname}: ${fx.originalname}`);
+
+			let data;
+			try {
+				data = zlib.gunzipSync(fx.buffer).toString('utf-8');
+			} catch (error) {
+				data = fx.buffer.toString('utf-8');
```
huss: This is treated differently when the gunzip fails for a log file. Should we be expecting a file that is not zipped? If not, this should be treated as in log upload.

src/server/test/web/obviusConfig.js
```diff
@@ -0,0 +1,78 @@
...
+// let { chai, mocha, expect, app, testDB, testUser } = require('../common');
+mocha = require('mocha');
+chai = require('chai');
+const ini = require('ini');
+const fs = require('fs');
+const Meter = require('../../models/Meter');
+const ConfigFile = require('../../models/obvius/Configfile');
+
+const { log } = require('../../log');
+
```
huss: I'm a little unclear about several aspects of the tests in this file. They seem to test functions but then output to the logging system. First, I don't know if doing that is what is desired and is not a good way to test results. Second, I'm not clear what trigger causes the test to fail other than an error. Finally, there are some commented out lines that should either be removed or integrated into the code.

huss:
This PR works fine. I want to thank @NoraCodes for important earlier work on the code in this pull request (and @lindavin who has done the recent work). I also note that @mm413 worked on this a little before.

I have merged in the current development. I made a couple of other edits while I did that but they should be minor. There are some issue(s) left that I hope will be resolved when earlier submitted but pending PRs are in development and merged here (waiting to do that). One is that Meter.js mapRow is missing identifier. I will do another round of full testing at that time.

Here are comments on code not changed in the PR that I cannot enter in that file/line:

- src/server/models/obvius/Configfile.js on line 85 has a comment marking the return value uncertain. Is there actually a return value?
- If you upload a config file twice it causes an error of rejected promise. I think an OED error would be best if the meter already exists, etc.
- There are 7 failing tests on my local machine. The first one ("Read Mamac log from a file: 
       rolls back correctly when it rejects:") is likely only on my machine due to a local postgres issue but wanted to know if others see it. The others seem related to obvius work. Can @lindavin  look at these? I can help as needed.

huss (changes requested):
Just a few files need updating and the general comment I left should be addressed.

huss:
I want to note that a few items will need to be addressed but are outside this PR:

- The code assumes every reading is 1 hour. We have had discussions on how to get the correct time for a reading when the start & end times are both not provided. Note this is more complex for the last reading and esp. if there is only one reading provided (as may happen with obvius).
- This code needs to be tried on a live server getting real Obvius data to confirm it works correctly.
- We are working on enhancing the types of users allowed. Once this is done then the password assumed will be removed.

huss (approved):

PR #529, opened 2021-01-04
Pipeline fix

huss:
- Fix JSDoc, esp. meter_id being a number (int).
- readCSV->readCsv to fix bug found by @lindavin and also make usage consistent in code.


huss:
Passing local testing (except known issues). Simple change so I'm approving and merging.

PR #528, opened 2020-12-19
Update database.js

vitaly-t:
Fixing incorrect migration to the latest version of pg-promise. See [event connect](http://vitaly-t.github.io/pg-promise/global.html#event:connect).

PR #530, opened 2021-01-04
fix pg-promise connect usage

huss:
Fix incorrect usage of connect after upgrade of pg-promise.
Many thanks to @vitaly-t for pointing this out.
This replaces PR #528.


huss:
Passes local testing. Basic fix suggested by pg-promise developer so approving.

PR #538, opened 2021-01-19
Graph coloring update

### LINKED ISSUE #495: improve graphic color choices
Similar colors are sometimes chosen for different meters/groups. It would be nice if the color was consistent in all graphics (currently does), stayed the same all the time (currently does), usually choose colors that are easy to tell apart, and somehow differentiated meters and groups.

cjpickle:
An improvement to the previous color-choosing algorithm and color array for graphs. For graphed data, there is now an easy visual distinction between meter data and group data.

This fixes #495.


src/client/app/containers/LineChartContainer.ts
```diff
@@ -21,6 +22,7 @@ function mapStateToProps(state: State) {
 			const readingsData = byMeterID[timeInterval.toString()];
 			if (readingsData !== undefined && !readingsData.isFetching) {
 				const label = state.meters.byMeterID[meterID].name;
+				const colorID = state.meters.byMeterID[meterID].id;
```
huss: I checked an meterID and colorID are the same. Thus, calling state.meters.byMeterID[meterID].id returns the value meterID. Thus, this can be changed to
const colorID = meterID;
The same applies to groupID below and both of these changes should be applied to BarChartContainer.ts.

huss (changes requested):
I've run checks and tests as well as tried the code. It all seems fine. Just suggesting a simple change. Note this will be part of version 0.8 so it will wait a little until merged. It will need to be rechecked after it is merged with that version.

src/client/app/containers/LineChartContainer.ts
cjpickle: Thanks for that recommendation. I've made the quick requested changes.

huss (approved):

huss:
Merged with development. One fix for maps and one change in array size. Runs fine.

PR #540, opened 2021-01-28
chore: enable noUnusedLocals for tsconfig and clean up unused variables

campionfellin:
Cleaning up some of the code as I go

Signed-off-by: campionfellin <campionfellin@gmail.com>


huss (approved):
I did not see any issues in the changes. I merged with development and it now works fine. My only regret is that @campionfellin could not magically make the changes after this was done avoid any linting errors :-)

PR #541, opened 2021-01-29
update info

huss:
Change .md files for new developer website.
Update email contact to our domain.


huss:
This is only changing info that I want into system now so I am going to merge.

PR #542, opened 2021-01-31
feat: set up github actions for testing

campionfellin:
Sets up initial framework for running test workflows on github actions

It seems that 4 tests are failing right now, so those may need to be updated

Signed-off-by: campionfellin <campionfellin@gmail.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,11 +37,11 @@ jobs:
       # see: https://github.com/actions/checkout/issues/335
       - name: upgrade git
         run: |
-          apt-get update && apt-get upgrade \
+          apt-get update && apt-get upgrade -y \
           && apt-get install -y software-properties-common \
-          apt-get update && apt-get upgrade \
+          apt-get update && apt-get upgrade -y \
           && add-apt-repository -y ppa:git-core/ppa \
-          apt-get update && apt-get upgrade \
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build
+on: [pull_request]
+
+jobs:
+  container-job:
+    env:
+      OED_DB_USER: test
```
campionfellin, line 7: One nice benefit to a lot of these now is instead of having them in plaintext in the `.travis.yml` file, they can be put into the Github secrets section and then inserted like `${{ secrets.OED_DB_TEST_USER }}`.

Then perhaps down the road a test postgres db can be set up permanently that all the tests connect to.

campionfellin:
Fixed it up so that the github actions workflow checks for the header in each file. Also changed that script to fail if it's checking only 0 files. Down to that 1 remaining test that seems to fail consistently.

huss (approved):
Everything seems good. I note a couple of related items:

- The fix in this PR also fixed the failing test on my machine. I assume it will be gone for everyone.
- I have removed the Webhook for Travis since those tests did not run (so failed) and the new GitHub actions are working fine. I have left the .travis.yml file for now. This should mean that PRs will get a green mark if all is okay. My short change to the PR did pass.

PR #549, opened 2021-02-05
Travis db fix

sasankthapa:
The issue that local builds and travis are having is that the ./postgres-data folder is being created as a non root user. So, the container is not able to access that folder.

Everything looks correct in docker-compose.yml file and I tried a bunch of other ways to create a ./postgres-data that didn't work.
One simple solution was to run the docker-compose up as root.

Maybe upgrading docker-compose version in the .yml to 3.x would work.


sasankthapa:
@huss This build did work when I first committed it and travis ran, but it seems to have broken again. 👎

PR #552, opened 2021-02-11
Adding export raw.

### LINKED ISSUE #101: Data Dump Export
Extend #22 (PR #83) by allowing users to export all available data.

sasankthapa:
This PR fixes  #101. 

Note that the export graph and the export raw will have different time with this PR until the merge for fixing time to UTC is pulled in.

This does not work for groups. @huss should open an issue regarding exactly what is "raw data" for groups, an idea would be to export each meter data that is recursive in a group.


huss:
I merged in development and all runs fine. The code worked as expected and is nice work. It is ready to merge.

huss (approved):
Nothing of significance. I made a few modest changes since @sasankthapa is busy with other things.

PR #563, opened 2021-02-22
plotly/moment localization

### LINKED ISSUE #473: Internationalization on time display
OED currently displays all date/time information in a single format. We should probably make it language/region aware. <https://momentjs.com/docs/> gives information on `moment.locale` that might be useful in doing this. It is unclear how this relates to Plotly and its formatting.

### LINKED ISSUE #523: get plotly to honor internationalization language choice
The items associated with plotly (date labels, menu hovers) do not show the language selected for OED. I believe plotly can change languages so this should be changed.

Note this relates to #473 about international time display.

LukeLavan:
# Description

fixes #473 and fixes #523 

The user-selected language choice changes the localization that Plotly and Moment use for graphs.

Each element of `xData` for line charts and bar charts uses Moment to format the date and time, so simply updating Moment's global localization setting via Moment.locale() is sufficient. This is done when the "Default Language" preference is changed (via `updateDefaultLanguage()` in src/client/app/actions/admin.ts). 

The PlotlyChart itself has its own locale specification that controls the formatting of the dates on the x-axis, as well as the utility buttons in the top-right of the chart ('Zoom In', 'Zoom Out', 'Download plot as a png', etc). This is controlled via the `config` object that's connected to the PlotlyChart and defined uniquely to each type of ChartContainer (the `DisplayModeBar` entry differs between the line and bar charts, for example). As a result, the Plotly localizations must be imported for each ChartContainer, hardcoded into the `config` object's `locales` entry, and then compared against `state.admin.defaultLanguage` (which contains the currently preferred language) to determine which value `locale` should take. These changes are made to src/client/app/containers/LineChartContainer.ts and src/client/app/containers/BarChartContainer.ts.

## Type of change

## Checklist

## Limitations

- Since the Plotly locales require import, they needed to be declared as modules - I've done so in a new file at src/client/containers/types/locales.d.ts but I'm not sure if this is the best place for these declarations.
- There is a potential for discrepancies among how OED, Moment, and/or Plotly chooses to name its localizations. For the currently implemented languages - 'en', 'fr', and 'es' - the names match. In the future, new languages should be sure to follow [the ISO standard](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for language names to avoid conflict.
- Moment is used to format Moment objects into readable strings, but the format specified may not match how the locale would prefer to format the Moment. For example, the BarChartContainer uses the string `'MMM DD, YYYY'`, but this date format is [not universal](https://en.wikipedia.org/wiki/Date_format_by_country). Moment offers [locale-aware formats](https://momentjs.com/docs/#/parsing/string-format/) and [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) [compliant strings](https://momentjs.com/docs/#/displaying/as-iso-string/) that could make these formatted dates properly internationalized. This is an easy technical change, but represents a design/presentation choice that I feel should be discussed.
- The `format()` function of Moment objects is also used for exporting data (in src/client/app/components/ExportComponent.tsx and src/client/app/utils/exportData.ts), and this will use Moment's global locale as set above. The `format()` function is also used in some javascript files (src/server/data/generateTestingData.js, src/server/services/obvius/listConfigfiles.js, and src/server/test/generateTestDataTests.js). It's unclear to me how these formattings will be affected by Moment's locale change.
- Moment's global locale is only set when the defaultLanguage is updated, so users may need to toggle between another language to get `updateDefaultLanguage()` to trigger in order to see this change reflected.


LukeLavan:
- Moment's locale-aware formatting may work with the browser data and not the locale specified? 
- The same changes should be made for the new Map graphs

I will re-open this pull-request when I've looked into these points.

LukeLavan:
I've made the necessary changes to `MapChartContainer.ts` and `CompareChartContainer.ts` for the PlotlyChart to recognize the `state.admin.defaultLanguage` as the desired locale. This was done in the same manner as was previously done for the Line and Bar charts, so the same points above apply.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -9,6 +9,8 @@ import getGraphColor from '../utils/getGraphColor';
 import { State } from '../types/redux/state';
 import PlotlyChart, { IPlotlyChartProps } from 'react-plotlyjs-ts';
 import {TimeInterval} from '../../../common/TimeInterval';
+import * as es from 'plotly.js/lib/locales/es';
+import * as fr from 'plotly.js/lib/locales/fr';
```
huss: The changes for all graphic types are the same (line, bar, compare, map). Thus, I wonder if an export/import could be created so it is all kept in one place so that future language additions would be easier to make. Not sure if it will work with all the items but something to look into.

huss (changes requested):
This works and is nice. A few overall thoughts:

- src/client/app/components/ExportComponent.tsx controls graph and raw export. src/client/app/utils/exportData.ts uses Data for raw and moment for graph. This seems to be the reason that raw export is not internationalized. This is a bit outside this PR but could it easily be fixed (or a new issue)?
- As noted in the PR, you need to change language for this to take effect. One place this causes an issue is that when you load OED, the language choice is not honored for all items. It seems the x-axis label on line is internationalized but not on bar. The hover is not. If you change the language and put it back then they are all internationalized. Since non-admins cannot change the language (yet) and you would want the language to be correct on initial load, we should see if this can be fixed.
- As for the comment on the code sometimes specifying the specific format that may not match the internationalization of moment. I think seeing what it would be like if we internationalized so moment did it the way it thought best for the region. Can this be done where the type of info displayed would be similar to now? Could you see and let us know what it might look like?

LukeLavan:
Responding to the above points, in respective order:

- It looks like the `downloadRawCSV()` function in `src/client/app/utils/exportData.ts` that was responsible for the raw export never touched Moment, which explains why the global Moment localization wasn't affecting this specifically. I've updated the function to use Moment to bring it in line with the way all other localizations are handled.
- Previously, the Moment global localization was only changed when the language preference was changed in the admin panel, which explains why it needed to be toggled between to update properly. I've added another update to the global localization in the `fetchPreferences()` function of the same `src/client/app/actions/admin.ts` file; this is called to fetch the preferences from the database when the website is initially loaded, so toggling between languages is no longer necessary if the desired language is already set in the database.
- I've updated some of the format strings passed into `Moment.format()` to use [locale-aware formatting](https://momentjs.com/docs/#/parsing/string-format/) where it was previously using hardcoded formatting. I tried to make sure that I only changed the format of strings that are actually displayed - there are a number of interally used Moments that seem to not to need internationalization as they are strictly internal values, some of which internationalizing breaks functionality for anyway (for example, the `xData` fields in `LineChartContainer` are responsible for determining the highlighted range for zoomed in redrawing, so internationalizing these strings breaks that functionality). If any such strings are deemed to be needing of internationalization (and it won't break anything in the process), the change is small enough that a later pull request can quickly address it.
- I have made the process of importing all the currently defined localizations easier via a central module in `locales.ts`, so that as more localizations are added to OED no changes should be necessary to the five containers that use `PlotlyChart`. The code that actually selects the `locale` for the `PlotlyChart`s' `config` remains spread out over each of these containers, so if any other containers are added in the future that use `PlotlyChart` they will have to implement the same method of updating `props.config.locale` to be the value of `state.admin.defaultLanguage` to honor localization. Note that if the `PlotlyChart` cannot find a good match for the specified `locale` in the given set of `locales` to select from, it will default to English. Care should be taken that the value of `state.admin.defaultLanguage` - whose possible values are (currently) derived from the hardcoded radio buttons in `src/client/app/components/admin/PreferencesComponent.tsx` - sufficiently matches a `plotly.js` localization that's enumerated in `locales.ts`. Perhaps in the future, the PreferencesComponent could be refactored to use localizations defined in `locales.ts` or both files could be refactored to reference some more central file in order to avoid any future issue arising from the mismatch of localization names.

LukeLavan:
Put some finishing touches on this, like making the CSV output of raw/nonraw exports have the same outcome format.

Note that the times output by raw export and by the regular export are NOT the same on my machine - they seem to be off by a few hours. I don't think this is something as simple as forgetting a `utc()` call somewhere, because the values of `props.exportVals.datasets[0].exportVals[0].x` (the first timestamp of the first reading of the exported data) and `props.timeInterval.getStartTimestamp()` (the beginning timestamp of the highlighted interval to be exported) are not the same. The reading values are not the same, either - the values for the raw export seem to be rounded to the nearest integer, whereas the values for the normal export are much more precise. This seems deserving of its own issue, and is beyond the bounds of internationalizing strings.

huss:
The difference between raw and graph export was my fault. I suggested a utc() shift that is not needed.

LukeLavan:
I've removed the redundant `utc()` calls. Additionally, it was elsewhere noted that the day of the week should be included with the timestamps in exported CSVs, so that has been added as well (via the longform `'dddd'` token). The filenames were also showing some inconsistency based on the specific localization in use, so I've updated the regex used to format the timestamps to be more controlling (specifically making sure commas are always omitted since this breaks the CSV formatting and ensuring that common special characters are replaced with an underscore).

huss:
The latest version does everything correctly. I verified that the raw times and daily average agree.

huss (approved):

PR #568, opened 2021-02-26
Changes to move OED to version 0.6.0

huss:
This has a few last minute fixes:

- Allow meter update when timezone not set.
- Fix unit labels on graphics.

An the final changes to move to version 0.6.0:

- Finalize SQL migrations.
- Label this as OED version 0.6.0.

The plan is to allow a few days for people to try this out before we think about making it the official version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "Open-Energy-Dashboard",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "private": false,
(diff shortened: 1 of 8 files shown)


huss:
This passes all tests and the changes are not too significant. I'm too excited to get to version 0.6.0 to wait for someone else to review. So, I'm abusing my privilege a little to merge this PR. GitHub won't let me do a review since it is my PR :-(

Issue #524, opened 2020-12-11
page layouts


huss:
As we have expanded what OED can do, we are having issues with the layout of the page. For example, the new map choice means the options often go beyond the area set aside. Maybe making this a drop-down would be the correct fix. We should make sure the docked menus do similar things. (Fixed by PR #598)

On a related item, when we change languages (see French as an example) the translated text can exceed the area reserved for the text. We should see if there is a general way to fix this.

huss:
PR #598 did the dropdown menu. The issue of it exceeding the allowed area is still there.

huss:
Pushed back to 1.0 since critical issue of map choice was fixed in PR #598.

huss:
Many of the items are scaling as expected. It appears that the ones above and below the graphic on the right side of the page do not scale correctly as show by (gotten by shrinking the horizontal size):

![Screen Shot 2022-01-20 at 2 03 10 PM](https://user-images.githubusercontent.com/987062/150413756-55268598-b2e7-4a7a-a266-4ba92a16d761.png)

where the type and language menus are above/below the the graph area. This is probably okay but would be nice to clean up and might relate to issue #466 about graphics not scaling nicely. 

The click areas to choose the days for bar graphics do not wrap/scale as other items. While not related to French, it is easier to see as the name is longer (gotten by shrinking the horizontal size):

![Screen Shot 2022-01-20 at 2 07 49 PM](https://user-images.githubusercontent.com/987062/150414281-71a55d5d-6232-4b83-af5f-281547dc4be0.png)

Another little glitch was when the horizontal size was made larger it messed up by bringing the menu below up:

![Screen Shot 2022-01-20 at 2 10 37 PM](https://user-images.githubusercontent.com/987062/150414650-e2e1d5bf-ff44-4500-b041-645e19d10d62.png)

This self corrected when the horizontal size was tweaked.

huss:
Another issue is on the compare graphic. The text rotated to be on the diagonal when the bar labels are longer but the one on the right can get cut off as show in this figure:

![Screen Shot 2022-01-23 at 11 48 34 AM](https://user-images.githubusercontent.com/987062/150691188-ca07e8aa-da1e-4b0e-bf1b-d8f4e96611cd.png)

huss:
Issue #611 is another example of this.

huss:
PR #1113 changed the page layout. Overall, some of these issues are resolved or better. However, the issue of the time selection on bar/compare/map is still there.

[...]

PR #598, opened 2021-03-03
Page layout update

LouisLu705:
The home page now features a dropdown menu to navigate chart selection in place of the old group of buttons. By clicking 'Select a graph type' a dropdown menu will appear and users can choose to navigate to 'line', 'bar', 'compare' or 'map'.

Partly address #524 

## Type of change

## Checklist

- [x] Remove text in ( ) from the issue request

## Limitations

The page layout will need to be updated in the future as more features are implemented into OED. We should also add a 'highlight feature' so that when users open the dropdown menu their current graph type is highlighted.


huss (dismissed):
removed

src/client/app/components/ChartSelectComponent.tsx
```diff
@@ -40,32 +50,37 @@ export default class ChartSelectComponent extends React.Component<ChartSelectPro
...
-					<Button
-						outline={this.props.selectedChart !== ChartTypes.map}
-						onClick={() => this.handleChangeChartType(ChartTypes.map)}
-					>
-						<FormattedMessage id='map' />
-					</Button>
-				</ButtonGroup>
+				<Dropdown isOpen={this.state.dropdownOpen} toggle={this.toggleDropdown}>
+					<DropdownToggle caret>
+						<FormattedMessage id='Select a graph type' />
```
huss: The "Select a graph type" is not language aware so it does not change with selected language.

src/client/app/components/ChartSelectComponent.tsx
```diff
@@ -40,32 +50,37 @@ export default class ChartSelectComponent extends React.Component<ChartSelectPro
 				<p style={labelStyle}>
 					<FormattedMessage id='graph.type' />:
```
huss, line 51: I'm not sure we need this label above the dropdown since it has wording inside. Maybe use the graph.type internationalized message to hold the "Select a graph type" below and replace with that.

huss (changes requested):
The dropdown menu has a gray background. All other menus are white with darker text. It partly seems okay but you also have to select a meter/group to get a first graphic so the difference is not clear to me (and maybe users). Overall, I'm inclined to remove the background unless that is difficult or others think this is better.

Overall, this does what was asked and testing indicates it works fine. Note I merged in the current development so all is fine with that.

src/client/app/components/ChartSelectComponent.tsx
LouisLu705: Please take a look at this: 

https://imgur.com/a/3XpVE6e/all

I've gone ahead and taken screenshots of what the dropdown menu would look like by itself and when users click on them with and without the label among other changes such as making the background white. Note the last two pictures are how the dropdown menu currently looks. We can proceed with any of these or I can try other suggestions as well.

src/client/app/components/ChartSelectComponent.tsx
LouisLu705: I have updated the pull request to reflect the FormattedMessage so it is language aware. 

Also, I've kept the style the same for now, but I've posted screenshots to show how the dropdown may look with the proposed ideas. Take a look at the lower conversation for the link.

src/client/app/components/ChartSelectComponent.tsx
huss: Thanks @LouisLu705. Having looked at all the options I think I like "dropdown w/ Outline and Graph Type label". This means I'm sticking with not using a background on the menu (before clicked) but changing so we keep the label. I think the consistency is valuable until we run low on vertical window space.

If you don't know, you can attach files/images with drop-n-drag to keep everything on the GitHub record.

huss (approved):

huss:
Good work and addressed everything asked. It is ready to go.

PR #603, opened 2021-03-07
Linux Compatibility

alicebeebow:
- Added a warning about use on 64-bit Linux operating systems
- Added documentation on supported Linux distros

# Description

This is a documentation addition; you should not see any conflicts with the original documents, and there are
no dependencies that need to be changed or fixed.

This did not fix any documented issue, but it does fix one I have found in my research as a student in RCOS:

I have been trying to install `docker-compose` on RHEL 7.8, so I might be able to build and test whether or not
OED is compatible with the s390x architecture.  I discovered that `docker-compose` is only compatible with 
64 bit operating systems, which is important information to know before trying to build OED.  I also included a
link to Linux distros that support `docker-compose`.

## Type of change

## Checklist

## Limitations

The list is not definitive, but it does include enough information to augment your current documentation.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -43,6 +43,14 @@ If you want to install OED so you can contribute code, see the "Development" sec
 you are setting up OED in order to gather data from power meters and display it, see the
 "Production" section.
 
+### On Linux ###
+
+In order to run OED with Docker on Linux, your operating system must be `docker-compose`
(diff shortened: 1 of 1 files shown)


huss (approved):
This seems fine and I appreciate the clarification. As noted at the top of this page, this information will be transferred when the new pages are complete.

PR #607, opened 2021-03-15
Version footer

### LINKED ISSUE #497: show OED version on admin page

limh0228:
# Description

Addresses isssue [#497](https://github.com/OpenEnergyDashboard/OED/issues/497)

This PR adds a version number of the client's currently running OED version. This was achieved by:
- changing the existing route to send out the version string instead of the version object since there's no need for the whole object
- creating a version API to retrieve the running OED version via routes from the backend
- writing a classic react-redux codepath for storing and retrieving stuff from the store
        - wrote reducers to trigger actions in the store (like writing and retrieving the version string)
        - wrote action types that define behavior when accessing the store
        - wrote actions to call the API I wrote
- wrapping the existing footer component in a footer container since it now has properties interacting with the store (as per react convention...see the HeaderComponent and HeaderContainer for examples)
        - defined dispatch and state mappings to the footer properties
- redefining the footer component as a child of the React.Component class since it now has class properties interacting with the store
- Adding relevant version strings to translations

Was all of this necessary just to show the version footer?...Well yes. These codepaths are the only way to update the redux store and keep it updated (which we will want, since the version may change). 

Fixes [#497](https://github.com/OpenEnergyDashboard/OED/issues/497)

## Type of change

## Checklist

## Limitations

N/A


huss (approved):
NIce work. This turned out to be more work than I thought and this work fixes everything needed.

PR #612, opened 2021-03-22
Added user selected language under homepage options

### LINKED ISSUE #525: user selected language
We have an admin feature where the site language is set. I think it would be valuable to allow the user to select the language they prefer. Given we don't have a login for a user, it will not be something stored in the DB. We could see if setting a cookie would allow it to last across sessions (refresh). If not, then the user will need to reset on each visit.

limh0228:
# Description

I added a dropdown underneath the graph options on the left side of the home page where the user can select their preferred language. Though language can be set from the admin page, this option is for non-admin users. The language preference does not last across refreshes since it does not log to the database.

Addresses [Issue #525](https://github.com/OpenEnergyDashboard/OED/issues/525)

Fixes #525 

## Type of change

## Checklist

## Limitations

N/A


limh0228:
Possible change: see if you can decide the language used to render the language names in the native language

limh0228:
Note to self: insert a line space above the Language selector

huss (approved):
This works as desired and has now issue. It is good work.

PR #613, opened 2021-03-22
fixed meter id showing for non-admins

### LINKED ISSUE #558: Hide meter id if not an admin
The current meter page shows the id of the meter. This is an internal value for the database. It should not be displayed if the user is not logged in as an admin. There are other values not shown in this case so this is similar.

miriammjr:
# Description

OED used to show meter ids to non-admins but now it won't.

Fixes #558 

## Type of change

## Checklist

## Limitations

None


huss (approved):
Looks good. Local testing shows no issues. Does what is expected.

PR #618, opened 2021-03-30
might have fixed the plotly issue?

### LINKED ISSUE #587: small change plotly graphic usage
Relates to LineChartContainer.ts in src/client/app/containers. https://plot.ly/javascript/hover-text-and-formatting/ has info. Seems to use line (and marker) color and wonder if we should use line since no markers seen as in:
line: {
	shape: 'spline',
	width: 3,
	color: getGraphColor(label)
},
// marker: {color: getGraphColor(label)}

miriammjr:
# Description

Uses line instead of marker for the hovering in charts. 

Fixes #587 

## Type of change

## Checklist

### DIFF
--- a/src/client/app/containers/BarChartContainer.ts
+++ b/src/client/app/containers/BarChartContainer.ts
@@ -57,8 +57,7 @@ function mapStateToProps(state: State) {
 						shape: 'spline',
 						width: 3,
 						color: getGraphColor(colorID, DataType.Meter)
-					},
-					//marker: {color: getGraphColor(colorID, DataType.Meter)}
+					}
 				});
 			}
 		}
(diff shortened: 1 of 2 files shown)


huss:
The checks failed due to linting issues. Hopefully it will pass after fixing them. These should be addressed before the PR will be reviewed.

src/client/app/containers/BarChartContainer.ts
```diff
@@ -53,7 +53,11 @@ function mapStateToProps(state: State) {
 					text: hoverText,
 					hoverinfo: 'text',
 					type: 'bar',
-					marker: {color: getGraphColor(colorID, DataType.Meter)}
+					line: {
```
huss: Since this is a bar graphic, I'm not certain we should be using the line attribute. Was there a specific reason to do this? Note the color of the bars differs from the same item on the line graphic. I don't think it used to do that.

Applies below as well.

huss (changes requested):
Does what was expected and seems to work fine. Just one comment on bar graphics to think about.

src/client/app/containers/BarChartContainer.ts
huss: fixed

huss (approved):

Issue #543, opened 2021-02-01
eliminate north map orientation


huss:
Maps in PR #502 assumes the map is oriented to north.  @huss has worked out a fairly easy fix. The planned steps are:
- [x] Change code to allow arbitrarily oriented maps. @huss will do this.
- [x] Modify DB to hold orientation angle for each map.
- [x] Create route to new values.
- [x] Modify map creation page to allow input of angle.
- [x] Modify map page to show and allow edit of angle.

huss:
@huss has a generally working version but it has the angle hard coded. Once it is in the DB this can be changed. Once an admin can set the angle on map creation this should be ready to go. Looking for someone to do the interface part.

huss:
Note the version now seems stable. It is in the mapRotation branch for anyone thinking about this. Once the proper interfaces are there I am willing to modify the code to incorporate or someone else can do this.

Issue #574, opened 2021-02-27
internationalize strings on maps


huss:
Some strings are not internationalized for maps
- [x] All on the create map page
- [ ] popup on page when upload complete on create map page
- [x] all on calibration page
- [ ] on map page: the word "edited" added to first column when a change is made
- [ ] popup to confirm delete of a map
- [x] map button for map page and map button for graphic type on main page

patrick9958:
Hi Steve (@huss),

Just a quick update. Nicole and I just finished the first and third bullet point. We are running the test suite and we're thinking of submitting a pull request tomorrow :)

Best,
Patrick & Nicole

PR #620, opened 2021-04-01
Map rotation

### LINKED ISSUE #571: map circle size
The circle size is not well determined on the maps. Should decide how to do this and implement. I think there are at least three considerations:

1. How to get all circles to be normalized to the maximum size. We could see the value for all points displayed and then make the largest value the max and scale the others. This means the circle sizes change as you change what is displayed. I think this may be okay given the scale changes on other graphics as you change what is displayed.
2. Should the circle size be by the radius or the area? I'm not sure which would make the mos sense to the user.
3. What should the maximum size circle be? We could make it a fraction of the map size which would be constant since every map has an underlying grid of 300x500. Choosing that fraction is a little trickier. It seems to me it might relate the how large the features are on the map and how close they are to each other so circles are not too small nor do they overlap. Maybe we should have a default value and then allow the admin to choose another value on the map page?

huss:
# Description

This allows for maps with arbitrary rotations and fixes the circle size issue. In more detail:

- Maps can now be rotated by an arbitrary angle from true north.
- The circle size is scaled by area and made a fraction of the smaller map dimension.
- The calibration error calculation was found to be flawed. It is now a relative error percent.
- Significant documentation was added.

Fixes #571 . Partly Addresses #543 .

## Type of change

## Checklist

## Limitations

- Both the angle and maximum circle size are hard coded. These need to be changed to come from the database/Redux state. The angle is currently set to 0 so it assumes the map is true north. The value is set in calibration.ts and must be changed manually if the map is not true north. That will go away once the DB value is used. The max circle size is set to be 0.15 for now. This value might need to be tweaked and will become the DB default value.
- The code will not work if the map crosses the line on the globe where it goes from -180 to +180 in longitude. This is a discontinuity that messes up the assumed linear nature of the GPS coordinates. It may not be too hard to fix this up but looking at the globe indicates very few parts of the world's land mass cross this point. For now it is not addressed and should be noted in the map documentation.


limh0228 (commented):
The code looks good to me. Regarding line 41, I pulled up the code with no map, and it gave a blank screen as expected. 

Perhaps the function gpsToUserGrid can be moved to calibration.ts since it is imported and referenced from multiple files, and is mostly a helper function. (Up to you, you know the code best).

I double checked the coordinate math performed for map rotation starting on line 311 of calibration.ts.

(Very much appreciating the thorough documentation). This code needs a review from someone more acquainted with maps though, as I don't understand all of this code, particularly, the changes to the function meterMapInfoOk, which seem to contain significant differences, and the code calculating scale on line 263 of calibration.ts

huss:
I refactored function as suggested by @limh0228. Good idea.

src/client/app/utils/calibration.ts
```diff
@@ -86,25 +122,82 @@ export function isValidGPSInput(input: string) {
...
+	const userGrid: CartesianPoint = rotateShift(size, gridTrueNorth, 1, -trueNorthAngle.angle);
+	return userGrid;
+}
+
+/**
+ * @param calibrationSet All the points clicked by the user for calibration.
+ * @param imageDimensions The dimensions of the original map to use from the user.
+ * @returns The error and the origin & opposite point in GPS to use for mapping.
+ */
+export function calibrate(calibrationSet: CalibratedPoint[], imageDimensions: Dimensions): CalibrationResult {
```
lindavin, line 173: I suppose I'm being nitpicky here, but I think there's supposed to be a short summary before the params like for the other functions.

src/client/app/utils/calibration.ts
```diff
@@ -200,3 +324,103 @@ export function normalizeImageDimensions(dimensions: Dimensions) {
...
+	// Now rotate the centered image by the angle given.
+	const rotated: CartesianPoint = rotate(angleDeg, point);
+	// Note you typically shift because the rotation must occur around the center of the map but the plotly coordinates
+	// have the origin at the bottom, left corner. This shift is normally the opposite of the shiftRotate one since
+	// going the other way.
+	const shifted: CartesianPoint = shift(size, rotated, direction);
+	return shifted;
+}
+
+/**
```
lindavin, line 399: For consistency, you might want to include a one line summary of the purpose of this function.

src/client/app/utils/calibration.ts
```diff
@@ -200,3 +324,103 @@ export function normalizeImageDimensions(dimensions: Dimensions) {
...
+ * @returns The origin coordinates on the true north map
+ */
+export function trueNorthOrigin(size: Dimensions): CartesianPoint {
+	// Origin coordinate on user map is (0, 0).
+	const origin: CartesianPoint = { x: 0, y: 0 };
+	// Shift this value to center and then rotate to put on true north map.
+	return shiftRotate(size, origin, -1, trueNorthAngle.angle);
+}
+
+/**
```
lindavin, line 416: For consistency, you might want to include a one line summary of the purpose of this function.

src/client/app/utils/calibration.ts
lindavin, line 403: For consistency, you might want to include a one line summary of the purpose of this function.

src/client/app/utils/calibration.ts
```diff
@@ -86,25 +122,82 @@ export function isValidGPSInput(input: string) {
...
+		x: originTrueNorth.x + (gps.longitude - originGPS.longitude) / scaleOfMap.degreePerUnitX,
+		y: originTrueNorth.y + (gps.latitude - originGPS.latitude) / scaleOfMap.degreePerUnitY
+	};
+	// Now rotate to user map and shift so origin at bottom, left. Do - angle since going from
+	// true north to user map.
+	const userGrid: CartesianPoint = rotateShift(size, gridTrueNorth, 1, -trueNorthAngle.angle);
+	return userGrid;
+}
+
+/**
```
lindavin, line 168: I suppose I'm being nitpicky here, but I think there's supposed to be a short summary before the params like for the other functions.

src/client/app/containers/MapChartContainer.ts
```diff
@@ -88,18 +158,23 @@ function mapStateToProps(state: State) {
...
 			data.push(traceOne);
+			// }
 		}
 	}
 
 	// set map background image
 	const layout: any = {
+		// Either the actual map name or text to say it is not available.
 		title: {
 			text: (map) ? map.name : 'There\'s not an available map'
```
lindavin, line 179: I'm not sure if this text ('There\\'s not an available map') is displayed in the actual UI anywhere, but it might have to be internationalized?

lindavin (changes requested):
This looks could to me.  I left some minor comments about documentation and another one about internatlizing a string.  Like @limh0228 I am not well aquainted with maps, but I don't see any glaring issues. 
Thank you for the work!

src/client/app/containers/MapChartContainer.ts
huss: Issue #574 will address this as there are a number of places that maps do not use internationalized strings.

src/client/app/utils/calibration.ts
huss: Done. Thanks.

src/client/app/utils/calibration.ts
huss: Done. Thanks.

src/client/app/utils/calibration.ts
huss: Done. Thanks.

huss:
This was reviewed by both @limh0228  and @lindavin but neither was marked as approving. It probably happened since both were doing this. Given no major issues were found and I addressed all the comments, I am going to merge this PR (I cannot approve my own review so doing as admin).

Issue #265, opened 2018-01-26
Allow administrator users to reset their passwords


NoraCodes:
Probably via email, as usual. This will require some kind of mail service integration.

huss:
We are in the process of upgrading so we have a user has a role and there will be multiple users. We will be adding a user page under admin to allow an admin to edit user info (email, password & role). This will address this issue and seems a better choice than doing it via email.

huss:
Changing a password probably will not happen often. The new admin user pages (see PR #621) allows one to do this by deleting a user and then recreating. For changing your admin password, you need to create another one, login in as that new admin, delete the other one and then recreate the one just deleted (and can delete the new admin if you want). While this is a little clunky, it was deemed fine given its limited usage.

Issue #266, opened 2018-01-26
Rate Limiting


NoraCodes:
The API executes a great deal of code on the server, and could easily be leveraged to cause a DoS condition. We should add:
- [ ] a configurable limit to API access rate, possibly 
- [ ] bypassable with a token available from the admin panel

johndiiorio:
Recommend using express-rate-limit for this purpose.

huss:
PR #633 also has a rate limiting warning from CodeQL that should be fixed.

PR #621, opened 2021-04-01
Add users

### LINKED ISSUE #544: allow multiple users with roles
We are adding the ability to receive data from Obvius and CSV files along with downloading the raw meter data. We need to have a user and password associated with these actions to make sure only authorized people do them. Having them linked to the admin risks that password and limits granularity of access. Thus, users need to be enhanced to have a role. We need an admin page to edit users and modify the current code to change login check for admin role. Also, the current setup in Obvius, CSV and export will all check this user (for export it will only be for large downloads). This is related to issues #265 and #231.

On a related note, it would also be nice to allow the installer to set the email & password for the initial admin rather than using defaults that are then reset after the install. Maybe this can be address as part of this or made a separate issue if that does not happen.

lindavin:
# Description

This PR adds user roles into OED. It adds the following user roles: Admin, Obvius, Csv, and Export to regulate usage within OED. For user role requirements, we add the:
- Admin role requirement where the user wants to POST via the groups, maps, meters, preferences, users API.
- Obvius role requirement on the obvius API.
- Csv role requirement on the fileProcessing API.
- Export role requirement on the export API.
 
We also added an interface for Admins to create new users, delete users, and modify user roles.

Fixes #544

## Type of change

## Checklist

## Limitations

- This PR does not provide a direct way for Admins to reset user passwords, which corresponds to issue #265. However, it can be indirectly achieved by creating new users and deleting old ones. For example, to change the test user's password, an Admin would login as the test user and create a new Admin user. Then they would log in with the new Admin account and delete the old test account.
- Future test code may want a set of tests that ensure that other user roles can perform a specific action. For instance, to ensure that only Admins can delete a user we add mocha tests that loop over the other user roles to perform the request and observe a failure (status code 403). It would be better if this could be generalized into one reusable function.
- We hard code the 403 HTTP Status code to indicate an insufficient role. It would be better if these status codes were standardized somewhere in OED.


src/client/app/utils/api/UsersApi.ts
```diff
@@ -0,0 +1,48 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import ApiBackend from './ApiBackend';
+import { User, UserRole } from '../../types/items';
+
+interface NewUser extends User {
+	password: string;
+}
+export default class UsersApi {
```
huss, line 15: Should have blank line before class.

src/server/models/User.js
```diff
@@ -83,6 +107,28 @@ class User {
...
+	 * @param conn the connection to use
+	 * @return {Promise<void>}
+	 */
+	static createUserTypesEnum(conn) {
+		return conn.none(sqlFile('user/create_user_types_enum.sql'));
+	}
+}
+
+/**
+ * If enum needs to be kept in sync with the src/server/sql/create_user_types_enum.sql and the UserRoles enum in src/client/types/items.ts
```
huss: Is this "If enum" or "enum" since I think this should be true.

src/server/routes/authenticator.js
```diff
@@ -15,23 +21,150 @@ authMiddleware = (req, res, next) => {
...
+ * is executed. Otherwise, the server responds with a 400 error.
+ */
+function credentialsRequestValidationMiddleware(req, res, next) {
+	const validParams = {
+		type: 'object',
+		required: ['email', 'password'],
+		properties: {
+			email: {
+				type: 'string',
+				minLength: 3,
```
huss: Just being picky but wouldn't a@b.c be the min or 5 characters?

src/server/routes/authenticator.js
```diff
@@ -15,23 +21,150 @@ authMiddleware = (req, res, next) => {
...
+				maxLength: 254
+			},
+			password: {
+				type: 'string',
+				minLength: 3
```
huss: I think OED requires at least 8 character passwords so maybe that should be the min? I would fail later so not essential.

src/server/services/user/editUser.js
```diff
@@ -22,6 +23,12 @@ const { getConnection, dropConnection } = require('../../db');
 		terminateReadline('Password must be at least eight characters, user\'s password not modified');
 	}
 
+	role = await ask('Role: '); // untested
```
huss, line 25: Why does it say "untested"? Does this need to be done via a script (sorry, did not check)?

src/server/services/user/editUser.js
```diff
@@ -11,6 +11,7 @@ const { getConnection, dropConnection } = require('../../db');
 (async () => {
```
huss, line 10: extra blank line above (not @lindavin edit but nice to fix up).

src/server/services/user/editUser.js
```diff
@@ -22,6 +23,12 @@ const { getConnection, dropConnection } = require('../../db');
 		terminateReadline('Password must be at least eight characters, user\'s password not modified');
 	}
 
+	role = await ask('Role: '); // untested
+	role = role.toUpperCase();
+	if(User.role[role] === undefined){
```
huss: Spacing issues.

src/server/test/web/groups.js
```diff
@@ -5,9 +5,11 @@
 /* This file tests the groups API. */
 
 const { chai, mocha, expect, app, testUser, testDB } = require('../common');
+const bcrypt = require('bcryptjs');
```
huss, line 8: I like the new tests. However, here and in other files, they tend to not check a valid request from an admin. I thought one way to do it would be to have an else on the check for the admin role in the loops. Another would be to put that check inside each check so less code needs to be repeated. Just and idea of how to code and don't really care how it happens.

src/server/test/web/usersTest.js
```diff
@@ -0,0 +1,121 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* This file tests the API for retrieving meters, by artificially
+ * inserting meters prior to executing the test code. */
+
+const { chai, mocha, expect, app, testDB, testUser } = require('../common');
+const User = require('../../models/User');
+const bcrypt = require('bcryptjs');
+mocha.describe('Users API', () => {
```
huss, line 12: Blank line between code and import.

src/server/util/userRoles.js
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const jwt = require('jsonwebtoken');
```
huss, line 5: Is the jwt import needed?

src/server/util/userRoles.js
```diff
@@ -0,0 +1,47 @@
...
+const User = require('../models/User');
+const promisify = require('es6-promisify');
+const jwtVerify = promisify(jwt.verify);
+const { getConnection } = require('../db');
+const secretToken = require('../config').secretToken;
+
+/** Checks if a role has the authorization capabilities as the requested role. */
```
huss: All the functions should have JSDoc with @param and @return. Esp. important since others are likely to use these.

src/server/util/userRoles.js
```diff
@@ -0,0 +1,47 @@
...
+		return false;
+	}
+}
+
+/** Checks if a user is authorized. */
+/**
+ * 
+ * @param {User} user 
+ * @param requestedRole 
+ */
```
huss, line 49: Missing return and the comment "description" should be within JSDoc as a single comment.

huss (changes requested):
This is good work and implements what is needed. A few general comments:

- When I log in in a role other than admin, I can still get to the admin, user, meter, group & map page. They show all the same items/table headers as when in the admin role but it fails when you try to do a save of changes. Do I have an issue with my install or is this a more general issue? Note if not logged in then the pages appear as expected.
- There are 5 CodeQL tests that fail. These are due to not rate limiting. This is a larger issue with OED that needs to be addressed outside this PR. Hopefully we can easily disable these until the needed changes are made so PRs will stay okay for the checks.
- I still need to check that Obvius, csv & export does what is expected in those roles but am waiting on that.

src/client/app/utils/api/UsersApi.ts
lindavin: Thanks!

src/server/test/web/usersTest.js
lindavin: Thanks!

src/server/util/userRoles.js
lindavin: Thanks!

src/server/util/userRoles.js
lindavin: Yes. We use the promisify package to turn jwt#verify into a promise.

src/server/services/user/editUser.js
lindavin: It's untested because I didn't formally check that this works (by running it or writing a mocha test). I am fairly confident that it should work as intended though. Do you think I should add a mocha test?

src/server/models/User.js
lindavin: Thanks for the catch.

src/server/services/user/editUser.js
huss: I think this is fine but the comment confused me.

src/server/util/userRoles.js
huss: Sorry, my mistake. I missed the other reference.

src/server/routes/authenticator.js
lindavin: Thanks done.

src/server/routes/authenticator.js
lindavin: Thanks done.

src/server/test/web/groups.js
lindavin: The testUser imported on line 7 is an admin by default so the other checks are an implicit check on the admin role. However, I can make an explicit check for the admin role. I like the first suggestion of the else check. I don't see how putting that check inside each check would mean that less code is repeated. Wouldn't I be adding a line to each code so it's more repetition? I'll think about it some more, but will probably go with the first suggestion.

src/server/util/userRoles.js
lindavin: Thanks done.

src/server/test/web/groups.js
lindavin: I have taken the first suggestion for groups.js. For the other tests, I rewrote 'authorized role' -> 'admin role' to be more explicit. Let me know if this is okay; I put this work in the commit 'Write explicit tests for Admin role'.

lindavin:
I have added resolved most of the comments from the review by @huss. I setup internationalization of the text, but we will need to have actual translations. I marked the text that need to be translated with a "(Needs <language>)" in the data.json. As for testing checking that a request from the admin is valid, I have now added a test for than in groups.js.  Additionally, since the testUser is an admin user, the existing tests should now fill the role of checking a valid request from an Admin.

src/server/services/user/editUser.js
huss: fixed.

src/server/services/user/editUser.js
huss: fixed

lindavin:
Regarding the issue of failing when you try to do a save of changes when not an admin, this is working as expected because only admins should. Should the interface change so that you cannot get certain pages if you are not an admin? I allowed access to those pages because ultimately the user cannot save any changes unless they are an admin.

huss:
> Regarding the issue of failing when you try to do a save of changes when not an admin, this is working as expected because only admins should. Should the interface change so that you cannot get certain pages if you are not an admin? I allowed access to those pages because ultimately the user cannot save any changes unless they are an admin.

I was looking at this right now. src/client/app/utils/token.ts has hasToken() and that is used in many places to decide if items should be displayed. I think in all these places the idea was that only an admin could see that info. I think we need to change them to use your new admin authentication so it does the same thing. Otherwise, anyone logged in with any role can see the info.

I understand that your thinking was anyone could get to the page but only admins could edit. However, I have two concerns with this. First, it exposes information to non-admins that they really should not see. Second, I think it can be confusing to the user. Yes, they would need to be logged in as some user but I don't think non-admins should see this info. This is what the help pages now show.

Is this something you want to do or should someone else take this on? I think it needs to be an update to this PR either way.

huss:
One more item. The lint checks are failing from the edit to src/server/test/web/groups.js.

lindavin:
I can take on the task of having the UI change depending on the role. Plus I want to learn a bit of redux as I don't know it. This is what I'm thinking for the implementation, I'm open to suggestions:
1) On first load (i.e. before the app mounts), check whether or not there is a valid token in `localStorage` via `checkToken`. If there is get the role via the `/users/token` route and put it into the store. Otherwise, we wait until the user has logged in to store the role into the redux store.
2) Components will check the store to see if the user is an admin and change accordingly.
3) If the roles has changed while the user is logged in, they would reload their page or log out then log in.

lindavin:
I've gone ahead with this strategy with some success.

> I can take on the task of having the UI change depending on the role. Plus I want to learn a bit of redux as I don't know it. This is what I'm thinking for the implementation, I'm open to suggestions:
> 
>     1. On first load (i.e. before the app mounts), check whether or not there is a valid token in `localStorage` via `checkToken`. If there is get the role via the `/users/token` route and put it into the store. Otherwise, we wait until the user has logged in to store the role into the redux store.
> 
>     2. Components will check the store to see if the user is an admin and change accordingly.
> 
>     3. If the roles has changed while the user is logged in, they would reload their page or log out then log in.

lindavin:
It took more work than anticipated, but now the UI appears the same to both authenticated non-Admin users and unauthenticated users. Only admins should have access to the admin, user, meter, group & map page with full information. The links to these pages are hidden when not an admin and attempts to access these pages via the url will redirect the user to the home page. When I attempted to access pages restricted to just admins as an authenticated non-Admin, I could still see the page for a split second. I do not know how to fix this. Open to suggestions, unless we are okay with that for now.

I did not replace all instances of hasToken because it is still a useful method to check whether that the user is not authenticated.

src/client/app/utils/hasPermissions.ts
```diff
@@ -0,0 +1,19 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { UserRole } from '../types/items';
+
+/**
+ * Checks if the role A has the permission of the role B.
+ */
+export function hasPermissions(A: UserRole, B: UserRole): boolean {
+	return A === UserRole.ADMIN || A === B;
```
huss: I might add this comment:
// Admins always have any other role.

src/client/app/utils/hasPermissions.ts
```diff
@@ -0,0 +1,19 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { UserRole } from '../types/items';
+
+/**
+ * Checks if the role A has the permission of the role B.
+ */
+export function hasPermissions(A: UserRole, B: UserRole): boolean {
```
huss: I can see why you named these A & B but normally only classes, etc. are capitalized. Could it be either "a" or "userA" (I prefer the second one).

src/client/app/reducers/currentUser.ts
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { CurrentUserAction, CurrentUserState } from '../types/redux/currentUser';
+import { ActionType } from '../types/redux/actions';
+
+/*
+* Defines store interactions when version related actions are dispatched to the store.
```
huss, line 9: Should this be profile related changes?

huss (changes requested):
There are a couple of small comments to clear up.

As for seeing the page briefly when not an admin, I needed to manually access that page by entering the URL. This seems unlikely and no security information is displayed. I think it is okay for now.

My only other comment is that when I'm logged in as a user other than admin and go to the meters page I still see other columns of data and the column headers are off. I think this may be related to using hasToken() instead of checking for admin role in src/client/app/components/meters/MeterViewComponent.tsx. However, in src/client/app/components/meters/MetersDetailComponent.tsx it uses loggedInAsAdmin to decide what headers to put there and this seems correct.

src/client/app/utils/hasPermissions.ts
lindavin: Thanks! Done.

src/client/app/reducers/currentUser.ts
lindavin: Thanks! Done.

src/client/app/utils/hasPermissions.ts
lindavin: Thanks! Done.

lindavin:
Thanks for the comments and the catch for the meters view! This is fixed now.

huss:
This is now ready to merge. I thank @lindavin for working on more areas than originally expected and following through to get this all done.

huss:
I want to leave a comment that this PR is causing CodeQL failures for rate limiting. This has been a known issue (see #266). CodeQL has suggested fixes that are not too hard so it is hoped to fix this soon. However, it will be done outside of this PR so it can move forward.

huss (approved):

Issue #352, opened 2018-10-04
Allow renaming meters for display


NoraCodes:
That is, meters should have a "real" name (name as reported by the hardware) and a display name. Display name should be the same as "real" name by default (or just use real name if display name is not set) but admins should be able to set the display name.

huss:
The Obvius changes now include an identifier that is designed to serve this purpose. The graphic code needs to be changed to use this name. It isn't clear if groups need this as the name is assigned by the admin when created (unlike meters that are often automatically assigned). If we want to do for parallelism that is okay but we could also add a note to groups add comments about them. Note maps have notes so that would be similar.

huss:
I thought about one more issue. The name of a meter is currently unique since it is used to look up the meter in several places. If we are going to use the identifier to identify meters (or groups) then it is likely we want to make it unique as well. This is a DB change in constraints on that column.

huss:
Previous work made the identified unique in the DB. The new CSV/pipeline allow for updates so that is one way to change the name or identifier of a meter. Issue #644 address this directly so this issue is being closed.

PR #622, opened 2021-04-04
Add 'identifier' to redux state and Meters page

### LINKED ISSUE #605: get identifier available for code and admin
### Describe the bug

@huss added identifier to Meter model but did not add it to the redux state. This needs to be changed.

This is not a bug but we need to add identifier to the meter page so admins can edit it. While we are at it, we should remove name if you are not an admin.

LukeLavan:
# Description

Adds the `identifier` field to the redux state. Previously it was added only in the Meter model.

Adds the 'identifier' column to the Meters page. If the client is not logged in, this column replaces the Name column. If the client is logged in, this is a separate column. Additionally, logged in clients are able to update the identifier in the database, similar to how other edits (eg, GPS and timezone) can be made.

Fixes #605 

## Type of change

## Checklist

## Limitations

- I noticed that the buttons on the Meters page use hardcoded strings like 'show' and 'edit'; these should probably be changed to used localized strings. (This goes beyond the scope of my issue)
- As noted in the issue, the 'name' attribute should not be sent to non-logged in clients. That is, in every instance that OED displays the meter Name to the client, it should more than likely be displaying the Identifier instead. This would require going through every usage of the `meter.name` attribute and changing the functionality to use the identifier instead - this seems like an issue that should be opened in the event that this PR is merged. (This goes beyond the scope of my issue)
- The functionality regarding being logged in or not uses the current `hasToken()` like other places do in the code. However, this will need to be changed if/when #621 gets merged. If that gets merged first then I will push changes that update what's added here to use the new tokens; otherwise, I suspect this will be covered in the scope of a different issue that will go through and change all such instances of `hasToken()`. (This goes beyond the scope of my issue)


huss:
- I added issue #623 about meter strings.
- Issue #352 is already present about renaming meters. It is worded differently but for the change to show meters identified not name.

huss (approved):
Everything looks fine, does what is expected and worked when tested. Good work.

PR #624, opened 2021-04-17
remove many calls to dropConnection

huss:
# Description

The new version of pgp does not require us to drop connections. OED recovers by doing a new connection but that is slower. Previously, many of these calls were removed but some remained. This removes most of the remaining ones except during testing where we want to reset the DB each time.

## Type of change

## Checklist

## Limitations

None known. I tried to test manually all the functions that were impacted and found no problems.

### DIFF
--- a/src/server/routes/obvius.js
+++ b/src/server/routes/obvius.js
@@ -26,7 +26,7 @@ const streamBuffers = require('stream-buffers');
 const loadLogfileToReadings = require('../services/obvius/loadLogfileToReadings');
 const middleware = require('../middleware');
 const obvius = require('../util').obvius;
-const { getConnection, dropConnection } = require('../db');
+const { getConnection } = require('../db');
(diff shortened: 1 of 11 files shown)


LukeLavan (commented):
I was able to install this on my machine fine; as usual, though, the first install failed while subsequent installs boot normally. 

Attached is the my terminal output from clean install to running.
[docker-compose-up-output.txt](https://github.com/OpenEnergyDashboard/OED/files/6332269/docker-compose-up-output.txt)

limh0228:
Same results as @LukeLavan, did not fix the issue of the initial install failing, but as usual, installing with install_args="--keep_node_modules" worked on subsequent boots.

LukeLavan (approved):

huss:
The review is not marked as approved  since reviewer does not have write access. I am accepting the review and merging as administrator.

PR #626, opened 2021-04-20
Javascriptdisabled

### LINKED ISSUE #559: warn user if JavaScript not enabled
While not common, the user can disable JavaScript in their web browser. When this is done, OED loads fine but you cannot see an meters, groups, etc. and it will not work correctly. I was thinking it might be nice to warn the user (maybe just on initial load if possible) that JavaScript is disabled. We should only do this if it does not impact the speed of OED for other users. I did add a note in the main help page saying OED need JavaScript but it seems less likely that someone will see this if they cannot get OED to run.

miriammjr:
# Description

Instead of a blank screen, OED will show a warning that JS is disabled.

Fixes #559 

## Type of change

## Checklist

## Limitations

FormattedMessage is a JS tag, so it will not work in this, so it is only in English.


miriammjr:
I don't know why edits I made on other branches were committed too or how to get rid of them

huss (approved):
Does what is asked and seems to work correctly. Thanks.

PR #627, opened 2021-04-20
Internationalizes strings on map page and calibration page

patrick9958:
# Description

This pull request allows those in the international community to more effectively use the map page and the calibration page. Specifically, this pull request internationalizes all strings on the create map page and calibration page, excluding the Choose File button on the calibration page.

Partly Addresses #574

## Type of change

## Checklist

## Limitations

The choose file button is still hard coded in English. By default you cannot change the text in an `<input type='file'/>` HTML tag. We found one possible solution that could potentially work by making the choose file button invisible and having another button’s onClick function directly communicate with the invisible choose file button. However, this would most likely require the import of react-intl’s ‘intl’ package and we have only found a possible way to do this with React Functional components. More research is needed to see how this can be done without React Functional components.


huss (approved):
This looks good and testing indicates it works as expected.

I will soon update the linked issue to indicate what items are now done.

Thank you for a good first PR.

Issue #564, opened 2021-02-23
Continued Obvius work


huss:
Now that basic Obvius works, we need to do the following:

- [ ] Test with a live system.
- [x] The code assumes each reading is 1 hour. This is not a valid assumption. Note this may be addressed in a more general way, including the upcoming CSV file acceptance, for all systems that send only one timestamp. (PR #629 fixes this)

huss:
This has been around for a while. Until a site volunteers to do this, we cannot easily test on a live system. Thus, I am closing the issue.

PR #629, opened 2021-04-26
Pipeline Update

### LINKED ISSUE #510: redundate pipeline file
This relates to the merging of PR #482. There are old files in src/server/services/ that relate to new files in pipeline. Some are: addMamacMeters.js, loadFromCsvStream.js, loadMamacReadingsFromCsvFile.js, readCSV.js, readMamacData.js, readMamacMeters.js, readMetasysData.js, updateMeters.js & also consider testing files (insertMamacReadingsFromCSVTests, metasysTests, etc.). We need to decide what should be done with these seemingly related versions.

### LINKED ISSUE #512: use of Metasys pipeline in OED code
This relates to the merging of PR #482. It seems insertMetasysData is only used in testing now. Should it be used in the actual OED code?

### LINKED ISSUE #513: first value in cumulative data
This relates to the merging of PR #482. handleCumulative drops the logical first value. We need to think about how we deal with sorted, reverse sorted & cumulative to get what we want. Dropping a value (esp. without warning) seems undesirable. This may be addressed as part of issue #370.

### LINKED ISSUE #515: missing meter on Metasys inport of data
This relates to the merging of PR #482. What happens if the meter does not exist when OED tries to create it in insertMetasysData.js? I'm concerned it will cause an issue where the code encounters but does not deal with an error. I know there are places we don't check for this, however, here this is typically data coming from another source where we are using the filename to get the meter name. Unlike the internal uses where we control the name, this seems to be a place where a mistake can really happen. I do know the original code did not check for this but I'm seeing it now.

### LINKED ISSUE #546: fixing endTime on readings when input
We have systems that send us data were they provide only one time stamp for each reading but OED stores readings with a start and end time stamp. The code often assumes each reading is 1 hour long because that is what the original system did where we received data. As we are bringing on more systems, this is not the case and must be undone. Here are some quick thoughts on this but others might be better (or there are gaps in what I say):
- Each meter could store the expected time range for a reading. This would make it easy to figure out the other time stamp for each reading. However, it would not work with a system with variable times for readings. We could note this with a special value in the meter and then try other measures to deal with it. I don't know if this is a likely situation. Also note we could allow admins to input the value and/or try to figure out from techniques below.
- We can assume the time to the next reading is the time length for the reading. This would cause mistakes if there is a gap in time between readings. It also does not work for the last reading where there is no next reading. The next bullet might help with this.
- The new pipeline has the function checkIntervals that sees if the time between readings varies. We could use a variant to figure out the interval seen in the submitted readings and use that value for all readings. It is trickier than storing in the meter but would cover the case where we don't know the meter interval and would allow some checking if the data seems to not fit the meter interval. This is likely to be less reliable if there are limited readings submitted. The extreme case of only one reading would not allow this to work at all. If it is the latest reading ever then the only way to make it work would be to use a known value. If that did not exist we would have to decide to either reject the reading (and the user might lose data) or to store it in some way to be fixed up by the admin and then added for use. This would add complexity to our code so we probably want to know if it is really needed.
- We could go to the DB to check if the gap makes sense. This is going to make getting meter values more expensive and would not work for the latest time value. I'm not sure if it is better than saving a meter value to use. However, it would allow us to see if the first reading received is not consistent in time gap with other values since we can look back to a previous reading (assuming the new readings are later in time).

Whatever we do, we need to go back and fix up existing code to use the new system. For example, the current Obvius code assumes 1 hour and that is not right.

This has become an important issue that needs to be figured out and addressed. I welcome thoughts from others.

### LINKED ISSUE #565: other meter type
We now have several meter types. Also, soon people can upload data via CSV. This means OED will logically have meters that are not one of the types we know. For now, maybe we should have a type of OTHER to represent this. We also need to think about what happens if such a meter is updateable (same for obvius) since they cannot be updated by OED. We could not allow or could just ignore this field. The later may be easier.

### LINKED ISSUE #641: CSV upload should start with DB data

PR #633 added a web page to upload CSV files with data. The form has many parameters to include in the request. The values should start as the ones stored for the requested meter rather than default values.

This only applies when uploading readings. Currently the meter name is entered at the top of the page. Some change will be needed so once the name is known the rest of the values are populated with the DB values. The checkbox that allows for creation of the meter might want to be moved to where the name input is because a created meter will not get values from the DB. There is not a set plan in place on how to update the CSV page but a complete fix will include getting the meter values into Redux state (also needed elsewhere so it might happen in another way).

Currently the page uses default values.  This is okay but not as desirable.

LouisLu705:
# Description

This pipeline change implements processing CSV files for meter and readings data in one place. This new process will check that the readings data does not violate expected constraints and outputs warnings as necessary if errors occur. Currently, this is tested on MAMAC data and works for CSV files which may come with several parameters to indicate how OED should process a certain CSV file such as: 

onlyEndTime - CSV files with onlyEndTime values given
cumulative - CSV files with readings which are cumulative
Tgap - the allowed time gap in milliseconds that a gap may occur between two readings 
Tlen - the allowed time variation in milliseconds that two readings may deviate from each other
cumulativeReset - true if the cumulative data can reset
resetStart - the start time of the allowed time range a cumulative CSV file may reset
resetEnd -  the end time of the allowed time range a cumulative CSV file may reset

Fixes #513 and Fixes #546

See @huss comments below on changes that fixes #510, fixes #512, fixes #515, fixes #565 and fixes #641.

## Type of change

## Checklist

## Limitations

This merges in the [csvPipeline](https://github.com/lindavin/OED/tree/csvPipeline). branch creataed by @lindavin. 

Currently, we expect that dates are passed in as one of two formats:

(YYYY-MM-DD H:mm) or as (MM-DD-YYYY H:mm) == (2020/4/5 0:30) or as (4/5/2020 0:30)

If the dates are in neither of these forms there may be errors. There is also a slight performance decrease due to checking both formats so if we can default to one specific moment format we may improve performance slightly. The difference I found was 11 seconds longer when testing on OneMinuteTestingData.

Also, we need to connect the user interface to the pipeline so that users can decide if their data has certain parameters such as specific cumulativeResetStart and cumulativeResetEnd times as well as onlyEndTime. Note that currently onlyEndTime is set to be false and if no cumulatveResetStart or cumulativeResetEnd times are given the values respectively default to 0:00:00.000 and 23:59:99.999 indicating that a cumulativeReset may occur from the beginning of any day to the millisecond before the day ends and the next day begins.

We may also consider counting how many errors occur and the client may use a parameter to create a maximum cap before too many errors occur and we drop the CSV file and do not insert anything into the database.

Also, we may make the error messages more descriptive for example letting the user know which reading causes the error including that particular reading value, meter id, start timestamp and end timestamp.

There is now also redundant code in the pipeline that will need to be cleared and testing will need to be done to see if we can use this new pipeline process to handle different types of meters besides Mamac such as Obvius. 

Here are some statistics from running the generatedTestData on my local machine with 8 GB of RAM.

fourDayFreqTestData:

    real    0m0.395s
    user    0m0.087s
    sys     0m0.095s

generateFourHourTestingData:

    real    0m0.922s
    user    0m0.007s
    sys     0m0.006s
generateTwentyThreeMinuteTestingData:

    real    0m5.593s
    user    0m0.000s
    sys     0m0.017s
generateFifteenMinuteTestingData:

    real    0m8.075s
    user    0m0.010s
    sys     0m0.011s
generateCosineTestingData:

    real    0m5.260s
    user    0m0.015s
    sys     0m0.000s
generateSineSquaredTestingData:

    real    0m0.181s
    user    0m0.000s
    sys     0m0.013s
generateCosineSquaredTestingData:

    real    0m0.149s
    user    0m0.000s
    sys     0m0.011s
generateVariableAmplitudeTestingData:

 - 15Freq1AmpSineTestData:
 
        real    0m16.320s
        user    0m0.000s
        sys     0m0.043s
 - 15Freq2AmpSineTestData:
 
        real    0m16.065s
        user    0m0.001s
        sys     0m0.048s
 - 15Freq3AmpSineTestData:
 
        real    0m16.329s
        user    0m0.010s
        sys     0m0.036s
 - 15Freq4AmpSineTestData:
 
        real    0m15.876s
        user    0m0.012s
        sys     0m0.034s
 - 15Freq5AmpSineTestData:
 
        real    0m15.612s
        user    0m0.001s
        sys     0m0.046s
 - 15Freq6AmpSineTestData:
   
        real    0m16.351s
        user    0m0.011s
        sys     0m0.031s
 - 15Freq7AmpSineTestData:
 
        real    0m15.882s
        user    0m0.010s
        sys     0m0.035s

generateOneMinuteTestingData:

	real    2m1.651s
	user    0m0.020s
	sys     0m0.222s


src/server/services/pipeline-in-progress/handleCumulativeReset.js
```diff
@@ -0,0 +1,39 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const moment = require('moment');
+
+/**
+ * @param {boolean} cumulativeReset true if the client expects to handle cumulativeReset
```
huss, line 9: It might be nice to tell the allowed date/time format in the JSDoc.

src/server/services/pipeline-in-progress/handleCumulativeReset.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {moment} startTimestamp start time of a current reading
+ * @returns {boolean} returns true if the current reading can be reset else returns false
+ */
+function handleCumulativeReset(cumulativeReset, resetStart, resetEnd, startTimestamp) {
+	if (!cumulativeReset) {
+		// If isCumulative is false then it is not possible to handleCumulativeReset
+		return false;
+	}
+	else {
+		let momentFormat = '';
```
huss: I think some comments here would help. The fact you are choosing between formats (I think) and what the moment creation is doing below with that format.

src/server/services/pipeline-in-progress/loadArrayInput.js
```diff
@@ -22,11 +19,18 @@ const handleCumulative = require('./handleCumulative');
  */
 async function loadArrayInput(dataRows, meterID, mapRowToModel, isCumulative, cumulativeReset, readingRepetition, conditionSet, conn) {
 	readingsArray = dataRows.map(mapRowToModel);
-	if (isCumulative) {
-		readingsArray = handleCumulative(readingsArray, readingRepetition, cumulativeReset, meterID);
-	}
-	readings = convertToReadings(readingsArray, meterID, conditionSet);
-	return await Reading.insertOrIgnoreAll(readings, conn);
+
+	// Temporary values for params
```
huss: Are these placeholders until the pipeline is able to pass them along? If so, maybe a TODO comment and a function declaration with the expected parameters with these default values?

src/server/test/web/csvPipeline/README
```diff
@@ -0,0 +1,4 @@
+This folder is need by src/server/data/automatedTestingData.js for placing
```
huss: Should this be in the directory db/data/automatedTests? It is that directory that needs to be present (I think).

Also, there are a number of csv files in that directory that were added. I believe they came from running the automated test data script. We were not planning to provide with OED and store in git but have people generate locally as needed. Thus, I think they should be removed unless there is a new usage I am not aware of.

src/server/services/csvPipeline/uploadReadings.js
```diff
@@ -0,0 +1,54 @@
...
+		false,
+		areReadingsCumulative,
+		cumulativeReset,
+		readingRepetition,
+		undefined,
+		hasHeaderRow,
+		conn
+	); // load csv data
+	// TODO: If unsuccessful upload then an error will be thrown. We need to catch this error.
+	require ('../refreshReadingViews');
```
huss: First, a comment here about the refreshReadingViews would be helpful. Second, I thought there was going to be a parameter to decide if a refresh should happen.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+ * @param {boolean} cumulativeReset true if the cumulative data can reset
+ * @param {string} resetStart a string representation of the start time a cumulativeReset may occur after. The default resetStart time is 0:00:00
+ * @param {string} resetEnd a string representation of the end time a cumulativeReset may occur before. The default resetEnd time is 23:59:99
+ * @param readingRepetition value is 1 if reading is not duplicated. 2 if repeated twice and so on (E-mon D-mon meters)
+ * @param {boolean} onlyEndTime true if the data only has an endTimestamp
+ * @param {number} Tgap the allowed time variation in milliseconds that a gap may occur between two readings 
+ * @param {number} Tlen the allowed time variation in milliseconds that two readings may deviate from each other
+ * @param {dict} conditionSet used to validate readings (minVal, maxVal, minDate, maxDate, interval, maxError)
+ */
+
```
huss: remove blank line.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+function processData(rows, meterID, isCumulative, cumulativeReset, resetStart = '0:00:00.000', 
+					resetEnd = '23:59:99.999', readingRepetition, onlyEndTime = false, Tgap, Tlen, conditionSet) {
+
+	// If processData is succesfully finished then return result = [R0, R1, R2...RN]
```
huss: successfully

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const moment = require('moment');
+const { log } = require('../../log');
+const Reading = require('../../models/Reading');
+const handleCumulativeReset = require('./handleCumulativeReset');
+
+
```
huss, line 16: remove second blank line.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+			prevReading = currentReading;
+		}
+	}
+	if (conditionSet !== undefined && !validateReadings(result, conditionSet)) {
+		log.error(`REJECTED ALL READINGS FROM METER ${ipAddress} DUE TO ERROR WHEN VALIDATING DATA`);
+		return null;
+	}
+	return result;
+}
+function isFirst(t) {
```
huss, line 410: Blank line before function. It should also have JSDoc.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+ *    reading #0 may be the cumulative value from unknown readings that may or may not have been inserted before
+ * @param {object[[]]} rows
+ * @param {number} meterID meter id being input
+ * @param {boolean} isCumulative true if the data is cumulative
+ * @param {boolean} cumulativeReset true if the cumulative data can reset
+ * @param {string} resetStart a string representation of the start time a cumulativeReset may occur after. The default resetStart time is 0:00:00
+ * @param {string} resetEnd a string representation of the end time a cumulativeReset may occur before. The default resetEnd time is 23:59:99
+ * @param readingRepetition value is 1 if reading is not duplicated. 2 if repeated twice and so on (E-mon D-mon meters)
+ * @param {boolean} onlyEndTime true if the data only has an endTimestamp
+ * @param {number} Tgap the allowed time variation in milliseconds that a gap may occur between two readings
```
huss: I'm thinking it makes more sense to make Tgap & Tlen be in seconds. The code can convert to milliseconds. The user can give values past the decimal if they want less than a second.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+	// If processData is succesfully finished then return result = [R0, R1, R2...RN]
+	const result = [];
+
+	let errMsg = '';
+
+	let meterReading = 0;
+	let meterReading1 = 0;
+	let meterReading2 = 0;
+
+	let startTimestamp = moment(0);
```
huss: Maybe comment on why this initial value.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+	let endTimestamp = moment(0);
+
+	let currentReading = new Reading(meterID, -1, startTimestamp, endTimestamp);
```
huss: I think a comment on what current and previous readings represent and how used would be very helpful.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	if (conditionSet !== undefined && !validateReadings(result, conditionSet)) {
+		log.error(`REJECTED ALL READINGS FROM METER ${ipAddress} DUE TO ERROR WHEN VALIDATING DATA`);
+		return null;
+	}
+	return result;
+}
+function isFirst(t) {
+	const E0 = moment(0);
```
huss: I think it might be better to declare this constant outside both functions and then use it here and to initialize the first times above.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+
+	// If processData is succesfully finished then return result = [R0, R1, R2...RN]
+	const result = [];
+
+	let errMsg = '';
+
+	let meterReading = 0;
+	let meterReading1 = 0;
+	let meterReading2 = 0;
+
```
huss: I think the plan was to use a value stored on the meter as the first previous reading. If this is not done because the DB is not ready then it would be good to have TODO items for each place that a change will be needed.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+			}
+			else {
+				startTimestamp = moment(rows[index - readingRepetition][1], ['YYYY/MM/DD HH:mm', 'MM/DD/YYYY HH:mm']);
+				endTimestamp = moment(rows[index - readingRepetition][2], ['YYYY/MM/DD HH:mm', 'MM/DD/YYYY HH:mm']);
+			}
+			if (isCumulative && isFirst(prevReading.endTimestamp)) {
+				readingOK = false;
+				errMsg = 'The first reading must be dropped when dealing with cumulative data.';
+			}
+			if (onlyEndTime && isFirst(prevReading.endTimestamp)) {
```
huss: Should this be an else if since the first point may have been dropped from cumulative? If not, then the errMsg should probably be +=.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+						> Tlen && !isFirst(prevReading.endTimestamp)) {
+						errMsg = 'The previous reading has a different time length than the current reading.';
+					}
+					currentReading = new Reading(meterID, meterReading, startTimestamp, endTimestamp);
+					result.push(currentReading);
+				}
+			}
+			if (!readingOK) {
+				// An error occurred, for now log it, let the user know and continue
+				log.error(errMsg);
```
huss: I think it would be good to give the reading values in the errMsg by adding here. Also, nice if told the reading number so easier to find in the file.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+					errMsg = 'The reading is not after the previous reading with only end time given so we must drop the reading.';
+				}
+				else if (startTimestamp.isBefore(prevReading.endTimestamp)) {
+					if (isCumulative) {
+						readingOK = false;
+						errMsg = 'The reading start time is before the previous endTime and cumulative so OED cannot use this reading.'
+					}
+					else if (!isFirst(prevReading.endTimestamp)) {
+						//Only treat this as a warning since the readings may be sent in a different order.
+						errMsg = 'The current reading startTime is not after the previous reading\'s end time. This is expected on the first reading ever.';
```
huss: I don't know if you need the second sentence since the first reading is excluded by the if. Should probably add that only warning since possible reading are not sent in time order.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+					meterReading1 = rows[index - 2 * readingRepetition][0]; // use this if regular
+					meterReading2 = rows[index - readingRepetition][0];
+					meterReading = meterReading2 - meterReading1; // use this ifCumulative
+				}
+				else {
+					meterReading = rows[index - readingRepetition][0];
+				}
+				// Reject negative readings
+				if (isCumulative && meterReading1 < 0) {
+					// If the value isCumulative and the first reading is negative reject unless cumulativeReset is true
```
huss: I think it might be confusing to future people because "first reading" is used in two ways. Above, when it is the first reading ever received and here where it is the prior reading for cumulative data. I might call this one prior.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+				if (readingOK) {
+					if (Math.abs(prevReading.endTimestamp.diff(prevReading.startTimestamp) - endTimestamp.diff(startTimestamp)) 
+						> Tlen && !isFirst(prevReading.endTimestamp)) {
+						errMsg = 'The previous reading has a different time length than the current reading.';
+					}
+					currentReading = new Reading(meterID, meterReading, startTimestamp, endTimestamp);
+					result.push(currentReading);
+				}
+			}
+			if (!readingOK) {
```
huss: I think this can be an else since it could be the follow up to if (readingOK) above.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+				if (isCumulative && !onlyEndTime || onlyEndTime && !isCumulative || isCumulative && onlyEndTime) {
+					meterReading = rows[index - readingRepetition][0];
+					currentReading = new Reading(meterID, meterReading, startTimestamp, endTimestamp);
+					// This currentReading will become the previousReading for the first reading if isCumulative is true
+				}
+			}
+			prevReading = currentReading;
+		}
+	}
+	if (conditionSet !== undefined && !validateReadings(result, conditionSet)) {
```
huss, line 373: I'm unsure how this function is being used since I think it comes from ./validateReadings.js but is not in a require statement. Not sure what I am missing.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+				}
+			}
+			if (readingOK && Math.abs(startTimestamp.diff(prevReading.endTimestamp)) > Tgap) {
+				if (isCumulative) {
+					readingOK = false;
+					errMsg = 'The end of the previous reading is too far from the start of the next readings in cumulative data so drop the readings.';
+				}
+				else if (!isFirst(prevReading.endTimestamp)) {
+					//Only treat this as a warning since we expect a gap in the first ever reading.
+					errMsg = 'There is a gap in time between this reading and the previous reading.';
```
huss: Should probably include that this is only a warning. Also, maybe += since I cannot convince myself this is independent from the warning above.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+				}
+			}
+			if (readingOK) {
+				//This reading can be used
```
huss: It is possible the errMsg is not empty but readingOK is true. We need to log that in this case. May need to do here or use more complex logic since there is one more warning below that happens after other possible errors and a return. However, I think it may be okay to move that check up here to make it simpler.

Also, maybe make comment that reading okay up to this point since about to do more tests.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+					if (Math.abs(prevReading.endTimestamp.diff(prevReading.startTimestamp) - endTimestamp.diff(startTimestamp)) 
+						> Tlen && !isFirst(prevReading.endTimestamp)) {
+						errMsg = 'The previous reading has a different time length than the current reading.';
+					}
+					currentReading = new Reading(meterID, meterReading, startTimestamp, endTimestamp);
+					result.push(currentReading);
+				}
+			}
+			if (!readingOK) {
+				// An error occurred, for now log it, let the user know and continue
```
huss: We are going to need some value returned so the user knows when values were dropped. This is going to be trickier since we loop over all readeings. Maybe a counter of number dropped?

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+					meterReading = rows[index - readingRepetition][0];
+				}
+				// Reject negative readings
+				if (isCumulative && meterReading1 < 0) {
+					// If the value isCumulative and the first reading is negative reject unless cumulativeReset is true
+					log.error(`DETECTED A NEGATIVE VALUE WHILE HANDLING CUMULATIVE READINGS FROM METER ${meterID}, ` +
+						`ROW ${index - readingRepetition}. REJECTED ALL READINGS`);
+					return [];
+				}
+				// To handle cumulative readings that resets at midnight
```
huss: I don't think this check is for midnight but in the allowed time range.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+				if (meterReading < 0) {
+					// if meterReading is negative and cumulative check that the times fall within an acceptable reset range
+					if (handleCumulativeReset(cumulativeReset, resetStart, resetEnd, startTimestamp)) {
+						meterReading = meterReading2;
+					}
+					else {
+						//cumulativeReset is not expected so throw an error
```
huss: This does not throw an error but logs it and causes the reading to be ignored.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+						errMsg = ('A negative meterReading has been detected but either cumulativeReset is not enabled, or the start time and end time of this reading is out of the reset range.')
+						readingOK = false;
+					}
+				}
+				if (readingOK) {
+					if (Math.abs(prevReading.endTimestamp.diff(prevReading.startTimestamp) - endTimestamp.diff(startTimestamp)) 
+						> Tlen && !isFirst(prevReading.endTimestamp)) {
+						errMsg = 'The previous reading has a different time length than the current reading.';
```
huss: include that this is a warning.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -0,0 +1,169 @@
...
+					}
+					currentReading = new Reading(meterID, meterReading, startTimestamp, endTimestamp);
+					result.push(currentReading);
+				}
+			}
+			if (!readingOK) {
+				// An error occurred, for now log it, let the user know and continue
+				log.error(errMsg);
+				readingOK = true;
+				if (isCumulative && !onlyEndTime || onlyEndTime && !isCumulative || isCumulative && onlyEndTime) {
```
huss: This is complex logic. Please add a comment to what it is trying to do. Then I can decide if it is correct.

I wonder about the second two or clauses since they both have onlyEndTime and check the other value and not the other value.

huss (changes requested):
I thank @LouisLu705 for good work on a feature with changing needs. I know this is only in a draft state but I wanted to get in some comments now to keep the process moving. @LouisLu705 and I have talked about moving the files in src/server/test/web/csvPipeline/ to be similar to the test/db/data/ directory and excluding them from the header check to avoid that issue. Some CI test issues are due to including the upcoming CSV upload work and will be dealt with in that upcoming PR.  I note that I have only done very limited testing of the pipeline and need to do more types of CSV. Finally, this PR will not be merged until the CSV upload PR has cleared due to dependency on that code.

src/server/services/pipeline-in-progress/handleCumulativeReset.js
LouisLu705: For clarification @huss do you mean the allowed date/time format of the readings data or the allowed date/time format that we expect reset time specifications to be passed in as or maybe both?

huss: I was thinking of the reset time since that is an argument to this function but more documentation is fine :-)

src/server/test/web/csvPipeline/README
LouisLu705: Thank you for this catch. This entire automatedTests directory should not be in there. Instead, csv pipeline test files should be in the src/server/services/test/web/csvPipeline directory.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: @huss Could you elaborate with a small example? I'm confused by what "the value stored on the meter" means. Is this the last reading value for a particular meter id or something else?

huss: The recently merged PR #632 added attributes to the meters table. They include reading, start_timestamp & end_timestamp. These three values are supposed to represent the last reading stored for that meter from the pipeline. As such, they should be the first value used as the previous reading for the next value(s) that come to the pipeline for that meter. The last value processed by the pipeline should always be stored back to the DB so it is ready for the next batch. Note the reading for cumulative data is the raw, cumulative value so it still works for the next time. (The value in the readings table will be the net usages for that time frame so it will differ.) The default value when a meter is created sets the value to 0, 0001-01-01 00:00:00, 0001-01-01 00:00:00. The plan was that such a date/time (so far back in time) would mean the pipeline would still work correctly for the first point. It will generate a warning due to the time gap and not use the first ever cumulative value due to that gap but that is what is desired. Note the Meter.js model should now return these values so it can be used to get these three values from the DB and store new ones back. We really should put them into Redux with other meter state but we can skip that for now if you want. Just leave a TODO if you do that. Let  me know if this is not clear or the default values do no work as hoped.

src/server/services/pipeline-in-progress/handleCumulativeReset.js
LouisLu705: Noted and done.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Done and noted.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Corrected the typo thank you.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Done and noted.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've added specifications and added a line thank you.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Ok the logic accepts ms due to how moment calculates a moment.diff(). Therefore I've gone ahead and just multiplied the value passed in by 1000. This should work if not then we may need to dig deeper.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've added comments explaining this.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've added a large comment explaining how these values are used and calculated. It should be clear but if not I will revise as deemed necessary.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I agree. I've made it a constant value outside of the function.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Done and noted. The way it's done now means that if the data is the first ever meter readings passed in for cumulative data and onlyEndTimes are used then we append the first drop error message so the user sees: 

"The first reading must be dropped when dealing with cumulative data. The first reading must be dropped when dealing with only endTimestamps."

I've also considered potentially just using the first one so that user doesn't see the second sentence but I think it makes more sense that the user sees as much information as possible. This can be changed if deemed necessary.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've gone and done this. The user will now see:

"Error parsing Reading # {}. Reading value gives {} with error message: {}".

The first curly brace is the row that the user will see the reading such as in the csv file. the second curly brace is the actual meter reading value and then the last curly brace is the error message on why this reading was dropped.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: You're right. I've changed it so that the second sentence was removed and the user will instead see:

"The current reading startTime is not after the previous reading's end time. Note this is treated only as a warning since readings may be sent out of order"

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I agree it was confusing and might be difficult for future developers. I've added more comment which hopefully make it clear what each meterReading value actually represents. 

So for example in the case of cumulative,

meterReading1 // In cumulative this is the previous raw reading
meterReading2 // In cumulative this is the current raw reading
meterReading // This is the net reading value.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: This is actually a little tricky because even though the final readingOK may be true it might be the case that the meterReading still errors out in which case the reading is dropped. 

So for example start times and end times might pass all checks and drop into if (readingOK). However, at this point we might find out that meterReading turns out to be negative and cumulative in which we case we must drop the reading and we still want to let the user know that the reading dropped even though it was ok up until this point. 

That's why I still think it's correct to leave it as an if !readingOK since if it occurs that readingOK is not true there is an accompanying error message that we should always let the client know,

HOWEVER, with the new changes to the meter model I've changed the logic so that meterReading is calculated along with start and end times in the beginning so it might make sense to move the checks out of the current if block and move it above the current if block. If we do this we can definitely do all the checks regarding meterReadings, start timestamps and end timestamps and then finally there are only two possibilities, either there is an error (readingOK == false )or no error (readingOK == true) in which case we can use an if else block. I hope this is clear but if it's not I can definitely try to make it clearer.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Apologies this an oversight. I forgot to import this js file. I've updated it so it should work correctly now. It should be used to check that the final result array which contains all the readings is well-formed.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've gone ahead and done this. However, I have some doubts because it currently never calls this warning if the reading is the first reading which is the only case (I think) we expect an unexpected gap.

Therefore, I think this warning is only called in the case there is an actual unexpected gap in readings which I think is a valid error that we might want to consider doing something with.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've moved the comment to the final place when all checks have been completed. As for the empty errMsg it should be the case that an errMsg can't be empty with readingOK not false unless the errMsg is a warning. I've added an if statement to handle this warning case here. The logic now is that:

all checks pass
 if (! (errMsg === '')){
     log (errMsg);
} 
I think this should satisfy any possible warnings and in the case there is a real error it will also be logged since all errors are couple with readingOK === false which is checked at the end of given reading number.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: You are right. I have updated this comment.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Ok thank you, I've updated this so it returns an empty array which is equivalent to processing no readings. I think this makes sense since if there are unexpected net negative meter readings that cannot be handled it would be better to not insert any into the db and let the user know so they can check their data.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Done

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I've added this comment to make it clear why this logic exists.

/*
If the data is cumulative then regardless of if it comes with end timestamps only or both end timestamps and start timestamps the first reading ever should become the previous reading. This is necessary because there are no previous readings in the db yet so we must drop the first point ever and use this first point as the first previous reading in order to begin calculating net readings since the data is cumulative.

If the data is not cumulative but there are only end timestamps then we still must drop the first reading ever in order to use that reading as the first start timestamp for the next reading. All following readings can then use the previous end timestamps as the current readings start timestamp until all further readings have been processed. 
*/

I think now though after writing this comment out, it may make more sense to instead use isFirst() to check the first reading since this is the only case this logic is necessary.

src/server/services/pipeline-in-progress/loadArrayInput.js
LouisLu705: Done.

src/server/services/pipeline-in-progress/handleCumulativeReset.js
LouisLu705: Done.

src/server/services/csvPipeline/uploadReadings.js
LouisLu705: Done.

Edit: I've noticed after some more testing that this functionality actually was recently added to a different part of the pipeline. So I am removing th refreshReadingViews from UploadReadings since it is redundant.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: I think I've completed this. I've left a TODO for the redux state since I'm not sure how to go about doing this.

src/server/services/pipeline-in-progress/processData.js
LouisLu705: Alright, I've added a loop to iterate over the reading numbers which were dropped. The idea is that in the file we might have 

HEADER ROW
READING VALUE # 1, START TIME, END TIME (READING # 0)
READING VALUE # 2, START TIME, END TIME (READING # 1)
... (READING # 3)
and so on

and if we drop for example READING #0 because the data is cumulative and this is the first-ever data added to the db we let the user know:

Dropped Reading #0

and also the user will know ahead of time when parsing the data 

"Error Parsing Reading # 0. Reading value gives READING VALUE # 1 with error message: The first reading must be dropped when dealing with cumulative data."

So the full log might be something like,

"Error Parsing Reading # 0. Reading value gives READING VALUE # 1 with error message: The first reading must be dropped when dealing with cumulative data."

Dropped Reading # 0

Now I don't know how this will work with other meter formats but I think it should be fine since ultimately each meter will contain readings and it should just be a matter of how we identify the readings which in my example is in canonical order of the CSV. Also this implicitly counts how many erros occur so we can do something with counting errors if necessary as well.

LouisLu705:
Thanks for all the comments. I have resolved most of them and left a few open since I'm not sure if some things are satisfactory as is. Some new things to note:

The pipeline now uses the previous values from the meter table to form the first previous reading which slightly increases the time it takes to parse a CSV file. The difference in time was negligible on my local machine but more testing would be nice. I did not do a heavy stress test with at most having ~20 meters in the meter table.

I also added a new parameter to take in timeSort. I noticed for the metasys tests that the CSV files were sorted in descending order instead of the expected ascending order which threw off cumulative values. In order to fix this, I've taken in the timeSort parameter to determine what to do for cumulative depending on ascending vs descending timeSort.

Also, some other things I noticed were incorrect parsing of query parameters. For example, we expected readingRepetitions (which is the number of duplicates) to be a number value but it was instead a string. Similarly, the CSV API was using the refreshReadings parameter check as a boolean when it was instead a string. I've changed it so that the tests pass and the functionality works as intended based on existing comments but there may be other instances where the query parameters are not behaving as expected.

I've also slightly changed how the CSV API deletes files. From testing it seemed to be that the files were being deleted multiple times because there is an fs.unlink call in the CSV API which is called twice regardless if the file was gzipped or not (due to the structure of the code), and also once more in the UploadReadings.js. This was causing the file to be deleted once after it was created, once more because it thought the file was unzipped, and then once more in upload readings after the CSV file is inserted into the db. It seems that even if the file is deleted multiple times it is fine in terms of the data being correctly inserted into the db but I have made some slight changes that hopefully fixes this issue.

Also, I've changed how the pipeline handles duplicates so instead of the complex logic previously it is instead a single for loop which increments a counter by the duplicate value starting from 0. This works for all my tests but more testing is desirable. I noticed in the metasys test that with a duplicate value of 2, the first reading is actually only seen once and then all the other readings are duplicated and seen twice. I believe that the for loop will still work in such a case but it has not been tested on very large duplicate values. So for example,

Unique first reading, start time, end time (0)
second reading, start time, end time (1)
duplicate second reading ... (2)
third reading, start time, end time (3)
duplicate third reading ... (4)
...
The idea is we start at 0 and if readingRepetition = 2 then we have itr 0 at 0. itr 1 at 0+2=2. itr 2 = 2+2 = 4.... and so on. If the readingReptition is 1 (there are no duplicates) then it will increment by 1 which also works. I haven't been able to come up with an example where this doesn't work yet.

Finally, I added a bunch of comment updates as well as a slight change to the output errors at the end. Big thanks to @huss for the detailed comments and observations.

If anything is confusing or unclear let me know and I will explain in further detail.

huss:
First, I want to thank @LouisLu705 for all the work on the pipeline. I don't have any additional comments on their code.

There are some CodeQL issues that will need to be addressed at some point. It might be after this PR is merged.

I'm pushed a lot of changes/commits to the PR. I kept making changes to finish tasks and it got somewhat out of hand where changes in related code outside the original PR is now included. The positive is that the pipeline and related code should now be fully integrated and working with all options. Given the scope of the work, my changes should be reviewed before they are merged with the PR. The changes are discussed in each commit but here are the highlights:

- The curl & web page now have the same abilities. This required many changes including unspecified values now default to what is stored on the meter (when appropriate) so Fixes #641; the database was modified so it has values for every value that makes sense in the pipeline; some new abilities to handle all the options; update option now correctly works with DB so functional.
- Messages from the pipeline are now returned and displayed on a web page popup or as the return HTML for curl requests. The messages were enhanced to make them more useable. A maximum size for the returned message so won't report too many issues but all go to the log.
- MAMAC and Obvius now go through the pipeline. The duplicate capability from previous work is now gone. Previous code for the MAMAC drop zone was removed. Note that the reading length assumption of the old Obvius code is now gone as pipeline can figure out reading length. Fixes #510. Partly addresses #564.
- A new meter type of other was added so automatically created meters of unknown origin have a reasonable type. Fixes #565.
- The stream input option has been removed. The new pipeline seems to handle reasonable size CSV files so stream was not worth upgrading. Note testing shows that large files can fail if you don't give JS enough memory. This seems more a developer rather than production server issue.
- The old Metasys import code was removed. The new pipeline does everything it did and more. Fixes #512. Fixes #515 as making it obsolete.
- The developer script for inputting test data now goes through the pipeline.
- Made DB have same default start/end date/time as moment(0) so they are consistent. Also changed some of the tests that happened to use this date/time and was causing failure. Many were left as not causing issue.
- A script with test CSV files is included that runs many tests against the pipeline. Note the hope is that many can be converted to Mocha/Chai tests so they become part of the continuous integration.
- Many other changes including removal of obsolete tests, few bug fixes, better error checking on some parameters, enhanced commenting, etc.

lindavin (approved):
This looks really good. I'm very glad that mamac and obvius utilize the same pipeline code now. I don't see any obvious JS errors. I also like the addition of the enums TimeSortTypesJS and BooleanTypesJS in validateCSVUploadParams. In some places we checked if a variable is equal to string 'true', for example, line 150 of /src/server/routes/csv.js. I think this could be changed to utilize BooleanTypeJS as a later TODO.

huss:
After my review of the work by @LouisLu705 and the review of my changes by @lindavin, I think this is ready to merge. The CodeQL issues will be addressed at a later time.

huss (approved):

PR #633, opened 2021-05-02
Csv pipeline

### LINKED ISSUE #370: accepting CSV data via HTTPS
There are users who want to directly load a CSV file with meter readings. In this case it is from exporting data from Johnson Controls Metasys. Note this data comes in reverse chronological order with a time stamp that is the end time of the reading and a meter value. We need to make the following changes:

- [ ] Allow user to send the information needed and CSV file via HTTPS. The request will have parameters for user name, password,  meter name, if the data is cumulative or not (not by default), if reading are duplicated and # of times (this may become obsolete) (default of not duplicated), and time interval of each reading (default means calculate from file)

The following are changes that are needed to the current code for importing Metasys CSV files:

- [ ] The previous point's time stamp is the start time for the next point. The last point in the file (oldest in these files) without a previous point will be assigned a start time by subtracting the interval (end - start) of the next to last reading from the end time of this point.  We should warn the admin of points that have different intervals. I propose do this by comparing the current interval to the previous one and warn if it is different. This is a simple and easy method and seems adequate given we don't expect variation unless something went wrong and should not be common. We should not get files with only one point but will do the same technique as the next item for cumulative data to set the interval. If the user provides the interval then this is not done and that value is used. See next item if doing cumulative data since that is also different for the last point.

- [ ] If the values are cumulative then we need to deal with the last point where we don't have a previous point in the file. In this case we will go to the DB to find the closest previous time point and use its value to subtract from the current one. We will also use the difference in time with the DB point as the interval for this point and warn if the time interval is different from the previous one (as usual). If there is no previous value then we should drop this point (this should only happen, in general, if this is the first upload) and warn admin.

Thoughts on this are welcome.

### LINKED ISSUE #398: allow admins to input new meters
We have several sites that want to enter meter data on a regular basis via CSV. Issue #370 will allow this to happen via a URL and we already allow via the admin panel. With MAMAC and Obvius we create the new meters before first acquisition of data in an automated way. It seems a good idea to allow an admin to create new meters so they can then feed in CSV data. I think this should be compatible with the new interface to allow admins to hide meters, indicate if they are automatically read, etc. Then a site should be able to create meters and input via CSV on their own. This is just the start of the idea and we may already have some code to do parts of this.

lindavin:
# Description

This PR implements the CSV Pipeline for OED. This allows OED users to upload CSV files for meter or readings data.

Fixes #370 and fixes #398. 

I would like to thank @LouisLu705 for helping me at various points in this PR.

## Type of change

## Checklist

## Limitations
- The page for uploading CSVs have inputs for all the documented parameters for the CSV Pipeline. At the moment, several of them are not functioning; however, I believe these parameters are implemented in the recent PR #629. These are:
     - Cumulative
     - Cumulative Reset
     - Cumulative Reset Start
     - Cumulative Reset End
     - Length
     - Length Variation
     - Update
- The error messages and success messages appear as pop up notification. These disappear after a period of time and it may be better to have them appear within a static error box rather than as temporary pop ups.
- The error messages from validation the parameters are generated by server depending on the type of error that has occurred. These are more useful for cURL requests, but the messages are full length strings that are concatenated together. The limitation here is that they are not translated and only in English.
- Additionally, the messages from the server are not 'ids' found in `data.json` and therefore not translatable by react-intl. 
- Many new strings are added to `data.json` that now need translation.
- At the end, the form is quite long and if more parameters are added in the future, it may get long. I think it could be better to be a multistage form.

### New CSV Page
image


src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+import * as React from 'react';
+import { Button, Col, Input, Form, FormGroup, Label } from 'reactstrap';
+import { ReadingsCSVUploadProps, TimeSortTypes } from '../../types/csvUploadForm';
+import { showErrorNotification, showSuccessNotification } from '../../utils/notifications';
+import FormFileUploaderComponent from '../FormFileUploaderComponent';
+import translate from '../../utils/translate';
+import { FormattedMessage } from 'react-intl';
+
+/** A range of values, inclusive lower bound and exclusive upper bound. */
```
huss: Might be nice to use JSDOC and state it returns an array of these values.

src/client/app/containers/RouteContainer.ts
```diff
@@ -9,18 +9,22 @@ import { State } from '../types/redux/state';
 import { changeOptionsFromLink, LinkOptions } from '../actions/graph';
 import { clearCurrentUser } from '../actions/currentUser';
 import { isRoleAdmin } from '../utils/hasPermissions';
+import { UserRole } from '../types/items';
 
 function mapStateToProps(state: State) {
 	const currentUser = state.currentUser.profile;
 	let loggedInAsAdmin = false;
+	let role: UserRole | null = null;
```
huss, line 17: spacing issue not from current PR but lets fix it (line below)

src/client/app/containers/csv/UploadCSVContainer.tsx
```diff
@@ -0,0 +1,283 @@
...
+	state = {
+		activeTab: MODE.readings,
+		uploadMetersPreferences: {
+			gzip: false,
+			headerRow: false,
+			update: false
+		},
+		uploadReadingsPreferences: {
+			createMeter: false,
+			cumulative: false,
```
huss: In src/server/sql/meter/create_meters_table.sql (lastest one in development, not yet in this PR), it has default values for these preferences for the DB. I think it might be good to make these the same.

src/client/app/containers/csv/UploadCSVContainer.tsx
```diff
@@ -0,0 +1,283 @@
...
+				duplications: value
+			}
+		}))
+	}
+	private selectTimeSort(value: TimeSortTypes) {
+		this.setState(previousState => ({
+			...previousState,
+			uploadReadingsPreferences: {
+				...previousState.uploadReadingsPreferences,
+				duplications: value
```
huss: Should this be setting the duplications state of something with time sort?

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -0,0 +1,35 @@
...
+	constructor(backend: ApiBackend) {
+		this.backend = backend;
+	}
+
+	public async submitReadings(uploadPreferences: ReadingsCSVUploadPreferencesItem, readingsFile: File): Promise<void> {
+		const formData = new FormData();
+		for (const [preference, value] of Object.entries(uploadPreferences)){
+			formData.append(preference, value.toString());
+		}
+		formData.append('csvfile', readingsFile); // It is important for the server than the file is attached last.
```
huss: than -> that

src/server/middleware/validatePassword.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { CSVPipelineError } = require('../services/csvPipeline/CustomErrors');
+
+// STUB, TODO: Validate password
+async function validatePassword(password){
```
huss: formatting/spacing

This does not seem to be used anywhere and it does not seem to be something that should be here unless a real password is used. I suspect it was from testing and should be removed but not certain.

src/server/routes/csv.js
```diff
@@ -0,0 +1,139 @@
...
+				const token = request.headers.token || request.body.token || request.query.token;
+				if (token) {
+					// If a token is found, then we will check authentication and validation via the token.
+					(await isTokenAuthorized(token, csvRole)) ? cb(null, true): cb(new Error('Invalid token'));
+				} else {
+					// If no token is found, then the request is mostly like a curl request. We require an 
+					// email and password to be supplied for curl requests.
+					const { email, password } = request.body;
+					const verifiedUser = await verifyCredentials(email, password, true);
+					if(verifiedUser){
```
huss: formatting

src/server/routes/csv.js
```diff
@@ -0,0 +1,139 @@
...
+	}
+});
+
+router.post('/readings', validateReadingsCsvUploadParams, async (req, res) => {
+	try {
+		let fileBuffer = fs.readFileSync(req.file.path);
+		if (req.body.gzip === 'true') {
+			fileBuffer = zlib.gunzipSync(fileBuffer);
+		}
+		const filepath = await saveCsv(fileBuffer, 'meters');
```
huss: Should the file be named meters or readings?

src/server/services/csvPipeline/saveCsv.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const crypto = require('crypto');
+const { CSVPipelineError } = require('./CustomErrors');
+const fs = require('fs').promises;
+const streamBuffers = require('stream-buffers');
```
huss: This and the next imports do not seem to be used.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -0,0 +1,49 @@
...
+const { log } = require('../../log');
+const Meter = require('../../models/Meter');
+const readCsv = require('../pipeline-in-progress/readCsv');
+
+/**
+ * Middleware that uploads meters via the pipeline. This should be the final stage of the CSV Pipeline.
+ * @param {express.Request} req 
+ * @param {express.Response} res 
+ * @param {filepath} filepath Path to meters csv file.
+ * @param conn
```
huss: Should say what conn is

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -0,0 +1,49 @@
...
+ * Middleware that uploads meters via the pipeline. This should be the final stage of the CSV Pipeline.
+ * @param {express.Request} req 
+ * @param {express.Response} res 
+ * @param {filepath} filepath Path to meters csv file.
+ * @param conn 
+ * @returns 
+ */
+async function uploadMeters(req, res, filepath, conn) {
+	try {
+		// const columns = Object.keys(new Meter()).slice(1); // used for the shape of the csv
```
huss: Should this commented out code remain?

src/server/services/csvPipeline/validateCsvUploadParams.js
```diff
@@ -0,0 +1,181 @@
...
+		update: 'false'
+	},
+	meters: {
+	},
+	readings: {
+		createMeter: 'false',
+		cumulative: 'false',
+		cumulativeReset: 'false',
+		cumulativeResetStart: '0:00:00',
+		cumulativeResetEnd: '23:59:59',
```
huss: The DB create_meters_table.sql has slightly different defaults. We should use those since they were the final ones decided. Sorry for the change while you were doing this work.
cumulative_reset_end TIME DEFAULT '23:59:59.999999',
I don't know if these defaults makes sense here but we do need to figure out how they related to DB defaults:
    reading_length TIME DEFAULT '00:00:00',
    reading_variation TIME DEFAULT '23:59:59.999999',

src/server/services/csvPipeline/validateCsvUploadParams.js
```diff
@@ -0,0 +1,181 @@
...
+		duplications: '1',
+		length: '',
+		lengthVariation: '',
+		refreshReadings: 'false',
+		timeSort: 'increasing'
```
huss: I assume this relates to TimeSortTypes but since it is js and not ts cannot be used. If so, a comment that they must (?) be the same would be useful.

src/server/test/web/csvPipelineTest.js
```diff
@@ -0,0 +1,107 @@
...
+		);
+
+		const conn = testDB.getConnection();
+		const dbMeters = await Promise.all(csvMeters.map(async (meter, idx) => {
+			const dbMeter = await Meter.getByName(meter.name, conn);
+			csvMeters[idx].id = dbMeter.id;
+			return meter;
+		}));
+		expect(dbMeters).to.deep.equal(csvMeters);
+		expect((await Meter.getAll(conn)).length).to.equal(3);
```
huss: Might be better to check the length before the deep.equal to avoid longer error message and the length is a more basic test.

src/server/test/web/csvPipelineTest.js
```diff
@@ -0,0 +1,107 @@
...
+const readingsPath = `${__dirname}/csvPipeline/sampleReadings.csv`;
+const readingsBuffer = fs.readFileSync(readingsPath);
+const zippedReadingsBuffer = zlib.gzipSync(readingsBuffer);
+const metersPath = `${__dirname}/csvPipeline/sampleMeters.csv`;
+const metersBuffer = fs.readFileSync(metersPath);
+const zippedMetersBuffer = zlib.gzipSync(metersBuffer);
+const metersPathWithHeader = `${__dirname}/csvPipeline/sampleMetersWithHeader.csv`;
+const metersWithHeaderBuffer = fs.readFileSync(metersPathWithHeader);
+const zippedMetersWithHeaderBuffer = zlib.gzipSync(metersWithHeaderBuffer);
+
```
huss, line 29: The tests are fine. Would it be good to add basic test for unzipped file (meter and readings)?

huss (changes requested):
I have done testing of the new code and it works fine. I did not test all the options on upload since the new pipeline is needed for many of them. This is good work. A few overall comments:

- The /tmp/uploads/ files are not deleted. I cannot recall if I said to keep them but I don't think it is a good idea. We don't want to fill up the disk with CSV files. Thus, they should be deleted.
- I'm not positive but I think files may also be left in that directory when a failure occurs. For example, I saw “sampleReadings.csv-1620914667189.octet-stream”. I have not looked more into this.
- I merged with development since there was a conflict. A few changes were made to the test CSV files and pushed back to this PR.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
lindavin: Thanks done.

src/client/app/containers/RouteContainer.ts
lindavin: Thanks done.

src/client/app/utils/api/UploadCSVApi.ts
lindavin: Thanks, done.

src/server/middleware/validatePassword.js
lindavin: Thanks for the catch. I have deleted it.

src/server/routes/csv.js
lindavin: Thanks, done.

src/server/routes/csv.js
lindavin: It should be named readings. Thanks for the catch.

src/server/services/csvPipeline/saveCsv.js
lindavin: Thanks I have deleted them.

src/server/services/csvPipeline/uploadMeters.js
lindavin: Thanks, done.

src/server/services/csvPipeline/uploadMeters.js
lindavin: Thanks, I have removed it.

src/server/services/csvPipeline/validateCsvUploadParams.js
lindavin: Okay done.

src/client/app/containers/csv/UploadCSVContainer.tsx
lindavin: You're right, thanks for the catch. I have fixed it.

src/server/services/csvPipeline/validateCsvUploadParams.js
lindavin: Thanks, done.

lindavin: As for the DB defaults of reading_length and reading_variation, I suppose these would correspond to the parameters length and lengthVariation. However, I have not change the default values for length and lengthVariation since the devDocs has specified certain defaults.

lindavin: I'm not so sure what to do about reading_length and reading_variation. Wouldn't they appear in the meters csv file and so we do not have to specify them as parameters here?

src/server/test/web/csvPipelineTest.js
lindavin: Thanks for the suggestion! Done.

src/server/test/web/csvPipelineTest.js
lindavin: Okay done.

src/client/app/containers/csv/UploadCSVContainer.tsx
lindavin: Thanks, done.

lindavin:
Thank you for the review. I have addressed and resolved all of the comments. I left one of the threads unresolved because although I believe I fixed it, I'm not sure what to do about readings_length and reading_variation. I also don't know how to address the CodeQL alerts/errors.

src/server/services/csvPipeline/validateCsvUploadParams.js
huss: I agree that the meter will often have the values. Thinking about that, it would probably be best to load the meter values into the form but I think that is something for the future given it would be an update once the meter name is known. Given the DB will either have a set value or the default value, we should let it be used unless a specific value is given in the form/curl request. Thus, I have changed my mind and now think we should not have default values in the form for these values so the pipeline will use the DB defaults unless a new value is given. Thus, src/client/app/utils/csvUploadDefaults.ts would not set a default for cumulative for those 4 values. I think the other defaults are fine. Unfortunately we don't have duplication in the DB but we may change how it works so that is probably okay at this point (making default 1). Does this seem okay as a change?

huss:
If I am uploading a non-zip file, then I get message such as:

> web_1       | POST /api/csv/meters 200 234.403 ms - 55
> web_1       | [INFO@2021-05-17T17:03:19.621Z] Successfully deleted the uploaded file /usr/src/app/src/server/tmp/uploads/csvPipeline/2021-05-17T17:03:19.526Z-1d76-sampleMeters.csv.
> web_1       | 
> web_1       | [ERROR@2021-05-17T17:03:19.627Z] Failed to remove the file /usr/src/app/src/server/tmp/uploads/csvPipeline/2021-05-17T17:03:19.526Z-1d76-sampleMeters.csv.
> web_1       | Stacktrace: 
> web_1       | Error: ENOENT: no such file or directory, unlink '/usr/src/app/src/server/tmp/uploads/csvPipeline/2021-05-17T17:03:19.526Z-1d76-sampleMeters.csv'

It happens for readings and meters; also for form or curl. For zip files, it seems to delete two different files and works fine. I have not tried to track this down any more than running the code.

huss:
I'm having an issues (I feel as if it was not happening before the recent changes but not certain) where the refresh of readings does not seem to be happening. Not sure what is up.

huss:
I have reviewed and tested the recent changes. All seems good. Just a few comments already posted.

src/server/services/csvPipeline/validateCsvUploadParams.js
lindavin: This change sounds okay. To clarify, does this mean that we won't have the cumulative values in the request anymore as it would be known in the database?

huss: Sorry for the confusion. What I think might be best is to allow the user to input a value (to override what is in the DB) but not to give default values in the form. Thus, I expect the pipeline to use the DB  values if nothing is given in the form but use any provided values as a override. Another consequence of this change is that the form and curl requests will act similarly for these values.

lindavin:
I have fixed the issue regarding the following comment:

> If I am uploading a non-zip file, then I get message such as:
> 
> > web_1       | POST /api/csv/meters 200 234.403 ms - 55
> > web_1       | [INFO@2021-05-17T17:03:19.621Z] Successfully deleted the uploaded file /usr/src/app/src/server/tmp/uploads/csvPipeline/2021-05-17T17:03:19.526Z-1d76-sampleMeters.csv.
> > web_1       |
> > web_1       | [ERROR@2021-05-17T17:03:19.627Z] Failed to remove the file /usr/src/app/src/server/tmp/uploads/csvPipeline/2021-05-17T17:03:19.526Z-1d76-sampleMeters.csv.
> > web_1       | Stacktrace:
> > web_1       | Error: ENOENT: no such file or directory, unlink '/usr/src/app/src/server/tmp/uploads/csvPipeline/2021-05-17T17:03:19.526Z-1d76-sampleMeters.csv'
> 
> It happens for readings and meters; also for form or curl. For zip files, it seems to delete two different files and works fine. I have not tried to track this down any more than running the code.

Thanks for the catch.

As for the refresh of readings, I noticed a mistake in an `if` statement and have corrected it in the recent commit afe5035. There is probably a better way to do branching with parameters than to check against `'true'`, but maybe we can leave that as a future task. This _should_ fix any issues with refreshReadings and call it when specified. 

Moreover, I might be be misunderstanding the purpose of refreshReadings, but I think I can still view uploaded data for previous days even when I do not specify `refreshReadings=true` in my curl request.

src/server/services/csvPipeline/validateCsvUploadParams.js
lindavin: Thanks for clarifying. This is interesting because that means that the cumulative options are optional. Now that I think about it, checkboxes may give the illusion that not ticking means that option is ignored, but they are implicitly saying `false`. I will think more on whether or not I should use other button types. Maybe a radio button for true/false or a select button to _really_ ignore the option and rely on server/db defaults.

lindavin: This is interesting because it means that the cumulative checkboxes need to represent three values (true,false,meter value) rather than just two values (true,false). I might have to change the button type to more appropriately serve this purpose. I will think more about it.

lindavin: I decided not to change the check boxes. It would easier to just tick the check boxes depending on the value of the meter served from the DB, which would be a future task. The default value is undefined for parameters whose default should be read from the DB. I elaborate in `/src/server/services/csvPipeline/validateCsvUploadParams.js` in the recent commit 22a2252. Otherwise, done.

lindavin:
I have addressed the comments since the second review. I left some TODOs in 22a2252, regarding how we should read `cumulativeResetStart` and `cumulativeResetEnd` from the DB and how the cumulative values should be served from the database and populate the form.

huss:
I pushed fixes for a couple of small items. Everything seems to now be working fine and all comments were addressed. The CodeQL issues will be addressed outside this PR. Thanks to @lindavin for the good work and getting everything ready.

huss (approved):

Issue #547, opened 2021-02-04
translate new strings


huss:
As we have added items to the user pages, we added strings. We need to verify that the translations are correct and missing ones are provided. Known ones that are coming with 0.7 are:

- [x] maps (note more are coming as we international all the map strings)
- [ ] help
  - [x] including the link text of "here" in tooltips (and the other text)
- [x] timezone
- [x] user page (see src/client/app/components/admin/UsersDetailComponent.tsx)

We are happy to help anyone with how to do these translations. Doing it is fairly easy if you know both languages.

huss:
Note the Spanish translations are done (PR #658) so only French is still needed.

huss:
I'm close this as we have lots of new strings for v1.0 so this is now a different issue.

PR #648, opened 2021-08-18
version with tooltip

huss:
# Description

To keep the help system correct and easy to edit new versions, the tooltips now incorporate the version. For example, what used to be help/index.html is not help/<version>/index.html where <version> is v0.6.0, v0.7.0 or some later version to be created. Users are automatically pointed to the version of OED that they are running. The version of OED is bumped to 0.7.0 since this more accurately reflects the code status for help.

A few new tooltips were added for new items in v0.7.0.

Note the web pages were already updated to handle this change.

Thanks to @lindavin who contributed to getting this working as indicated in the commit message.

## Type of change

## Checklist

## Limitations

None known


src/client/app/components/TooltipHelpComponentAlternative.tsx
```diff
@@ -5,54 +5,109 @@
...
+						if (dataTip === null) {
+							return;
+						}
+						// Create links
+						const values = helpLinks[dataTip] || {}; // This is in case the help tip does not have any links.
+						const links: Record<string, JSX.Element> = {};
+						Object.keys(values).forEach(key => {
+							const link = values[key];
+							links[key] = version ? (<a target='_blank' rel='noopener noreferrer' href={link}>
+								here
```
lindavin, line 94: I just realized this as a mistake but 'here' will need to be internationalized.

lindavin (approved):
This is good work. The links work well. The only thing is that the 'here' text for the link should be internationalized. This seems minor and I can put this in later as a TODO, so I approve this PR.

src/client/app/components/TooltipHelpComponentAlternative.tsx
huss: Good point. There already was issue #547 about translating the changes for v0.7. It included about the help changes but I added an explicit sub-bullet about the tooltip help and the link.

Issue #650, opened 2021-08-28
should readings in DB be 32 or 64 bit real number?


huss:
### Is your feature request related to a problem? Please describe.

PR #357 changed the stored readings values from integer to real. It uses real (32-bit values) whereas other places in OED used float (64-bit values). Except for space and access speed changes, it would seem float is better. One concern is that we actually truncate readings input into OED if they exceed the real accuracy. A second is that we do calculations on the data so the final error in what is shown could be larger. This is an issue to start a discussion of whether having only 6 decimal digits of reading accuracy is an issue for sites, and, if so, to investigate making the change to float that gives 15 digits.

### Describe the solution you'd like

In terms of space, each reading now takes up 4 + 2 * 8 (two dates at 8 bytes each) = 20 and it would become 8 + 2 * 8 = 24 or around a 20% increase. Given raw readings are usually not accessed too often and disk space is not a premium at many sites, this might be reasonable to do. However, testing of speed and disk space should be done. Also, others may see potential consequences to the change. If we decide this is a good idea, the actual change is not bad as it means changing the reading table creation and creating a migration (see below on that).

### Describe alternatives you've considered

As indicated above, this is not directly breaking anything. We could leave as is.

### Additional context

PR #651 changes the meter table to float but this takes very little space and isn't accessed too often.

Note that the change will have to be done carefully for the migration because a quick test found that Postgres will not let you alter the numeric since it is used in view calculations.

huss:
I note that the daily readings view used double precision (64-bit) and the new hourly readings view does the same. Given this, I think the readings should change to 64-bit/double precision unless someone gives a reason not to. The reading table is in src/server/sql/reading/create_readings_table.sql.

huss:
This was completed with PR #776 but did not close.

PR #651, opened 2021-08-28
change meter reading from real to float

huss:
# Description

The meter table added for v0.7.0 stores the last seen reading uploaded to OED . This value was a real. It is being changed from a real (32-bit) to a float (64-bit) for several reasons:

1. Most other places in OED involving readings use float.
2. It was much more likely that OED would truncate the value for cumulative data since this is the one place where the raw meter value is stored and cumulative data gets larger value. The next reading would then subtract from this one so the final error could be much larger. For example, it rounded to 6 digits (decimal) to store in DB and when it subtracted suppose the first 4 digits are the same as the next reading so the 6th digit is the 2nd in the final answer. This means the error could now be 5 parts in 100 or 5%.  Any large impact is exceedingly unlikely with 64-bit/15 digit accuracy.
3. Unfortunate differences could occur depending on how whether values were in a single CSV file or two that split at the value in question for cumulative data. The JS code should be using 64-bit arithmetic as it goes through readings so it acts in the way OED would if it was float in the DB. However, with the current real in the DB, if the last value of a CSV is stored on the meter and rounded to 32-bit it leads to what is described above. This means the next value input will have different arithmetic vs. if all values were in a single CSV upload.  Changing to float eliminates this.
4. It isn't common to access the reading on the meter and it isn't in time critical code. Thus, any slowdown is not significant. Also, the space used in the DB is trivial given the number of meters.

## Type of change

- [X] Note merging this changes the database configuration.

## Checklist

## Limitations

Note that no migration is being provided since this is all within the v0.7.0 release. A note will be posted to developers about what to do if they are inflight.

While this fix addresses the issue on meter readings, it does not address the parallel issue that reading values are real and not float. That is more complex but at least does not involve raw/cumulative values being stored. Issue #650 addresses that.


lindavin (approved):
Think looks good! Thanks for the work!

PR #652, opened 2021-08-30
Gps input

huss:
# Description

To properly load GPS values with meter CSV you need to validate and switch the latitude, longitude. This does that.

## Type of change

- [X] This change requires a documentation update

## Checklist

## Limitations

None. Note I will soon add this to the help documentation for v0.7.0.


huss:
Note this included the changes from PR #651 (oops) so it should be merged after that PR is merged.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -64,4 +79,39 @@ async function uploadMeters(req, res, filepath, conn) {
...
+ * @returns true if string is GPS and false otherwise.
+ */
+function isValidGPSInput(input) {
+	if (input.indexOf(',') === -1) { // if there is no comma
+		return false;
+	} else if (input.indexOf(',') !== input.lastIndexOf(',')) { // if there are multiple commas
+		return false;
+	}
+	// Works if value is not a number since parseFloat returns a NaN so treated as invalid later.
+	const array = input.split(',').map(value => parseFloat(value));
```
lindavin, line 96: I think it's cleaner to say `.map(parseFloat)` here instead of `.map(value => parseFloat(value)`.

lindavin (approved):
This looks good! I left a minor suggestion for one line of code. Thanks for the work!

src/server/services/csvPipeline/uploadMeters.js
huss: I think this is a valid point. I checked the code base and found that there are a number of places that use the syntax in this PR and not the other way. To keep it consistent, I am going to leave it here but they all could be changed in the future.

Issue #600, opened 2021-03-05
lose selected meter when go to map


huss:
### Issue

Selecting meters that cannot be shown on the map before selecting a meter on the map graph causes them to stop being graphed.

### To reproduce

1. Select a meter on any graphic type page other than map (line, bar, compare) that cannot be displayed on the map.
2. Go to the map graphic page.
3. Select a meter that can be displayed.
4. Go back to another graphic type page and only the meter that was shown on the map appears and the other meter is gone.

### Note

If you select these two meters before going to the map page then they remain. The critical step appears to be selecting a meter while on the map page. Also, groups are not impacted but that may be because we currently cannot graph them on maps.

### Expected behavior
Selected meters on any graphic should remain even if they cannot be shown on another graphic.

jina2k:
Attempted to reproduce after changing code with pull request #657 
Steps taken:
1. Select a meter on line/bar that cannot be displayed on an example map
2. Change to map display instead of line/bar and select the map that was chosen above.
3. ^Actually step 2 ends up removing the meter so this shows that the issue is still there, before continuing onto selecting another meter.

huss:
I think the meter is still there in step 3 but is grayed out as expected since it cannot be shown on the map. If you select another meter then when you go back to link it is not there. The desired behavior is that selected meters  would be grayed out on maps if cannot be shown but would still return when you go to a different graphic. We will have to figure this out later (not a high priority since unusual and does not mean the user cannot see what they want with extra effort).

huss:
See PR #718 about how this bug is gone but the menus are not working quite as we would like.

PR #657, opened 2021-10-08
Maps now work with groups

### LINKED ISSUE #569: maps should work with groups
It should work with groups in the same way it does with meters and similarly to other graphics. Note that src/client/app/containers/ChartDataSelectContainer.ts checks disabled meters but not groups so this also needs to be added.

### LINKED ISSUE #579: groups sometimes fail to show up
This is a little tricky to see. It only seems to happen for some groups when the meter and group overlap in the sense that the meter is in the group. It does not happen for all groups of this type and, so far, it has only been for one of the meters of this type. When it does happen, the group is not displayed on the bar graphic. It does not seem to matter if you add other meters but it seems important that you choose the meter first or the group that is already displayed remains. You can remove any meter other than the special one and the group does not show. However, as soon as your remove that meter from graphing the group reappears no matter when in the process of removing the meters it happens.

jina2k:
# Description
Fixes #569 
So previously, displaying a map would only work with meters, and now maps can work with groups as well, as long as you put GPS coordinates into the database directly. 
Fixes #579 , I fixed this bug by changing it to accurately pull using groupID instead of meterID
## Type of change

## Checklist

## Limitations
As mentioned above, this implementation can only be seen visually by inserting gps values directly into the database; ie; updating gps point values in the groups table. In the future, we will need to implement UI for the admin to be able to edit GPS values in the database for the groups from the website.

### DIFF
--- a/src/client/app/actions/barReadings.ts
+++ b/src/client/app/actions/barReadings.ts
@@ -44,7 +44,7 @@ export function shouldFetchMeterBarReadings(state: State, meterID: number, timeI
  * @returns {boolean} True if the readings for the given group, time, and duration are missing; false otherwise.
  */
 export function shouldFetchGroupBarReadings(state: State, groupID: number, timeInterval: TimeInterval, barDuration: moment.Duration): boolean {
-	const readingsForID = state.readings.bar.byMeterID[groupID];
+	const readingsForID = state.readings.bar.byGroupID[groupID];
 	if (readingsForID === undefined) {
 		return true;
 	}
--- a/src/client/app/actions/groups.ts
+++ b/src/client/app/actions/groups.ts
@@ -282,7 +282,9 @@ export function submitGroupInEditingIfNeeded() {
 			const group = {
 				name: rawGroup.name,
 				childGroups: rawGroup.childGroups,
-				childMeters: rawGroup.childMeters
+				childMeters: rawGroup.childMeters,
+				gps: rawGroup.gps,
+				displayable: rawGroup.displayable
 			};
 			if (creatingNewGroup(getState())) {
 				return dispatch(submitNewGroup(group));
--- a/src/client/app/containers/ChartDataSelectContainer.ts
+++ b/src/client/app/containers/ChartDataSelectContainer.ts
@@ -10,8 +10,9 @@ import { State } from '../types/redux/state';
 import { Dispatch } from '../types/redux/actions';
 import { ChartTypes } from '../types/redux/graph';
 import { SelectOption } from '../types/items';
-import { CartesianPoint, Dimensions, normalizeImageDimensions, calculateScaleFromEndpoints, meterDisplayableOnMap, meterMapInfoOk } from '../utils/calibration';
+import { CartesianPoint, Dimensions, normalizeImageDimensions, calculateScaleFromEndpoints, meterDisplayableOnMap, itemMapInfoOk } from '../utils/calibration';
 import { gpsToUserGrid } from './../utils/calibration';
+import { DataType } from '../types/Datasources';
 
 
 /* Passes the current redux state of the chart select container, and turns it into props for the React
@@ -68,7 +69,7 @@ function mapStateToProps(state: State) {
 				const scaleOfMap = calculateScaleFromEndpoints(origin, opposite, imageDimensionNormalized);
 				// Convert GPS of meter to grid on user map. See calibration.ts for more info on this.
 				const meterGPSInUserGrid: CartesianPoint = gpsToUserGrid(imageDimensionNormalized, gps, origin, scaleOfMap);
-				if (!(meterMapInfoOk({ gps, meterID: meter.value }, state.maps.byMapID[state.maps.selectedMap]) &&
+				if (!(itemMapInfoOk(meter.value, DataType.Meter, state.maps.byMapID[state.maps.selectedMap], gps) &&
 					meterDisplayableOnMap(imageDimensionNormalized, meterGPSInUserGrid))) {
 					meter.disabled = true;
 					disableMeters.push(meter.value);
--- a/src/client/app/containers/ChartDataSelectContainer.ts
+++ b/src/client/app/containers/ChartDataSelectContainer.ts
@@ -10,7 +10,7 @@ import { State } from '../types/redux/state';
 import { Dispatch } from '../types/redux/actions';
 import { ChartTypes } from '../types/redux/graph';
 import { SelectOption } from '../types/items';
-import { CartesianPoint, Dimensions, normalizeImageDimensions, calculateScaleFromEndpoints, itemDisplayableOnMap, itemMapInfoOk, MapScale } from '../utils/calibration';
+import { CartesianPoint, Dimensions, normalizeImageDimensions, calculateScaleFromEndpoints, itemDisplayableOnMap, itemMapInfoOk } from '../utils/calibration';
 import { gpsToUserGrid } from './../utils/calibration';
 import { DataType } from '../types/Datasources';
 
@@ -59,17 +59,14 @@ function mapStateToProps(state: State) {
 		// causes TS to complain about the unknown case so not used.
 		const origin = state.maps.byMapID[state.maps.selectedMap].origin;
 		const opposite = state.maps.byMapID[state.maps.selectedMap].opposite;
-		var scaleOfMap: MapScale;
-		if (origin !== undefined && opposite !== undefined) {
-			// Get the GPS degrees per unit of Plotly grid for x and y. By knowing the two corners
-			// (or really any two distinct points) you can calculate this by the change in GPS over the
-			// change in x or y which is the map's width & height in this case.
-			scaleOfMap = calculateScaleFromEndpoints(origin, opposite, imageDimensionNormalized);
-		}
 		sortedMeters.forEach(meter => {
 			// This meter's GPS value.
 			const gps = state.meters.byMeterID[meter.value].gps;
 			if (origin !== undefined && opposite !== undefined && gps !== undefined && gps !== null) {
+				// Get the GPS degrees per unit of Plotly grid for x and y. By knowing the two corners
+				// (or really any two distinct points) you can calculate this by the change in GPS over the
+				// change in x or y which is the map's width & height in this case.
+				const scaleOfMap = calculateScaleFromEndpoints(origin, opposite, imageDimensionNormalized);
 				// Convert GPS of meter to grid on user map. See calibration.ts for more info on this.
 				const meterGPSInUserGrid: CartesianPoint = gpsToUserGrid(imageDimensionNormalized, gps, origin, scaleOfMap);
 				if (!(itemMapInfoOk(meter.value, DataType.Meter, state.maps.byMapID[state.maps.selectedMap], gps) &&
@@ -87,6 +84,7 @@ function mapStateToProps(state: State) {
 		sortedGroups.forEach(group => {
 			const gps = state.groups.byGroupID[group.value].gps;
 			if (origin !== undefined && opposite !== undefined && gps !== undefined && gps !== null) {
+				const scaleOfMap = calculateScaleFromEndpoints(origin, opposite, imageDimensionNormalized);
 				const groupGPSInUserGrid: CartesianPoint = gpsToUserGrid(imageDimensionNormalized, gps, origin, scaleOfMap);
 				if (!(itemMapInfoOk(group.value, DataType.Group, state.maps.byMapID[state.maps.selectedMap], gps) &&
 					itemDisplayableOnMap(imageDimensionNormalized, groupGPSInUserGrid))) {
--- a/src/client/app/containers/ChartDataSelectContainer.ts
+++ b/src/client/app/containers/ChartDataSelectContainer.ts
@@ -81,7 +81,6 @@ function mapStateToProps(state: State) {
 			}
 		});
 		// filter groups - this is not yet done as groups do not show up on maps yet.
-		
 		sortedGroups.forEach(group => {
 			const gps = state.groups.byGroupID[group.value].gps;
 			const origin = state.maps.byMapID[state.maps.selectedMap].origin;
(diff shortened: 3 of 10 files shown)


src/client/app/actions/groups.ts
```diff
@@ -282,7 +282,9 @@ export function submitGroupInEditingIfNeeded() {
 			const group = {
 				name: rawGroup.name,
 				childGroups: rawGroup.childGroups,
-				childMeters: rawGroup.childMeters
+				childMeters: rawGroup.childMeters,
+				gps: rawGroup.gps,
```
huss, line 286: While this is fine for the long-term, I just want to note that I think that neither gps nor displayable are currently set in the group create/edit so they will remain empty until we create a group edit page.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
 			}
 		});
 		// filter groups - this is not yet done as groups do not show up on maps yet.
+		sortedGroups.forEach(group => {
```
huss, line 84: The comment above about not being done yet seems out of date. I might also add that it follows the logic for meters above so see those comments.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
...
+			const gps = state.groups.byGroupID[group.value].gps;
+			const origin = state.maps.byMapID[state.maps.selectedMap].origin;
```
huss: You did not start this but looking at the logic I think that origin & opposite cannot change while this function is running. Thus, they should probably be calculated outside the loops (after imageDimensions) so not redone with each group. If done, the same change should be applied to meter loop above.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
...
+			const opposite = state.maps.byMapID[state.maps.selectedMap].opposite;
+
+			if (origin !== undefined && opposite !== undefined && gps !== undefined && gps !== null) {
```
huss, line 86: In a similar light, the first two parts of the if on origin and opposite could be used outside the loop and then do scaleOfMap. The last two parts on gps stay here to protect the rest of the calculation.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
...
+				const scaleOfMap = calculateScaleFromEndpoints(origin, opposite, imageDimensionNormalized);
+				const meterGPSInUserGrid: CartesianPoint = gpsToUserGrid(imageDimensionNormalized, gps, origin, scaleOfMap);
```
huss: Should this be named groupGPSInUserGrid instead of meterGPSInUserGrid since I think it involves the group?

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
...
+				if (!(itemMapInfoOk(group.value, DataType.Group, state.maps.byMapID[state.maps.selectedMap], gps) &&
+					meterDisplayableOnMap(imageDimensionNormalized, meterGPSInUserGrid))) {
+					group.disabled = true;
+					disableGroups.push(group.value);
+				}
+			} else {
+				// Lack info on this map so skip. This is mostly done since TS complains about the undefined possibility.
```
huss: I'm wondering if the comment on mostly done for TS is now obsolete since disabledGroups is used below.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
...
+		sortedGroups.forEach(group => {
+			const gps = state.groups.byGroupID[group.value].gps;
+			const origin = state.maps.byMapID[state.maps.selectedMap].origin;
+			const opposite = state.maps.byMapID[state.maps.selectedMap].opposite;
+
+			if (origin !== undefined && opposite !== undefined && gps !== undefined && gps !== null) {
+				const scaleOfMap = calculateScaleFromEndpoints(origin, opposite, imageDimensionNormalized);
+				const meterGPSInUserGrid: CartesianPoint = gpsToUserGrid(imageDimensionNormalized, gps, origin, scaleOfMap);
+				if (!(itemMapInfoOk(group.value, DataType.Group, state.maps.byMapID[state.maps.selectedMap], gps) &&
+					meterDisplayableOnMap(imageDimensionNormalized, meterGPSInUserGrid))) {
```
huss: I think here and in function declaration in calibration.ts, meterDisplayableOnMap should become itemDisplayableOnMap since now used for both meter & groups. This makes it parallel to change for itemMapInfoOk.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -79,15 +81,39 @@ function mapStateToProps(state: State) {
...
 	}
 
-	const selectedGroups = state.graph.selectedGroups.map(groupID => (
-		{
-			label: state.groups.byGroupID[groupID] ? state.groups.byGroupID[groupID].name : '',
-			value: groupID,
-			disabled: false
-		} as SelectOption
-	));
+	const selectedGroups = state.graph.selectedGroups.map(groupID => {
```
huss: The group logic should come after meters so same order as all above.

src/client/app/utils/calibration.ts
```diff
@@ -65,16 +66,18 @@ export interface Dimensions {
...
  */
-export function meterMapInfoOk(meterInfo: { gps?: GPSPoint, meterID: number }, map: MapMetadata): boolean {
+export function itemMapInfoOk(itemID: number, type: DataType, map: MapMetadata, gps?: GPSPoint): boolean {
 	if (map === undefined) { return false; }
-	if ((meterInfo.gps === null || meterInfo.gps === undefined) || map.origin === undefined || map.opposite === undefined) { return false; }
-	if (!isValidGPSInput(`${meterInfo.gps.latitude},${meterInfo.gps.longitude}`)) {
-		logToServer('error', `Found invalid meter gps stored in database, id = ${meterInfo.meterID}`);
+	if ((gps === null || gps === undefined) || map.origin === undefined || map.opposite === undefined) { return false; }
+	if (!isValidGPSInput(`${gps.latitude},${gps.longitude}`)) {
+		logToServer('error', `Found invalid ${type === DataType.Meter ? 'meter' : 'group'} gps stored in database, id = ${itemID}`);
```
huss: During testing I realized that there is a bug that was not due to any change in the PR. The logToServer call returns a function. We need () at the end so it becomes:
logToServer('error', `Found invalid ${type === DataType.Meter ? 'meter' : 'group'} gps stored in database, id = ${itemID}`)();
so the function returned is executed and the logging occurs. I was able to validate this works properly for both meter & group by setting at least one of the GPS values to be too large, i.e., 400, and then running the code. It causes a popup warning too (since it is reusing the calibration functions) but I think that is okay for now since this really should not happen (a GPS input error should be caught before it is entered into the DB). It also logs to log.txt as was the desire.
Also, src/client/app/components/maps/MapCalibrationInitiateComponent.tsx has the same issue on line 93 which needs to become:
logToServer('error', `Error, map source image uploading: ${err}`)();
I searched and all other uses of logToServer are in actions that dispatch them so I think all the others are okay.

src/client/app/containers/MapChartContainer.ts
```diff
@@ -123,6 +123,59 @@ function mapStateToProps(state: State) {
 				}
 			}
 
+			for (const groupID of state.graph.selectedGroups) {
```
huss, line 126: Most of the comments from meter loop similar to this one are included. The ones on the first few lines are missing (and edit as needed for meters).

src/client/app/containers/MapChartContainer.ts
```diff
@@ -123,6 +123,59 @@ function mapStateToProps(state: State) {
...
+				const byGroupID = state.readings.bar.byGroupID[groupID];
+				const gps = state.groups.byGroupID[groupID].gps;
+				if (gps !== undefined && gps !== null) {
+					// Convert the gps value to the equivalent Plotly grid coordinates on user map.
+					// First, convert from GPS to grid units. Since we are doing a GPS calculation, this happens on the true north map.
+					// It must be on true north map since only there are the GPS axis parallel to the map axis.
+					// To start, calculate the user grid coordinates (Plotly) from the GPS value. This involves calculating
+					// it coordinates on the true north map and then rotating/shifting to the user map.
+					const meterGPSInUserGrid: CartesianPoint = gpsToUserGrid(imageDimensionNormalized, gps, origin, scaleOfMap);
```
huss: I think the variable with meter should become group here.

src/client/app/actions/barReadings.ts
```diff
@@ -44,7 +44,7 @@ export function shouldFetchMeterBarReadings(state: State, meterID: number, timeI
  * @returns {boolean} True if the readings for the given group, time, and duration are missing; false otherwise.
  */
 export function shouldFetchGroupBarReadings(state: State, groupID: number, timeInterval: TimeInterval, barDuration: moment.Duration): boolean {
-	const readingsForID = state.readings.bar.byMeterID[groupID];
+	const readingsForID = state.readings.bar.byGroupID[groupID];
```
huss, line 47: I think this is an okay change but it does not seem to relate to the map group changes. (Same thought for compareReadings.ts) Is it correct that this fixes a mistake that has been in the code for a while? I did some quick testing (not extensive) and it may be the case that issue #579 is fixed by this change. Would you be willing to see if it seems gone for you and link that issue if it is the case? (Also add comment to issue that think okay but it has had to find so closing for now.)

huss:
I think these changes may have fixed issue #600. Could you see if that is the case for you? If so, please link that issue to this PR.

huss (changes requested):
This works as expected and is good work. I have made some comments to address ideas in the code and some that I saw while reviewing the PR but not really part of it. Could you please look them over and make the changes as is appropriate.

jina2k:
> I think these changes may have fixed issue #600. Could you see if that is the case for you? If so, please link that issue to this PR.

After attempting to replicate this issue, it appears that this issue is still not fixed, so I will leave it unresolved.

huss:
> > I think these changes may have fixed issue #600. Could you see if that is the case for you? If so, please link that issue to this PR.
> 
> After attempting to replicate this issue, it appears that this issue is still not fixed, so I will leave it unresolved.

Okay. If you know how to reproduced the issue then please put the directions into issue #600 so it will be easier to continue to work on it. Thanks for being careful to determine it is still there.

src/client/app/actions/groups.ts
jina2k: Noted, I will leave it as is for now as the group edit page will be added eventually

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: changed comment to account for above code

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: took out origin and opposite for both meters and groups

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: scaleOfMap was left in the if statement as it still required origin and opposite to be defined, and I figured const variables were better than using let variables. It's also possible to have the entire foreach loop for both meters and groups to be in an if statement, and then define it this way, but I left it as is for now. This could potentially simplify the else statement where origin,opposite, or gps are undefined. Not sure which method looks better to you.

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: renamed variable name to groupGPSInUserGrid

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: I think this piece of code is still necessary, though I'm not sure the comment accurately displays the use. If the GPS coordinates are still undefined as such with groups/meter, I suppose it is still valid behavior to disable the group/meter from being displayed.

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: renamed function to itemDisplayableOnMap

src/client/app/containers/ChartDataSelectContainer.ts
jina2k: moved the code to account for the meter -> group order

src/client/app/utils/calibration.ts
jina2k: added () to both function calls as mentioned

src/client/app/containers/MapChartContainer.ts
jina2k: added comments and edited it for groups

src/client/app/containers/MapChartContainer.ts
jina2k: renamed variable name

src/client/app/actions/barReadings.ts
jina2k: issue #579 is fixed supposedly, if my recreation of the bug is right. I will add it into this pull request as well.

huss:
> > > I think these changes may have fixed issue #600. Could you see if that is the case for you? If so, please link that issue to this PR.
> > 
> > 
> > After attempting to replicate this issue, it appears that this issue is still not fixed, so I will leave it unresolved.
> 
> Okay. If you know how to reproduced the issue then please put the directions into issue #600 so it will be easier to continue to work on it. Thanks for being careful to determine it is still there.

I think I was just lucky or got it wrong, but I agree the bug is still there.

huss (approved):

huss:
The changes are great and this PR is ready to go. Thanks for another good PR.

PR #660, opened 2021-10-14
Fast pts

### LINKED ISSUE #583: API for sub day data
fast-pt creates a materialized view of the day data. This makes it very fast when longer time frames are desired. It switches to another view (see issue #519 for another issue on this) when the day data does not have enough points. I think OED would benefit from an API call that can deal with non-day averaging of data in a rapid way. The idea is that you determine the full days involved and use the day data for that. Any remaining hours would be handled with the raw meter (or hourly) data. For example, if you needed the usage from 9:00 on 1/2 to 15:00 on 1/21, you would use the day data to get the values for 1/3-1/20 and then calculate 9:00-24:00 on 1/2 and 0:00-15:00 on 1/21 separately and then combine the result. Since the worst case is two days of raw readings (if both the edge days are virtually full days), you should not be looking at many raw meter or hourly reading values. Note you can take it a step further by first using the hourly table for the extra time and then raw for the little bit left over. There have been some proposals to uses this type of data but I don't think OED yet has one that is available to users. Thus, this isn't yet a high priority item.

lindavin:
# Description

This PR addresses the performance issues of switching between views by materializing the hourly view and replacing the minute views with serving raw data. This removes the need to use intermediate views (4-hour, 15-minute, etc...) The rationale behind this is that the improvement in speed outweighs the increased space costs. The motivating data are located in the DevDocs under fast-ptAnalysis/fastPtSystemBenchmarkData.md.

Fixes #583 

## Type of change

## Checklist

## Limitations
The remaining work is also located DevDocs under fast-ptAnalysis/fastPtSystemBenchmarkData.md under section **Remaining Work** and are restated below:

- Though not recorded here, we observed a somewhat **linear** increase in time depending on the **amount of data** in the database. This may suggest a problem in the indexing of the readings or meters tables.
- Closely check the impact on size of materialized view. When does size become an issue? The space data up by an empty is 57.3 MB.
- Rerun these tests on:
  - (Edge case) A large set of data (i.e. 10 years of data)
  - Different reading frequencies (23-minute, daily, etc...). We do not expect to see significant differences, but this should be done for completeness.
- Check the amount of time it takes to refresh (materialized) hourly readings. This is important because hourly readings view will have 24 times the data in the daily readings view. We are considering to use an hourly CRON job to update the hourly view, but a different strategy may be needed If refreshing takes a long time.
- Design and experiment with an intelligent data serving strategy that takes into account the expected length of readings, the max/expected reading length variations, and whether data is prone to having holes.


src/client/app/containers/csv/UploadCSVContainer.tsx
```diff
@@ -189,6 +190,16 @@ export default class UploadCSVContainer extends React.Component<{}, UploadCSVCon
 		}
 	}
 
+	private toggleRefreshHourlyReadings() {
+		this.setState(previousState => ({
+			...previousState,
+			uploadReadingsPreferences: {
+				...previousState.uploadReadingsPreferences,
+				refreshReadings: !previousState.uploadReadingsPreferences.refreshHourlyReadings
```
huss: I think this should this be refreshHourlyReadings:. I made the change and the click button worked properly but would not click before that.

src/server/migrations/0.7.0-0.7.1/sql/readings/create_materialized_hourly_readings.sql
```diff
@@ -0,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+DROP VIEW IF EXISTS hourly_readings;
```
huss, line 5: This drops the view in migraation but src/server/migrations/0.3.0-0.5.0/sql/fast-pt-compression/create_compressed_reading_views.sql for 0.5.0 for the original version did a create or replace. Curious about the choice and if you think one is better.

src/server/models/Reading.js
```diff
@@ -95,6 +96,17 @@ class Reading {
 		return conn.none('REFRESH MATERIALIZED VIEW daily_readings');
 	}
 
+	/**
+	 * Refreshes the compressed hourly readings view.
+	 * Should be called at least once a day, preferably in the middle of the night.
```
huss: Maybe this one should say:
Should be called at least once a day but need to do hourly if want zooming in to see hourly data as it is available. This can take more time than refreshing the daily readings so be sure calling it more frequently does not impact the server response time. If only call once a day then probably best in the middle of the night as suggested for daily refresh.

src/server/sql/reading/create_compressed_reading_views.sql
```diff
@@ -149,31 +155,60 @@ CREATE INDEX if not exists idx_daily_readings ON daily_readings USING GIST(time_
...
+		which translates to 0.60 megabytes per meter. 
+		
+
+		+------+-------------------
+		| TODO |
+		+------+
+		- Note that the above documentation does not account for the increase in data when packaged in JSON format.
+		- This function is complicated by the fact that groups can contain meters with different levels of granularity.
+		We address this now by having sites pick a site-level reading frequency. (Not done)
+		- When resource generalization is implemented in OED, this function will need to be done in a
```
huss: The design for resource generalization is nearing completion. The current design might add parameters for the transformation desired but I don't think it will necessarily go to meter by meter since we can group meters with the same reading unit together for getting the values. Also, a single meter is a special case of a set of meters and it isn't clear it slows this down to treat it like that. Another option is to continue to get the data the same way and convert on the server. It is unclear which will be better. Finally, I'm unclear what the "different levels of granularity means". Do you mean different reading rates? If so, I'm not sure that impacts the generaliztion but maybe I'm missing something. Mostly, I want to understand your thinking so we get resource generalization correct.

As for the comment, maybe just say resource generalization will may impact this function and leave the ideas to the development document.

src/server/test/db/newCompressedReadingsTests.js
```diff
@@ -119,13 +125,16 @@ mocha.describe('Compressed Readings 2', () => {
...
 				const duration = moment.duration(reading.end_timestamp.diff(reading.start_timestamp));
-				expect(duration.asMilliseconds()).to.equal(moment.duration(1, 'minute').asMilliseconds());
+				// Since the interval is one day, compressed readings returns raw data.
+				expect(duration.asMilliseconds()).to.equal(moment.duration(1, 'day').asMilliseconds());
+				// 100 kWh over a day is 100/24 kW.
 				expect(reading.reading_rate).to.be.closeTo(100 / 24, 0.00001);
 			}
 
-			expect(readingsForMeter.length).to.equal(24 * 60); // minutes in a day
+			expect(readingsForMeter.length).to.equal(1); // minutes in a day
```
huss: I think the comment needs to be updated since this is 1 point for the day now.

src/server/test/db/newCompressedReadingsTests.js
```diff
@@ -187,10 +224,101 @@ mocha.describe('Compressed Readings 2', () => {
...
+				const end = moment(startDate).clone();
+				end.add(61, 'days');
+				const allReadings = await Reading.getNewCompressedReadings([meter1.id], start, end, conn);
+				expect(allReadings['2'].length).to.greaterThan(61); // more data will be returned because of shift down to the hour view.
+			});
+
+			mocha.it('Hour resolution', async () => {
+				const start = moment(startDate);
+				const end = moment(startDate).clone();
+				end.add(61, 'days');
```
huss, line 242: Should this be 31 instead of 61? I thought the test was seeing if less than 61 days switches to hours (and below checks for 31 \*24\). If not, I'm unsure how it differs from the test above it in terms of the data it expects.

Note this same comment applies to 23\-min test below.

If this is correct, why is it now passing? I just want to be sure the test is doing what is expected.

src/server/test/db/newCompressedReadingsTests.js
```diff
@@ -139,10 +148,19 @@ mocha.describe('Compressed Readings 2', () => {
 
 			await Reading.refreshCompressedReadings(conn);
 
+			// We need to refresh the hourly readings view because it is materialized.
+			await Reading.refreshCompressedHourlyReadings(conn);
+
 			const meterReadings = await Reading.getNewCompressedReadings([meter.id], dayStart, dayStart.clone().add(1, 'minute'), conn);
 
-			expect(meterReadings[meter.id].length).to.equal(1);
+			// TODO: We want to return at least some value for this interval instead of nothing.
```
huss: Yes, it would be fixed if we could set how to get data for this meter. I think the bigger issue is that there are so few points \(one in a day time range\) that the new raw reading change means this test does not have a value. I think your tests below are much better. In the longer\-term, OED needs to test the results of getting back a nontrivial number of points. This would mean changing the test below to look at each returned point and check its value \(a little tricky given the reading times cross the returned points interval but easier for raw readings\). We could do a simpler test by using a square wave so the value was constannt in the desired range. I think we can put this off but it would be nice to check that at least one point is correct. I'm open to ideas but here is a thought. What if the reading for each of these tests is changed to span 15 minutes. Then ask for that 15 minutes in the range for the readings desired. I think it would return that one point so it would be easy to check the number and value. This is the square wave idea where there is only one point so it is less of a wave and not many readings on the one value. What do you think?

docker-compose.yml
```diff
@@ -70,6 +72,7 @@ services:
                 # condition: service_healthy
         # Lets docker-compose up work right
         # If environment variable install_args is not set then it becomes blank without warning user.
+        #command: ["bash", "./src/scripts/installOED.sh", "--keep_node_modules"] #, "${install_args:-}"]
```
huss: Is this a left over, commented out line?

package.json
```diff
@@ -25,6 +25,8 @@
     "editUser": "node ./src/server/services/user/editUser.js",
     "sendLogEmail": "node ./src/server/services/sendLogEmail.js",
     "refreshReadingViews": "node -e 'require(\"./src/server/services/refreshReadingViews\").refreshReadingViews()'",
+    "refreshDailyReadingViews": "node -e 'require(\"./src/server/services/refreshReadingViews\").refreshReadingViews()'",
```
huss, line 29: I think leaving the old refreshReadingViews alone probably makes sense and I like the alias for daily. What do you think about having a refreshAllReadingViews that does both daily and hourly? This would make it easier on developers (and some sites) to do them all at once. It also means any change in the future could be put into this combined command.

src/server/data/automatedTestingData.js
```diff
@@ -303,6 +303,7 @@ function generateVariableAmplitudeTestingData() {
 }
 
 module.exports = {
+	DEFAULT_OPTIONS,
```
huss: I tried to figure out why this needs to be exported but seem to have missed it. Could you explain it for me?

src/server/sql/reading/determineMinPoints.js
```diff
@@ -0,0 +1,43 @@
...
+ * be replaced when we later implement meter-by-meter logic for compressed readings.
+ * @returns [minimumHourPoints, minimumDailyPoints]
+ */
+function determineMinPoints(){
+	// Minimum daily points is set such that when the interval requested is under 60 days or 
+	// about two months then the compressed_readings_2 algorithm should to the hourly view.
+	const minimumDailyPoints = 61;
+
+	// Minimum daily points is set such that when the raw reading rate is every 15 minutes and
+	// then the compressed_readings_2 algorithm should to the raw data view when the interval
```
huss: should to -> should go to

huss (changes requested):
I have tested the graphing and CSV upload via the web page and curl. Everything works as expected outside the comments. I have a few comments/suggestions that I hope are easy to address. Let me know if you want help with it.

I want to note that the size of this PR does not reflect the complexity of the work done. It took a lot of designing, working with me and testing to come up with an efficient way that also supports our sites/users. I thank @lindavin for all this work and the great outcome.

src/server/data/automatedTestingData.js
lindavin: I had planned to use it in one of the tests, but I didn't end up needing to. I will remove this export.

src/server/sql/reading/determineMinPoints.js
lindavin: Thanks!

docker-compose.yml
lindavin: Yes, thanks for the catch! I'll remove it.

src/server/test/db/newCompressedReadingsTests.js
lindavin: Okay

package.json
lindavin: Done

src/server/models/Reading.js
lindavin: Done

src/server/migrations/0.7.0-0.7.1/sql/readings/create_materialized_hourly_readings.sql
lindavin: I thought 'CREATE OR REPLACE' would not work because we would be replacing an unmaterialized view with a  materialized view? I did this in two steps because I thought I had heard this somewhere, perhaps in one of our meetings. I think this is fine though, but I can change it if you think it'll work in one step.

src/client/app/containers/csv/UploadCSVContainer.tsx
lindavin: Done

src/server/test/db/newCompressedReadingsTests.js
lindavin: I had picked 31 days because that's how much data I had chosen to put into the testDB since otherwise refreshing causes the mocha beforeEach hook to timeout. I have now fixed this by extending the timeout and uploading enough data into the testDB.

src/server/sql/reading/create_compressed_reading_views.sql
lindavin: Yes, by "different levels of granularity" I mean different reading rates. I was using the language that I see in the eGauge documentation. I mainly pointed out resource generalization because different resource meters will likely have different reading rates, but you're correct that meters with different reading rates is not unique to resource generalization. I might have just been over-thinking this comment. I have updated the comment.

src/server/test/db/newCompressedReadingsTests.js
lindavin: I think its easier to change the underlying data rate to be 15 minutes. I updated the test to reflect this and to also check that the reading value is correct.

src/server/migrations/0.7.0-0.7.1/sql/readings/create_materialized_hourly_readings.sql
huss: Yes, this seems reasonable. It is unlikely someone would run this twice (not sure if our migrations can all do that) so it just means recreating the data which won't take that long.

huss (approved):

huss:
Thanks to @lindavin for doing all the request items and also catching a few other things. This was going to be the last PR for version 0.8.0 but now we have to do a couple of translations and an automated package update. Sorry you don't get to be the ultimate one.

PR #661, opened 2021-10-25
Map changes

### LINKED ISSUE #543: eliminate north map orientation

### LINKED ISSUE #606: sort map names alphabetically

Meters & groups dropdown menus are sorted by name. The map one is not.

Sort maps by name.

### LINKED ISSUE #609: auto select map if only one available

It seems a common case that a site would only have one map. In this case it would be nice to autoselect this map when you go to the map graphic page. The dropdown for maps would still be there but the one item on the list would have already been selected.

See above.

Clearly the user can click on the map an select it. This is what happens for meters and groups. However, only having one map seems much more likely so only change in this case.

### Additional context

None

jina2k:
# Description

Fixes #609 : If only one map exists, automatically display map.
Fixes #606 : Sorted map names alphabetically
Fixes #543 : Implemented north angle into calibration. Instead of using the default 0.0 angle, now we use the provided true north angle from the user during calibration. Most of the code was left untouched as it was mostly implemented already, the parameters were just changed to account for this new angle.

## Type of change

Note: for the documentation update it should provide info on figuring out the angle (ie; it should correctly calibrate within 3 points, or show a picture side by side to represent 90 degree rotation) and also mentioning that the north angle is no longer defaulted to 0.0. In addition, the test documentation should maybe remove using x,y provided by the grid to find the points, where you would normally divide by 10 and reverse the x,y to y,x. This use case would not work with rotated maps, and should either be removed or modified to address these changes.

## Checklist

## Limitations
n/a


src/client/app/components/maps/MapCalibrationInitiateComponent.tsx
```diff
@@ -70,9 +88,30 @@ class MapCalibrationInitiateComponent extends React.Component<MapInitiatePropsWi
...
+
+	private handleAngle(event: any) {
+		event.preventDefault();
+		const pattern = /^\d+(\.\d+)?$/;
+		if (!pattern.test(this.state.angle)) {
+			this.notifyBadNumber();
+			return false;
+		}
+		else {
+			if (parseFloat(this.state.angle) > 360) {
```
huss, line 111: I think the check should also make sure it is at least zero. The bad number check gets negative values but the popup says it is not a number which isn't really the case. I think this check should catch that case if not too hard to do.

src/client/app/containers/MapChartSelectContainer.ts
```diff
@@ -16,13 +16,18 @@ function mapStateToProps(state: State) {
...
+	// If there is only one map, selectedMap is the id of the only map. ie; display map automatically if only 1 map
+	if (Object.keys(sortedMaps).length === 1) {
+		state.maps.selectedMap = sortedMaps[0].value;
+	}
 	const selectedMap = {
 		label: state.maps.byMapID[state.maps.selectedMap] ? state.maps.byMapID[state.maps.selectedMap].name : '',
 		value: state.maps.selectedMap
 		};
 
+
```
huss: extra blank line?

src/client/app/translations/data.json
```diff
@@ -133,6 +133,8 @@
 		"manage": "Manage",
 		"map": "Map",
 		"maps": "Maps",
+		"map.bad.number": "Not a number, please change angle to a number between 0 and 360",
```
huss, line 136: The three new strings are in en & fr but missing from es.

huss (changes requested):
I have checked that map rotations work with my test maps I used when that feature was added. Everything worked fine. The other changes also seem to work fine from testing. This is good work where a few small changes are requested.

src/client/app/components/maps/MapCalibrationInitiateComponent.tsx
jina2k: added test for negative values

src/client/app/containers/MapChartSelectContainer.ts
jina2k: removed blank line

src/client/app/translations/data.json
jina2k: readded the strings, it was removed during merging with the translations, and also added additional strings for the test case for negative values

huss:
I just find one more item. If there is only one map and it is autoselected then all the meters show up as graphable when you first go to the map graphic. If you select any meter then the meter menu corrects to gray out any ungraphable meters. Note this only seems to happen when you first reload OED. Sorry I missed this the first time around.
Also, I have not checked that this works with groups but it might be the same issue.

jina2k:
> I just find one more item. If there is only one map and it is autoselected then all the meters show up as graphable when you first go to the map graphic. If you select any meter then the meter menu corrects to gray out any ungraphable meters. Note this only seems to happen when you first reload OED. Sorry I missed this the first time around. Also, I have not checked that this works with groups but it might be the same issue.

Fixed it to show only meters/groups that are displayable on map

huss (approved):
Everything looks good.

PR #663, opened 2021-10-29
Add notification for uncalibrated maps

### LINKED ISSUE #575: map enabling when not calibrated
This is a small point but you can enable a map that is not calibrated. It shows grayed out on the list in the map graphic but it might confuse users. Should we stop or warn an admin if they try to enable a map that is not calibrated?

ngan-cassie:
# Description

Fixes #575  : If a map is uncalibrated and user tries displaying it, send out a notification to tell them that calibration is needed before display.

## Type of change

## Checklist


src/client/app/translations/data.json
```diff
@@ -244,6 +245,7 @@
 		"bar": "Bande",
 		"bar.interval": "Intervalle du Diagramme à Bandes",
 		"bar.stacking": "Empilage de Bandes",
+		"calib.needed": "(Need French) Calibration needed before display",
```
huss: I think calib.needed in fr and es are left over and not needed.

src/client/app/translations/data.json
```diff
@@ -172,6 +172,7 @@
 		"name": "Name",
 		"navigation": "Navigation",
 		"need.more.points": "Need more points",
+		"notify.calibration.needed": "Calibration needed before display",
```
huss: Lets rename this to map.notify.calibration.needed and move it into the map section. Need to do for all 3 languages.

src/client/app/components/maps/MapViewComponent.tsx
```diff
@@ -127,9 +128,18 @@ class MapViewComponent extends React.Component<MapViewPropsWithIntl, MapViewStat
...
 		let toggleButton;
 		if (hasToken()) {
+			// throw out alert if the admin wants to display uncalibrated map
+			if (!(this.props.map.origin && this.props.map.opposite)) {
+				toggleButton = <Button style={this.styleToggleBtn()} color='primary' onClick={this.notifyCalibrationNeeded}>
+				<FormattedMessage id={buttonMessageId} />
+				</Button>;
+			}
+			// if map is already calibrated, the button will allow it to be displayed
+			else {
```
huss, line 215: When I format in Visual Studio it changes some of the indenting such as making the else statements indented. If you can to it great. If not, I could change the formatting.

huss (changes requested):
This works well. I only have a few comments to address that should be easy. Congratulations to @ngan-cassie for their first PR and nice work.

src/client/app/translations/data.json
huss: I think src/client/app/components/maps/MapViewComponent.tsx needs to be updated with the new name so the translation works properly.

huss:
A recent PR caused a merge conflict. I merged in the latest development and resolved the conflict. The code still seems to work as desired but @ngan-cassie should double check me. You will need to pull the changes into your local clone to see the changes and work from that.

huss (approved):

PR #664, opened 2021-11-01
added functionality for changing circle size

### LINKED ISSUE #571: map circle size

jina2k:
# Description

Fixes #571 

## Type of change

## Checklist

## Limitations

Note: While this does not change the database configuration, I have edited some sql queries and some variable names. It's likely that pre-existing maps do not contain a circlesize, so it would be best to reupload maps to the database or insert the circle size directly


src/client/app/components/maps/MapViewComponent.tsx
```diff
@@ -48,6 +52,8 @@ class MapViewComponent extends React.Component<MapViewPropsWithIntl, MapViewStat
 		this.toggleNoteInput = this.toggleNoteInput.bind(this);
 		this.handleNoteChange = this.handleNoteChange.bind(this);
 		this.toggleDelete = this.toggleDelete.bind(this);
+		this.handlecSizeChange = this.handlecSizeChange.bind(this);
```
huss: Should the function name be handleSizeChange (not handlecSizeChange)?

src/server/routes/maps.js
```diff
@@ -156,7 +156,8 @@ router.post('/create', adminAuthenticator('create maps'), async (req, res) => {
 					origin,
 					opposite,
 					req.body.mapSource,
-					req.body.northAngle
+					req.body.northAngle,
+					0.15
```
huss: The value 0.15 is used in several places (find by searching for that value). I think it might be best if we had a constanst in one place to specify the value. I'm not sure of the best place (maybe src/server/models/Maps.js ?) but I think adding it would be good. It does cross client and server so needs to be included in both places (so is src/common/ best?).

src/client/app/components/maps/MapViewComponent.tsx
```diff
@@ -39,7 +41,9 @@ class MapViewComponent extends React.Component<MapViewPropsWithIntl, MapViewStat
 			nameFocus: false,
 			nameInput: this.props.map.name,
 			noteFocus: false,
-			noteInput: (this.props.map.note) ? this.props.map.note : ''
+			noteInput: (this.props.map.note) ? this.props.map.note : '',
+			circleFocus: false,
+			circleInput: (this.props.map.circleSize.toString()) ? this.props.map.circleSize.toString() : '0.15'
```
huss: I seems legacy systems (and maybe the current system) can have max_circle_size_fraction be null in the database. (It might be nice to trace down why at some point.) When this happens, this line of code causes an exception and nothing is displayed. I fixed this by changing to:

circleInput: (this.props.map.circleSize !== null && this.props.map.circleSize.toString()) ? this.props.map.circleSize.toString() : '0.15'

Also, in src/server/models/Map.js for mapRow, I had to make it return the same value so the mapping worked with the default (lines 80-1):

return new Map(row.id, row.name, row.displayable, row.note, row.filename, 
row.modified_date, row.origin, row.opposite, row.map_source, row.north_angle, (row.max_circle_size_fraction !== null ? row.max_circle_size_fraction : 0.15));

This all seemed to work fine. Adding comments about why the null check is needed would be good. It may be another change is better to address this and it seems unfortunate to have it in two places.

I considered the alternative of doing a DB migration but that seemed more painful.

huss (changes requested):
The changes seem good. A few comments to address.

src/client/app/components/maps/MapViewComponent.tsx
jina2k: changed function name to handleSizeChange

src/server/routes/maps.js
jina2k: used req.body.circleSize instead, and modified src/server/models/Map.js to what was shown below

src/client/app/components/maps/MapViewComponent.tsx
jina2k: added null test, and modified Map.js to what was mentioned

src/server/routes/maps.js
huss: I just pushed some code that reduces where 0.15 shows to only the server and added a const for it.

src/client/app/components/maps/MapViewComponent.tsx
huss: I just pushed some code to force the route to save the default value into the DB.

huss:
I just proposed some changes since I checked them out and thought that was easiest. Can @jina2k please review and let me know if you think they are okay or need changes.

jina2k:
> I just proposed some changes since I checked them out and thought that was easiest. Can @jina2k please review and let me know if you think they are okay or need changes.

Looks good. I can see that DEFAULT_CIRCLE_SIZE is now all pulled from src/server/models/Map.js now, which is easier to edit if need be in the future.

huss (approved):

huss:
Everything worked as expected and all changes have been reviewed. Time to merge.

PR #666, opened 2021-11-08
added maps to preferences

### LINKED ISSUE #573: option to make maps the default graphic
The admin page needs to have option to make maps the default graphic as it does for all the other ones.

jina2k:
# Description

Fixes #573 

## Type of change

## Checklist

## Limitations
I have created a migrations folder with the relevant files to move to 0.7.1 as a step before 0.8.0. Documentation needs to be updated where there is an issue of opening localhost:3000 and seeing default map be "displayed" where it shows a blank image (occurs every time browser is reopened). Not that big of an issue as users can reload the page or click a random meter and it will correct itself.


src/server/migrations/0.7.0-0.7.1/sql/preferences/alter_preferences_type.sql
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+DO $$ BEGIN
+    ALTER TYPE graph_type ADD VALUE 'map' AFTER 'compare';
```
huss: I believe there is an issue that you cannot do this in a transaction which is what the migration code does. Thus, it must be done as in src/server/migrations/0.6.0-0.7.0/sql/meter/add_meter_type.sql.

huss (changes requested):
The code works as desired and is good work. Just need to clean up the migration code (unless you show the current way works).

Also, would you like to open a new issue on the blank map or do you want someone else to do that? It just needs to be looked at sometime.

src/server/migrations/0.7.0-0.7.1/sql/preferences/alter_preferences_type.sql
jina2k: changed lines to reflect add_meter_type.sql

jina2k:
> The code works as desired and is good work. Just need to clean up the migration code (unless you show the current way works).
> 
> Also, would you like to open a new issue on the blank map or do you want someone else to do that? It just needs to be looked at sometime.

Someone else can open an issue for now, I will work on the group page in the meantime

src/server/migrations/0.7.0-0.7.1/sql/preferences/alter_preferences_type.sql
huss: Migrations are something we only recently got working correctly, esp. for changing an enum. Thus, I have pushed changes to get this to work as expected. I tested on my machine and the migration worked correctly. Note I did not push the version number change in package.json that is needed as this will be done when the release is made.

huss:
> > The code works as desired and is good work. Just need to clean up the migration code (unless you show the current way works).
> > Also, would you like to open a new issue on the blank map or do you want someone else to do that? It just needs to be looked at sometime.
> 
> Someone else can open an issue for now, I will work on the group page in the meantime

I opened issue #667 to address this.

huss (approved):

PR #668, opened 2021-11-15
show end time in bar chart export #489

### LINKED ISSUE #489: bar chart export only gives start time
When you export the data on a bar chart it has a column with the start time. We should have a second column with the end time for each point as a bar represents a time range. For many points the end time is the next start time. However, there is no next point and the user would need to infer it from the previous spacing. Also, while I have yet to see gaps in bar charts it may be possible and these would be an issue. We should help the user by giving the end time as the last column of the CSV.

truongdd03:
# Description

Added a variable to store the end time for data. The node version is changed to make the code runnable.

Fixes #489

## Type of change

## Checklist

## Limitations

There is no limitation.


huss:
I have made changes to the Obvius test so it now works. I finally realized it was due to two causes. Coincidentally, a recent PR (#672) found an issue in how Obvius users are tested. This meant that on development machines you got a password error returned. Now there is the question of why a development machine had a different code/message than in GitHub Actions during CI testing. The reason is that on GitHub it does not set the environment variable that was used (before PR #672) to check the password so it got a default value that matched what is used during testing. A change in the development setup (and production) in the recent past reset this value to something to let production sites know they should change the value. Thus, the password began to fail in these environments. When the fix was applied the issue went away so we can now check the status and the expected string. It passed all tests on my machine and on GitHub (commit 2813065)

I am about to remove the fix from PR #672 since that code should be added from that PR. This is going to make the two Obvius tests fail again but that is now acceptable for this PR.

This was beyond the issue addressed in this PR so I have addressed it. I thank @truongdd03 for finding it and helping out.

src/client/app/utils/exportData.ts
```diff
@@ -15,17 +15,33 @@ import { UserRole } from '../types/items';
...
-	let csvOutput = 'Label,Readings,Start Timestamp\n';
-	items.forEach(set => {
-		const data = set.exportVals;
-		const label = set.label;
-		data.forEach(reading => {
-			const info = reading.y;
-			const startTimeStamp = moment(reading.x).utc().format('dddd LL LTS').replace(/,/g,''); // use regex to omit pesky commas
-			csvOutput += `"${label}",${info},${startTimeStamp}\n`; // TODO: add column for units
+	// Export line chart
+	if (items[0].exportVals[0].z === undefined) {
```
huss, line 19: This is fine for now. Presumably we will be adding the end date/time to all exports (2 line ones are only ones left) and then we don't need the first case.

huss (approved):
This PR addresses the issue and works fine. There are no issues that need to be addressed. Thanks to @truongdd03 for this good work. Congratulations on their first approved PR!

PR #670, opened 2021-11-18
X point value for line graph

### LINKED ISSUE #477: Line/bar/compare graph x point value and label

ngan-cassie:
# Description

Made the x-axis display the time in the middle of start and end timestamps. 

Partly addresses #477 

## Type of change

## Checklist

## Limitations

None


huss (approved):
The PR is fine and does what is needed.  It addresses the line part of the linked issue. I am unclear on why one of the CodeQL checks is still queued but I don't think the PR could cause an issue so ignoring. Congratulations to @ngan-cassie for their first PR and it being approved.

PR #671, opened 2021-11-20
Implemented new groups page and edit groups

### LINKED ISSUE #551: group page similar to meters
We now have a meter page that allows users to see values for each meter and allows admins to edit these value. As we are adding more information associated with both meters and groups, we need another, similar page for groups. Before implementation can begin, we need to decide the layout and values on the group page and the relationship to the current group page that displays elements of each group.

jina2k:
# Description

Fixes #551 

Added GPS, displayable, note, and area to group page for both creating a new group and editing an existing one.

## Type of change

## Checklist

## Limitations
I have deleted the rename_group.sql file, which was replaced with the generalized update_group.sql file. I have also edited test cases to reflect this, and added displayable (set to false) for some tests. In addition, I have passed in id, name, gps, displayable, note, area all in the routes/groups.js so that it can be seen when editing. This could be changed depending on whether the user is an admin or not, but mostly group information cannot be seen by a non-admin regardless.

Future updates may revolve around simplifying the whole action/reducer React-related code such that it all can be done in a single action as opposed to 5. This can be seen in MapViewContainer/Component which uses editMapDetails, passing in the entire Metadata.

Also, documentation may be necessary if we decide to remove the ability to submit when nothing has changed (in edit container, when no gps or area exist, triggers a dirty status for displayable instead) so that those with pre-existing groups that are able to do this (one-time due to displayable) are notified that it is just a temporary fix (or we can make a migration file to update displayable for all groups instead)


src/client/app/components/groups/CreateGroupComponent.tsx
```diff
@@ -107,8 +166,52 @@ class CreateGroupComponent extends React.Component<CreateGroupPropsWithIntl, {}>
...
+					const gPoint: GPSPoint = {
+						longitude: parseFloat(parseGPS[0]),
+						latitude: parseFloat(parseGPS[1])
+					};
+					this.props.editGroupGPS(gPoint);
+				}
+				this.props.submitGroupInEditingIfNeeded();
+			}
+			else {
+				window.alert(this.props.intl.formatMessage({id: 'group.gps.error'}));
```
huss: isValidGPSInput creates a popup if there is an issue witht he GPS format. This line also does that so you get a second popup. I tried to think of a case where removing this popup would mean no message but I could not quickly get one. If this is correct, then I think it best to remove this one.

src/client/app/components/groups/EditGroupsComponent.tsx
```diff
@@ -209,6 +283,23 @@ class EditGroupsComponent extends React.Component<EditGroupsPropsWithIntl, EditG
...
+	private handleGPSChange(e: React.ChangeEvent<HTMLInputElement>) {
+		this.setState({ gpsInput: e.target.value });
+	}
+
+	private handleDisplayChange(e: React.ChangeEvent<HTMLInputElement>) {
+		this.props.editGroupDisplayable(e.target.value === 'true');
+	}
+
+	private handleNoteChange(e: React.ChangeEvent<HTMLTextAreaElement>) {
+		this.props.editGroupNote(e.target.value);
```
huss, line 296: I'm not sure why but I cannot edit the note. The characters typed do not show but the first one typed shows up when the changes are saved. I tried with an exising and new group and saw it in both cases. The creation of a group works fine.

src/client/app/components/groups/CreateGroupComponent.tsx
```diff
@@ -61,12 +87,45 @@ class CreateGroupComponent extends React.Component<CreateGroupPropsWithIntl, {}>
...
+								<FormattedMessage id='displayable' />:
+							</p>
+							<Input type='select' name='displayselect' onChange={this.handleDisplayChange}>
+								<option value='true'> True </option>
+								<option value='false'> False </option>
+							</Input>
+						</div>
+						<div style={divBottomStyle} className='col-4'>
+							<p style={textStyle}>
+								<FormattedMessage id='map.note' />:
```
huss: Should this be map.note or group.note? Same question in EditGroupsComponent.

huss (changes requested):
Overall, this works as expected. Just a few questions to resolve.

src/client/app/components/groups/CreateGroupComponent.tsx
jina2k: removed else statement for both

src/client/app/components/groups/EditGroupsComponent.tsx
jina2k: added setstate for both note and display

src/client/app/components/groups/CreateGroupComponent.tsx
jina2k: I was going to use the map.note because it just says Note in general, so instead map.note is changed to note and group points to this as well

huss: For reasons I did not see quickly, the map page shows the header of map.note in all languages.

huss (approved):

Issue #599, opened 2021-03-05
enhance data export


huss:
OED just added the ability to export raw meter data. As part of that and checking out all exports, I want to request some changes:

- [x] Either export (graph or raw) only gives the start time of each point. We have the start and end time in the DB. The upcoming plan is to allow import with start/end time and this is better since gaps can be seen. I think we should change this for raw and, if not too hard, for the graph as well. We should verify that the exported data can be read back into OED with the new import features.
- [ ] 2. The export puts the unit on each point. This makes it hard to easily graph the spreadsheet in a standard program. Once we do resource generalization the unit may vary by meter so it is needed. I propose we put it as a separate column unless a better idea is found. We should then verify that the way we put out dates does not cause any issues in graphing in a spreadsheet.
- [x] src/client/app/utils/exportData.ts has 5/25 MB file size hard coded (TODO markers in file). Need to make it under admin control with these as the default values.
- [ ] The raw export does not support groups. My first thought is it would dump all the underlying meters since you cannot aggregate the data without averaging if points are not perfectly aligned in start/end time. See if everyone agrees and then do what is decided.

LukeLavan:
#563 removes the `kW` from the exported CSV files. I add a TODO comment where they used to be and specify that in the future the units should be its own separate column when that information becomes available on the `ExportDataSet` and `RawReadings` interfaces.

huss:
Groups do not export raw. That was the final decision.

PR #674, opened 2021-11-23
Enhance data export

truongdd03:
# Description

Shows end time and fixes the error in raw export.

Partly Addresses #599 

## Type of change

## Checklist

## Limitations

The export file still requires to remove the first column (meter's name) to be importable where each exported meter needs to be in its own file.


src/client/app/utils/exportData.ts
```diff
@@ -78,10 +64,11 @@ export default function graphExport(dataSets: ExportDataSet[], name: string) {
  */
 export function downloadRawCSV(items: RawReadings[], defaultLanguage: string) {
 	// note that utc() is not needed
-	let csvOutput = 'Label,Readings,Start Timestamp\n';
+	let csvOutput = 'Label,Readings,Start Timestamp,End Timestamp\n';
 	items.forEach(ele => {
-		const timestamp = moment(ele.startTimestamp).format('dddd LL LTS').replace(/,/g, ''); // use regex to omit pesky commas
-		csvOutput += `"${ele.label}",${ele.reading},${timestamp}\n`; // TODO: add column for units
+		const startTimestamp = ele.startTimestamp;
```
huss: I think the moment, format, replace that was removed for startTimestamp and endTimestamp should be put back as they are now for convertToCSV() so it works the same way when the language is changed.

huss (changes requested):
There is the only one comment I have on this draft pull request. This works as expected and cleans up the code. I appreciate @truongdd03 for being so through. Once any concerns the author has are resolved then the draft state can be removed.

src/client/app/utils/exportData.ts
truongdd03: Thank you for your comment. I have fixed it in the new commit.

huss (approved):
Looks great. Ready to merge.

PR #675, opened 2021-11-27
fix bar graph duration + bar chart link

### LINKED ISSUE #477: Line/bar/compare graph x point value and label

### LINKED ISSUE #578: chart link not working for bar graphs
If you try to use the chart link it causes a promise error in the console and you don't see any graphic.

ngan-cassie:
# Description

- Change the time reading/ bar duration of the bar graph down by 1 day.
- Fix the promise error when pasting bar chart link.

Fixes #578 
Partly addresses #477. The line case was previously addressed. This fixes the bar case. Compare remains.

## Type of change

## Checklist


src/client/app/containers/BarChartContainer.ts
```diff
@@ -78,8 +78,9 @@ function mapStateToProps(state: State) {
 				const hoverText: string[] = [];
 				const readings = _.orderBy(readingsData.readings, ['startTimestamp'], ['asc']);
 				readings.forEach(barReading => {
+					// calculate more precise bar duration
```
huss: I would also add this comment above for meters (this is for groups). It might also be good to add why you are subtracting 1 day (because the day ends at midnight of the next day so getting an extra day in the labels).

src/client/app/containers/ChartLinkContainer.ts
```diff
@@ -31,7 +31,7 @@ function mapStateToProps(state: State) {
 			linkText += `&barStacking=${state.graph.barStacking}`;
 			break;
 		case 'line':
-			linkText += `&serverRange=${state.graph.timeInterval.toString()}`;
+			// linkText += `&serverRange=${state.graph.timeInterval.toString()}`;
```
huss: I think the commented out code should be removed and a comment added that these is non code for this case but left so show what is going on. (You can leave the next two lines if you want since already there.)

huss (changes requested):
This addresses both issues and fixes them. Just a couple of ideas on commending. Good work.

huss:
PR #680 replaced this PR so this PR is being closed.

PR #676, opened 2021-11-27
Add access rate limit

### LINKED ISSUE #266: Rate Limiting

truongdd03:
# Description

This PR uses [express-rate-limit](https://www.npmjs.com/package/express-rate-limit) module to set two access limits. The first one (200 requests/ 5 seconds) limits the rate of overall requests to OED. At the moment, the current limit is high due to issue #681. After resolving it, testings are needed to adjust the rate limit. The second limit is specifically for raw export (5 requests/ 5 seconds).

Fixes #266 

## Type of change

## Checklist

## Limitations

Note that the rate limit may make future automatic tests return the value of 429 (too many requests). In that case, the limiters need to be increased.


src/server/app.js
```diff
@@ -30,7 +31,11 @@ const logs = require('./routes/logs');
 const obvius = require('./routes/obvius');
 const csv = require('./routes/csv');
 
-const app = express();
+const limiter = new rateLimit({
```
huss: Nice to add a comment about the general intent of this code (to limit the rate of overall requests to OED) and what the two values mean.

huss:
This is good work and does basic rate limiting. Testing indicates it works as expected. I generalized the test script written by @truongdd03 and put it in the testing directory for future use.

I have two thoughts on this PR:

1. Why is the overall limit set at 300/minute? It seems like it may be fine but was there testings to see if this would work for sites? Can a single user request may count as multiple requests toward the limit if it makes api calls internally?Just want to know for the record.
2. Do people know if changing the time frame for the reset (currently 1 minute) would help with accidentally hitting it with valid requests at a site?
3. I do think the overall limit cannot be too low as we want log api requests to succeed so log messages are not lost.
2. In the long-term, we might want to have a default value and allow the admin to modify that value. That is beyond this PR but just noting.
2. An overall limit seems good but I think specific routes should probably also have limits. For example, asking for the raw data is going to be much more draining on resources than asking for bar graphic data. The default size to download without a login is 5MB so at 300 requests/min that would be 1.5 GB/min which would certainly tax the system. We don't expect a lot of these requests so maybe limiting to 5-10 per minute might be better. I don't see any other obvious routes that doe not require a login that are an issue like this.

huss (changes requested):

huss (approved):

huss:
Good work. I appreciate the effort to test and get all the details working. This rate limits overall requests (allowing a lot) and raw download (only a few). I updated the test script a little to deal with items. It all tests out fine so ready to merge.

PR #677, opened 2021-11-29
Updated webpack v4 to v5

### LINKED ISSUE #548: update packages
After we finish version 0.7 we need to systematically update our packages.

jina2k:
# Description

Partially Addresses #548 
mostly a transition from webpack v4 to v5, nodemon v2, and chokidar v3, while also editing webpack.config.js to conform to the format displayed on the official documentation.

Other changes include:
axios
-> follow-redirects
glob-parent
-> is-glob
plotly
babel-loader
lodash-webpack-plugin
style-loader
-> css-loader
buffer
assert
stream-browserify
removing uglifyjs-webpack-plugin and using terser instead

I've also installed node-polyfill-webpack-plugin as there was an issue with OED displaying a blank page.

--legacy-watch flag was added according to the documentation https://github.com/remy/nodemon as files were not being changed

## Type of change

## Checklist

## Limitations
tslint still needs to be migrated to eslint

Documentation should be created for others for when they plan on updating packages themselves as well, such that they would get the correct npm audit output, as shown below, using nvm.

ie;
-> https://stackoverflow.com/questions/11177954/how-do-i-completely-uninstall-node-js-and-reinstall-from-beginning-mac-os-x?rq=1
-> https://nbanzyme.medium.com/easy-way-to-install-nvm-on-ubuntu-18-04-2cfb19ee5391

then nvm install 14 to get nodev14

Another limitation is that previously, deprecated packages were shown as warnings when docker compose up was triggered, but currently they do not show anymore so the only way to show deprecated packages is some form of wiping package-lock.json using rm -rf package-lock.json node_modules and then triggering npm i to see the deprecated packages, while also keeping a copy of the previous package-lock.json as the one created with npm i does not have the same output.


jina2k:
with the most recent commit, i have
uninstalled:
nodemailer-mailgun-transport
netmask
https-proxy-agent
http-proxy-agent
proxy-agent
pac-resolver
pac-proxy-agent
crypto-browserify
socks-proxy-agent

and the mailing function apparently functions as intended when testing with firefox, from gmail to outlook. Not sure why google chrome does not work for both the previous version of OED and for this one

huss:
Two questions on your recent comment:

1. I've been trying this version in Chrome and am not having issues. Could you let me know what the issue(s) were?
2. Do you know how the email was sent? I want to be sure we are not creating an implicit requirement for a mail program (probably smtp) that we do not state given nodemailer is now gone.

> with the most recent commit, i have uninstalled: nodemailer-mailgun-transport netmask https-proxy-agent http-proxy-agent proxy-agent pac-resolver pac-proxy-agent crypto-browserify socks-proxy-agent
> 
> and the mailing function apparently functions as intended when testing with firefox, from gmail to outlook. Not sure why google chrome does not work for both the previous version of OED and for this one

src/server/logMailer.js
```diff
@@ -87,26 +86,34 @@ async function logMailer(conn) {
...
 	if (config.mailer.method === 'none') {
 		return;
-	} else if (config.mailer.method === 'mailgun') {
-		transporter = nodemailer.createTransport(mg({
-			auth: {
-				api_key: config.mailer.credential,
-				domain: config.mailer.ident
-			}
-		}));
 	} else if (config.mailer.method === 'gmail') {
```
huss: This appears to lock us to using gmail for sending mail. OED should have an open source, non-vendor way to send the log email.
I'm open to other ways. I don't know if this was removed to get rid of vulnerabilities and if an upgrade to the version would help.

package.json
```diff
@@ -6,10 +6,10 @@
   "repository": "https://github.com/OpenEnergyDashboard/OED",
   "scripts": {
     "start": "node ./src/bin/www",
-    "start:dev": "./node_modules/.bin/nodemon --inspect=0.0.0.0 ./src/bin/www",
-    "webpack:dev": "webpack -d --watch --color --progress",
-    "webpack:build": "cross-env NODE_ENV=production webpack -p",
-    "webpack": "webpack -d --color --progress",
+    "start:dev": "./node_modules/.bin/nodemon --legacy-watch --inspect=0.0.0.0 ./src/bin/www",
+    "webpack:dev": "webpack watch --color --progress --node-env development",
```
huss: There are several sources of info but https://stackoverflow.com/questions/55259238/what-is-the-difference-between-webpack-env-production-and-mode-production discusses the need for --node-env development and cross-env. Since –mode=development seems to set values for a DefinePlugin (which we do in webpack.config.js at the bottom), I wonder if we can avoid the --node-env. I'm not clear we needed cross-env even before your changes. I have not verified it works as stated in the post.

package.json
```diff
@@ -49,33 +49,43 @@
     ]
   },
   "dependencies": {
-    "axios": "~0.21.1",
+    "@babel/core": "~7.9.0",
```
huss, line 52: This version is about 2 years old. How did it get added? When I was testing new packages I tried 7.15.8 (about 2 months old) but never fully tested. Wondering what version OED should use.
Also, it is in dev dependencies with a lower version.

package.json
```diff
@@ -49,33 +49,43 @@
...
+    "axios": "~0.24.0",
     "babel-plugin-react-intl": "~2.4.0",
     "bcryptjs": "~2.4.3",
     "body-parser": "~1.18.3",
     "bootstrap": "~4.3.1",
+    "chokidar": "~3.5.2",
+    "co": "~4.6.0",
     "cookie-parser": "~1.4.4",
-    "cross-env": "~5.0.5",
+    "cross-env": "~7.0.3",
```
huss: See above and then decide if needed.

package.json
```diff
@@ -49,33 +49,43 @@
     ]
   },
   "dependencies": {
-    "axios": "~0.21.1",
+    "@babel/core": "~7.9.0",
+    "agent-base": "~6.0.2",
```
huss: The dependency and several below are new and don't appear directly in the OED code base. I just want to check that they are needed.

package.json
```diff
@@ -49,33 +49,43 @@
...
     "cookie-parser": "~1.4.4",
-    "cross-env": "~5.0.5",
+    "cross-env": "~7.0.3",
     "csv": "~5.3.2",
-    "dotenv": "~2.0.0",
+    "debug": "~4.3.2",
+    "dotenv": "~9.0.0",
     "es6-promisify": "~5.0.0",
     "express": "~4.16.4",
+    "get-uri": "~3.0.2",
```
huss: This one is 2 years old (v4 is much newer). Should it be upgraded? Is this needed (see above)?

package.json
```diff
@@ -49,33 +49,43 @@
...
     "morgan": "~1.9.0",
     "multer": "~1.3.1",
+    "node-polyfill-webpack-plugin": "~1.1.4",
     "nodemailer": "~6.4.16",
-    "nodemailer-mailgun-transport": "~1.4.0",
     "package-lock": "~1.0.3",
     "pg-promise": "~10.6.1",
-    "plotly.js": "~1.42.5",
+    "plotly.js": "~2.6.3",
+    "process": "~0.11.10",
```
huss, line 85: I'm just noting for the record that this package has not been updated in 5 years. Might be okay for this simple package that is heavily used but should check it out.  Added to issue #548.

package.json
```diff
@@ -91,16 +101,17 @@
     "reactstrap": "~8.0.0",
     "redux": "~3.7.2",
     "redux-thunk": "~2.2.0",
-    "request": "~2.88.0",
+    "request": "~2.88.2",
```
huss, line 101: For the record, request is now deprecated. We are going to have to look at alternatives. Added to issue #548.

package.json
```diff
@@ -91,16 +101,17 @@
...
     "request-promise-native": "~1.0.7",
     "serve-favicon": "~2.5.0",
-    "uglifyjs-webpack-plugin": "~1.2.4",
-    "webpack": "~4.30.0",
-    "webpack-cli": "~3.1.2",
+    "terser-webpack-plugin": "~5.0.0",
```
huss, line 104: terser-webpack-plugin is about a year old with a number of newer versions. Should it be updated? For all the update questions it may be the case that a dependency is stopping it.

package.json
```diff
@@ -91,16 +101,17 @@
...
+    "watchpack": "~2.1.1",
```
huss: This version of watchpack is 10 months old.

package.json
```diff
@@ -91,16 +101,17 @@
...
+    "webpack": "~5.45.1",
```
huss: Only 5 months out of date but can we go newer?

package.json
```diff
@@ -118,26 +129,29 @@
...
     "babel-preset-env": "~1.7.0",
     "babel-preset-react": "~6.24.1",
+    "buffer": "~6.0.3",
     "chai": "~3.5.0",
     "chai-as-promised": "~6.0.0",
     "chai-http": "~4.2.1",
     "cryptiles": "~4.1.3",
-    "css-loader": "~2.1.1",
-    "mocha": "~5.2.0",
+    "css-loader": "~6.2.0",
```
huss: Onnly 5 months old but there are newer versions.

package.json
```diff
@@ -118,26 +129,29 @@
...
+    "mocha": "~9.1.3",
     "mock-require": "~2.0.2",
-    "nodemon": "~1.18.7",
+    "nodemon": "~2.0.15",
     "redux-devtools": "~3.5.0",
-    "sinon": "~4.0.2",
-    "source-map-loader": "~0.2.4",
-    "style-loader": "~0.19.1",
+    "sinon": "~10.0.0",
```
huss: 8 months old. Can we go newer?

webpack.config.js
```diff
@@ -12,55 +13,70 @@ const BUILD_DIR = path.resolve(__dirname, 'src/client/public/app');
...
-            { test: /\/jsx?$/, exclude: /node_modules/, loader: "babel-loader" },
+            { test: /\/jsx?$/, exclude: /node_modules/, use:[{loader: "babel-loader"}] },
 			// CSS stylesheet loader.
-			{ test: /\.css$/, loader: 'style-loader!css-loader' },
+			{ test: /\.css$/, use: [
+                {loader: "style-loader"},
+                {loader: "css-loader"}
+            ], },
+            {
+                test: /\.m?js/,
```
huss, line 53: I'm just curious what this is needed for since it was added. Would a comment help?

webpack.config.js
```diff
@@ -12,55 +13,70 @@ const BUILD_DIR = path.resolve(__dirname, 'src/client/public/app');
...
-	],
-	node: {
-		fs: 'empty'
-	}
+        new NodePolyfillPlugin(),
+        new CheckerPlugin(),
+	]
 };
 
 if (process.env.NODE_ENV === 'production') {
```
huss, line 73: See other comment about if this is needed.

package.json
```diff
@@ -91,16 +101,17 @@
...
     "request-promise-native": "~1.0.7",
     "serve-favicon": "~2.5.0",
-    "uglifyjs-webpack-plugin": "~1.2.4",
-    "webpack": "~4.30.0",
-    "webpack-cli": "~3.1.2",
+    "terser-webpack-plugin": "~5.0.0",
+    "watchpack": "~2.1.1",
+    "webpack": "~5.45.1",
+    "webpack-cli": "~4.9.0",
+    "webpack-dev-server": "~4.5.0",
```
huss, line 108: This comment is for an unchanged line below for @types/plotly.js. 1.41.0 is 3 years old. As long as we are updating plotly maybe we should update this too?

containers/web/Dockerfile
```diff
@@ -5,7 +5,7 @@
 # *
 
 # Dockerfile for NodeJS application portion
-FROM node:latest
+FROM node:14
```
huss, line 8: Just noting that issue #682 discusses this change and needed work.

huss (changes requested):
My testing does not find any problems with the updated packages. I appreciate the effort to continue to push through the issues in updating packages, many of which were way out of date. I have asked some questions and put in some thoughts that should be addressed.

jina2k:
> Two questions on your recent comment:
> 
> 1. I've been trying this version in Chrome and am not having issues. Could you let me know what the issue(s) were?
> 2. Do you know how the email was sent? I want to be sure we are not creating an implicit requirement for a mail program (probably smtp) that we do not state given nodemailer is now gone.
> 
> > with the most recent commit, i have uninstalled: nodemailer-mailgun-transport netmask https-proxy-agent http-proxy-agent proxy-agent pac-resolver pac-proxy-agent crypto-browserify socks-proxy-agent
> > and the mailing function apparently functions as intended when testing with firefox, from gmail to outlook. Not sure why google chrome does not work for both the previous version of OED and for this one

1. I'm actually not sure, because it never prompts me to do anything. I would clear cookies for chrome but I would have to re-login to some stuff, so I ended up just using Firefox instead. 
2. You mentioned using gmail in your comment. I think it's possible there might be some issues if I use something else like Yahoo, but I'm not entirely sure about this either. Outlook to the openenergydashboard default mail address works, so I'm assuming the server just sends a request including the details of "mailto@...", and this would just be dependent on logging into an email. I read some documentation about using 2fa or some authentication key, but I don't think it's necessary. I will look into it later in the meantime

huss:
> > Two questions on your recent comment:
> > 
> > 1. I've been trying this version in Chrome and am not having issues. Could you let me know what the issue(s) were?
> > 2. Do you know how the email was sent? I want to be sure we are not creating an implicit requirement for a mail program (probably smtp) that we do not state given nodemailer is now gone.
> > 
> > > with the most recent commit, i have uninstalled: nodemailer-mailgun-transport netmask https-proxy-agent http-proxy-agent proxy-agent pac-resolver pac-proxy-agent crypto-browserify socks-proxy-agent
> > > and the mailing function apparently functions as intended when testing with firefox, from gmail to outlook. Not sure why google chrome does not work for both the previous version of OED and for this one
> 
>     1. I'm actually not sure, because it never prompts me to do anything. I would clear cookies for chrome but I would have to re-login to some stuff, so I ended up just using Firefox instead.

I have Chrome version 96. Are you near that version? I have a few extensions but not much. I'm wondering why it works for me and not you. Maybe we should have some other people try if we cannot easily figure this out. We certainly want Chrome to work.

>     2. You mentioned using gmail in your comment. I think it's possible there might be some issues if I use something else like Yahoo, but I'm not entirely sure about this either. Outlook to the openenergydashboard default mail address works, so I'm assuming the server just sends a request including the details of "mailto@...", and this would just be dependent on logging into an email. I read some documentation about using 2fa or some authentication key, but I don't think it's necessary. I will look into it later in the meantime

I'm not sure I was clear. I expect the SMTP request to work if the receiving system knows about how to deal with the request sent. What I'm actually thinking about is that OED installed on a site server should be able to send mail without resorting to another provider. We also don't want to impact their server if at all possible. I thought the old setup let the OED server send the mail and act as SMTP but maybe I have the wrong impression. If we want to remove it then it would be nice to find an alternative.

src/server/logMailer.js
jina2k: mailgun was supposedly connected to nodemailer-mailgun-transport, which is deprecated as it uses netmaskv1 which is insecure. I will get back to you on supporting more than just gmail, will probably use a general else statement instead and see how that works. I've used documentation provided here: https://nodemailer.com/smtp/
nodemailer does not have any dependencies, so there should be no issue with future updates. Also, note that the service is not included, so in general SMTP is used alongside TLS, without the use of gmail.

package.json
jina2k: for line 10 and 12, we can avoid using --node-env development and replace it with --mode development. I assumed from documentation that --node-env supersedes both node env and mode, so I just set it to --node-env instead. If you want, I can change it back to --mode instead https://webpack.js.org/api/cli/

package.json
jina2k: babel-core is now deprecated as it uses v6 as opposed to @babel/corev7. I added it somewhere while installing packages because of babel-loader probably. I actually have another branch that updates @babel/core further and changes the config file as such. For now, this does not actually do anything, but later on when I merge my other branch @babel/core will replace babel-core

package.json
jina2k: After looking at https://www.npmjs.com/package/cross-env documentation and webpack cli documentation, it looks like it's not necessary as webpack cli provides flags for setting node-env. As such, I will uninstall cross-env.

package.json
jina2k: As mentioned above, @babel/core will be used later on. Agent-base looks like it was related to *-proxy-agent which were removed, so I will remove this as well

package.json
jina2k: Uninstalled get-uri after not finding code that uses it

package.json
jina2k: Not sure about this one. I still think we need a few other packages to be updated first, especially migration to ESLint. I think the main reason I did not upgrade it was due to devDependencies involving typescript and webpack, those of which can be further updated later on. Again, in a different branch that I have been working on, I was moving towards replacing awesome-typescript-loader which used an older version of typescript. Maybe we can revisit this later on.

webpack.config.js
jina2k: the formatting was done to align with documentation for webpackv5. Without this, webpack v5 would not run, as it would not parse properly. As for the test: /\.m?js/, it had to do with babel: https://stackoverflow.com/questions/65922329/babel-cant-resolve-imports-in-it-its-own-source-code
Hopefully babel will later work? with babel v7, though I am not sure.

I will leave a comment just in case to indicate this, and maybe revisit later.

package.json
jina2k: moved to 2.2.0

package.json
jina2k: moved to 5.62.0

webpack.config.js
jina2k: from my understanding of your previous comment, I'm assuming node-env will stay production. If so, then I can try moving the code into plugins and seeing if this works

package.json
jina2k: moved to 6.4.0

package.json
jina2k: moved to 12.0.0

src/server/logMailer.js
jina2k: After using a general else statement, yahoo to gmail works fine. So, I assume this works as intended.

package.json
jina2k: moved to 1.54.12

webpack.config.js
jina2k: moving code into plugins seems to cause issues, sourcemap is also not a parameter for terser when attempting to do so, so I will leave it as is.

jina2k:
> 

Yes, I do have Chrome version 96. All it says in console is that a user gesture is required, though the page is completely empty.


SMTP request works with Yahoo as well after moving to a general else statement. We might need a try catch block, but I'm not exactly sure how this would be logged as this goes to a new tab. I actually think it is logged further down anyways, so this should be fine.

package.json
huss: I agree so leave as is.

package.json
huss: Okay. We can wait for the next update on another PR.

package.json
huss: This was updated to 1.54.12. That is okay but wondering why not to a later version. Were there issues?

huss:
> Yes, I do have Chrome version 96. All it says in console is that a user gesture is required, though the page is completely empty.

@jina2k and I discussed this. It was about the mailto: links within the application. The issue seems to be related to the specific setup of Chrome. There does not seem to be a general issue so this is considered done.

jina2k:
I tested smtp using my credentials for both gmail and yahoo. Note that for gmail to yahoo using gmail credentials, I had to allow less secure apps for my gmail. As for yahoo to gmail using yahoo credentials, I had to generate an app password, that of which was a 1-time use.

https://help.yahoo.com/kb/SLN15241.html
https://myaccount.google.com/lesssecureapps

Testing sequence is as follows:
- fill out credentials in docker-compose.yml and other info (smtp host, port, etc.)
- run oed
- go to csv page and submit with an error (ie; say theres no header when there is a header)
- run the command 'docker-compose run --rm web npm run sendLogEmail' in a separate terminal while oed is still running
- you should receive a 250 OK message.

From this most recent change, I confirm that the request works

jina2k:
reverted long lat as shown above, separating it from PR #684

huss (approved):
All changes look good and this is ready to merge. I note that I did not tests the log email sending since @jina2k did this. We do need to verify a live site works with the new setup after it is deployed.

PR #679, opened 2021-12-04
Add file size to admin panel

truongdd03:
# Description

Makes the warning file size and limit file size under admin control. The default values of them are 5 MB and 25 MB respectively.

Partly Addresses #599 

## Type of change

## Checklist

## Limitations

The PR still requires labels in French and Spanish for these two values.


src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -229,6 +259,14 @@ class PreferencesComponent extends React.Component<PreferencesPropsWithIntl, {}>
 	private handleSubmitPreferences() {
 		this.props.submitPreferences();
 	}
+
+	private handleDefaultWarningFileSizeChange(e: { target: HTMLInputElement; }) {
+		this.props.updateDefaultWarningFileSize(parseFloat(e.target.value));
```
huss, line 264: I think these two parseFloat calls could cause an error. In other places in OED it checks the input value to be a valid float so maybe so that?

src/server/sql/preferences/create_preferences_table.sql
```diff
@@ -9,5 +9,7 @@ CREATE TABLE IF NOT EXISTS preferences (
 	default_chart_to_render graph_type NOT NULL,
 	default_bar_stacking BOOLEAN NOT NULL,
 	default_language language_type NOT NULL,
-	default_timezone TEXT DEFAULT NULL 
+	default_timezone TEXT DEFAULT NULL,
+	default_warning_file_size VARCHAR(50) NOT NULL,
```
huss: I'm unclear on why these are a character type when it is a number.
Also, need a migration for the table change. I am pushing that out since OED does not have clear directions on doing this. Note it would need its type changed if it is changed here.

huss (changes requested):
Good and complete work. Testing did not find any issues. There are minimal comments to address. Also, src/server/test/web/preferencesTest.js test should be updated for new download values.

src/client/app/components/admin/PreferencesComponent.tsx
truongdd03: I have set the input type for these two values to be number (lines 211 & 222). Do you think we still need to check the inputs?

huss: No, you are right and I missed this. You cannot enter a non-number into the box so it is safe.

src/server/sql/preferences/create_preferences_table.sql
truongdd03: I've changed the type to float in the new commit. Thank you for helping me create the migration for the table change.

truongdd03:
Thank you for your review, @huss. I've updated the preferences test and resolved your comments. Please let me know if something else needs to be changed. Thank you.

huss (approved):
All changes look good and testing indicates they work. Note I tested the migration but did not do a full DB install to verify the preferences table creation works but it should.

PR #687, opened 2021-12-17
map chart duration same as bar chart

### LINKED ISSUE #570: Date ranges and what maps display
Maps should honor the date ranges selected in the line graphic. This one is only really valuable once the circles are not a simple 4 weeks. Maybe we should consider the average over the time range of the line graph? Maybe some options as in compare for predefined time ranges?

Note the code currently uses the bar value and issue #475 discusses how this is not correct at this time. If it did it similarly to compare it might work better but that also can have issues.

ngan-cassie:
# Description

Partly addresses #570 
- Add side options to select the duration for map chart.
- Change label for map chart.
- Fix some linting.

## Type of change

## Checklist


src/client/app/containers/BarChartContainer.ts
```diff
@@ -78,7 +78,7 @@ function mapStateToProps(state: State) {
 				const hoverText: string[] = [];
 				const readings = _.orderBy(readingsData.readings, ['startTimestamp'], ['asc']);
 				readings.forEach(barReading => {
-					// calculate more precise bar duration 
+					// subtracting one extra day caused by day ending at midnight of the next day
```
huss: This seems a duplicate of an earlier change that I already see in development branch. I think it should be removed from this PR.

huss:
Overall, this works as expected. I checked 1 & 2 maps available along with day, week & 4 week options. A few general thoughts:

- The slider is not working for me. Changes to the bar slider do not impact maps.
- I think the title of the PR should be more precise. Maybe something similar to: map chart duration same as bar chart
- The description should say it partly addresses issue #570
- I note that bar and map ranges are linked together. This is okay.
- The next step is getting the data linked to compare instead of bar. This could be an enhancement of this PR or a new PR.

huss (changes requested):

src/client/app/containers/BarChartContainer.ts
ngan-cassie: @huss how do I remove the change from the PR? Shall I go revert the comment and make a new commit? when I kept the old comment it said merge conflict because of this comment

huss: I believe the bigger issue here is that your branch is not up to date with the development branch on the OED GitHub account. Thus, it sees your comment (before the change) as changing the same line as something already in development so it sees a merge conflict. I just tried updating your PR with development and the change was no longer listed. While I can push my changes to your PR, it is probably best if you do this so your workflow will be okay in the future. There are lots of websites that describe how to do this but I think these are the basic steps you need to follow:

1. Get your clone of your fork of OED on your local machine to know about the OED repo. If you do `git remote -v` on your local machine in your clone of OED, I suspect you will see only your account on GitHub listed as origin (and not the OED repo on that account as upstream). If so, do `git remote add upstream https://github.com/OpenEnergyDashboard/OED.git`. Now if you do `git remote -v` you should see your the OED repo as upstream.
2. Get the changes from upstream (OED repo) into your work. There are several ways to do this but `git fetch` should work fine. (You can also pull from the OED repo.) There can be merge conflicts but should not be in this case. Now, you won't see your comment as a change. In the future, you can avoid the edit by doing this so the change will come from development branch on OED. Also note that you must push this out to get your GitHub account (origin) synced with your local clone that was just changed.

As I was finishing this I looked and it seems you have already brought development into your PR so you probably did something like this. I'm just leaving this for the record and you can comment if my steps are off.

huss:
The recent changes fix the items mentioned. I note two things:

1. The slider was removed from the map graphic page. That is okay for this PR. We should probably revisit this when the planned update to what maps display is done.
2. I added a link to the map help page in the new tooltip for time ranges since changes for v0.8.0 help will include information on the time range choices. Thanks to @ngan-cassie for catching the need for the tooltip.

I appreciate @ngan-cassie finishing this up and it is ready to merge.

huss (approved):

PR #690, opened 2021-12-23
Warn users about unsaved changes

### LINKED ISSUE #499: Warn users about unsaved changes before they close the page
Since react will not remember the state when a page is refreshed or another page is opened, a Pop-up menu should be displayed before a user closes OED, listing all pages with unsaved changes. The user could then make the following choices about these changes: 

- save: save all edits to database;
- cancel: drop the changes and leave;

issue #493 lists an example of this problem.

truongdd03:
# Description

This PR updates react-router to v5. Some changes need for the new version:
- [query-string](https://www.npmjs.com/package/query-string) is installed to get queries from links. 
- [react-router-dom](https://www.npmjs.com/package/react-router-dom/v/5.3.0) is installed to use Link.

This PR also adds unsaved changes warnings when users leave the page.

Fixes #499 

## Type of change

## Checklist

## Limitations

- Needs labels in French and Spanish.


package.json
```diff
@@ -64,7 +67,7 @@
     "es6-promisify": "~5.0.0",
     "express": "~4.16.4",
     "express-rate-limit": "~5.5.1",
-    "history": "~2.1.2",
+    "history": "~4.7.2",
```
huss, line 70: Most of the packages that were changed are either the latest or fairly new versions. The history package is 4 years old where 5.1.0 or 5.2.0 are recent. I don't know if there are issues in using the newer versions but it would be nice to use them if possible.

src/client/app/components/maps/MapViewComponent.tsx
```diff
@@ -74,6 +77,37 @@ class MapViewComponent extends React.Component<MapViewPropsWithIntl, MapViewStat
...
+		});
+	}
+
+	private submitUnsavedChangesFunction(successCallback: () => void, failureCallback: () => void) {
+		// This function is called to submit the unsaved changes
+		store.dispatch(submitEditedMaps()).then(successCallback, failureCallback);
+	}
+
+	private updateUnsavedChanges() {
+		// Notifiy that there are unsaved changes
```
huss: Notifiy -> Notify

src/client/app/components/meters/MeterViewComponent.tsx
```diff
@@ -66,6 +69,30 @@ class MeterViewComponent extends React.Component<MeterViewPropsWithIntl, MeterVi
...
+		});
+	}
+
+	private submitUnsavedChangesFunction(successCallback: () => void, failureCallback: () => void) {
+		// This function is called to submit the unsaved changes
+		store.dispatch(submitEditedMeters()).then(successCallback, failureCallback);
+	}
+
+	private updateUnsavedChanges() {
+		// Notifiy that there are unsaved changes
```
huss: Notifiy -> Notify

src/client/app/components/RouteComponent.tsx
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import * as React from 'react';
-import { RedirectFunction, Route, Router, RouterState } from 'react-router';
+import { Route, Router, Switch, Redirect } from 'react-router-dom';
```
huss, line 6: I just want to note that the changes seem to work but I'm not up on the auth code.

src/client/app/components/UnsavedWarningComponent.tsx
```diff
@@ -0,0 +1,153 @@
...
+					window.onbeforeunload = () => undefined;
+					// Navigate to the path that the user wants
+					this.props.history.push(this.state.nextLocation);
+				});
+			});
+		}
+	}
+
+	/**
+	 * Called when successfully submiting the unsaved changes
```
huss: submiting -> submitting

huss (changes requested):
I thank @truongdd03 for the thoroughness of this work. I tested and found no issues. I have a couple of simple comments that would be nice to address but this PR seems about ready to merge.

package.json
truongdd03: The history package is related to react-router. Our current react-router version is only compatible with history v4. I have googled and found that we need to update react-router to v6 to have history v5 (https://github.com/remix-run/history/issues/804). So I think that we cannot use a newer version of history at the moment.

huss: Thanks to @truongdd03 for checking this out. I'm good with the version in the PR.

huss (approved):

PR #710, opened 2022-02-04
Display start and end times for compare graph

### LINKED ISSUE #477: Line/bar/compare graph x point value and label

ngan-cassie:
# Description

Fixes #477 but completing the last item to do in the issue.

## Type of change

## Checklist

## Limitations

I tried loading test data but it threw an error at the end when I checked whether the readings were moved successfully. Perhaps you could help me with this. Here's the terminal output:
[issue with using test data.txt](https://github.com/OpenEnergyDashboard/OED/files/8000084/issue.with.using.test.data.txt)

Hopefully this compare graph fix can close the issue.


huss:
Thanks to @ngan-cassie for this pull request. During review, I realized there were two issues that we found but not part of the original PR work. I have pushed changes to address them:

1. I should have asked for two date ranges instead of one because the last and this bars in each comparison graph have different time ranges. I also adjusted the look to be consistent with bar graphs (not what @ngan-cassie was asked to do so my fault).
2. There was a performance and bug issue with the end time in Redux state. The changes made the bug be clear. See the last commit for what was up.

I ask that @ngan-cassie review my changes to see if they approve or think more changes would be appropriate.

huss:
For unknown reasons, my last two pushes to ngan PR branch are not showing up here. They were just attempts to get the tests to fire. Also, GitHub Actions did not trigger on the last commit that is showing up. The checks do work on my machine. Unsure why this is all happening.

ngan-cassie:
@huss  the changes look good to me. I don't have any other suggestion. Thank you for the code!

huss (approved):
The changes look good and run fine for me. The ones I made were looked at by @ngan-cassie. The PR is fine and ready to merge.

PR #711, opened 2022-02-08
New models and DB tables

truongdd03:
# Description

Fixes create-all-new-DB-tables and create-models-to-interface-to-DB-changes.

This PR adds new models and changes existing ones preparing for the upcoming generalization. Some tests related to these changes are also added.

## Type of change

## Checklist

## Limitations

In src/server/models/Unit.js, there are two TODO comments needs to resolve. I haven't yet decided which special value we should return.

### DIFF
--- a/src/server/models/Conversion.js
+++ b/src/server/models/Conversion.js
@@ -25,7 +25,7 @@ class Conversion {
 	}
 
 	/**
-	 * Return a promise to create the conversions table.
+	 * Returns a promise to create the conversions table.
 	 * @param {*} conn The connection to use.
 	 * @returns {Promise.<>}
 	 */
@@ -34,7 +34,7 @@ class Conversion {
 	}
 
 	/**
-	 * Create a new conversion from the row's data.
+	 * Creates a new conversion from the row's data.
 	 * @param {*} row The row from which the conversion will be created.
 	 * @returns The new conversion object.
 	 */
--- a/src/server/models/Conversion.js
+++ b/src/server/models/Conversion.js
@@ -1,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const database = require('./database');
+const Unit = require('./Unit');
+const sqlFile = database.sqlFile;
+
+class Conversion {
+	/**
+	 * @param {*} sourceId The unit id of the source.
+	 * @param {*} destinationId The unit id of the destination.
+	 * @param {*} bidirectional Is this conversion bidirectional?
+	 * @param {*} slope The slope of the conversion.
+	 * @param {*} intercept The intercept of the conversion.
+	 * @param {*} note Comments by the admin or OED inserted.
+	 */
+	constructor(sourceId, destinationId, bidirectional, slope, intercept, note) {
+		this.sourceId = sourceId;
+		this.destinationId = destinationId;
+		this.bidirectional = bidirectional;
+		this.slope = slope;
+		this.intercept = intercept;
+		this.note = note;
+	}
+
+	/**
+	 * Return a promise to create the conversions table.
+	 * @param {*} conn The connection to use.
+	 * @returns {Promise.<>}
+	 */
+	static createTable(conn) {
+		return conn.none(sqlFile('conversion/create_conversions_table.sql'));
+	}
+
+	/**
+	 * Create a new conversion from the row's data.
+	 * @param {*} row The row from which the conversion will be created.
+	 * @returns The new conversion object.
+	 */
+	static mapRow(row) {
+		return new Conversion(row.source_id, row.destination_id, row.bidirectional, row.slope, row.intercept, row.note);
+	}
+
+	/**
+	 * Returns the conversion associated with source and destination. If the conversion doesn't exist then return null.
+	 * @param {*} source The source unit id.
+	 * @param {*} destination The destination unit id.
+	 * @param {*} conn The connection to use.
+	 * @returns {Promise.<Conversion>}
+	 */
+	static async getBySourceDestination(source, destination, conn) {
+		const row = await conn.oneOrNone(sqlFile('conversion/get_by_source_destination.sql'), {
+			source: source,
+			destination: destination
+		});
+		return row === null ? null : Conversion.mapRow(row);
+	}
+
+	/**
+	 * Inserts a new conversion to the database.
+	 * @param {*} conn The connection to use.
+	 */
+	async insert(conn) {
+		const conversion = this;
+		await conn.none(sqlFile('conversion/insert_new_conversion.sql'), conversion);
+	}
+
+	/**
+	 * Updates an existed conversion in the database.
+	 * @param {*} conn The connection to use.
+	 */
+	async update(conn) {
+		const conversion = this;
+		await conn.none(sqlFile('conversion/update_conversion.sql'), conversion);
+	}
+
+	/**
+	 * Deletes the conversion associated with source and destination from the database.
+	 * @param {*} source The source unit id.
+	 * @param {*} destination The destination unit id.
+	 * @param {*} conn The connection to use.
+	 */
+	static async delete(source, destination , conn) {
+		await conn.none(sqlFile('conversion/delete_conversion.sql'), {
+			source: source,
+			destination: destination
+		});
+	}
+}
+
+module.exports = Conversion;
--- a/src/server/models/Conversion.js
+++ b/src/server/models/Conversion.js
@@ -3,7 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const database = require('./database');
-const Unit = require('./Unit');
 const sqlFile = database.sqlFile;
 
 class Conversion {
--- a/src/server/models/Group.js
+++ b/src/server/models/Group.js
@@ -43,7 +43,7 @@ class Group {
 	 * @return {Promise.<void>}
 	 */
 	async insert(conn) {
-		// Shallow copies the group object so that the original group's unitId will not be changed to null.
+		// Shallow copies the group object so that the original group's defaultGraphicUnit will not be changed to null.
 		const group = {...this};
 		if (group.id !== undefined) {
 			throw new Error('Attempt to insert a group that already has an ID');
@@ -186,7 +186,7 @@ class Group {
 	 * @returns {Promise.<>}
 	 */
 	async update(conn) {
-		// Shallow copies the group object so that the original group's unitId will not be changed to null.
+		// Shallow copies the group object so that the original group's defaultGraphicUnit will not be changed to null.
 		const group = {...this};
 		if (group.id === undefined) {
 			throw new Error('Attempt to update a group with no ID');
--- a/src/server/models/Group.js
+++ b/src/server/models/Group.js
@@ -186,7 +186,8 @@ class Group {
 	 * @returns {Promise.<>}
 	 */
 	async update(conn) {
-		const group = this;
+		// Shallow copies the group object so that the original group's unitId will not be changed to null.
+		const group = {...this};
 		if (group.id === undefined) {
 			throw new Error('Attempt to update a group with no ID');
 		}
--- a/src/server/models/Group.js
+++ b/src/server/models/Group.js
@@ -16,14 +16,16 @@ class Group {
 	 * @param gps Location in format of GIS coordinates, default null
 	 * @param note Note about the group
 	 * @param area Area of the group, default null
+	 * @param defaultGraphicUnit The foreign key to the unit table represents the preferred unit to display this group.
 	 */
-	constructor(id, name, displayable, gps, note, area) {
+	constructor(id, name, displayable, gps, note, area, defaultGraphicUnit) {
 		this.id = id;
 		this.name = name;
 		this.displayable = displayable;
 		this.gps = gps;
 		this.note = note;
 		this.area = area;
+		this.defaultGraphicUnit = defaultGraphicUnit;
 	}
 
 	/**
@@ -41,9 +43,14 @@ class Group {
 	 * @return {Promise.<void>}
 	 */
 	async insert(conn) {
-		const group = this;
+		// Shallow copies the group object so that the original group's unitId will not be changed to null.
+		const group = {...this};
 		if (group.id !== undefined) {
 			throw new Error('Attempt to insert a group that already has an ID');
+		}
+		// Switches the value of defaultGraphicUnit to null if it's equal to -99.
+		if (group.defaultGraphicUnit === -99) {
+			group.defaultGraphicUnit = null;
 		}
 		const resp = await conn.one(sqlFile('group/insert_new_group.sql'), group);
 		// resp = { id: 42 }, hence this line
@@ -56,7 +63,12 @@ class Group {
 	 * @returns {Group}
 	 */
 	static mapRow(row) {
-		return new Group(row.id, row.name, row.displayable, row.gps, row.note, row.area);
+		// Switches the value of defaultGraphicUnit to -99 if it's equal to null.
+		var defaultGraphicUnit = row.default_graphic_unit;
+		if (defaultGraphicUnit === null) {
+			defaultGraphicUnit = -99;
+		}
+		return new Group(row.id, row.name, row.displayable, row.gps, row.note, row.area, defaultGraphicUnit);
 	}
 
 	/**
@@ -88,6 +100,16 @@ class Group {
 	 */
 	static async getAll(conn) {
 		const rows = await conn.any(sqlFile('group/get_all_groups.sql'));
+		return rows.map(Group.mapRow);
+	}
+
+	/**
+	 * Returns a promise to retrive all groups that displayable is equal to true.
+	 * @param {*} conn The connection to be used
+	 * @returns {Promise.<Array.<Group>>}
+	 */
+	static async getDisplayable(conn) {
+		const rows = await conn.any(sqlFile('group/get_displayable.sql'));
 		return rows.map(Group.mapRow);
 	}
 
@@ -164,10 +186,15 @@ class Group {
 	 * @returns {Promise.<>}
 	 */
 	async update(conn) {
-		if (this.id === undefined) {
+		const group = this;
+		if (group.id === undefined) {
 			throw new Error('Attempt to update a group with no ID');
 		}
-		await conn.none(sqlFile('group/update_group.sql'), this);
+		// Switches the value of defaultGraphicUnit to null if it's equal to -99.
+		if (group.defaultGraphicUnit === -99) {
+			group.defaultGraphicUnit = null;
+		}
+		await conn.none(sqlFile('group/update_group.sql'), group);
 	}
 
 	/**
--- a/src/server/models/Meter.js
+++ b/src/server/models/Meter.js
@@ -5,6 +5,7 @@
 const moment = require('moment');
 const database = require('./database');
 const Reading = require('./Reading');
+const Unit = require('./Unit');
 
 const sqlFile = database.sqlFile;
 
@@ -33,11 +34,13 @@ class Meter {
 	 * @param reading The value of reading, default 0.0
 	 * @param startTimestamp Start timestamp of last reading input for this meter, default '1970-01-01 00:00:00'
 	 * @param endTimestamp  End timestamp of last reading input for this meter, '1970-01-01 00:00:00' 
+	 * @param unitId The foreign key to the unit table. The meter receives data and points to this unit in the graph
+	 * @param defaultGraphicUnit The foreign key to the unit table represents the preferred unit to display this meter
 	 */
 	constructor(id, name, ipAddress, enabled, displayable, type, meterTimezone, gps = undefined, identifier = name, note, area,
 		cumulative = false, cumulativeReset = false, cumulativeResetStart = '00:00:00', cumulativeResetEnd = '23:59:59.999999',
 		readingGap = 0, readingVariation = 0, readingDuplication = 1, timeSort = 'increasing', endOnlyTime = false,
-		reading = 0.0, startTimestamp = moment(0), endTimestamp = moment(0)) {
+		reading = 0.0, startTimestamp = moment(0), endTimestamp = moment(0), unitId, defaultGraphicUnit) {
 		// In order for the CSV pipeline to work, the order of the parameters needs to match the order that the fields are declared.
 		// In addition, each new parameter has to be added at the very end.
 		this.id = id;
@@ -63,6 +66,8 @@ class Meter {
 		this.reading = reading;
 		this.startTimestamp = startTimestamp;
 		this.endTimestamp = endTimestamp;
+		this.unitId = unitId;
+		this.defaultGraphicUnit = defaultGraphicUnit;
 	}
 
 	/**
@@ -115,7 +120,7 @@ class Meter {
 			row.default_timezone_meter, row.gps, row.identifier, row.note, row.area, row.cumulative, row.cumulative_reset,
 			row.cumulative_reset_start, row.cumulative_reset_end, row.reading_gap, row.reading_variation,
 			row.reading_duplication, row.time_sort, row.end_only_time,
-			row.reading, row.start_timestamp, row.end_timestamp);
+			row.reading, row.start_timestamp, row.end_timestamp, row.unit_id, row.defaultGraphicUnit);
 	}
 
 	/**
@@ -157,6 +162,20 @@ class Meter {
 	static async getEnabled(conn) {
 		const rows = await conn.any(sqlFile('meter/get_enabled_meters.sql'));
 		return rows.map(Meter.mapRow);
+	}
+
+	/**
+	 * For the given meter id, gets the associated unitId. 
+	 * Then, returns the unitIndex (the row/column id in the Cij/Pij table) of that unitId.
+	 * @param {*} meterId The meter id.
+	 * @param {*} conn The connection to use.
+	 * @return {Promise.<Int>}
+	 */
+	static async getUnitIndex(meterId, conn) {
+		const resp = await conn.one(sqlFile('meter/get_unit_id.sql'), { meterId: meterId });
+		const unitId = resp.unit_id;
+		const unit = await Unit.getById(unitId, conn);
+		return unit.unitIndex;
 	}
 
 	/**
--- a/src/server/models/Meter.js
+++ b/src/server/models/Meter.js
@@ -40,7 +40,7 @@ class Meter {
 	constructor(id, name, ipAddress, enabled, displayable, type, meterTimezone, gps = undefined, identifier = name, note, area,
 		cumulative = false, cumulativeReset = false, cumulativeResetStart = '00:00:00', cumulativeResetEnd = '23:59:59.999999',
 		readingGap = 0, readingVariation = 0, readingDuplication = 1, timeSort = 'increasing', endOnlyTime = false,
-		reading = 0.0, startTimestamp = moment(0), endTimestamp = moment(0), unitId, defaultGraphicUnit) {
+		reading = 0.0, startTimestamp = moment(0), endTimestamp = moment(0), unitId = -99, defaultGraphicUnit = unitId) {
 		// In order for the CSV pipeline to work, the order of the parameters needs to match the order that the fields are declared.
 		// In addition, each new parameter has to be added at the very end.
 		this.id = id;
@@ -116,11 +116,13 @@ class Meter {
 	 * @returns Meter from row
 	 */
 	static mapRow(row) {
-		return new Meter(row.id, row.name, row.ipaddress, row.enabled, row.displayable, row.meter_type,
-			row.default_timezone_meter, row.gps, row.identifier, row.note, row.area, row.cumulative, row.cumulative_reset,
-			row.cumulative_reset_start, row.cumulative_reset_end, row.reading_gap, row.reading_variation,
-			row.reading_duplication, row.time_sort, row.end_only_time,
-			row.reading, row.start_timestamp, row.end_timestamp, row.unit_id, row.default_graphic_unit);
+		var meter = new Meter(row.id, row.name, row.ipaddress, row.enabled, row.displayable, row.meter_type, row.default_timezone_meter, 
+						row.gps, row.identifier, row.note, row.area, row.cumulative, row.cumulative_reset, row.cumulative_reset_start, 
+						row.cumulative_reset_end, row.reading_gap, row.reading_variation, row.reading_duplication, row.time_sort, 
+						row.end_only_time, row.reading, row.start_timestamp, row.end_timestamp, row.unit_id, row.default_graphic_unit);
+		meter.unitId = Meter.convertUnitValue(meter.unitId);
+		meter.defaultGraphicUnit = Meter.convertUnitValue(meter.defaultGraphicUnit);
+		return meter;
 	}
 
 	/**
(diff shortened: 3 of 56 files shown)


src/server/sql/conversion/create_conversions_table.sql
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+CREATE TABLE IF NOT EXISTS conversions (
+    source_id INTEGER,
```
huss: This and destination_id needs to be a foreign key:
source_id INT NOT NULL REFERENCES units(id)

src/server/sql/conversion/create_conversions_table.sql
```diff
@@ -0,0 +1,12 @@
...
+    destination_id INTEGER CHECK (source_id != destination_id),
+    bidirectional BOOLEAN NOT NULL,
+    slope FLOAT,
+    intercept FLOAT,
+    note TEXT
+);
```
huss, line 14: Need a primary key:
PRIMARY KEY (source_id, destination_id)

src/server/sql/group/create_groups_tables.sql
```diff
@@ -8,7 +8,8 @@ CREATE TABLE IF NOT EXISTS groups (
 	displayable BOOLEAN,
 	gps POINT DEFAULT NULL,
 	note TEXT,
-	area REAL DEFAULT NULL
+	area REAL DEFAULT NULL,
+	default_graphic_unit INTEGER
```
huss: Needs to be a foreign key:
default_graphic_unit INT REFERENCES units(id)

src/server/sql/group/get_all_groups.sql
```diff
@@ -1,4 +1,4 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-SELECT id, name, displayable, gps, note, area FROM groups;
+SELECT id, name, displayable, gps, note, area, default_graphic_unit FROM groups;
```
huss: I know this is following what the code already had but let's fix it to be like meters and make future changes easier:
SELECT * FROM groups;

src/server/sql/group/get_displayable.sql
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+SELECT *
```
huss: While this layout is fine, other functions in OED make it a single line when it is simple like this.

src/server/sql/group/get_group_by_id.sql
```diff
@@ -1,4 +1,4 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-SELECT id, name, displayable, gps, note, area FROM groups WHERE id=${id};
+SELECT id, name, displayable, gps, note, area, default_graphic_unit FROM groups WHERE id=${id};
```
huss: As with get_all_groups, let's use * to get columns.

src/server/sql/group/get_group_by_name.sql
```diff
@@ -1,4 +1,4 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-SELECT id, name, displayable, gps, note, area FROM groups WHERE name=${name};
+SELECT id, name, displayable, gps, note, area, default_graphic_unit FROM groups WHERE name=${name};
```
huss: Again, let's use *.

src/server/sql/meter/create_meters_table.sql
```diff
@@ -51,5 +53,7 @@ CREATE TABLE IF NOT EXISTS meters (
     end_only_time BOOLEAN DEFAULT false,
     reading FLOAT DEFAULT 0.0,
     start_timestamp TIMESTAMP DEFAULT '1970-01-01 00:00:00',
-    end_timestamp TIMESTAMP DEFAULT '1970-01-01 00:00:00'
+    end_timestamp TIMESTAMP DEFAULT '1970-01-01 00:00:00',
+    unit_id INTEGER NOT NULL,
```
huss: Need foreign key:
unit_id INT REFERENCES units(id)

Also applies to default_graphic_unit but with NOT NULL.

src/server/sql/meter/get_meter_by_id.sql
```diff
@@ -2,9 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-SELECT id, name, ipaddress, enabled, displayable, meter_type, default_timezone_meter, gps, identifier,
-    note, area, cumulative, cumulative_reset, cumulative_reset_start, cumulative_reset_end,
-    reading_gap, reading_variation, reading_duplication, time_sort, end_only_time,
-    reading, start_timestamp, end_timestamp
+SELECT *
```
huss: Since you nicely made this much shorter, let's make it a single line.

src/server/sql/meter/get_meter_by_name.sql
```diff
@@ -2,9 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-SELECT id, name, ipaddress, enabled, displayable, meter_type, default_timezone_meter, gps, identifier,
-    note, area, cumulative, cumulative_reset, cumulative_reset_start, cumulative_reset_end,
-    reading_gap, reading_variation, reading_duplication, time_sort, end_only_time,
-    reading, start_timestamp, end_timestamp
+SELECT *
```
huss: Since you nicely made this much shorter, let's make it a single line.

src/server/sql/meter/get_unit_id.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT unit_id
```
huss: Format to one line so consistent with other OED SQL.

src/server/sql/unit/create_units_table.sql
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+CREATE TABLE IF NOT EXISTS units (
+	id SERIAL PRIMARY KEY,
+	name VARCHAR(50) UNIQUE NOT NULL,
```
huss: Why don't we do what is done in group and add:

UNIQUE NOT NULL CHECK (char_length(name) >= 1)

same applies to identifier.

If you want to be helpful, this can also be added to meter table creation

src/server/sql/unit/create_units_table.sql
```diff
@@ -0,0 +1,16 @@
...
+	identifier VARCHAR(50) UNIQUE NOT NULL,
+	unit_type unit_type NOT NULL,
```
huss: Missing unit_represent & sec_in_rate columns above this one. That was a later addition to the design document.

The name of the column should be type_of_unit. I don't think it is a good idea to have a type and column name be the same.

src/server/sql/unit/create_units_table.sql
```diff
@@ -0,0 +1,16 @@
...
+	unit_index INTEGER NOT NULL,
```
huss: Since units are created before the Cij table is created, this can be null at the beginning so remove NOT NULL.

src/server/sql/unit/create_units_table.sql
```diff
@@ -0,0 +1,16 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+CREATE TABLE IF NOT EXISTS units (
+	id SERIAL PRIMARY KEY,
+	name VARCHAR(50) UNIQUE NOT NULL,
+	identifier VARCHAR(50) UNIQUE NOT NULL,
+	unit_type unit_type NOT NULL,
+	unit_index INTEGER NOT NULL, 
+	suffix VARCHAR(50) DEFAULT '',
+	displayable_type displayable_type NOT NULL,
```
huss: The name of the column should be displayable.

src/server/sql/unit/create_units_table.sql
```diff
@@ -0,0 +1,16 @@
...
+	is_primary BOOLEAN NOT NULL,
```
huss: While I agree is_primary might be a better name than primary for a boolean, the design document specifies primary so we should do that. Note that everywhere else we use displayable, etc. without is_ so this is consistent.

src/server/sql/unit/create_units_table.sql
```diff
@@ -0,0 +1,16 @@
...
+	note TEXT,
+	unit_represent_type unit_represent_type NOT NULL
```
huss: I think the design document confused you. I think this will work (never actually tried in Postgres):

UNIQUE (type_of_unit, unit_index)

src/server/sql/unit/get_by_id.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
```
huss: Format as single line.

src/server/sql/unit/get_by_name.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
```
huss: Format as single line.

src/server/sql/unit/get_by_unit_index_meter.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT id
+FROM units 
+WHERE unit_type = 'meter' AND identifier = ${unitIndex};
```
huss: Is it possible to use unit_type.meter?  Column name changed.

I think this should compare to unit_index and not identifier.

src/server/sql/unit/get_by_unit_index_unit.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT id
+FROM units 
+WHERE unit_type = 'unit' AND identifier = ${unitIndex};
```
huss: Is it possible to use unit_type.unit?

I think this should compare to unit_index and not identifier.

src/server/sql/unit/get_suffix.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
```
huss: Format as single line.

src/server/sql/unit/get_type_meter.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
```
huss: Format as single line.

src/server/sql/unit/get_type_meter.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT * 
+FROM units 
+WHERE unit_type = 'meter';
```
huss: Is it possible to use unit_type.meter? Column name changed.

src/server/sql/unit/get_type_unit.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
+FROM units
+WHERE unit_type = 'unit';
```
huss: Is it possible to use unit_type.unit? Column name changed.

src/server/sql/unit/get_type_unit.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
```
huss: Format as single line.

src/server/sql/unit/get_visible_meter.sql
```diff
@@ -0,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
+FROM units
+WHERE unit_type = 'meter'
```
huss: Is it possible to use unit_type.meter? Column name changed.

src/server/sql/unit/get_visible_meter.sql
```diff
@@ -0,0 +1,10 @@
...
+-- Returns unit with properly displayable_type. 
+-- If user is admin then return units with displayable of all or admin if user is admin.
+AND (displayable_type = ${user} OR (displayable_type = 'all' AND ${user} = 'admin'));
```
huss: I think the logic for the first clause should be: displayable = displayable_type.all. This means any unit that is visible to everyone. The second part should be: displayable = displayable_type.admin AND ${user} = 'admin'. This means only show ones for admin if this user is an admin.

src/server/sql/unit/get_visible_unit_or_suffix.sql
```diff
@@ -0,0 +1,10 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+SELECT *
+FROM units
+WHERE (unit_type = 'unit' OR unit_type = 'suffix')
```
huss: Is it possible to use unit_type.unit & unit_type.suffix? Column name changed.

src/server/sql/unit/get_visible_unit_or_suffix.sql
```diff
@@ -0,0 +1,10 @@
...
+-- Returns unit with properly displayable_type. 
+-- If user is admin then return units with displayable of all or admin if user is admin.
+AND (displayable_type = ${user} OR (displayable_type = 'all' AND ${user} = 'admin'));
```
huss: I think the logic for the first clause should be: displayable = displayable_type.all. This means any unit that is visible to everyone. The second part should be: displayable = displayable_type.admin AND ${user} = 'admin'. This means only show ones for admin if this user is an admin.

src/server/sql/unit/insert_new_unit.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+INSERT INTO units(name, identifier, unit_type, unit_index, suffix, displayable_type, is_primary, note, unit_represent_type)
```
huss: The names and columns need to be changed per request in create_units_table.sql

src/server/sql/unit/update_unit.sql
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+UPDATE units
+	SET name = ${name},
```
huss, line 6: The names and columns need to be changed per request in create_units_table.sql

src/server/models/Conversion.js
```diff
@@ -0,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const database = require('./database');
+const Unit = require('./Unit');
```
huss: I don't think this import is used.

src/server/models/Group.js
```diff
@@ -41,10 +43,15 @@ class Group {
 	 * @return {Promise.<void>}
 	 */
 	async insert(conn) {
-		const group = this;
+		// Shallow copies the group object so that the original group's unitId will not be changed to null.
```
huss: Is it unitId or defaultGraphiUnit that can be changed? Same for update below.

src/server/models/Meter.js
```diff
@@ -33,11 +34,13 @@ class Meter {
...
 	 * @param startTimestamp Start timestamp of last reading input for this meter, default '1970-01-01 00:00:00'
 	 * @param endTimestamp  End timestamp of last reading input for this meter, '1970-01-01 00:00:00' 
+	 * @param unitId The foreign key to the unit table. The meter receives data and points to this unit in the graph, default 0
+	 * @param defaultGraphicUnit The foreign key to the unit table represents the preferred unit to display this meter, default 0
 	 */
 	constructor(id, name, ipAddress, enabled, displayable, type, meterTimezone, gps = undefined, identifier = name, note, area,
 		cumulative = false, cumulativeReset = false, cumulativeResetStart = '00:00:00', cumulativeResetEnd = '23:59:59.999999',
 		readingGap = 0, readingVariation = 0, readingDuplication = 1, timeSort = 'increasing', endOnlyTime = false,
-		reading = 0.0, startTimestamp = moment(0), endTimestamp = moment(0)) {
+		reading = 0.0, startTimestamp = moment(0), endTimestamp = moment(0), unitId = 0, defaultGraphicUnit = 0) {
```
huss: I'm wondering about the default value of 0 for the two new parameters. For unitID it should never be used if it was not set so maybe -1 would be better since this should cause an error. For defaultGraphicUnit, it isn't clear what it should be but it needs to be valid. Usually the ids start at 1 so I would use that but it really isn't safe. I think it might be best to not have a default or get a valid, random one from the units table.

src/server/models/Meter.js
```diff
@@ -159,6 +164,20 @@ class Meter {
...
 	}
 
+	/**
+	 * For the given meter id, gets the associated unitId. 
+	 * Then, returns the unitIndex (the row/column id in the Cij/Pij table) of that unitId.
+	 * @param {*} meterId The meter id.
+	 * @param {*} conn The connection to use.
+	 * @return {Promise.<Int>}
+	 */
+	static async getUnitIndex(meterId, conn) {
```
huss, line 176: I agree with the name change from the design document. I'll change it there but in the future we need to be careful that they stay the same as this function is used in several places in that document.

src/server/models/Unit.js
```diff
@@ -0,0 +1,225 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const database = require('./database');
+const sqlFile = database.sqlFile;
+
+class Unit {
+	/**
```
huss, line 9: The variable names need to be updated for changes proposed in create_units_table.sql. Also, add new columns.

src/server/models/Unit.js
```diff
@@ -0,0 +1,225 @@
...
+	 * @param {*} id This unit's ID.
+	 * @param {*} name This unit's name used internally and by the admin.
+	 * @param {*} identifier This unit's identifier displayed to the user. 
+	 * @param {*} unitType This unit's type. Can be meter, unit, or suffix.
+	 * @param {*} unitIndex The unique number for row/column index in conversion table for this unit.
+	 * @param {*} suffix This unit's suffix.
+	 * @param {*} displayable Can be none, all, or admin. Restrict the type of user that can see this unit.
+	 * @param {*} primary True if this unit is always displayed. If not, the user needs to ask to see (for future enhancement).
+	 * @param {*} note Note about this unit.
+	 * @param {*} unitRepresentType Tells how the data is fetched for readings (only need for meter type unit).
```
huss: This was removed.

src/server/models/Unit.js
```diff
@@ -0,0 +1,225 @@
...
+	 * @param {*} conn The connection to use.
+	 * @returns {Promise.<>}
+	 */
+	static createUnitRepresentTypesEnum(conn) {
+		return conn.none(sqlFile('unit/create_unit_represent_types_enum.sql'));
+	}
+
+	/**
+	 * Returns all units of meter type that are visible to user.
+	 * @param {*} user Can be all or admin.
```
huss: Is the user of type string? Same question on next function.

src/server/test/db/conversionTests.js
```diff
@@ -0,0 +1,56 @@
...
+	expect(actual).to.have.property('bidirectional', expected.bidirectional);
+	expect(actual).to.have.property('slope', expected.slope);
+	expect(actual).to.have.property('intercept', expected.intercept);
+	expect(actual).to.have.property('note', expected.note);
+}
+
+mocha.describe('Conversions', () => {
+	mocha.it('can be saved and retrived', async () => {
+		const conn = testDB.getConnection();
+		const conversionPreInsert = new Conversion(0, 1, true, 0, 0, '');
```
huss: Three ideas on these values. Frist, I think using 0 when another value is valid is risky since many things initialize to zero by default. Second, I think it is best to use unique values when you can so you can detect if an incorrect variable name/location is used. Third, I think a non-empty string is better. This applies to all these tests.

src/server/test/db/meterTests.js
```diff
@@ -41,14 +42,16 @@ function expectMetersToBeEquivalent(expected, actual) {
...
+	expect(actual).to.have.property('defaultGraphicUnit', expected.defaultGraphicUnit);
 }
 
 mocha.describe('Meters', () => {
 	mocha.it('can be saved and retrieved', async () => {
 		const conn = testDB.getConnection();
 		const meterPreInsert = new Meter(undefined, 'Meter', null, false, true, Meter.type.MAMAC, 'UTC',
 		gps,'Identified', 'notes', 33.5, true, true, '05:05:09', '09:00:01', 0, 0, 1, 'increasing', false,
-		25.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10');
+		25.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', 0, 0);
```
huss: See conversionTests.js for using 0 as values.

src/server/test/db/unitTests.js
```diff
@@ -0,0 +1,182 @@
...
+const { mocha, expect, testDB } = require('../common');
+const Unit = require('../../models/Unit');
+
+/**
+ * Compares the expected and actual units.
+ * @param {*} expected The expected unit.
+ * @param {*} actual The actual unit.
+ */
+function expectUnitToBeEquivalent(expected, actual) {
+	expect(actual).to.have.property('id', expected.id);
```
huss, line 14: Fix up for new proposed names and columns.

src/server/test/db/unitTests.js
```diff
@@ -0,0 +1,182 @@
...
+		expectUnitToBeEquivalent(unitTypePreInsert, unitTypePostInsertById);
+		// Gets unit by name.
+		const unitTypePostInsertByName = await Unit.getByName('Unit', conn);
+		expectUnitToBeEquivalent(unitTypePreInsert, unitTypePostInsertByName);
+		// Gets unit by index.
+		const idUnitTypePostInsertByIdentifier = await Unit.getByUnitIndexUnit('Index', conn);
+		expect(idUnitTypePostInsertByIdentifier).to.be.equal(1);
+	});
+
+	mocha.it('meter type can be retrieved by index', async () => {
```
huss: Maybe change the description so index is unit index.

huss:
I appreciate all the work that @truongdd03 did to get this to work, esp. given this is their first work with the OED database code. Overall, it is fine. I have noted a number of suggested changes to be addressed. If @truongdd03 feels it is then ready the please convert from a draft.

huss (changes requested):
I have suggested some changes.

huss (changes requested):
I have suggested some changes.

src/server/models/Group.js
truongdd03: My bad. It should be defaultGraphicUnit.

src/server/models/Unit.js
truongdd03: Do you mean changing unitRepresentType into unitRepresent? I still see this variable in the document so I'm not sure if it should be removed.

src/server/sql/unit/create_units_table.sql
truongdd03: I forgot to make a note about this. I tried to create the primary column in the database but Postgres treats it as the PRIMARY keyword. To fix this I've changed this line into `"primary" BOOLEAN NOT NULL`. Note that `primary` needs to be enclosed by `""`.

huss: Good point. I think it best not to have an attribute named primary. I've updated the resourceGeneralization doc to make this named alwaysDisplay so the DB name would be always_display. (Will update on website soon.) Could you please make this change?

src/server/models/Unit.js
huss: unit_represent_type is the type/enum in the database and unit_represent is the name of the column/attribute in the DB. So, in the model, the variable name would be unitRepresent.

src/server/sql/unit/create_units_table.sql
truongdd03: Sure! I changed it so I'll make this conversation resolved.

src/server/models/Unit.js
truongdd03: Thank you for your explanation. I've changed the variable name to unitRepresent.

src/server/sql/unit/create_units_table.sql
truongdd03: I've tested and it worked. Thank you for your suggestion!

src/server/sql/unit/get_by_unit_index_meter.sql
truongdd03: I've tried to use `unit_type.meter` but it failed. [The Postgres documentation about enum](https://www.postgresql.org/docs/9.1/datatype-enum.html) just does the same as the code above.

However, for safety, I think we can replace `'meter'` with `'meter'::unit_type`. This will cast `meter` to `unit_type` and will throw an error if `'meter'` is not a value of the `unit_type` enum. What do you think about this?

huss: That sounds fine. We don't expect the OED code to ever pass anything else but checks are always good.

truongdd03: I've changed other places as well so I'll mark them as resolved.

src/server/models/Meter.js
truongdd03: I agree with you that having default values is quite risky. I will remove them.

truongdd03:
Thank you so much for your review, @huss. I've resolved all the comments and addressed previous issues. The unitId and defaultGraphicUnit of meters are also allowed to be null now.

src/server/models/Meter.js
```diff
@@ -33,11 +34,13 @@ class Meter {
 	 * @param reading The value of reading, default 0.0
 	 * @param startTimestamp Start timestamp of last reading input for this meter, default '1970-01-01 00:00:00'
 	 * @param endTimestamp  End timestamp of last reading input for this meter, '1970-01-01 00:00:00' 
+	 * @param unitId The foreign key to the unit table. The meter receives data and points to this unit in the graph, default 0
```
huss: The default values changed in the code for unitID & defaultGraphicUnit.

src/server/test/db/groupTests.js
```diff
@@ -8,40 +8,75 @@ const { mocha, expect, testDB } = require('../common');
...
+	const unitB = new Unit(undefined, 'Unit B', 'Unit B Id', Unit.unitRepresentType.UNUSED, 2000, 
+							Unit.unitType.UNIT, 2, 'Unit B Suffix', Unit.displayableType.ALL, true, 'Unit B Note');
+	const unitC = new Unit(undefined, 'Unit C', 'Unit C Id', Unit.unitRepresentType.UNUSED, 3000, 
+							Unit.unitType.UNIT, 3, 'Unit C Suffix', Unit.displayableType.ALL, true, 'Unit C Note');
+	const groupA = new Group(undefined, 'GA', true, gps, 'notes GA', 33.5, 1);
+	const groupB = new Group(undefined, 'GB', false, gps, 'notes GB', 43.5, 2);
+	const groupC = new Group(undefined, 'GC', true, gps, 'notes GC', 53.5, -99);
+	await Promise.all([unitA, unitB, unitC].map(unit => unit.insert(conn)));
 	await Promise.all([groupA, groupB, groupC].map(group => group.insert(conn)));
 	const meterA = new Meter(undefined, 'MA', null, false, true, Meter.type.MAMAC, null, gps,
```
huss, line 29: I think using 1-3 for the unit id will always work since the DB starts empty. However, in general, it is probably safer to use the ids returned on unit.insert. Same in meterTest.js and others.

src/server/models/Meter.js
```diff
@@ -229,6 +266,44 @@ class Meter {
...
+	/**
+	 * Makes the given meter valid.
+	 * @param {*} meter The meter.
+	 */
+	static makeMeterDataValid(meter) {
+		if (meter.unitId === -99) {
+			// If there is no unitId, set defaultGraphicUnit to null and displayable to false.
+			if (meter.defaultGraphicUnit !== -99) {
+				meter.defaultGraphicUnit = -99;
+				log.warn('defaultGraphicUnit has been removed since there is no unitId.');
```
huss: For this message and the similar ones that follow, I think more information should be included in the message. The log file does not have the context of the operation going on. Thus, I think the message should include the meter name.

huss (changes requested):
I thank @truongdd03 for continued, good work on this. The updates all seem good. I've asked for two small changes and noted one item on unit ids that is optional. This is very close to being merged.

truongdd03:
Thank you for your review, @huss. I have resolved all the comments and replaced unitId in every test.

huss (approved):
Everything looks good and seems to work fine. I thank @truongdd03 for tackling this large issue. This is ready to merge.

PR #718, opened 2022-02-20
fixing previous pr bugs with getDerivedStateFromProps

### LINKED ISSUE #715: bar custom slider not working
### Describe the bug

The bar custom slider moves when you click day, week or 4 week buttons but you cannot drag to move it.

### To Reproduce

Click the "Toggle custom slider" button. You cannot more it from its current position.

### Expected behavior

The slider should move and set the length of each bar.

### Additional context

This works in v0.8.0. My first guess is that the recent package updates changed something but this has not been tested.

### LINKED ISSUE #717: cannot select a meter in group creation
### Describe the bug

If you go to the group page and click to create a new group, you cannot select any meters to add to the group. It may be the case you cannot select a group but this is unknown as you cannot add one to see. Other meter menus seem to work fine.

### Expected behavior

This worked as of v0.8.0 but does not work now. It should work. It isn't yet known which merged PR caused the issue.

jina2k:
# Description

Fixes #717  & Fixes #715 

These bugs had to do with the state not changing due to getDerivedStateFromProps. 
I've checked every file previously edited from componentWillReceiveProps that turned into getDerivedStateFromProps, and concluded that it wasn't necessary in general, after testing. I removed these functions in the PR.

Also, I moved to package-lock version 2 as it's backwards compatible with v1 and supports npm v5 v6 and v7. We are currently using npm v8 but I see no issue in doing this. Once v6 is no longer relevant, it is stated in the documentation that we should move to v3. For now, v2 seems fine.

## Type of change

## Checklist

## Limitations
n/a

### DIFF
--- a/containers/web/Dockerfile
+++ b/containers/web/Dockerfile
@@ -5,7 +5,7 @@
 # *
 
 # Dockerfile for NodeJS application portion
-FROM node:14
+FROM node:16.13.2
 
 # Create the directory for the app
 RUN mkdir -p "/usr/src/app"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15301 @@
 {
   "name": "Open-Energy-Dashboard",
   "version": "0.8.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "Open-Energy-Dashboard",
+      "version": "0.8.0",
+      "license": "MPL-2.0",
+      "dependencies": {
+        "@babel/plugin-proposal-object-rest-spread": "~7.16.0",
+        "@babel/preset-react": "~7.16.0",
+        "@formatjs/intl": "~1.17.0",
+        "@types/history": "~4.7.9",
+        "@types/react-router": "~5.1.17",
+        "@types/react-router-dom": "~5.3.2",
+        "attr-accept": "~2.2.1",
+        "axios": "~0.24.0",
+        "babel-core": "~7.0.0-bridge.0",
+        "babel-plugin-formatjs": "~10.3.12",
+        "bcryptjs": "~2.4.3",
+        "body-parser": "~1.18.3",
+        "bootstrap": "~4.3.1",
+        "chokidar": "~3.5.2",
+        "co": "~4.6.0",
+        "cookie-parser": "~1.4.4",
+        "csv": "~5.3.2",
+        "debug": "~4.3.2",
+        "dotenv": "~9.0.0",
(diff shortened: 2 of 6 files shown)


huss (commented):
This does fix the two issues. Thanks. A few notes:

- Going to the map page when you have selected meters/groups now has a different behavior. All meters/groups that could not be displayed used to be grayed out. Now meters/groups that were already selected but cannot be placed on the map are still shown as selected but do not show on the map. If you unselect one of them then it goes onto the dropdown menu grayed out. The negative I see is that a user might be confused because a selected item does not show on the graphic. A positive is it fixes issue #600 but not exactly as was expected. If it can go back to the old behavior that would be nice but if not then we can leave it and change the documentation on how it works.
- The package-lock.json file lists new dependencies (and removes some). Thus, I checked the box indicating that the node_modules were updated. Let me know if you think this is wrong.
- I put the word "Fixes" in from of the other fixed issue so that GitHub properly linked to it.

jina2k:
> This does fix the two issues. Thanks. A few notes:
> 
> * Going to the map page when you have selected meters/groups now has a different behavior. All meters/groups that could not be displayed used to be grayed out. Now meters/groups that were already selected but cannot be placed on the map are still shown as selected but do not show on the map. If you unselect one of them then it goes onto the dropdown menu grayed out. The negative I see is that a user might be confused because a selected item does not show on the graphic. A positive is it fixes issue #600 but not exactly as was expected. If it can go back to the old behavior that would be nice but if not then we can leave it and change the documentation on how it works.
> * The package-lock.json file lists new dependencies (and removes some). Thus, I checked the box indicating that the node_modules were updated. Let me know if you think this is wrong.
> * I put the word "Fixes" in from of the other fixed issue so that GitHub properly linked to it.

- I wasn't sure about the package-lock.json changing dependencies from transitioning to version 2 but it looks fine regardless. I think all it did was declare a separate variable where we list all dependencies and their related versions

- I think this is fine for now. We can add details onto the issue for now, and later on it can be fixed via changing various files. I find that to implement this change to provide same results as before, we would need to combine the ChartDataSelectComponent with the MapChartSelectComponent, so that if we changed the map, we could directly handle changing meters/groups. I tried implementing the componentDidUpdate for SingleSelect and MultiSelect as it didn't have recursion issues, but it didn't do anything, leading me to believe the UIOptionsComponent is where it handles removing of values. Therefore, this next change would require changing quite a few files, but since the current issues seem of higher priority, we can let it go for now.

huss (approved):
Everything looks good. Thanks @jina2k for the work.

PR #720, opened 2022-02-23
Create graph

truongdd03:
# Description

Fixes create-graph-from-units-&-conversions-DB-tables

This PR installs graph packages (ngraph.graph and ngraph.path) and creates a new function to build the graph. The function name is "createConversionGraph" since "createGraph" has already been the name of the library.

## Type of change

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "Open-Energy-Dashboard",
-  "version": "0.7.0",
+  "version": "0.8.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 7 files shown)


huss (approved):
Thanks to @truongdd03 for doing this quickly. The changes seem fine but I cannot tell if the graph creation is correct. At some point we need to check that the created graph is correct. That can likely happen by tests of other fucntions that are coming but I wanted to note this.

I checked that the PR changes node modules. I also don't know why CodeQL did not run but not too worried about it for the PR.

Issue #662, opened 2021-10-27
update React (and Redux)


huss:
### Describe the bug

We are behind on the version of React (and Redux). This means we cannot use the latest features (such as useSelector). A quick attempt to upgrade indicated that it may cause issues with current code so this needs to be checked out.

PR #722, opened 2022-02-25
React-redux v5 to v7

jina2k:
# Description

Partially Addresses #662 

React-redux is updated to v7, allowing us to use useSelector as described in the issue. React has not been updated as of yet, currently doing incremental changes for now. 
Updated babel as well, along with redux-devtools.

## Type of change

## Checklist

## Limitations
Currently uses a workaround for store.dispatch where you explicitly cast any. This is due to type issues between thunk and dispatch. In the future, this should be replaced.


src/client/app/types/redux/actions.ts
```diff
@@ -116,8 +116,10 @@ export enum ActionType {
 
 /**
  * The type of the redux-thunk dispatch function.
+ * Uses the overloaded version from Redux-Thunk.
  */
-export type Dispatch = Disps<State>;
+// export type Dispatch = ThunkDispatch<State, void, Action<any>>;
```
huss: Can this commented line be removed?

huss (changes requested):
I thank @jina2k for taking on another package update. The react-dedux one has been delayed due to unresolved issues that were taken on for this important package. Overall, it looks good. I have one minor comment in the code. I also noticed these while trying to extensively test:

- If you go to the map page and select any meter or group the page goes white and seems to hang. If a meter was preselected that can be displayed on the map then it seems to show up. There are no errors in the console. Does this happen for you?
- Choosing items from pages now seems to have a lag between the choice and its happening that did not happen before. The more meters or groups selected, the longer the lag seems to be.
- On the group edit page, when you select either meter or group and hit the arrow to use or unuse, nothing seems to happen (used to go to the other menu). However, when you save the group, the change shows up.

src/client/app/types/redux/actions.ts
jina2k: removed commented line

jina2k:
> I thank @jina2k for taking on another package update. The react-dedux one has been delayed due to unresolved issues that were taken on for this important package. Overall, it looks good. I have one minor comment in the code. I also noticed these while trying to extensively test:
> 
> * If you go to the map page and select any meter or group the page goes white and seems to hang. If a meter was preselected that can be displayed on the map then it seems to show up. There are no errors in the console. Does this happen for you?
> * Choosing items from pages now seems to have a lag between the choice and its happening that did not happen before. The more meters or groups selected, the longer the lag seems to be.
> * On the group edit page, when you select either meter or group and hit the arrow to use or unuse, nothing seems to happen (used to go to the other menu). However, when you save the group, the change shows up.

- fixed this first issue by checking for undefined variables.
- Later on we will need to replace some store.dispatch commands with the equivalent (whether it is through useStore(), useDispatch, etc.)
- I saw this as well. I tried testing to see if I see it in the current development build as well, and I see it there as well. I'm not sure when this occurred? So it's not related to this PR, but could be related to a previous PR.

jina2k:
I took another look, and componentDidUpdate is still necessary for single and multi select component. This should not pose any issues. Again, from the previous PR, no recursion issues should exist. We would still need to implement maps to work well with the selected meters/groups, but for groups in general this isn't a problem.

Edit: Apparently the meters are now automatically removed when being displayed onto a map where they cannot be displayed.

huss (approved):
Thanks to @jina2k for addressing the review. It seems good and ready to merge.

As for the slow down, I have been working to track it down. It does not appear to happen with consistency. I have seen it and then when I come back it does not happen. Thus, the cause and a clear path to reproduce is unclear. This will need more work to track down and is outside the PR.

Issue #471, opened 2020-04-10
Message in console on OED load


huss:
When you start OED as a developer and look at the console in the web browser you see a message about a `<div>`. Should decide if it is important and fix if needed.

huss:
We have other warning that are coming out when you load and do operations. They all should be looked into.

huss:
Also, we should review from of the log files to see if there are any messages there to deal with.

huss:
Someone earlier tried looking at this but did not figure it out. They felt one of the messages was discussed in https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html#changes-in-detail.

ngan-cassie:
Warnings about componentWill.. are likely due to some react libraries like react-select, react-dom, react-redux not being up-to-date. Upgrades might be needed to fix the warnings.

huss:
All of these issues are now gone so I am closing this. There are occasional messages but the code generally notes the issue.

PR #724, opened 2022-02-27
convert outline arguments from boolean to string

ngan-cassie:
# Description

Fix 1 warning: change the argument type in a div tag from boolean to string

Partly addresses #471 

## Type of change

## Checklist

## Limitations

Some React libraries packages might need updating to remove all warnings

### DIFF
--- a/src/client/app/components/ChartSelectComponent.tsx
+++ b/src/client/app/components/ChartSelectComponent.tsx
@@ -56,25 +56,25 @@ export default class ChartSelectComponent extends React.Component<ChartSelectPro
 					</DropdownToggle>
 					<DropdownMenu>
 						<DropdownItem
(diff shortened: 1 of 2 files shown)


huss:
I am trying to replace this PR with #725. While the proposed change by @ngan-cassie removed the warning, it probably was not clear from the code what had been intended. The code should be indicating the current choice and outline is not the way to do it with this type of element. Thus, I created the other PR to change the behavior. This also fixes a desire to let users know what the current graphic choice since that was harder to see if there was no data to display. (The same is true of the language choice but that is easy to see from the page.) I appreciate the efforts of @ngan-cassie in figuring out this issue and leading to the ultimate solution. If PR #725 is accepted then this PR will be closed.

PR #727, opened 2022-03-05
Input example units and conversions

truongdd03:
# Description

This PR inputs the example units/conversions in the document. Besides that, it also contains some minor changes in unit/conversion models/tests.

Fixes automated-input-of-example-units-and-conversions

Note this brings in the development branch via a merge. This is why the node_modules must be updated. Developers need to add --build the first time they bring OED up to get node to update (docker compose up --build).

## Type of change

## Checklist


src/server/data/automatedTestingData.js
```diff
@@ -302,6 +305,68 @@ function generateVariableAmplitudeTestingData() {
 	}
 }
 
+/**
```
huss, line 308: You cannot execute the conversion insertion unless the units have been done. Thus, I was thinking a new function that combined them might be good. It would first call units and then call conversions. Then, the package.json would be modified to just have a script to call this new function. This would eliminate the ability to insert just one of these but I'm not sure that is of real interest (and that would still be in the code).

src/scripts/installOED.sh
```diff
@@ -117,6 +127,9 @@ while [ $create_error == 0 ]; do
 		# There was no createdb error so assume database ready for use so stop process
 		create_error=1
 		printf "%s\n" "  database creation had no errors so assume schema creation worked."
+		# Insert special units and conversions
```
huss: I don't think we want this here. The special units are for developers only. Thus, they should be added manually as they do with all other test data. If we leave this then production sites would also get these units.

src/server/data/automatedTestingData.js
```diff
@@ -302,6 +305,68 @@ function generateVariableAmplitudeTestingData() {
...
+		['US_dollar', 'US $', Unit.unitType.UNIT, '', Unit.displayableType.ALL, true],
+		['Euro', '€', Unit.unitType.UNIT, '', Unit.displayableType.ALL, true],
+		['kg CO2', 'kg CO2', Unit.unitType.UNIT, 'CO2', Unit.displayableType.ALL, false],
+		['Trash', '', Unit.unitType.METER, '', Unit.displayableType.NONE, false]
+	];
+
+	for (let i = 0; i < specialUnits.length; ++i) {
+		const unitData = specialUnits[i];
+		if (await Unit.getByName(unitData[0], conn) === null) {
+			await new Unit(undefined, unitData[0], unitData[1], Unit.unitRepresentType.UNUSED, 3600,
```
huss: I think the 4th parameter should be Unit.unitRepresentType.QUANTITY for each one of type METER. At this point the examples only do quantity type readings. It is okay to apply to all since it is not used if not a type meter.

I would just leave the 5th blank instead of 3600 since the code will default to 3600 and would still be okay if that is changed in the future.

src/server/data/automatedTestingData.js
```diff
@@ -302,6 +305,68 @@ function generateVariableAmplitudeTestingData() {
...
+		['Euro', '€', Unit.unitType.UNIT, '', Unit.displayableType.ALL, true],
+		['kg CO2', 'kg CO2', Unit.unitType.UNIT, 'CO2', Unit.displayableType.ALL, false],
+		['Trash', '', Unit.unitType.METER, '', Unit.displayableType.NONE, false]
+	];
+
+	for (let i = 0; i < specialUnits.length; ++i) {
+		const unitData = specialUnits[i];
+		if (await Unit.getByName(unitData[0], conn) === null) {
+			await new Unit(undefined, unitData[0], unitData[1], Unit.unitRepresentType.UNUSED, 3600, 
+				unitData[2], null, unitData[3], unitData[4], unitData[5], null).insert(conn);
```
huss: Should an empty note be null or "". I'm uncertain what other code will do if it is null. In any case, maybe it should be "special unit" so it is clear in the DB that this is the case.

src/server/models/Conversion.js
```diff
@@ -32,6 +33,31 @@ class Conversion {
...
+	 */
+	static async insertStandardConversions(conn) {
+		// The table contains standard conversions' data.
+		// Each row contains: sourceName, destinationName, bidirectional, slope, intercept, note.
+		const standardConversions = [
+			['kWh', 'MJ', true, 3.6, 0, 'kWh → MJ'],
+			['MJ', 'M3_gas', true, 2.6e-2, 0, 'MJ → M3_gas'],
+			['MJ', 'BTU', true, 947.8, 0, 'MJ → BTU'],
+			['kg', 'Metric_ton', true, 1e-3, 0, 'kg → Metric ton'],
+			['Fahrenheit', 'Celsius', true, 1.8, 32, 'Fahrenheit → Celsius']
```
huss: I think the temperature conversion is backward. It should be:
['Celsius', 'Fahrenheit', true, 1.8, 32, 'Celsius → Fahrenheit']

src/server/models/Unit.js
```diff
@@ -80,6 +80,33 @@ class Unit {
...
+			['kg', 'kg', Unit.unitType.UNIT, '', Unit.displayableType.ALL, false],
+			['Metric_ton', 'Metric ton', Unit.unitType.UNIT, '', Unit.displayableType.ALL, false],
+			['Fahrenheit', 'Fahrenheit', Unit.unitType.UNIT, '', Unit.displayableType.ALL, false],
+			['Celsius', 'Celsius', Unit.unitType.UNIT, '', Unit.displayableType.ALL, false]
+		];
+
+		for (let i = 0; i < standardUnits.length; ++i) {
+			const unitData = standardUnits[i];
+			if (await Unit.getByName(unitData[0], conn) === null) {
+				await new Unit(undefined, unitData[0], unitData[1], Unit.unitRepresentType.UNUSED, 3600,
```
huss: See comments in automatedTestingData.js for similar changes here.

huss (changes requested):

huss:
Thanks again to @truongdd03 for good work. There are a few comments to address but overall this looks good and testing found no issues. I did edit the PR description a little to clarify about merging in developement.

src/server/models/Conversion.js
truongdd03: I think the current one is correct since` Fahrenheit = 1.8 * Celsius + 32`. The order is similar to line 47, where `kg = 1e-3 * Metric_ton`.

huss: Maybe I have this wrong but I still have a concern. I think the correct conversion for line 47 is
kg = 1000 * Metric_ton
because there are 1000 kg in a metric ton. Thus, the conversion table is specifying the other way around of
Metric_ton = 1e-3 * kg
If that is true then the temperature conversion is backward.

truongdd03: My bad. You are correct. The conversion should be reversed. Thank you for your explanation.

truongdd03:
Thank you so much for your review, @huss. I have addressed the issues and resolved all the comments. Please let me know if there's something else that needs to be changed. Thank you.

huss (approved):
Changes all look good. This is ready to merge.

Issue #577, opened 2021-03-02
sort meter page by name not ID


huss:
Since we are not going to show the meter id on the meter page anymore (for non-admin), sorting by this value is confusing. It also does not really make sense since it is not something people know about. Thus, we should change this page to sort by name instead.

huss:
A good place to start is src/client/app/components/meters/MeterViewComponent.tsx and src/client/app/containers/maps/MapsDetailContainer.tsx. The code for sorting maps displayed is in src/client/app/containers/MapChartSelectContainer.ts and can serve as an example of what to do. The sorting of the meter dropdown menu in src/client/app/containers/MeterDropdownContainer.ts is another example.

huss:
PR #773 sorts the cards by meter identifier and completes this task.

PR #728, opened 2022-03-08
Sorted meters by identifier issue[#577]

ecluster:
# Description
Meters are now sorted by identifier rather than id/name. Added an algorithm that maps ownProp id to sorted indices. Then the sorted meters are retrieved and passed in as a single meter. 

Fixes #[577]

## Type of change

## Checklist

## Limitations
The meter page loads in each meter individually rather than loading an array of meters. If this changes in the future and meters are loaded in differently, this code needs to be revisited and fixed appropriately. 

## Additional context
../src/client/app/containers/meters/MeterViewContainer.tsx (file that takes care of sorting)

Demo of sorted meters:
Screen Shot 2022-03-07 at 9 18 33 PM


src/client/app/components/meters/MetersDetailComponent.tsx
```diff
@@ -34,7 +34,7 @@ export default class MetersDetailComponent extends React.Component<MetersDetailP
 
 	public render() {
 		const loggedInAsAdmin = this.props.loggedInAsAdmin;
-
+
```
huss, line 37: Please remove the extra tabs at the start of the line

src/client/app/containers/meters/MeterViewContainer.tsx
```diff
@@ -10,23 +10,44 @@ import { editMeterDetails } from '../../actions/meters';
 import { MeterMetadata } from '../../types/redux/meters';
 import { logToServer } from '../../actions/logs';
 import { isRoleAdmin } from '../../utils/hasPermissions';
-
+import _ = require('lodash');
+var indexArray = 0;
```
huss: We should try to avoid global variables. I appreicate you want to have a value that stays between function calls but maybe there is another way?

There should be a blank line before the function.

src/client/app/containers/meters/MeterViewContainer.tsx
```diff
@@ -10,23 +10,44 @@ import { editMeterDetails } from '../../actions/meters';
...
 function mapStateToProps(state: State, ownProps: { id: number }) {
-	let meter = JSON.parse(JSON.stringify(state.meters.byMeterID[ownProps.id]));
+
```
huss: No blank line at stat of a function.

src/client/app/containers/meters/MeterViewContainer.tsx
```diff
@@ -10,23 +10,44 @@ import { editMeterDetails } from '../../actions/meters';
...
+	const meters = state.meters.byMeterID;
+	let sortedMeters = _.sortBy(_.values(meters).map(meter =>
```
huss: meter should have a type

src/client/app/containers/meters/MeterViewContainer.tsx
```diff
@@ -10,23 +10,44 @@ import { editMeterDetails } from '../../actions/meters';
...
+	let sortedMeters = _.sortBy(_.values(meters).map(meter => 
+	({ 
+		identifier: meter.identifier,
+		id: meter.id,
+		name: meter.name,
+		displayable: meter.displayable,
+		enabled: meter.enabled
+
+	})), 'identifier');
+	// Algorithm developed to keep sorted Array of meters's index in line with the current passed prop id.
```
huss: It is good there is a comment here but it would be better if it told how it works.

src/client/app/containers/meters/MeterViewContainer.tsx
```diff
@@ -10,23 +10,44 @@ import { editMeterDetails } from '../../actions/meters';
...
-		meter = JSON.parse(JSON.stringify(state.meters.editedMeters[ownProps.id]));
+		meter = JSON.parse(JSON.stringify(state.meters.byMeterID[i]));
 	}
-
 	const currentUser = state.currentUser.profile;
 	let loggedInAsAdmin = false;
 	if (currentUser !== null) {
 		loggedInAsAdmin = isRoleAdmin(currentUser.role);
 	}
+	// Keeps track of sortedMeter index. Since page is loaded multiple times resets when count hits the limit.
```
huss, line 41: Can you please explain how this works? I think this is called for each meter so when it has been called that many times it resets back to zero to start the process again.

src/client/app/types/redux/meters.ts
```diff
@@ -55,10 +55,19 @@ export interface MeterMetadata {
 	gps?: GPSPoint;
 }
 
+export interface MeterSortedByIdentifier {
```
huss, line 58: I'm unclear on where this is used.

src/client/app/types/redux/meters.ts
```diff
@@ -55,10 +55,19 @@ export interface MeterMetadata {
...
+	identifier: string;
+	name: string;
+}
+
 export interface MeterMetadataByID {
 	[meterID: number]: MeterMetadata;
 }
 
+export interface MeterMetadataByIdentifier {
```
huss: Is this used outside this file? If not, does it need the export?

src/client/app/containers/meters/MeterViewContainer.tsx
```diff
@@ -10,23 +10,44 @@ import { editMeterDetails } from '../../actions/meters';
...
+	// Needs to be visited when meter page is redesigned. 
+	if (indexArray != sortedMeters.length - 1){
```
huss: formatting off, e.g., "){"

huss:
First, thanks to @ecluster and @Joseph-M-Newman for doing this work and submitting their first PR! I've left some comments on specific lines of code. Here are some overall comments:

I have run into several places where the code is not doing what I would expect:

- If you are an admin on the meters page and click the "Enable", "Disable", "Show" or "Hide" buttons then the value displayed does not change. If you select a time zone then it is not displayed. After you  click "save meter updates" they do change.
- If you edit the meter identifier and save, its location on the list does not change. If you refresh the page it does move.
- If you log out and go to the meter page then meters the user should not see are displayed. Refresh fixes it.
- I believe that I have seen edits to the meters, then logging out do not show on the non-admin version of the page. Refreshing seems to fix. Once the page went white (nothing displayed) and only came back on refresh.

I'm not sure what is up but suspect that the state is not properly updated and that the mapping you have to sorted meters has an issue. It may be possible to fix this but I am concerned there may be a deeper issue here. I'm happy to discuss more.

Other general comments:

- It would be nice to have more comments. I know the current code does not but we are trying to be better on new code.
- The CI failed. Please see the issues indicated.

huss:
Can you please merge in development so you are up to date. You want to be sure it works with new versions of the software and changes to package versions (esp. react-redux).

huss (changes requested):

huss:
I finally had some time to look at this - sorry it took a little while (as we previously discussed). I believe I have a fix and it works on my branch. I changed src/client/app/containers/meters/MetersDetailContainer.tsx so the mapStateToProps modifies what it returns for meters. The complete return is:

```
	return {
		loggedInAsAdmin,
		// The state has all the meter info so map to get the id and identifier then sort that by
		// identifier and finally map so only have id as expected. This causes
		// MetersDetailComponent.tsx to process them in this sorted order
		// and pass them on to MeterViewContainer as desired.
		meters: _.sortBy(_.values(state.meters.byMeterID)
			.map(meter => ({ value: meter.id, label: meter.identifier.trim() })), 'label')
			.map(meter => meter.value),
		unsavedChanges: Object.keys(state.meters.editedMeters).length > 0
	};
```

where the comment explains what was done. Given you are actively working to modify the meter page with a complete revamp, I propose that you try to incorporate this into that work to see if it works. If so, we can close this PR and the fix will be in the PR from that effort. Please let me know if you have any questions or issues.

huss:
This was rolled into PR #773 so this PR is being closed.

PR #729, opened 2022-03-14
Conversion functions

truongdd03:
# Description

Fixes implement-conversion-functions

This PR adds necessary functions to create the Cik and Pik function. It also fixes a small bug when creating bidirectional edges in the conversion graph. A new function getTypeSuffix in the Unit model is also created.

## Type of change

## Checklist

## Limitations


src/server/services/graph/pathConversion.js
```diff
@@ -0,0 +1,105 @@
...
+	// Invert this equation to give:
+	// source_value = (1/slope) * destination_value - (intercept / slope)
+	const convertedSlope = 1.0 / slope;
+	const convertedIntercept = -(intercept / slope);
+	return [convertedSlope, convertedIntercept];
+}
+
+/**
+ * Returns the updated overall conversion given a new conversion to add at the end.
+ * @param {*} origSlope The current coversion's slope.
```
huss: coversion -> conversion

src/server/services/graph/pathConversion.js
```diff
@@ -0,0 +1,105 @@
...
+ * @param {*} origIntercept The current conversion's intercept.
+ * @param {*} newSlope The conversion's slope to add.
+ * @param {*} newIntercept The conversion's intercept to add.
+ * @returns 
+ */
+function updatedCoversion(origSlope, origIntercept, newSlope, newIntercept) {
```
huss: Coversion -> Conversion

src/server/services/graph/pathConversion.js
```diff
@@ -0,0 +1,105 @@
...
+	// is processed we get its conversion.
+	let slope = 1;
+	let intercept = 0;
+	let suffix = '';
+	for (let i = 0; i < path.length - 1; ++i) {
+		const sourceId = path[i].id;
+		const destinationId = path[i + 1].id;
+		const [newSlope, newIntercept, newSuffix] = await conversionValues(sourceId, destinationId, conn);
+		// Updates the path conversion for this new edge.
+		[slope, intercept] = updatedCoversion(slope, intercept, newSlope, newIntercept);
```
huss: Coversion -> Conversion

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -0,0 +1,119 @@
...
+	}
+}
+
+/**
+ * Adds new suffix units and conversions to the database and the conversion graph.
+ * @param {*} graph The conversion graph. 
+ * @param {*} conn The connection to use.
+ */
+async function handleSuffixUnits(graph, conn) {
+	const suffixUnits = await Unit.getSuffix(conn);
```
huss, line 94: The pseudocode in the design document had some more comment. I think it is worth having some of them in the actual code. I understand that it all seems clear right now but in the future it will help someone. I'm fine if you think editing them some would help with what they mean.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -0,0 +1,119 @@
...
+		unit.displayable = Unit.displayableType.NONE;
+		await unit.update(conn);
+	}
+
+	for (const p of paths) {
+		const secondUnit = await Unit.getById(p[1].id, conn);
+		// The paths to suffix units shouldn't be deleted.
+		if (secondUnit.unitType !== Unit.unitType.SUFFIX) {
+			// Removes the conversion from the conversion graph.
+			graph.removeLink(p[0].id, p[1].id);
```
huss, line 82: Is there any issue if this link does  not exist in the graph?

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -0,0 +1,119 @@
...
+			// Removes the conversion from the database.
```
huss: The design document did not call for removing the conversion. Could you explain to me the reasoning. It may be fine but I want to be sure.

I think the reasoning was that it stays in the database but the conversion cannot be used since it is no longer in the graph. The unusual case being considered is if the suffix unit goes to another unit and the admin adds a new conversion from that unit to another unit. Since the suffix remains, the new path will be added to the new unit. Also, this is why the code deals with the case of the unit already existing since it stayed in the database.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -0,0 +1,119 @@
...
+ * @param {*} conn The connection to use.
+ */
+async function hideSuffixUnit(unit, paths, graph, conn) {
+	// Hides the suffix unit since we added the units based on it if not previously done.
+	if (unit.displayable !== Unit.displayableType.NONE) {
+		unit.displayable = Unit.displayableType.NONE;
+		await unit.update(conn);
+	}
+
+	for (const p of paths) {
```
huss, line 77: Could you add back some of the explanation in the design document for what this is doing and why?

src/server/services/graph/createConversionArrays.js
```diff
@@ -0,0 +1,104 @@
...
+ * For source units, the index is the row id.
+ * For destination units, the index is the column id.
+ * @param {*} units The list of units.
+ * @param {*} conn The connection to use.
+ */
+async function assignIndex(units, conn) {
+	let id = 0;
+	// The old indices may cause conflicts while we are assigning new ones.
+	// Therefore, we need to reset them to null first.
+	for (let unit of units) {
```
huss, line 21: The design always had an issue that while Cik was being recreated any requests might return incorrect values. This was true until the DB settled and the new Cik was transferred to a client. Since this was rare we let it go. The concern I have here is that setting the index to null will cause an error if a client happens to use this value. I'm not sure that is worse than a wrong value but it might cause problems in the code. We need to think about what to do in that code. Maybe catch the exception, wait a second, and reload Cik. I don't think this is an issue to fix in this code but noting for future code to be added. Maybe a TODO comment here is desirable so we remember this.

src/server/services/graph/createConversionArrays.js
```diff
@@ -0,0 +1,104 @@
...
+				// All suffix units were dealt in src/server/services/graph/handleSuffixUnits.js
+				// so all units with suffix have displayable of none.
+				// This means this path has a suffix of "" (empty) so it does not matter.
+				// The name of any unit associated with a suffix was already set correctly.
+				// Thus, we can just use the destination identifier as the unit name.
+				c[source.unitIndex][destination.unitIndex] = [slope, intercept, destination.identifier];
+			}
+		}
+	}
+	// The table in the database for the logical Cik needs to be wiped and these values stored.
```
huss: I would put TODO at the start of this comment so we know to deal with it.

src/server/services/graph/createConversionArrays.js
```diff
@@ -0,0 +1,104 @@
...
+	// This code will be added once the database table for using it to get readings is set.
+	// At the moment, we just return the array.
+	return c;
+}
+
+/**
+ * Returns the Pik array.
```
huss: Maybe: Returns the Pik array which is true if there is a conversion in Cik.

src/server/services/graph/createConversionArrays.js
```diff
@@ -0,0 +1,104 @@
...
+	// Assigns new indices.
+	for (let unit of units) {
+		unit.unitIndex = id;
+		id += 1;
+		await unit.update(conn);
+	}
+}
+
+/**
+ * Returns the Cik array.
```
huss: Maybe: Returns the Cik which gives the slope, intercept and suffix name between each meter and unit where it is NaN, Nan, '' if no conversion.

The goal is for the JSDoc to tell what is going on without looking at code.

huss (changes requested):
As always, @truongdd03 has done nice work. There are a few items to address (moslty commenting) but overall this looks very good.

I note that @truongdd03 manually tested the result of Cik and it was correct based on the sample values in the design document. Hopefully this means the design was correct and the code is also correct. At some point these will become automated tests.

src/server/services/graph/handleSuffixUnits.js
truongdd03: The graph package will do nothing if the link doesn't exist. So I think it's fine.

src/server/services/graph/handleSuffixUnits.js
truongdd03: If I'm correct, we rebuild the conversion graph whenever OED starts. If not, I agree that we don't need to remove the conversion from the DB.

Take the sample values in the document to be the example. If we run the handleSuffixUnits function for the first time, "metric ton of CO2" and "kg of CO2" will be added to the DB. Then, we restart OED and rebuild the graph. However, note that the new graph will contain the conversion from "kg CO2" to "kg" (since this conversion is still in the DB). Even if we run handleSuffixUnits one more time, this conversion will not be removed (since all the suffix units related to "kg CO2" have been added so the handleSuffixUnits function will do nothing). This conversion will lead to the wrong Cik table.

src/server/services/graph/createConversionArrays.js
truongdd03: During the process of assigning new values, the old ids and new ids may be similar. This will cause an error since we have a constraint of unique unit index in the DB. So I think it's best to set everything to null. I agree with you that we should be careful during the recreating Cik array process. I've added a TODO comment about this.

src/server/services/graph/handleSuffixUnits.js
huss: The conversion graph will be rebuilt when OED starts and each time a conversion is modified/added.

I think I may be missing something because it seems to me the graph would remove the "kg CO2" to "kg" edge. Line 126 does:
await hideSuffixUnit(unit, paths, graph, conn);
for each suffix unit. Since "kg CO2" is still a suffix unit in the unit table (if you leave instead of removing), it should call this function with all paths from this suffix unit. I think there will again be three paths (to "kg", "kg of CO2" and "Metric ton of CO2"). The last two will be suffix and left alone. The first from "kg CO2" to "kg" will again be removed since it is not a suffix unit. Is this wrong?

Now for the other case I'm worried about. Suppose an admin adds the unit "lb" and a conversion from "kg" to "lb". If this had been done originally then OED would add a vertex in the graph from "kg CO2" to "lb" so we want this added now. The only way I see that happening is if there is a link from "kg CO2" to "kg" so it can then follow it to "lb". The path from "kg CO2" to "kg" already exists in the conversions so won't be added again. The new path from "kg CO2" to "lb" will be added to the conversions and also added to Cij. This is why I had the "kg CO2" to "kg" conversion staying in the conversion table but not being displayed.

Finally, removing a unit might confuse an admin as it was there before. Hiding it is easier to understand (at least from my perspective). This isn't absolutely necessary but nice.

I welcome your thoughts and thank you for helping think about these complex cases.

truongdd03: > I think I may be missing something because it seems to me the graph would remove the "kg CO2" to "kg" edge. Line 126 does:
await hideSuffixUnit(unit, paths, graph, conn);
for each suffix unit. Since "kg CO2" is still a suffix unit in the unit table (if you leave instead of removing), it should call this function with all paths from this suffix unit. I think there will again be three paths (to "kg", "kg of CO2" and "Metric ton of CO2"). The last two will be suffix and left alone. The first from "kg CO2" to "kg" will again be removed since it is not a suffix unit. Is this wrong?


I've changed the get_suffix.sql a little bit so it will not return units with displayable of type NONE. So when we run the second time, the suffixUnits in line 89 will not contain "kg CO2" hence the graph still has the conversion from "kg CO2" to "kg".

The reason why I changed the get_suffix.sql is that it causes a problem if we run handleSuffixUnits for "kg CO2" one more time. Since we already had conversions from "kg CO2" to "kg of CO2" and "Metric ton of CO2" from the first time, it will create new suffix units "kg of CO2 of CO2" and "Metric ton of CO2 of CO2". After reading your comment, I think changing get_suffix.sql is not a good idea.

Alternatively, I plan to check if the destination unit in line 103 is a suffix unit. I think that we only need to run addNewUnitAndConversion if the destination is not a suffix unit. So this way will prevent creating units such as "kg of CO2 of CO2" and the unnecessary conversions will be removed as well. Please let me know what do you think about this. Thank you.

huss: I get it now. The pseudocode in the design document checked to avoid recreating the new suffix unit again from the same path. However, it misses that the new path gives the unit a new name. Thanks for catching this and thinking about it.

I have thought about your proposed solution and I think it is fine. If a suffix unit is the final destination of a path then there are no units it converts into. If such a vertex existed in the graph originally then no new unit/conversion would be created by the suffix process. I cannot think of a use case at this time to have such an item in the graph where it is placed by an admin. Thus, ignoring them as proposed seems fine. I say do that and don't ignore internal suffix units as the current code does.

truongdd03:
Thank you for your review, @huss. I have resolved all the issues and updated the sources and destinations in src/server/services/graph/createConversionArrays.js as we have discussed. Please let me know if there is something else that needs to be changed. Thank you.

huss (approved):
Everything seems good. Thanks to @truongdd03 for quickly implementing the final changes. This is ready to merge.

PR #733, opened 2022-03-19
Easier way to find all meters in a group

### LINKED ISSUE #591: easier way to find all meters in a group
The group page allows people to see the meters and groups that were added to when the group was created. However, a typical user may want to know all the unique meters that are part of a group to see what is being included. If a group contains groups then you have to recursively check out each group to figure this out and then remove duplicates. It seems a good idea to list them for the user. This does not need to be fancy; a text list on a line would be fine.

1. The first step to see how to get this information from the DB. OED likely has this ability already. @wadoodalam is looking into this.
2. Need to add the info to the group page.

Onyi21:
# Description

All meters in a group/subgroup are displayed on a single line recursively.

Fixes #591

## Type of change

## Checklist

## Limitations

At this time, there is no work remaining to be done on this issue.


package-lock.json
```diff
@@ -124,9 +124,9 @@
       }
     },
     "node_modules/@babel/cli": {
-      "version": "7.16.7",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.16.7.tgz",
-      "integrity": "sha512-0iBF+G2Qml0y3mY5dirolyToLSR88a/KB6F2Gm8J/lOnyL8wbEOHak0DHF8gjc9XZGgTDGv/jYXNiapvsYyHTA==",
+      "version": "7.16.8",
```
huss: The commit message says "npm fix" but several say this. (I think you mean a fix for issues found in the checks in the others but am uncertain here.) I'm unclear on why package-lock.json was changed, esp. given no changes to package.json. Can you please explain why and if this is needed. Note this would mean that your PR changes node modules.

src/client/app/components/groups/GroupViewComponent.tsx
```diff
@@ -68,6 +77,15 @@ export default class GroupViewComponent extends React.Component<GroupViewProps,
 						<FormattedMessage id='edit.a.group' />
 					</Button>
 				</Link>
+					<div>
+						<p style={groupAllMeters}>
+						<FormattedMessage id='group.all.meters' />:
+						</p>
+							{this.props.deepMeterNames.map ((item, index) => (
+								<span key={`d_${index}`} style={deepMeters}>{(index ? ', ': '') + item}</span>
```
huss: If there are a lot of child meters then the second line onward start with ", ". It would be good if the comma came after the previous item so in this case it was on the preceding line.

src/client/app/components/groups/GroupViewComponent.tsx
```diff
@@ -46,6 +47,14 @@ export default class GroupViewComponent extends React.Component<GroupViewProps,
 			display: renderEditGroupButton ? 'inline' : 'none',
 			paddingLeft: '5px'
 		};
+		const groupAllMeters: React.CSSProperties = {
```
huss, line 50: I don't know for sure if this is where it is happening and this is a small point. Would it be possible to move the "All Meters:" label slightly downward so it is a little farther from the "Edit a group" button? Maybe around the vertical spacing between the "All Meters:" label and the list of meters below that.

huss (changes requested):
Thanks to @AndrewMei99 and @Onyi21 for doing this work. Overall, it works well where I made a few comments to look at before it is approved.

package-lock.json
Onyi21: We didn't change anything in the package.json file. Not sure if it's needed

huss: package-lock.json is showing a fair number of changes. I looked at the current version of the development branch and it seems that these are real changes from the current code. Thus, if you didn't intend to change the packages then this change should be removed so the code can be tested with those versions of the packages. This also has to happen to allow merging of the PR since it can only have desired changes. If you need help then let me know.

src/client/app/components/groups/GroupViewComponent.tsx
huss: This was not done but can be fixed at a future time.

huss (approved):

huss:
Changes look good and addressed the needed items. Testing indicates that it works properly.

PR #734, opened 2022-03-20
Update react to v17

### LINKED ISSUE #662: update React (and Redux)

jina2k:
# Description

Fixes #662 , Partially Addresses #548
Previous package update already allowed us to use useSelector. This package update doesn't provide much new functionality except for updating react-select from v1 to v5. 
I followed the migration guides linked below:
https://react-select.com/upgrade-to-v2
https://react-select.com/upgrade

react-select now takes in options or array of options only, so we use TimeZoneOption instead of string. React-select no longer has a separate types library, and css shouldn't be imported from node_modules. Change was somewhat required as react-select v1 required reactv16 and lower.

Packages installed are:
react@17.0.2
react-dom@17.0.2
react-dropzone@12.0.4
react-intl@5.22.0
react-notification-system@0.4.0
react-tooltip@4.2.20
reactstrap@8.10.0
webpack@5.67.0
@types/react@17.0.35
@types/react-dom@17.0.10
@types/react-notification-system@0.2.41
react-select@5.2.0

## Type of change

## Checklist

## Limitations

I haven't updated react-router and react-router-dom.
Currently, this would require removing history which affects multiple files. The migration guide also indicates that react-router v6 is not backwards compatible yet with v5. In this case, it would be preferred to wait until v6 is backwards compatible for seamless upgrading.

I would also note that react-notification-system is pretty old, but it's still usable so there's no issue. We could transition into using a different notification system if possible, as the peer-deps for this require react v16 and lower. 
ie; https://www.npmjs.com/package/react-notifications, https://github.com/fkhadra/react-toastify


src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -14,6 +14,7 @@ import {
 	UpdateDefaultWarningFileSize,
 	UpdateDefaultFileSizeLimit
 } from '../../types/redux/admin';
+import { TimeZoneOption } from '../../types/timezone';
```
huss: I'm not sure where in the code this is happening but the current choice of timezone is not showing as the selected one on the dropdown menu. It used to do this. This is happening in preferences and on the meter page.

src/client/app/components/meters/MeterViewComponent.tsx
```diff
@@ -11,6 +11,7 @@ import TimeZoneSelect from '../TimeZoneSelect';
 import { updateUnsavedChanges } from '../../actions/unsavedWarning';
 import { fetchMetersDetails, submitEditedMeters, confirmEditedMeters } from '../../actions/meters';
 import store from '../../index';
+import { TimeZoneOption } from '../../types/timezone';
```
huss: Not sure where in the code this is happening but when you change a meter timezone and save it goes from "edited" to "submitting" but then does not change to nothing on the page. The change is not submitted to the database. Also see issue noted in preferences on current value not displayed.

huss (changes requested):
Thanks to @jina2k for continued work on packages. Overall, this looks good. Testing showed a couple of items to look at.

src/client/app/components/admin/PreferencesComponent.tsx
jina2k: isolated and fixed issue to draw only string using filter

src/client/app/components/meters/MeterViewComponent.tsx
jina2k: see above, also fixed.

huss:
Thanks. This works well. I'm only having one issue. For timezones that existed before the update, the menu does not show them. Once you set to a new value then it works fine. From looking at the database, I think the issue might be that the name being saved is different. Previously, I had a meter that shows "Asia/Ashgabat" as the default_timezone_meter and it does not show up in the dropdown menu. I changed a different meter to be this timezone and it has "Asia/Ashgabat (+05) PT5H" and it works fine. Do you think there is an easy fix or do we need to migrate old timezone names to new ones as part of this change?

jina2k:
> Thanks. This works well. I'm only having one issue. For timezones that existed before the update, the menu does not show them. Once you set to a new value then it works fine. From looking at the database, I think the issue might be that the name being saved is different. Previously, I had a meter that shows "Asia/Ashgabat" as the default_timezone_meter and it does not show up in the dropdown menu. I changed a different meter to be this timezone and it has "Asia/Ashgabat (+05) PT5H" and it works fine. Do you think there is an easy fix or do we need to migrate old timezone names to new ones as part of this change?

I assumed from development that the timezones were using the label which contained everything rather than just the name. Upon looking in the database it seems they use value instead, so I've changed it to match the value.

huss (approved):
It works great. Thanks for your attention to all the details. It is ready to merge.

PR #696, opened 2022-01-17
Updating packages #2

jina2k:
# Description

Partially Addresses #548 ,
#662 and #463 

npm audit fix to resolve new commits
Packages changed:
### babel v6 to v7.
.babelrc -> babel.config.json with new params due to new @babel/core

babel params in package.json as well

babel-plugin-transform-object-rest-spread -> @babel/plugin-proposal-object-rest-spread
babel-preset-react -> @babel/preset-react
babel-plugin-react-intl -> babel-plugin-formatjs
babel-preset-env -> @babel/preset-env
react-dropzone update
attr-accept
source-map-resolve
@babel/cli
@babel/core
babel-core bridge

### Transition to ts-loader from awesome-typescript-loader

https://github.com/s-panferov/awesome-typescript-loader
as said in documentation here, awesome-typescript-loader aims to support only typescript@2.x and webpack@2.x.
As such, we will use the recommended ts-loader instead, even though there will be performance degradation.
This is mentioned in https://webpack.js.org/guides/typescript/

I've also fixed some glaring issues revolving around syntax errors that somehow made it to the build (only made obvious after transition from commonjs to es6), files listed are: app/types/redux/actions.ts, app/actions/admin.ts, app/actions/groups.ts

### React-intl v2 to v5

react-intlv2->react-intlv5.21.0
@formatjs/intl

## Type of change

- [x] Note merging this changes the node modules
- [ ] Note merging this changes the database configuration.
- [x] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [x] I have removed text in ( ) from the issue request

## Limitations
Cannot upgrade to typescript v4.5.2 without changing the script test case as well as declaring types in multiple areas of the code. https://github.com/formatjs/formatjs/issues/3276
I attempted to use better-typescript-lib but found it easier to revert to 4.4.4 as this is not a new project.

PR #736, opened 2022-03-27
TSLint to ESLint

### LINKED ISSUE #463: migrating to ESLint
We currently use tslint and also a separate typescript check. tslint is now deprecated and a new ESLint is available. ESLint can check both TypeScript and JavaScript. We should transition to the new tool, verify it does what we want and see about additional testing it provides. See <https://github.com/palantir/tslint/issues/4534> for a discussion of this transition.

jina2k:
# Description

Fixes #463 , Partially Addresses #548

## Type of change

## Checklist

## Limitations

Note that previously we ignored files in src/server which contained commonjs syntax (with TSLint). Here we do the same, though later in the future it might be suggested to somehow allow for these files. (which might also involve renaming all of these files from .js -> .cjs)  This would also require some other rules as I previously had issues in pr #696 . ie; no-var-requires, no-undef, etc.


huss (approved):
Many thanks to @jina2k for making this work including all the new linting issues that were uncovered by the change to ESLint and further requests. Testing indicates it works fine and is ready to merge.

PR #739, opened 2022-03-31
Determining compatible units and transfer Pik to the client-side

truongdd03:
# Description

Fixes determine-compatible-units and transfer-pik-on-client-arrival

This PR transfers Pik, all the units, and unitID and defaultGraphicUnit of meters to the client-side. The functions to determine compatible units were also added.

Note 1: the function metersInGroup in src/client/app/utils/determineCompatibleUnits.ts returns a promise. This is different from the signature function that I created at first.

Note 2: this PR also includes changes from the development branch. The list of files directly relating to this PR is below:

- src/client/app/actions/units.ts
- src/client/app/components/InitializationComponent.tsx
- src/client/app/containers/InitializationContainer.ts
- src/client/app/reducers/index.ts
- src/client/app/reducers/units.ts
- src/client/app/types/conversionArray.ts
- src/client/app/types/redux/actions.ts
- src/client/app/types/redux/meters.ts
- src/client/app/types/redux/state.ts
- src/client/app/types/redux/units.ts
- src/client/app/utils/api/ConversionArrayApi.ts
- src/client/app/utils/api/index.ts
- src/client/app/utils/api/unitsApi.ts
- src/client/app/utils/determineCompatibleUnits.ts
- src/server/app.js
- src/server/data/automatedTestingData.js
- src/server/routes/conversionArray.js
- src/server/routes/meters.js
- src/server/routes/units.js
- src/server/services/createDB.js

## Type of change

## Checklist


src/client/app/utils/determineCompatibleUnits.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @param meters The set of meter ids.
+ * @returns
+ */
+export function unitsCompatibleWithMeters(meters: Set<number>): Set<number> {
+	const state = store.getState();
+	// The first meter processed is different since intersection with empty set is empty.
+	let first = true;
+	// Holds current set of compatible units.
+	let compatibleUnits = new Set<number>();
+	// Loops over all emters.
```
huss: emters -> meters

src/client/app/utils/determineCompatibleUnits.ts
```diff
@@ -0,0 +1,146 @@
...
+import { fetchGroupChildrenIfNeeded } from '../actions/groups'
+import { GroupDefinition } from 'types/redux/groups';
+
+/**
+ * The intersect operation of two sets.
+ * @param setA
+ * @param setB
+ * @returns
+ */
+function intersect(setA: Set<number>, setB: Set<number>): Set<number> {
```
huss: I think the ability to intersect two sets will be needed in other places (such as graphic unit menu). Could we rename this setIntersect and export it?

src/client/app/utils/determineCompatibleUnits.ts
```diff
@@ -0,0 +1,146 @@
...
+export function unitsCompatibleWithMeters(meters: Set<number>): Set<number> {
+	const state = store.getState();
+	// The first meter processed is different since intersection with empty set is empty.
+	let first = true;
+	// Holds current set of compatible units.
+	let compatibleUnits = new Set<number>();
+	// Loops over all emters.
+	meters.forEach(function (meterId: number) {
+		// Gets the meter associated with the meterId.
+		const meter = _.find(state.meters.byMeterID, function (o: MeterMetadata) { return o.id == meterId; }) as MeterMetadata;
```
huss: See below on comment about how get a matching meter.

src/client/app/utils/determineCompatibleUnits.ts
```diff
@@ -0,0 +1,146 @@
...
+import * as _ from 'lodash';
+import { MeterMetadata } from '../types/redux/meters';
+import { ConversionArray } from '../types/conversionArray';
+import { UnitData, UnitType } from '../types/redux/units';
+import { fetchGroupChildrenIfNeeded } from '../actions/groups'
+import { GroupDefinition } from 'types/redux/groups';
+
+/**
+ * The intersect operation of two sets.
+ * @param setA
```
huss: Could you please fill in the param and return here and in other functions.

src/client/app/utils/determineCompatibleUnits.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @param groupId The groupId.
+ * @returns
+ */
+export async function metersInGroup(groupId: number): Promise<Set<number>> {
+	// Fetch group children if needed.
+	await store.dispatch<any>(fetchGroupChildrenIfNeeded(groupId));
+
+	const state = store.getState();
+	// Gets the group associated with groupId.
+	const group = _.find(state.groups.byGroupID, function (o: UnitData) {
```
huss: In other parts of OED it does it as  
state.meters.byMeterID[meterID].name  
when it needs to get a meter by its id. Could you do something similar here?

huss (changes requested):
Thanks to @truongdd03 for completing this. I could only do limited testing (looked at state, pik) but it seems fine. The real test of the new functions will be when they are used for the new menus. Overall, this looks good. I just made a couple of small comments to look at.

src/client/app/utils/determineCompatibleUnits.ts
truongdd03: I tried to do so but got the following error:

```Element implicitly has an 'any' type because expression of type 'number' can't be used to index type 'Dictionary<NamedIDItem> | MeterMetadataByID'.```

The state that we use in other places has the type of State, while it's an object here. This may be the reason. I have done some google searches but still couldn't find a way to make it work. Please let me know if you have a way to work around this.

Alternatively, I tried: `_.get(state.meters.byMeterID, meterId)` and it works properly. Please let me know what you think about this.

huss: Okay. There may be a way to modify the type definition but I don't feel it is important enough to do it. Your alternative seems a little more straightforward so why not use that (but the original way is okay too).

truongdd03:
Thank you for your review, @huss. I've resolved all the comments. Please let me know if you think something else needs to be changed. Thank you.

huss (approved):
Everything looks good. Thanks to truongdd03 for quickly finishing this. It is ready to merge.

PR #742, opened 2022-04-04
Hide meters/groups that aren't displayable for groups page

### LINKED ISSUE #414: Allow hiding groups
It's kind of strange to allow hiding meters from displaying on the website, but not groups. This could probably be added to the large blank space on the right side of the groups page.

### LINKED ISSUE #699: blank page/app hangs if try to view a group that has underlying meter displayable of false if not admin
### Describe the bug

If any meter in a group has displayable as false then a non-admin clicking on it on the groups pages causes the page to go blank. The only way out appears to be to reload OED in the web browser.

### To Reproduce

1. Set a meter to have displayable as false.
2. Create a group (if it does not yet exist) that contains this meter.
3. Log out as admin.
4. Visit the groups page and click on this group. The page should go blank.

### Expected behavior

OED allows non-admins to graph groups that contain meters that are not displayable. Given this, OED should show the underlying meters of a group whether they are displayable or not. It is assumed that this will eliminate the blank page issue.

The changes planned for resource generalization will include honoring the recently added displayable field for groups. As such, an admin can choose to hide such a group. The admin group help page should be changed to discuss users seeing all the meters and also that a user can likely determine the underlying non-displayable meter readings by graphing this group and all the other meter in the group.

### Additional context

None.

jina2k:
# Description

Fixes #699, Fixes #414 

Hides meters/groups that aren't displayable to avoid blank page error. Uses the returned meter/group names in mapStateToProps to distinguish hidden vs displayable meters/groups. All groups/meters must have at least 1 character in their
name to be created, so I used an empty string for those that were hidden. 
I then noticed groups were not hidden even when displayable was set to false, and saw issue #414 which addresses this. Changes were then made to allow for this functionality, where it follows meters being displayable or not. In general, this change is minimal and does not hinder the resource generalization work, as it doesn't touch the front-end code for displaying. 

## Type of change

## Checklist

## Limitations
n/a

### DIFF
--- a/src/client/app/components/ListDisplayComponent.tsx
+++ b/src/client/app/components/ListDisplayComponent.tsx
@@ -33,10 +33,23 @@ export default function ListDisplayComponent(props: ListDisplayProps) {
 		paddingRight: '12px'
 	};
 
+	const hiddenStyle: React.CSSProperties = {
+		opacity: 0.5
+	};
+
 	return (
 		<div className='list-wrapper' style={listWrapperStyle} >
 			<ul id='meterList' style={listStyle} >
-				{props.items.map((item: any) => <li key={item.toString()}>{item.toString()}</li>)}
+				{props.items.map((item: any, index: number) => {
+					// If the item, ie; meter/group had a name of length 0, this meant it was a hidden name.
+					if (item.toString().length !== 0) {
+						return <li key={item.toString()}>{item.toString()}</li>
+					}
+					else {
+						// Each list item needs a unique key, so we use index to differentiate hidden elements.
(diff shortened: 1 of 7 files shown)


jina2k:
Note that if need be, hiding groups can be commented out relatively easily, and implemented at a later time where we would just return all groups regardless of if it were hidden or not. 

Another method of getting hidden groups/meters is passing in an extra parameter to props to indicate the true size of meters/groups in the group. This would involve eliminating the hidden meter/group "names" and comparing the size of meter/group names with the size, and then modifying the list to have a single hidden field instead of multiple hidden fields.

huss (approved):
This all looks good and works well. Thanks to @jina2k for fixing both issues and covering all the cases. This is ready to merge.

PR #744, opened 2022-04-05
Checking for valid group names

### LINKED ISSUE #416: Can't backspace the first character of the group name
When editing a group's name, you cannot backspace the first character or in any way fully empty the text field. This is really bizarre from a user's perspective. Instead, it should let you empty the field and then just give an error like "Group must have a name" when they try to save it. (was #411)

### LINKED ISSUE #496: space or emtpy Group name
If you Create a Group on the Group page you can have two issues. First, if you leave the name blank then a read popup shows up but it is empty (the group is not created). Second, if you just put in a space then a group is created that is on the list but the line appears empty. We should check that a name is not just spaces/blank and not allow but warn the admin with a good popup.

jina2k:
# Description

Fixes #496 , Fixes #416 

Fixes both of these issues, one through a window alert popup, and the other through modifying the onChange function.
I also reimplemented #742 using what I had described previously, so that the hidden field would show up at the bottom instead of top (due to sorting). This also allows for displaying multiple hidden fields if need be in the future, though it would probably be better to "mask" the total number of hidden meters.

## Type of change

## Checklist

## Limitations
n/a


huss (approved):
Good work, no issues found and testing indicates it work as expected. Ready to merge.

PR #745, opened 2022-04-06
Fix #723 by removing duplicate meter ids when generating compressed readings

### LINKED ISSUE #723: Group line value wrong if another group contains that group
### Describe the bug

If you line graph a single group then you get the correct values. If you add a group to the line graphic that contains the first group then the values are wrong. It seems to happen whether or not other meters and groups are in second group but this was not heavily tested. This may be part of some larger bug/issue but that is unknown. How long this is been happening is unknown but an issue was seen a while back so it probably isn't knew but is now well defined.

### To Reproduce

Using the developer test data, create the group s2c2 that contains testSqSin and testSqCos. Create a group (here named issue) that only has s2c2 included. Now create a third group (here named issue2) that also only contains s2c2. Now do a line graphic of testSqCos, testSqSin and s2c2. You should get the correct graph of daily line points where s2c2 always equal 2.5 as seen in this image:

Screen Shot 2022-02-25 at 12 08 21 PM

Now add group issue to line graphic and the values for both s2c2 and issue will increase to 5

Screen Shot 2022-02-25 at 12 08 47 PM

If you now add issue2 the values for all groups will go to 7.5. Thus, it seems that the values are being added as the groups are being added.

Screen Shot 2022-02-25 at 12 09 46 PM

When the groups are created (before or after start graphing) does not seem to matter.

Note I looked at the redux state and the points are wrong there so it seems to be an issue in getting the data, probably from the database.

### Expected behavior

Groups should not impact each other unless the members of the included group are changed. The values should stay at 2.5 for all these groups.

### Additional context

None

rkalnins:
# Description

Fix #723 by removing duplicate meter ids when generating compressed readings. `array_agg` introduces duplicate meter ids in the `meter_ids` array. `compressed_readings_2` returns summed values when duplicate meter ids are provided.

Fixes #723 

## Type of change

## Checklist

## Limitations

None.

### DIFF
--- a/src/server/sql/reading/create_compressed_reading_views.sql
+++ b/src/server/sql/reading/create_compressed_reading_views.sql
@@ -248,7 +248,7 @@ AS $$
 		meter_ids INTEGER[];
 	BEGIN
 		-- First get all the meter ids that will be included in one or more groups being queried
-		SELECT array_agg(gdm.meter_id) INTO meter_ids
+		SELECT array_agg(DISTINCT gdm.meter_id) INTO meter_ids
 		FROM groups_deep_meters gdm
 		INNER JOIN unnest(group_ids) gids(id) ON gdm.group_id = gids.id;
 


huss (approved):
This is an easy, elegant fix for the issue. Testing indicates it works correctly and solves the problem. I thank @christianjw18 and @rkalnins for effort to track down what was causing the issue to allow for this to be fixed. I believe it is ready to merge if the authors are willing to convert from a draft and they think that is okay. I also acknowledge that this is the first PR from both these authors - congratulations!

huss:
Just for the record, I thought about why this was happening with line and not bar graphics. The seemingly subtle difference between getting the meters outside the main joins makes a difference in this case. At some future date we can consider if it is possible to make the logic comparable between these. This is something best done after the resource generalization changes to these functions are complete.

huss:
I just want to note that this fix was rolled into the resourceGeneralization branch since it is doing the same thing (for both line and bar) and had the bug. The code is in new functions for units so it had to be manually added. Thus, this fixed this bug and some new ones in upcoming code.

PR #752, opened 2022-04-15
Reset to no time zone implementation

### LINKED ISSUE #566: reset to no timezone once set
For both meters and preferences, once you set the timezone you can only change it to another timezone. We should allow it to be reset to no value as it is before it is set. This will allow admins to undue one that was set but not desired.

AndrewMei99:
# Description
The user is able to reset a timezone once it's set with the option "No timezone" in the dropdown of the timezones on the meter page.

Fixes #566

## Type of change

## Checklist

## Limitations
There is no work remaining on this issue for now.


huss (changes requested):
Thanks to both of you for doing this. Overall, it looks good. A few comments:

- On the admin page, when you select "no timezone" then that shows up on the menu. It also shows up if that is what was in the database when the page is loaded. On the meter page this does not happen. I think it should work as it does on the admin page.
- The string "no timezone" should be internationalized. The basic steps are:  
Add the key timezone.none to the translation/data.js with the appropriate words as the value (for all languages). Then change  
const resetTimeZone = [{value: null, label: 'No timezone'}];
to  
const resetTimeZone = [{value: null, label: <FormattedMessage id='timezone.none' />}];  
If you change the language then it should change the messge to that language.

huss:
I saw that @Onyi21 pushed a commit several days ago. I have not reviewed again since it is failing the CI tests. I just want you to know this in case you were not aware.

huss (approved):
Thanks to @Onyi21 and @AndrewMei99 for finishing this up. It works as expected. The French translations are an added bonus (and I hope they are correct as I don't know French :-). I note this PR is well timed as the PR for eGauge will be the first time that a meter will be using its timezone.

I pushed a commit to translate "no timezone".

This PR should be ready to merge.

PR #754, opened 2022-04-18
Unit dropdown

andrew-cline:
# Description
Adds a dropdown selection menu on the graphing page for units. The menu has logic for enabling and disabling units depending on compatibility with meters and groups that have been selected.
## Type of change

## Checklist

## Limitations
When selecting a meter the units menu does not update with compatible units. This can lead to issues where a meter is selected when an incompatible unit is selected or accidentally selecting the wrong unit for a meter upon first selection, both of which cause a crash on the front end.


src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+ * Determines the compatibility of units in the redux state for display in dropdown
+ * @param {State} state - current redux state
+ * @return {SelectOption[]} an array of SelectOption
+ */
+export function getUnitCompatibilityForDropdown(state: State) {
+	// Holds all units that are compatible with selected meters/groups
+	const compatibleUnits = new Set<number>();
+	// Holds all units that are not compatible with selected meters/groups
+	const incompatibleUnits = new Set<number>();
+
```
huss: extra blank lines

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+		let first = true;
+		let units = new Set<number>();
+		const M = new Set<number>();
+
+		state.graph.selectedMeters.forEach(meter => {
+			M.add(meter);
+		})
+		// Get for all meters
+		M.forEach( () => {
+
```
huss: remove blank line

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+		})
+	} else {
+		// Some meter or group is selected
+		// Holds the units compatible with the meters/groups selected.
+		// The first meter or group processed is different since intersection with empty set is empty.
+		let first = true;
+		let units = new Set<number>();
+		const M = new Set<number>();
+
+		state.graph.selectedMeters.forEach(meter => {
```
huss, line 185: It would seem that the two meter loops could be made one as is done for groups.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+			if (first) {
+				// First meter/group so all its units are acceptable at this point
+				units = newUnits;
+				first = false;
+			} else {
+				// Do intersection of compatible units so far with ones for this meters
+				units = setIntersect(newUnits, units);
+			}
+		})
+		getVisibleUnitOrSuffixState(state).forEach(o => {
```
huss, line 210: The comments from the design document are mising from here on. It is fine to modify them but there should be comments in the code.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+	return finalUnits;
+}
+
+/**
+ * Creates an array from the state filled with all units that are not meter or not displayable
+ * @param {State} state - current redux state
+ * @return {UnitData[]} an array of UnitData
+ */
+export function getVisibleUnitOrSuffixState(state: State) {
+	const visibleUnitsOrSuffixs = _.filter(state.units.units, function (o: UnitData) {
```
huss: Suffixs is misspelled.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+		})
+	}
+	const finalUnits = getUnitCompatibility(compatibleUnits, incompatibleUnits, state);
+	return finalUnits;
+}
+
+/**
+ * Creates an array from the state filled with all units that are not meter or not displayable
+ * @param {State} state - current redux state
+ * @return {UnitData[]} an array of UnitData
```
huss, line 229: The words just mirror the type declaration before it. Could it be more descriptive (maybe similar to the function comment above)? This is also true in next function.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+ */
+export function getVisibleUnitOrSuffixState(state: State) {
+	const visibleUnitsOrSuffixs = _.filter(state.units.units, function (o: UnitData) {
+		return o.typeOfUnit != UnitType.meter && o.displayable != DisplayableType.none;
+	})
+	return visibleUnitsOrSuffixs;
+}
+
+/**
+ * Converts two sets of unitIDs into one sorted array of SelectOptions for the dropdown
```
huss: I think the plan is that this functioin will be used in several places (why it is exported). Thus, could the comments be less specific to units and more general if this is the case. However, the function sets label and value and also loops based on unit. Is it possible to make this more general? If not, then I don't know if it should be exported for general use by meters and groups.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+	const finalUnits: SelectOption[] = [];
+	compatibleUnits.forEach(unit => {
+		finalUnits.push({
+			label: state.units.units[unit].identifier,
+			value: unit,
+			isDisabled: false
+		} as SelectOption
+		)
+	})
+
```
huss: extra blank line here and below.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+			label: state.units.units[unit].identifier,
+			value: unit,
+			isDisabled: true
+		} as SelectOption
+		)
+	})
+
+
+	return _.orderBy(finalUnits, ['isDisabled', 'label'], ['asc', 'asc']);
+}
```
huss: OED is going to standardize on having a return on the last line so could you please add one?

src/client/app/components/ChartDataSelectComponent.tsx
```diff
@@ -40,8 +45,10 @@ class ChartDataSelectComponent extends React.Component<ChartDataSelectPropsWithI
 		const messages = defineMessages({
 			selectGroups: { id: 'select.groups' },
 			selectMeters: { id: 'select.meters' },
+			selectUnit: { id: 'select.unit' },
 			helpSelectGroups: { id: 'help.home.select.groups' },
-			helpSelectMeters: { id: 'help.home.select.meters' }
+			helpSelectMeters: { id: 'help.home.select.meters' },
+			helpSelectUnit: { id: 'help.home.select.unit' }
```
huss: I appreciate you starting to add the tooltip we want for the new unit menu. However, it is not added later in this file for the TooltipMarkerComponent, in that component nor in the translation file. I would be great if you added it in all these places but, if not, then please remove this.

src/client/app/containers/ChartDataSelectContainer.ts
```diff
@@ -125,21 +129,142 @@ function mapStateToProps(state: State) {
...
+		finalUnits.push({
+			label: state.units.units[unit].identifier,
+			value: unit,
+			isDisabled: true
+		} as SelectOption
+		)
+	})
+
+
+	return _.orderBy(finalUnits, ['isDisabled', 'label'], ['asc', 'asc']);
```
huss: This was not in the design document but I noticed that we have units that are both capitalized and not. It would be nice if they were sorted to ignore this. For example, kW and kWh would not be past Metric ton.

huss (changes requested):
Thanks to @andrew-cline for this work and PR. Overall, it looks good. The final testing cannot happen until this is integrated with the rest of the new resource generalization code but it seems to work fine. I did notice that when I arbitrary set the selected unit (state.graph.selectedUnit = 101;) in src/client/app/containers/DashboardContainer.ts at the top of mapStateToProps, the initial load of the page showed that as the selected unit but all units were grayed out (including that one) on the unit menu. I'm hoping this is due to how I set the value and it will go away when it is selected normally.

There are a couple of minor/modest things that are requested to be addressed.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Removed blank lines in the most recent commit.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Removed blank lines in the most recent commit.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Combined the meter loops into one in most recent commit.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Added missing comments from the design document in the most recent commit.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Fixed misspelling in the most recent commit.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Changed the description for the function in the most recent commit. Let me know if you want it changed further.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: I'm not really sure how to generalize the function for the other menus. Since the function has to pull from that specific type's redux state.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Removed blank lines in the most recent commit.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: Added return on the last line in the most recent commit.

src/client/app/components/ChartDataSelectComponent.tsx
andrew-cline: Added lines to both the translation file and TooltipMarkerComponent. Also added the appropriate TooltipMarkerComponent call into ChartDataSelectComponent. I was not really sure what link to put in TooltipHelpComponentAlternative so I copied the format from the other ones. If that should be removed let me know.

src/client/app/containers/ChartDataSelectContainer.ts
andrew-cline: I figured out a way to get Lodash to sort in that order in the most recent commit.

huss:
Thanks to @andrew-cline for the quick addressing of the review. They all look good and testing indicates all works fine. This PR is ready to merge.

huss (approved):

Issue #586, opened 2021-03-02
replace moment


huss:
The moment project has decided to wind down and recommends replacements (https://momentjs.com/docs/#/-project-status/recommendations). There are other packages and we could wait for JS to formalize its own version. While moment works, we need to change sometime.

huss:
PR #760 demonstrated some of the changes needed to convert to day.js. It has been closed since we need to decide what will be used as a replacement. However, that PR is useful in thinking about this and for the actual change.

spearec:
After doing some research, native javascript support for date/time, as well as durations, is a currently proposed change in the form of Temporal (https://github.com/tc39/proposal-temporal).

Temporal is currently in the stage 3 proposal phase, which means that there are unlikely to be changes to the API before it gets released. Using a polyfill, we could use this package now, even without web browser support.

Since moment still works and this feature is still in the proposal phase, there are no urgent reasons to switch over now.
However, it also makes no sense to switch to another package, such as day.js or Luxon, when support for these features is close to being supported natively.

huss:
I'm just updating the this still seems to be in the final proposal stages. It would be great to wait until it is fully in ECMAScript but OED could move forward now if someone wants to pursue this. It will touch many parts of the code.

huss:
This is to note that node now supports this. Thus, it is probably safe to move forward. If someone wants to do a proof of concept of the different uses in OED that would be great. Given the large number of changes to OED for time-varying conversions, the actual change to the code base should probably wait until that is closer to ready since this will also change a lot of files.

Also, OED currently has code to convert to/from the database to deal with moment. It is unclear if that needs updating or can go given temporal. See src/server/models/patch-moment-type.js.

PR #760, opened 2022-04-23
Conversion from moment.js to day.js Issue #586

jbmayor4:
# Description

Begun migration from moment.js to day.js library. The motivation for migration is that moment.js is "done" and in legacy status. Day is the closest library to moment and thus was chonse

Partly Addresses #586 

## Type of change

- [x ] Note merging this changes the node modules

## Checklist

- [x ] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas

## Limitations

Only partial migration further steps for migration need to be done.


huss:
I appreciate @jbmayor4 for looking into this. Unfortunately, OED cannot merge this work at this time because moment is one of the central libraries used in OED. The project needs to have a discuss about which package to use before moving to something new. We have also considered waiting for the new ECMA that may allow for what we need directly in JavaScript without a library. This work is a good demonstration that Day.js is likely to meet our needs but more testing is needed be sure it works in all the uses. Also, PR #755 (currently in review) changes some uses of moment and we need to be sure that these are incorporated into anything we change to. OED will hold this code for future consideration for incorporation and to help guide our discussions. As such, it is a useful contribution to the project. It is being placed as a Draft PR for the reasons given. If anyone thinks this is a bad decision then please let me know.

I have also edited the description so it will properly show up in the related issue.

huss:
As indicated, the project needs to decide how to move forward on this. The linked issue has been update to note this work. As such, this PR is being closed.

PR #761, opened 2022-04-23
Create migrations

truongdd03:
# Description

Fixes create-DB-migrations. This PR creates migrations for new changes in the database. 

## Type of change

## Checklist

### DIFF
--- a/src/server/migrations/0.8.0-1.0.0/sql/group/add_default_graphic_unit.sql
+++ b/src/server/migrations/0.8.0-1.0.0/sql/group/add_default_graphic_unit.sql
@@ -1,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ALTER TABLE groups
+    ADD COLUMN IF NOT EXISTS default_graphic_unit INTEGER REFERENCES units(id) DEFAULT id_of_units('kWh');
--- a/src/server/migrations/0.8.0-1.0.0/sql/group/add_default_graphic_unit.sql
+++ b/src/server/migrations/0.8.0-1.0.0/sql/group/add_default_graphic_unit.sql
@@ -2,5 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+-- Add default_graphic_unit to the groups table.
 ALTER TABLE groups
-    ADD COLUMN IF NOT EXISTS default_graphic_unit INTEGER REFERENCES units(id) DEFAULT id_of_units('kWh');
+    ADD COLUMN IF NOT EXISTS default_graphic_unit INTEGER REFERENCES units(id) DEFAULT id_of_units('Electric_utility');
--- a/src/server/migrations/0.8.0-1.0.0/sql/meter/add_default_units_and_conversions.sql
+++ b/src/server/migrations/0.8.0-1.0.0/sql/meter/add_default_units_and_conversions.sql
@@ -1,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Insert Electric_utility, which is the unit for all existing meters/groups.
+INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, unit_index, suffix, displayable, preferred_display, note)
+VALUES ('Electric_utility', 'Electric_utility', 'quantity'::unit_represent_type, 3600, 'meter'::unit_type, null, '', 'none'::displayable_type, FALSE, 'created during migration')
+ON CONFLICT DO NOTHING;
+
+-- Insert kWh.
+INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, unit_index, suffix, displayable, preferred_display, note)
+VALUES ('kWh', 'kWh', 'quantity'::unit_represent_type, 3600, 'unit'::unit_type, null, '', 'all'::displayable_type, TRUE, 'created during migration')
+ON CONFLICT DO NOTHING;
+
+-- Function to find the id associated with a unit name.
+CREATE OR REPLACE FUNCTION id_of_units(unit_name VARCHAR) RETURNS INTEGER AS $$
+    SELECT id FROM units WHERE name = unit_name;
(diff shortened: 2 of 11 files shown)


src/server/migrations/0.8.0-1.0.0/sql/group/add_default_graphic_unit.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Add default_graphic_unit to the groups table.
+ALTER TABLE groups
+    ADD COLUMN IF NOT EXISTS default_graphic_unit INTEGER REFERENCES units(id) DEFAULT id_of_units('Electric_utility');
```
huss: Two notes on this. First, I don't think we want to set the DEFAULT as this does not mirror the create_groups_table.sql. Instead, it should update the rows that exist in the groups table to use the desired default_graphic_unit after the alter to add that column. Second, I don't think current rows should be set to Electric_utility. This is the graphing unit and that cannot be a meter type. I think we want it to be kWh as OED assumed all graphics were that unit in the past.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_default_units_and_conversions.sql
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Insert Electric_utility, which is the unit for all existing meters/groups.
```
huss: Per other comment, this is the unit of meters but not groups.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_default_units_and_conversions.sql
```diff
@@ -0,0 +1,23 @@
...
+VALUES ('Electric_utility', 'Electric_utility', 'quantity'::unit_represent_type, 3600, 'meter'::unit_type, null, '', 'none'::displayable_type, FALSE, 'created during migration')
+ON CONFLICT DO NOTHING;
+
+-- Insert kWh.
+INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, unit_index, suffix, displayable, preferred_display, note)
+VALUES ('kWh', 'kWh', 'quantity'::unit_represent_type, 3600, 'unit'::unit_type, null, '', 'all'::displayable_type, TRUE, 'created during migration')
+ON CONFLICT DO NOTHING;
+
+-- Function to find the id associated with a unit name.
+CREATE OR REPLACE FUNCTION id_of_units(unit_name VARCHAR) RETURNS INTEGER AS $$
```
huss: While using a function is an elegant solution, I have two thoughts. First, you use this in other migration files and that means that this must happen first. I think any such dependency should be carefully documented. Second, this leaves a function in the database that does not exist in unmigrated systems. I don't think that is a good idea. While it could be removed after migrations are done, I think it might be cleaner to select the needed id within the insert commands below and in other files to avoid the need for a function that is only used for migrations.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_unit_id_default_graphic_unit.sql
```diff
@@ -0,0 +1,8 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Add unit_id and default_graphic_unit to the meters table.
```
huss, line 5: As in the comment on groups, I don't think we should use DEFAULT but alter the current rows. In this case it is correct that it should be Electric_utility.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_default_units_and_conversions.sql
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Insert Electric_utility, which is the unit for all existing meters/groups.
+INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, unit_index, suffix, displayable, preferred_display, note)
```
huss, line 9: The dependency that this must happen before other migrations (such as meters) should be documented here and in the index file.

huss (changes requested):
Thanks to @truongdd03 for doing this. I've made a few suggested changes. I believe the migrations are now working for you so maybe this should not be draft anymore. Also, it might be good to note that this completes the resource generalization task "create DB migrations" in the description. Finally, I did a quick check and the migrations seem to work fine. I'll do a more extensive test with meter and group data after the changes are made.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_default_units_and_conversions.sql
```diff
@@ -0,0 +1,21 @@
...
+ON CONFLICT DO NOTHING;
+
+-- Insert kWh, which is the default_graphic_unit for all existing meters/groups.
+INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, unit_index, suffix, displayable, preferred_display, note)
+VALUES ('kWh', 'kWh', 'quantity'::unit_represent_type, 3600, 'unit'::unit_type, null, '', 'all'::displayable_type, TRUE, 'created during migration')
+ON CONFLICT DO NOTHING;
+
+-- Insert the conversion from Electric_utility to kWh. The conversion's slope is 1 since all existing data is in kWh.
+INSERT INTO conversions(source_id, destination_id, bidirectional, slope, intercept, note)
+VALUES (id_of_units('Electric_utility'), id_of_units('kWh'), FALSE, 1, 0, 'Electric Utility → kWh')
```
huss: This is still using the deleted function of id_of_units.

huss (changes requested):
I have not carefully tested this yet but noticed this one item while looking at the code.

truongdd03:
Thank you for your review, @huss. I have resolved all the comments. Please let me know if you think there is something else that needs to be changed. Thank you.

huss (approved):
Thanks to @truongdd03 for the quick changes. Everything seems good. I did a full migration from 0.8.0 and it worked. I have updated the help pages to reflect some changes in the process. This PR is ready to merge.

Issue #635, opened 2021-05-05
update code for new meters, groups & maps


huss:
### Describe the bug

PR #632 modified the database for meters, groups & maps. There are a number of places in the code where these objects are created. Right now, they are not explicitly sending values for the new attributes of these objects. At some point it would be good to look through them to see which should be modified for the new values.

### Additional context

I did a quick search of the code and found these files are impacted (searched for "new meter(", etc.). Another careful check should be made when the work is done.
        ◦ Meters
            ▪ src/server/models/obvius/processConfigFile.js
            ▪ src/server/services/readMamacMeters.js
            ▪ src/server/services/obvius/loadLogfileToReadings.js
            ▪ src/server/services/pipeline-in-progress/insertObviusData.js
            ▪ and lots of tests
        ◦ groups
            ▪ src/server/routes/groups.js
            ▪ and lots of tests
        ◦ maps
            ▪ src/server/routes/maps.js
            ▪ and lots of tests

huss:
We may want to wait until the resource generalization database changes are done so they can be rolled into this.

huss:
I looked at all the listed files and found they are either fixed or don't do new meter. The code was swept earlier to try to get all the needed values set or use proper defaults. Thus, this issue does not seem needed.

PR #762, opened 2022-04-24
fix unit arguments

ngan-cassie:
# Description

Partly address #635 

## Type of change

## Checklist

- [x ] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas

## Limitations
- Some need other developers to look into and make changes if needed


src/server/models/obvius/processConfigFile.js
```diff
@@ -53,7 +55,9 @@ function processConfigFile(configFile) {
 			true,
 			undefined,
 			undefined,
-			undefined
+			undefined,
```
huss, line 67: It is not safe to assume the id number of an item in the database. You need to use Unit.getByName (for example, see src/server/data/automatedTestingData.js) to get the id associated with the unit name (and as you added to the test code).

src/server/services/obvius/loadLogfileToReadings.js
```diff
@@ -24,7 +24,7 @@ async function loadLogfileToReadings(serialNumber, ipAddress, logfile, conn) {
 			meter = new Meter(undefined, `${serialNumber}.${i}`, ipAddress, true, false, Meter.type.OBVIUS,
 				null, undefined, `OBVIUS ${serialNumber} COLUMN ${i}`, 'created via obvious log upload on ' +
 				moment().format(), undefined, undefined, undefined, undefined, undefined, undefined, undefined,
-				undefined, undefined, true, undefined, undefined, undefined);
+				undefined, undefined, true, undefined, undefined, undefined, undefined, undefined);
```
huss, line 30: I think you should add to the comment above about why these are undefined and that the unit cannot be determined.

src/server/services/readMamacMeters.js
```diff
@@ -49,12 +49,14 @@ async function reqWithTimeout(url, timeout, csvLine) {
...
 async function getMeterInfo(url, ip, csvLine) {
+	// TODO: get the unit name from the file
+	const unitName = 'kWh';
 	return reqWithTimeout(url, 5000, csvLine)
 		.then(raw => parseXMLPromisified(raw))
 		.then(xml => {
 			const name = xml.Maverick.NodeID[0];
 			return new Meter(undefined, name, ip, true, true, Meter.type.MAMAC, null, undefined, undefined,
-				'created via MAMAC meter upload on ' + moment().format());
+				'created via MAMAC meter upload on ' + moment().format(), unitId=1, defaultGraphicUnit=1);
```
huss: See elsewhere on getting unit id.

src/server/test/web/meters.js
```diff
@@ -80,6 +80,8 @@ function expectMetersToBeEquivalent(meters, length, offset) {
 			expect(meter).to.have.property('reading', null);
 			expect(meter).to.have.property('startTimestamp', null);
 			expect(meter).to.have.property('endTimestamp', null);
+			expect(meter).to.have.property('unitId', 1);
```
huss: This is inserted a unitId in the tests and taht is what it should equal. As elsewhere, you cannot assume 1.

huss (changes requested):
Thanks to @ngan-cassie for looking into this. I've made a few comments about the unit id that should be addressed.

src/server/services/obvius/loadLogfileToReadings.js
ngan-cassie: @huss could you give me more context on this part? I checked the note last time but didn't see or remember why we agreed to set the unit ID and default graphic unit to undefined.

huss: A part of resource generalization that should occur in the next few months is to get all the automatically created meters for ones OED automatically reads to figure out the actual unit of the meter. At this point in time, we are not doing that. Thus, we are setting the unit and graphic unit to be undefined to indicate it has no unit yet. Note the displayable is set to false so no one can graph this meter's data so having an undefined unit is okay. The admin will need to update the units and displayable manually for any such meters but current ones will be set to kWh as part of the migration. I hope this helps.

ngan-cassie (commented):
I fixed some code to get the right unit ID (basically just did the async await thing) in three files: 
src/server/test/web/meters.js
src/server/models/obvius/processConfigFile.js 
src/server/services/readMamacMeters.js

When I tried running docker-compose up, they seemed to work okay. What bothered me are:

1. For the meters test file:
- I don't know what unit ID non-admin roles will see, so I just put null in the test (line 86). However, tests with non-admin roles failed (see image attached). I assume this has something to do with authentication that I'm not really familiar with. 
Screen Shot 2022-05-05 at 4 25 27 PM

2. When I first ran docker-compose up, the console logged all the % webpack.Progress thing, but whenever I made changes and saved, it just kept showing this no matter whether the code is wrong or not. This made it difficult to check if I make any error. 
Screen Shot 2022-05-05 at 4 27 45 PM

@huss  could you have a look at it and let me know what you think?

ngan-cassie (commented):
@huss Actually I just realized somebody apparently already made changes to the meters test file (adding an additional parameter for the expectMetersToBeEquivalent function) so now my changes conflicted with that.

src/server/services/obvius/loadLogfileToReadings.js
huss: I updated the comment some.

src/server/services/readMamacMeters.js
huss: I changed the code a little but it is done.

huss:
For point 2. above, if you are changing only server code then webpack does not does not rebundle the app. In this case it only says it is listening on port 3000 when it is done. If something is wrong it should give you an error before this but server code errors might not show up.

As for the changes to src/server/test/web/meters.js: I have a correspondence between us about fixing that code up. My response was: "Yes. I have the email and will work on getting this fixed." to say that someone else would get this done. I edited your PR to make it cleaner for this code. I will do the merge soon and resolve what code to use.

huss (approved):

huss:
I made some changes to the code that are indicated in the merge comments including resolving the conflict holding up the merge. This addresses point 1 from @ngan-cassie  above. Please let me know if you think any of the changes are an issue. Otherwise it should be ready to merge.

PR #763, opened 2022-04-24
Tooltips fix

### LINKED ISSUE #732: help icons not working in some places
### Describe the bug

The ? help icons should pop up help text. This does not work on the bar, compare or map graphic page after the Meters: one and before the Toggle chart link one. It works fine on the line graphic page. I note that this is the area on the page where non-drop down menu items are located.

### To Reproduce

See above

### Expected behavior

In version 0.8 this worked and should work. It is unknown which change caused the issue but we might want to first look at the package updates.

### Additional context

None

lindavin:
# Description

Fixes the problem of tooltips not rendering in the right place. 

- To enable tooltips to render on markers (i.e. the circles') that mount later, we call the ReactTooltip#rebuild function. This addresses the problems of tooltips not rendering when switching to different chart types.
- To enable tooltips to appear on top of the modal (rather than behind it), we use the CSS via classname 'tip' in the css file styles/tooltip.css. This idea came from this discussion https://stackoverflow.com/questions/39982778/react-and-bootstrap-tooltip-inside-modal-appears-behind-modal-window.
- We also rename the TooltipContainerAlternative and TooltipComponentAlternative files to TooltipContainer and TooltipComponent respectively.
- We also removed the TooltipContainer from the LanguageSelectorComponent because it references page='home'. This causes problem because there should just be one TooltipContainer per page (i.e. there should not be two TooltipContainers with the same page attributes).

Fixes #732 

## Type of change

## Checklist


src/client/app/components/UIOptionsComponent.tsx
```diff
@@ -63,6 +64,12 @@ class UIOptionsComponent extends React.Component<UIOptionsPropsWithIntl, UIOptio
 		};
 	}
 
+	public componentDidUpdate(prev: UIOptionsProps) {
+		if (prev.chartToRender !== this.props.chartToRender) {
+			ReactTooltip.rebuild(); // This rebuilds the tooltip so that it detects the marker that disappear because the chart type changes.
```
huss, line 69: This is a small point but OED is trying to get comments on lines above that line and not on the same line, esp. if there is any length to the comment.

huss (changes requested):
I apreciate @lindavin looking into this. Testing indicates this works well. I don't know if @lindavin sees this but if I dock the menus, go to the menus and then click on a help icon then I cannot get the popup help to go away without clicking off the menu popup and then clicking again which closes it. Is there a way around this? Also, one help popup for the top, right menu choices does not go away when you click another tooltip when the menu is docked.

huss:
The issue with the tooltop that is in the top, right of the webpage staying open has been around for a while. Why is unknown. Clicking away from it cause it (and another open tooltip) to close. For now we are going to ignore this issue.

huss (approved):
Everything seems to work as expected so this is ready to merge.

PR #776, opened 2022-05-10
Set reading to float

### LINKED ISSUE #650: should readings in DB be 32 or 64 bit real number?

truongdd03:
# Description

Testing found that the space and access time are acceptable when changing the reading type from real (32-bits) to float (64-bit). This PR makes this change for the reading table and functions using these values. DB migration is also created. I have done a full migration and it works as expected on my machine.

Fixes #650 

## Type of change

## Checklist

### DIFF
--- a/src/server/migrations/0.8.0-1.0.0/index.js
+++ b/src/server/migrations/0.8.0-1.0.0/index.js
@@ -26,6 +26,8 @@ module.exports = {
 		await db.none(sqlFile('../migrations/0.8.0-1.0.0/sql/cik/create_cik_table.sql'));
 		await db.none(sqlFile('../migrations/0.8.0-1.0.0/sql/readings/drop_old_views.sql'));
 		// Note that old views should be dropped before the reading's type is changed to float.
+		// It is also important that the creation of new views below is done after this because
+		// you cannot alter a type that is used in a view.
 		await db.none(sqlFile('../migrations/0.8.0-1.0.0/sql/readings/set_reading_to_float.sql'));
 		await db.none(sqlFile('../migrations/0.8.0-1.0.0/sql/readings/set_function_get_compare_readings_to_float.sql'));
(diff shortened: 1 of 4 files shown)


huss:
I just want to leave a record that I searched all SQL uses of real and found these:

Screen Shot 2022-05-10 at 8 47 59 AM

Given their usage they are probably fine at 32-bit so they are not being changed now.

huss (approved):
This works fine and as intended. I did not do a full migration but tested each migration individually. I push a few small changes to clarify about order and needed operations before alters/creates. Can @truongdd03 look at them to make sure they are okay?

I note that I realized you have to refresh the reading views after the migration since there are new views. That will be dealt with separately.

I think this is ready to merge after my changes are checked.

truongdd03:
Thank you for your review and addition, @huss. New changes look good to me so I think the PR can be merged now.

PR #779, opened 2022-05-15
messages for empty graphs or without readings

### LINKED ISSUE #576: empty graphic
If there is no data for the comparison (because the meter/group data is missing for that time range) then you get an empty graph without any title for that graphic. It seems better if we did something better. Maybe a title but instead of the % change it could say "no data"? On a line and bar graphic you see an unusual graphic.

Maps say "There's not an available map" so the issue does not happen.

These empty graphics cause other issues such as the closed #616 that should be fixed by not showing these unusual graphics.

ngan-cassie:
# Description

Fixes #576 

## Type of change

## Checklist

## Limitations

- Changes in LineChartContainer break the page, I think especially line 23. 
- For CompareChart, if you select a meter without reading and then a meter with readings, nothing will get displayed but if you select a meter with readings first it will get displayed.


huss:
Thanks to @ngan-cassie for working on this. I have looked at the indicated issues and can see them. I looked at the code and the reason was not immediately clear. This has then sat for handful of days. So, I just wanted to note that I need to find some time to really look into this so there are no issues. The PR will need to wait for that (unless someone else figures this out).

In terms of the fix it seems fine modulo these issues.

huss:
This work was superseded by PR #965 so is being closed.

Issue #647, opened 2021-08-14
automate pipeline testing


huss:
### Is your feature request related to a problem? Please describe.

PR #629 has many tests in src/server/tests/web/csvPipeline/ that are CSV input file that are expected to give certain results. These tests should be automated and put into the tests done for checking.

### Describe the solution you'd like

Use Mocha/Chai examples already in OED to create appropriate tests. Some tests will require thinking how to automate.

### Describe alternatives you've considered

The current files allow testing but someone needs to verify the correct result is obtained. This does not meet the need to automate.

### Additional context

@huss created these tests so can be contacted to automate them. Also, someone could do some of them without doing them all.

huss:
In the src/server/tests/web/csvPipeline/ directory there are a couple of files that can help in understanding the desired tests:

- README.md: describes how the testing works (manually right now) and what each test has for settings. It also tells the expected outcome.
- pipelineCurl.sh: script that runs the tests manually at this time.
- pipelineCurlExpectedOutput.html: shows the expected output from running the test script. The current technique is to capture the output from script and diff it from this file to see if anything is off.

During testings manually, OED is inspected with the web interface to see if the desired meters and readings are created. This is why each test creates unique meters. If necessary the database is looked at. With automated testing, the mocha/chai tests should probe the database to see if the values are what were expected from the given CSV file.  The CSV upload would happen via curl commands or a CSV route. In the case testing a failed CSV upload, the test should verify that the bad values are not in the database.  It would also be good if the test looked at the string returned by the CSV upload request to verify it is what is expected in the standard output. src/server/test/web/obviusTest.js and src/server/test/routes/compressedReadingsRouteTests.js are some current tests that give ideas on testing responses from a web request. Also, the CSV upload changes the values in the database for the meter involved (reading, start_timestamp, end_timestamp) where they normally become the last reading uploaded. It may be easiest to see the expected value by looking at the database after manually loading the data. With complete testing, these should also be tested.

[...]

PR #780, opened 2022-05-23
Automation of the Pipeline

anggutierrez:
# Description
This issue aims to automate testing in the pipeline. This is important to have because it will speed up the development process for all developers contributing code to OED.

This change will use the input and expected files found at src/server/test/web/csvPipeline/automatedTests. These files were added to aid in the completion of this issue.

Partly Addresses #647 

## Type of change

## Checklist

## Limitations
A smart data structure needs to be created to store all of the corresponding parameters for each file. The main issue with this is the difference between files. This is because some files have headers whereas other files do not.

On top of this, more logic needs to be added to the loops themselves. Previously the pipeline was attaching to a file in multiple steps, but this needs to be changed so that it can be done in one step within a loop.


anggutierrez:
Previously, the database in its entirety would be downloaded locally for testing each time the a developer ran the test suite.

We were able to communicate and change this so that there would be expected and input files. This is important because it will save on developer time in the future. Unfortunately, this still has to be implemented, as the current test automation files still use the old system. 

---

After correctly using the input and expected tests, the next TODO with importance would be to find a smart data structure to hold all the parameters from the old .sh file. This is tricky because we have to come to an agreement on how we want the parameters to be parsed (either in a separate file or within the input file), as well as correctly storing them inside of a good data structure with key:value pairs.

---

Those are the two glaring issues that the team could not fix in the allotted time. The code for the testing itself already exists in the codebase, the issue would be to just transfer the logic over and use the DRY principle to work within loops.

huss:
I thank @anggutierrez for their efforts. This work will continue as a new effort so this PR is being closed. The PR is linked to the associated issue so others can find it.

Issue #644, opened 2021-07-05
update/upgrade meter page


huss:
### Is your feature request related to a problem? Please describe.

PR #629 means there are a lot of new meter values that admins should be able to edit/set. Basically everything that can be set via the CSV drop should be editable. Also, the name of a meter should be editable but is not on the current page.

### Describe the solution you'd like

We need to modify the meter page code to handle all of this. Might also look at the map page that has somewhat nicer edit abilities.

### Describe alternatives you've considered

An admin can now use a CSV file upload to edit the values (changing the meter name is a little tricky). However, we need to update the meter page to have a good interface.

### Additional context

None.

huss:
Note issue #577 proposes how to sort this page. It contains some information on where the code is located.

huss:
This is a proposal for how to modify the page. Thoughts and modifications based on difficulty in implementing are welcome.

If this is a non-admin then the page has a table with:
| Identifier | Unit | Default Graphic Unit | Enabled |
followed by one row for each meter with this information sorted by Identifier.
Unit and Default Graphic Unit will give the identifier of the unit associated with this id.

If this is an admin user then the page initially has a table with:
| Identifier | Unit | Default Graphic Unit | Displayable | Enabled | Edit/More Info |
followed by one row for each meter with this information sorted by Identifier.
Unit and Default Graphic Unit will give the identifier of the unit associated with this id.

The Edit/More info column has one button per row that is labeled Edit/More Info. If it is clicked then that meter row is expanded to show all 25 items associated with a meter along with ability to edit each one. If possible, it would be nice to show the expanded list in its same place in the table. If not, it could be done at the bottom of the page. The single row for the meter would become 5 rows (or 10 if the value is below) as follows:
| Identifier | Unit | Default Graphic Unit | Meter Type | GPS | Area |
| Displayable | Enabled | Meter Address | Timezone | note |
| Cumulative | Cumulative Reset | Reset Start | Reset End |
| Time Sort | End Only | Reading Gap | Reading Variation | Reading Duplication |
| Reading | Start Date/Time | End Date/Time | ID | Name | Save Button | Close/Cancel Button |
where the value for this meter would be with or right below the item.  The Save Button would become clickable if any meter value was edited. The other button would start as Close but switch to Cancel if an edit it done. It would be good to have a confirmation for Cancel so as not to accidentally lose edits.

Each item can be edited as follows:
[...]

PR #773, opened 2022-05-04
Update Meter Page, Sort Meters by Identifier, Display meters by identifier rather than name #577 #644, #700

Joseph-M-Newman:
# Description
Issue #644 
A complete overhaul of the previous meter page. Each meter gets its own card associated with the information. If you are logged in as an admin, an edit button appears attached with a Modal Card which houses all values. This was a complete redesign of the meter page. A new Module was added: React-Boostrap. 

Included is complete edit and save functionality for the identifier. This does not include all other values however the boiler code for additional changes is provided. This will require the route for saving to be changed when the other new values are implemented. The values currently included are as follows: Identifier, Name, GPS, displayable, Enabled, TimeZone. All others are currently not on the route code. src>server>routes>meter.js on line (186: 192).

Screen Shot 2022-05-03 at 2 05 21 PM

Screen Shot 2022-05-03 at 2 05 26 PM

# Description 
Issue: #577 and #700 

Meters on the meter page will now be displayed by meter identifier rather than the name. Meters will also be displayed in alphabetical sorted. An example is provided above in the pictures.

# Fixes

Fixes #644 #577 #700 

## Type of change
- [x] Note merging this changes the node modules
- [ ] Note merging this changes the database configuration.
- [x] This change requires a documentation update

## Checklist
- [x] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [x] I have removed text in ( ) from the issue request

## Limitations
TODO: 
- All other 24 values need to be wired up to the save function. Examples are provided and the boiler code is provided as well.
- The time zone null fixed provided by @AndrewMei99 @Onyi21 needs to be moved into the MeterModalEditComponent from MeterViewComponent and implemented into the isTimeZone function.
- Save route needs to be edited to allow for all values to be saved to the database rather than just Identifier, Name, GPS, displayable, Enabled, TimeZone. The file to be edited can be found at src>server>routes>meter.js on line (186: 192).

PR #786, opened 2022-07-21
Meter-page-modal-redux

wootent:
# Description

Created meter view, detail, edit modal & create modal interface w/ the use of Redux 

Fixes issue #644 and is the meter code discussed in PR #773.

## Type of change

## Checklist

## Limitations

1. GPS handlers and state update do not work as intended. Is most parts of the application GPS is handled as string and parsed and formatted before it is written to DB; however in some parts of the application this is not so.
2. create meter api does not work correctly, write fails.
3. form inputs need check for correctly formatted inputs.


huss (approved):
The code was reviewed. Some items were found but I updated the PR to address them. See commit messages for details. A reasonable level of testing was done and not issues remain. I do note that we should add more testing to the meter input to avoid incorrect values but that will be done later.

I thank those who made this substantial change to OED.

huss:
Local checks passed and there is an ongoing issue with GitHub Actions so it should be fine to merge.

PR #788, opened 2022-08-03
Conversions Page, Action Confirmation Modal Component

### LINKED ISSUE #700: use identifier not name for meters when displayed

OED added an identifier for meters. The idea was that the name, that is sometime automatically generated and relates to the name from the system sending the readings,  is not useful for a user. The identifier can be set by the admin and would serve as the name shown to users. While the database and admin changes were made, the identifier is not being used in menus and graphics.

The menu for meters should use the identifier. The label on all graphics should do the same. The group page should show identifier and not the name. The meter page already does this.

None.

### Additional context

The resource generalization effort will work on the meter menu and should use identifier. This change can go on in parallel but coordinate to see if that effort has/is doing the menu change.

There are a number of places the code needs to be changed. They are (list might not be complete):

- src/client/app/containers/MeterDropdownContainer.ts for meter dropdown menu
- src/client/app/containers/LineChartContainer.ts, src/client/app/containers/BarChartContainer.ts, src/client/app/containers/CompareChartContainer.ts (the name is in linked code) and src/client/app/containers/MapChartContainer.ts for the different graphic types
- src/client/app/containers/groups with code for groups

Deelane:
# Description

### Conversions Page

Conversions page built using the Units page from PR 787 as a template. Added delete functionality.

### Action Confirmation Modal Component

New modal component for confirming any action. This component can be used anywhere the developer needs confirmation before executing an action. Required props are a boolean "show" that controls the modal's visible state, a message to display in the modal, and 2 functions. One function is executed when the "confirm" button is pressed and one executes when the "reject" button is pressed. 

It is assumed that the developer implements their own functionality in the reject function for altering the "show" boolean to alter the visibility of the modal. Also, be sure to hide any parent modals too if you use the confirm modal as a nested modal. 

Optional props include confirm and reject text that will replace the default "yes" and "no" text for the buttons, and a title for the modal.

*Note: This modal component has very basic CSS, updates to the visuals to make it look nicer are welcome from any developer.

Fixes #700 to use identifier for meter.

## Type of change

## Checklist


src/client/app/components/ConfirmActionModalComponent.tsx
```diff
@@ -0,0 +1,94 @@
...
+	// usually alters the show props to control closing the modal
+	handleClose: () => void;
+	// Function to execute if action is confirmed
+	// Be sure to pass into any additional handleClose functions for hiding parent modals
+	// Also be aware that react boostrap does not support nested modals by default
+	// A good solution to this is to hide the parent modal when this modal is opened
+	actionFunction: () => void;
+}
+
+/**
```
huss, line 30: This seems a general way that is only used in EditConversionModalComponent. Was it intended to be used elsewhere?

src/server/routes/conversions.js
```diff
@@ -0,0 +1,188 @@
...
+		}
+	};
+
+	const validatorResult = validate(req.body, validConversion);
+	if (!validatorResult.valid) {
+		log.warn(`Got request to edit conversions with invalid conversion data, errors:${validatorResult.errors}`);
+		res.status(400);
+	} else {
+		const conn = getConnection();
+		try {
```
huss, line 74: Why do we need to get the conversion since all values overwritten? Is it to check if it exists?

src/client/app/types/redux/conversions.ts
```diff
@@ -0,0 +1,79 @@
...
+	slope: number;
+	intercept: number;
+	note: string;
+}
+
+export interface ConversionsState {
+	hasBeenFetchedOnce: boolean,
+	isFetching: boolean;
+	selectedConversions: number[];
+	// Use an array of ConversionData because we need the combination of source/destination id to give us a unique conversion
```
huss, line 67: I'm unclear on why this needs to be an array since each conversion has has both values. Can you help me understand?

src/client/app/utils/api/ConversionsApi.ts
```diff
@@ -0,0 +1,37 @@
...
+import { ConversionData, ConversionEditData } from '../../types/redux/conversions';
+export default class ConversionsApi {
+	private readonly backend: ApiBackend;
+
+	constructor(backend: ApiBackend) {
+		this.backend = backend;
+	}
+
+	public async edit(conversion: ConversionData): Promise<ConversionEditData> {
+		return await this.backend.doPostRequest<ConversionEditData>(
```
huss: The input is ConversionData but then it does the post with ConversionEditData.

src/client/app/types/redux/conversions.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface ConversionData {
+	sourceId: number;
+	destinationId: number;
+	bidirectional: boolean;
+	slope: number;
+	intercept: number;
+	note: string;
+}
+
+export interface ConversionEditData {
```
huss: Edit seems the same as ConversionData. Is there a reason we need both? I know some others have both since the information varies (but not here). This is used in src/client/app/utils/api/ConversionsApi.ts for what is returned by edit (and posted) but SubmitEditedConversionAction uses ConversionData in the action above.

huss (changes requested):
This PR looks good. I have a few thoughts that I would like to see if the authors can give insight. I thank everyone for doing this major task.

huss:
I want to apologize to the authors of the PR for taking so long. I was hoping to clear up the issue of updating Cik on edits. However, it turned into a bigger issue and also impacts other pages. Thus, I have opened issue #791 to deal with this. This PR is passing local tests and is failing on GitHub per the ongoing issue. Once the last items are addressed and some final testing is done this should be ready to merge and then be put int development.

src/client/app/components/ConfirmActionModalComponent.tsx
huss: Added TODO for future consideration.

src/server/routes/conversions.js
huss: Seems safe to assume conversion exists and the catch would return an error if fails. Changed to create a new conversion and update that without the first DB call.

src/client/app/types/redux/conversions.ts
huss: Removed ConversonEditData and replaced with ConversionData.

src/client/app/utils/api/ConversionsApi.ts
huss: Fixed when removed ConversionEditData.

src/client/app/types/redux/conversions.ts
huss: I'm not sure if this needs to be an array but it works so letting it go for now so the PR can be cleared.

huss (approved):

huss:
All comments have been addressed. Note I updated the PR in several ways that are given in the commit messages. Local tests all pass so continuing to ignore GitHub Action failure (ongoing issue).

Note that there are issues in certain actions on conversions (and units/meters). Issue #791 was created to address these problems. However, this PR is now being merged to move it forward.

PR #801, opened 2022-10-04
Bar Chart date x-axis and single-day mouseover

### LINKED ISSUE #688: bar location if different meters/groups have different time ranges
### Describe the bug

If the meters/groups in a bar graph have different date ranges then the x-axis date labels can be off. The values seem fine. The following figure shows that 2021 comes before 2020. I only tested data where the meters and groups had disjoint dates but it isn't know if that is important to reproduce the bug. I suspect it is because the meters are graphed first and then the groups so the dates are not interleaved/sorted properly. Thus, it may be important that the group time range is before the meter time range to see the issue. The line graphic does not seem to have this issue.

### To Reproduce

The image was created using the data for the OED website. Directions should soon be on the DevDocs on using this data. However, it is not believed to be linked to this specific data but the type of data.

### Expected behavior

Bars should display in increasing dates.

### Screenshots

Screen Shot 2021-12-18 at 11 47 53 AM

### Additional context

None.

spearec:
# Description
This changes the bar graph's x-axis to display using dates instead of string ranges, which is how the line graph currently displays data. 
Issue #688 was caused by displaying the date ranges as strings, which plotly does not sort. By changing the display type to dates, the x-axis values are automatically sorted.
This also changes the mouseover readout for bar and map data. When the display range is set to one day, it will only display that one day instead of a range.

Screenshot 2022-10-04 144909

Fixes #688 

## Type of change

## Checklist

### DIFF
--- a/src/client/app/containers/BarChartContainer.ts
+++ b/src/client/app/containers/BarChartContainer.ts
@@ -68,13 +68,19 @@ function mapStateToProps(state: State) {
 				const hoverText: string[] = [];
 				const readings = _.orderBy(readingsData.readings, ['startTimestamp'], ['asc']);
 				readings.forEach(barReading => {
-					// subtracting one extra day caused by day ending at midnight of the next day.
-					// Going from DB unit timestamp that is UTC so force UTC with moment, as usual.
-					const timeReading: string =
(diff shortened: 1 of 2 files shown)


huss (approved):
Thanks to @spearec for their first PR. Everything looks look. Local testing found no issues. I made a small change to one comment that is pushed. This is ready to merge.

PR #807, opened 2022-10-22
802 make page button non dynamic

### LINKED ISSUE #802: Make page buttons non-dynamic
Per @huss suggestion:

> When page buttons (to select what OED page you are on) were first added, they were made dynamic so
> the page you are on does not show. People don’t like the fact that the buttons move when you click one.
> Thus, we want to put them in a fixed order so the current page is still present. A nice addition would be
> to mark the current page button (such as grayed out). This requires working with React but it should
> not be too complex.

(We're going to assign this to a team of students already :) )

AriseTran:
# Description
As part of the CodeDay Microinternship, our group, @AriseTran @TrongQuocLe @tranchung163 @Cao1224, made changes to the page buttons so that the buttons are always displaying and will be greyed out depending on the tab that the user is on. 
Issue #802  was caused by the page buttons originally being made dynamic.

https://user-images.githubusercontent.com/104533320/198497982-e2804028-0a96-4e63-91ae-d1b83c459288.mp4

Fixes #802 

## Type of change

## Checklist

### DIFF
--- a/src/client/app/components/HeaderButtonsComponent.tsx
+++ b/src/client/app/components/HeaderButtonsComponent.tsx
@@ -40,13 +40,13 @@ export default class HeaderButtonsComponent extends React.Component<HeaderButton
 		const showOptions = getPage() === '';
 		const renderLoginButton = !hasToken();
 		const shouldHomeButtonDisabled = getPage() === '';
-		const renderAdminButton = loggedInAsAdmin && getPage() !== 'admin';
+		const renderAdminButton = loggedInAsAdmin && getPage() == 'admin';
 		const shouldGroupsButtonDisabled = getPage() === 'groups';
 		const shouldMetersButtonDisabled = getPage() === 'meters';
-		const renderMapsButton =loggedInAsAdmin && getPage() !== 'maps';
-		const renderCSVButton = role && hasPermissions(role, UserRole.CSV) && getPage() !== 'csv';
-		const renderUnitsButton = loggedInAsAdmin && getPage() !== 'units';
-		const renderConversionsButton = loggedInAsAdmin && getPage() !== 'conversions';
+		const renderMapsButton =loggedInAsAdmin && getPage() == 'maps';
+		const renderCSVButton = Boolean(role && hasPermissions(role, UserRole.CSV) && getPage() == 'csv');
+		const renderUnitsButton = loggedInAsAdmin && getPage() == 'units';
+		const renderConversionsButton = loggedInAsAdmin && getPage() == 'conversions';
 		const renderLogoutButton = hasToken();
 
 		const loginLinkStyle: React.CSSProperties = {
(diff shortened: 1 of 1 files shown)


huss:
First, thank you to these new developers for their first (draft) PR. While it is a draft, I took a quick look at the code. I had two thoughts:

1. Given we are no longer changing the order of the buttons, I think we should put them in alphabetical order based on English. The order may not be perfect when the user switches to another language but it is often pretty good.
2. You may know this but the buttons other than meter and group are no longer disappearing when someone is not logged in as an admin. That was the case before and we would like to continue to do that.

Please let me know if anything is not clear or you have any thoughts.

Cao1224:
Group member！

src/client/app/components/HeaderButtonsComponent.tsx
```diff
@@ -102,16 +78,28 @@ export default class HeaderButtonsComponent extends React.Component<HeaderButton
 				<div className={this.props.showCollapsedMenuButton ? 'd-none d-lg-block' : ''}>
 					<TooltipHelpContainer page='all' />
 					<TooltipMarkerComponent page='all' helpTextId='help.home.header' />
-					<Link style={adminLinkStyle} to='/admin'><Button outline><FormattedMessage id='admin.panel' /></Button></Link>
-					<Link style={groupsLinkStyle} to='/groups'><Button outline><FormattedMessage id='groups' /></Button></Link>
-					<Link style={metersLinkStyle} to='/meters'><Button outline><FormattedMessage id='meters' /></Button></Link>
-					<Link style={mapsLinkStyle} to='/maps'><Button outline><FormattedMessage id='maps' /></Button></Link>
-					<Link style={csvLinkStyle} to='/csv'><Button outline><FormattedMessage id='csv' /></Button></Link>
-					<Link style={homeLinkStyle} to='/'><Button outline><FormattedMessage id='home' /></Button></Link>
+					<Link
```
huss, line 81: OED has not been completely consistent in how it formats items with multiple properties. Lately we have been placing them on separate lines as you did in your new lines. I think that is good but I also think we should be consistent within a file. Thus, I would reformat the other Link items to also be multiline.

huss (changes requested):
Thanks to  @AriseTran, @TrongQuocLe ,@tranchung163 and @Cao1224 for their first PR. Testing indicates it works as desired. I appreciate the good work. I have made one formatting suggestion.

src/client/app/components/HeaderButtonsComponent.tsx
AriseTran: @huss I've reformated the link styles to be multilined. Does this suit the change you wanted?

huss: Thank you for the fast update. Everything looks good and about to be approved.

huss (approved):

PR #808, opened 2022-10-25
Redo Cik and refresh DB views on conversions/units updates

### LINKED ISSUE #791: Update Cik & DB views on conversion and other updates
### Describe the bug

PR #788 adds the conversion admin page. It works properly for creating/storing the new conversion. However, the Cik DB table is not properly updated. Also, Pik will be out of data and the reading DB views for daily and hourly may need to be refreshed. Refreshing the views can take time if the DB has lots of values. Thus, we only want to do this when necessary. Also, redoing Cik normally changes the indices for each unit in Cik/Pik so a client needs to refresh to see the changes. Given this, it is desirable to only do this when needed.

It may be possible to fix the admin Redux state on this change but it has to be done carefully given the changes. Note one reason this was not tried was because the units for graphing must be updated because Pik may have changed. We don't currently redo them on these types of changes because we did not expect Pik to update. Also, this is not a common action by an admin. Given this, it would be easier to just refresh the web browser page (such as window.location.reload()).

Certain changes to meters and maybe units can also require similar work. @huss worked on this but decided to move PR #788 forward and due to the complexity to do this right, an issue is now being opened. @huss will work with someone who want to do this to explain what was already tried. Note edit and create both have similar issues.

The following was tried or is suggested as a starting point:
- src/client/app/components/conversion/CreateConversionModalComponent.tsx adds code after the addConversion to do the needed redo and refresh. Since this can take time, the admin should be warned to wait until the refresh happens. @huss has code for this. It is important that the refresh does not happen until the redo of Cik is complete.
- Modify src/client/app/utils/api/ConversionArrayApi.ts to add an API to redo Cik and refresh the views (probably using arguments to decide which are done).
- src/server/services/graph/redoCik.js has functions for doing this work.

### Expected behavior

OED should work properly after admin changes. The resourceGeneralization document has info on when these types of updates are needed but it may need changes.

### Additional context

The issue of someone else having stale state is not addressed by this. That is a more general issue to fix in the future.

It would be nice to fix all the cases (units, conversions, meters & maybe groups) when this is done. Note changing groups for units is being done separately.

truongdd03:
# Description

Fixes #791 

This PR adds a new API to redo Cik and refresh DB views if needed after updating conversions or units. The web page is reloaded after the refresh happens to make sure the user has the correct redux state.

OED handles suffix units by adding other units/conversions automatically. When a unit's suffix changes, the additional conversions are removed while the additional units are hidden (set displayable = false). The code handling this logic can be found inside `src/server/services/graph/handleSuffixUnits.js`.

## Type of change

## Checklist


src/client/app/actions/admin.ts
```diff
@@ -129,3 +129,20 @@ export function submitPreferencesIfNeeded(): Thunk {
...
+
+function updateCikAndDBViews(): t.UpdateCikAndDBViews {
+	return { type: ActionType.UpdateCikAndDBViews };
+}
+
+/**
+ * Redo Cik and/or refresh reading views.
+ * This function is called when some changes in units/conversions affect the Cik table or reading views.
+ */
+export function updateCikAndDBViewsIfNeeded(shouldRedoCik: boolean, shouldRefreshReadingViews: boolean): Thunk {
```
huss, line 145: I'm wondering about the name of this function. I thought OED used IfNeeded in the name when it checks if it should be done. Here there is no check. So, should it be checking the update state set in UpdateCikAndDBViews or should the function have a different name?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -121,10 +121,14 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 				props.conversion.slope != state.slope ||
 				props.conversion.intercept != state.intercept ||
 				props.conversion.note != state.note);
+		// Need to redo Cik if slope, intercept, or bidirectional changes.
```
huss, line 117: I think this could be simplified by putting conversionHasChanges second and making it check for a note change or shouldRedoCik.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -73,6 +74,35 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+	//	2. There are no changes
+	const shouldUpdateUnit = (): boolean => {
+		// Check for case 1
+		if (props.unit.typeOfUnit === UnitType.meter && state.typeOfUnit !== UnitType.meter) {
+			// Get an array of all meters
+			const meters = Object.values(store.getState().meters.byMeterID);
+			const meter = meters.find(m => m.unitId === props.unit.id);
+			if (meter) {
+				// There exists a meter that is still linked with this unit
+				window.alert(`The unit of meter ${meter.name} needs to be changed before changing this unit's type`);
```
huss: This user string should be internationalized.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -81,30 +111,22 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
-			// Save our changes by dispatching the submitEditedUnit action
-			dispatch(submitEditedUnit(state));
-			// The updated unit is not fetched to save time. However, the identifier might have been
-			// automatically set if it was empty. Mimic that here.
-			if (state.identifier === '') {
-				state.identifier = state.name;
-			}
-			dispatch(removeUnsavedChanges());
-		}
+		if (!shouldUpdateUnit()) return;
```
huss: This is somewhat a personal perference but doing a return in the middle of a function when it can easily be avoided is not desirable because it makes the logic harder to follow. I would remove the ! and return and put the code below in braces.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -81,30 +111,22 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
 
+		// Need to redo Cik if the suffix, displayable, or type of unit changes.
+		const shouldRedoCik = props.unit.suffix !== state.suffix
+			|| props.unit.typeOfUnit !== state.typeOfUnit
+			|| props.unit.displayable !== state.displayable;
+		// Need to refresh reading views if unitRepresent or secInRate changes.
+		const shouldRefreshReadingViews = props.unit.unitRepresent !== state.unitRepresent || props.unit.secInRate !== state.secInRate;
```
huss: I think the secInRate only impact flow and raw units so we could check. On the other hand, admins should not set the value for the other units.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -81,30 +111,22 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
-				state.identifier = state.name;
-			}
-			dispatch(removeUnsavedChanges());
-		}
+		if (!shouldUpdateUnit()) return;
 
+		// Need to redo Cik if the suffix, displayable, or type of unit changes.
+		const shouldRedoCik = props.unit.suffix !== state.suffix
+			|| props.unit.typeOfUnit !== state.typeOfUnit
+			|| props.unit.displayable !== state.displayable;
```
huss: Change to displayable was not in the design document. I think you figured out an issue but I would like to understand how displayable impacts Cik.

src/server/routes/conversionArray.js
```diff
@@ -23,4 +24,18 @@ router.get('/', async (req, res) => {
 	}
 });
 
+/**
+ * Route for redoing Cik and/or refreshing reading views.
+ */
+router.post('/refresh', async (req, res) => {
+	const conn = getConnection();
```
huss: A really small point but why not put the conn inside the if so you avoid if no redoCik.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -42,12 +42,25 @@ async function addNewUnitAndConversion(sourceId, destinationId, slope, intercept
...
+		// The destination suffix unit exists but the conversion doesn't.
+		// Create a new conversion with desired values.
+		const newConversion = new Conversion(sourceId, destinationId, false, expectedSlope, expectedIntercept,
+			`${source.name} → ${destination.name} (created by OED for unit with suffix)`);
+		// Insert the new conversion to database and graph.
+		await newConversion.insert(conn);
+		graph.addLink(sourceId, destinationId);
+		return;
+	}
 	if (currentConversion.slope !== expectedSlope || currentConversion.intercept !== expectedIntercept) {
```
huss: Should this be an else if?

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -102,8 +115,8 @@ async function handleSuffixUnits(graph, conn) {
 			const destinationId = p[p.length - 1].id;
 			// The destination unit.
 			const destinationUnit = await Unit.getById(destinationId, conn);
-			// We don't need to create any new units/conversions if the destination unit has the type of suffix.
-			if (destinationUnit.typeOfUnit === Unit.unitType.SUFFIX) {
+			// We don't need to create any new units/conversions if the destination unit has the type of suffix or it's not displayed.
+			if (destinationUnit.typeOfUnit === Unit.unitType.SUFFIX || destinationUnit.displayable === Unit.displayableType.NONE) {
```
huss, line 131: I'm curious about the change to include displayable of none. I thought OED allows for passing through units that are not displayable. For example, kg CO2 is changed to displayable of none when it is first processed but if you add a conversion from kg to lbs then OED should add lbs of CO2. I just want to be sure this still does what we want and I may be missing something here.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -126,6 +139,32 @@ async function handleSuffixUnits(graph, conn) {
 	}
 }
 
+/**
+ * OED handles suffix units by adding conversions and units automatically.
+ * When a unit's suffix changes, these additional conversions and units need to be removed.
+ * Since it's complicated to remove units, we just set their displayable to false.
```
huss: Technically it is displayable to NONE.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -126,6 +139,32 @@ async function handleSuffixUnits(graph, conn) {
...
+ */
+async function removeAdditionalConversionsAndUnits(suffixUnit, conn) {
+	// Get all conversions from this suffix unit.
+	const conversions = (await Conversion.getAll(conn)).filter((conversion) => conversion.sourceId === suffixUnit.id);
+	conversions.forEach(async (conversion) => {
+		const destinationUnit = await Unit.getById(conversion.destinationId, conn);
+		// The units that OED adds are suffix unit.
+		if (destinationUnit.typeOfUnit === Unit.unitType.SUFFIX) {
+			// Delete the conversion.
+			await Conversion.delete(conversion.sourceId, conversion.destinationId, conn);
```
huss, line 171: I think the logic elsewhere makes the correct Cik and view refresh happen. Just in case someone else uses this function, it might be worth having a comment about needing to do this.

huss (changes requested):
I want to thank @truongdd03 for doing good work and carefully thinking through all the cases. My testing indicates it generally works fine. I have some comments in the code and two general thoughts:

- If the admin creates a suffix unit by adding/editing a unit to be suffix then you don’t see the OED created suffix units on unit page. I believe this happens because the analysis is done on the server so the Redux state is not updated. If possible, can the unit state be fetched in this case? I also suspect this happens when a new conversion is added that allows a new suffix unit. I don't know how to easily detect that on the client side without passing back the change from the server.
- Editing a unit to add suffix does not change displayable from none if an OED created suffix unit existed before. This can happen if it was set to none because the admin previously removed the suffix unit. It then does not reappear if the admin reverses this edit so not easily reversible. This is a rare, special case but maybe we should change displayable to what it is when any suffix unit is created. A negative is it will change it even if an admin previously changed the displayable after the initial creation but I think that is okay and consistent (since we did it on the original creation).

Both of these are unusual and somewhat complex so I welcome thoughts.

src/server/services/graph/handleSuffixUnits.js
truongdd03: I do a return in line 62 so it's not necessary to put an else here. Nevertheless, I removed the return and put an else here instead since it makes it clearer.

src/server/routes/conversionArray.js
truongdd03: Fixed it in the latest commit!

src/client/app/components/unit/EditUnitModalComponent.tsx
truongdd03: I got it and made the change.

src/client/app/components/conversion/EditConversionModalComponent.tsx
truongdd03: Fixed it. Thank you for the feedback!

src/client/app/actions/admin.ts
truongdd03: I didn't check the state since the loading page already prevents this function from being called multiple times. In other words, it's always called when the state is false. But to make it fit with naming standards, I put the check back.

src/server/services/graph/handleSuffixUnits.js
truongdd03: Fixed!

src/client/app/components/unit/EditUnitModalComponent.tsx
truongdd03: Fixed!

src/client/app/components/unit/EditUnitModalComponent.tsx
truongdd03: I found that the Cik table only contains visible units (displayable = ALL or ADMIN). Therefore, when displayable changes, I think we need to redoCik to be able to graph that unit. Please let me know what you think about this.

src/server/services/graph/handleSuffixUnits.js
truongdd03: This condition does not prevent OED from passing through hidden units. It only prevents OED from replacing the destination unit with a new suffix unit if the destination unit is hidden. So in your example, since `lbs` is visible and there's a path from `kg` to `lbs` (the suffix for this path is `CO2` since it goes through `kg CO2`), `lbs CO2` is still added. On the other hand, if `lbs`'s displayable is none, `lbs CO2` will not be added.

Theoretically, if displayable is none, the unit is considered nonexistent. Adding a suffix unit, which will not be used, to replace it seems to be redundant to me. That's why I added this condition. Please let me know if there's something unclear or if you have any thoughts. Thank you!

src/server/services/graph/handleSuffixUnits.js
truongdd03: This function is called when a unit's suffix is updated (not when we redoCik or refresh views). There is a comment in `src/server/routes/units.js` (line 92) where it's used. Is that what you mean?

src/client/app/components/unit/EditUnitModalComponent.tsx
truongdd03: I updated the condition here.

truongdd03:
Thank you so much for your review, @huss. I have resolved/replied to all comments. About your general thoughts:

> * If the admin creates a suffix unit by adding/editing a unit to be suffix then you don’t see the OED created suffix units on unit page. I believe this happens because the analysis is done on the server so the Redux state is not updated. If possible, can the unit state be fetched in this case? I also suspect this happens when a new conversion is added that allows a new suffix unit. I don't know how to easily detect that on the client side without passing back the change from the server.

I think we already do a refresh after editing/adding a conversion/unit, hence the redux state should be updated. When I tried to add a new suffix unit/conversion, it displays correctly on the page. Please let me know if I missed something here.

> * Editing a unit to add suffix does not change displayable from none if an OED created suffix unit existed before. This can happen if it was set to none because the admin previously removed the suffix unit. It then does not reappear if the admin reverses this edit so not easily reversible. This is a rare, special case but maybe we should change displayable to what it is when any suffix unit is created. A negative is it will change it even if an admin previously changed the displayable after the initial creation but I think that is okay and consistent (since we did it on the original creation).

I have added a step to compare the suffix unit's displayable with that of the destination unit. As you mentioned, it will also rewrite the displayable that an admin sets for that unit. But I agree with you that this is okay.

Please let me know if you find something else needed to be changed or if you have any other thoughts. Thank you so much!

src/client/app/components/unit/EditUnitModalComponent.tsx
huss: I now see the issue. src/server/services/graph/redoCik.js does several steps. src/server/services/graph/createConversionGraph.js gets all the units. src/server/services/graph/createConversionArrays.js gets unit and suffix type by excluding ones that have NONE as displayable. So, it should only change if displayable goes to/from NONE but your more general check is okay. Maybe we should add a comment if this is correct to indicate a more general check is used?

src/server/services/graph/handleSuffixUnits.js
huss: Seems fine. I overlooked that it was checking the destination unit and not the source unit. Thus, what you say makes sense.

src/server/services/graph/handleSuffixUnits.js
huss: Sorry I was not clear. What I was thinking is that a conversion is being deleted. When that happens you need update Cik and refresh the views. I agree this code is currently called for suffix unit change but I was contemplating that someone might try to use this code in another way and not do those needed steps. Thus, putting in a comment that they need to be done and are done for the current usage. Is that clearer or am I missing something?

huss:
> > * If the admin creates a suffix unit by adding/editing a unit to be suffix then you don’t see the OED created suffix units on unit page. I believe this happens because the analysis is done on the server so the Redux state is not updated. If possible, can the unit state be fetched in this case? I also suspect this happens when a new conversion is added that allows a new suffix unit. I don't know how to easily detect that on the client side without passing back the change from the server.
> 
> I think we already do a refresh after editing/adding a conversion/unit, hence the redux state should be updated. When I tried to add a new suffix unit/conversion, it displays correctly on the page. Please let me know if I missed something here.

I also tested this and could not see an issue. I recall seeing an issue but cannot reproduce so I assume it was not important/a glitch from my strange units/conversions during testing.

huss:
> > * Editing a unit to add suffix does not change displayable from none if an OED created suffix unit existed before. This can happen if it was set to none because the admin previously removed the suffix unit. It then does not reappear if the admin reverses this edit so not easily reversible. This is a rare, special case but maybe we should change displayable to what it is when any suffix unit is created. A negative is it will change it even if an admin previously changed the displayable after the initial creation but I think that is okay and consistent (since we did it on the original creation).
> 
> I have added a step to compare the suffix unit's displayable with that of the destination unit. As you mentioned, it will also rewrite the displayable that an admin sets for that unit. But I agree with you that this is okay.

Tested and works as desired. Thanks.

src/server/services/graph/handleSuffixUnits.js
truongdd03: I understand what you mean now. Thank you for the clarification. I added a comment for this function.

src/client/app/components/unit/EditUnitModalComponent.tsx
truongdd03: I agree with you that this check is more general. I added a comment to clarify here. Thanks.

huss:
Thanks to @truongdd03 for being so careful on a complex issue and doing the changes so quickly. Everything looks good and it is ready to merge.

huss (approved):

Issue #805, opened 2022-10-20
Changing database return for bar graph data


spearec:
### Is your feature request related to a problem? Please describe.

This encompasses general issues with the way bar graphs display meter data, especially when data are missing from time ranges.
This will directly fix #475 and #570, and some ideas here apply to #479 as well.

### Describe the solution you'd like

One solution is to change the database code to figure out the maximum number of bars that can fit within the requested range. These many bars are then requested from the database, and the graph is drawn as it currently is. This would prioritize recent data, so the last bar would be accurate, but some data at the beginning of the meter's range may be omitted.
For example, if a meter has data for 365 days, but the graph scale is 4 weeks (28 days), the bar graph will currently display 14 bars. However, only 13 of these bars will have data across their full range. This new method would ignore a small piece of data at the beginning of the meter readings in order to eliminate the misleading final bar.
Some edge cases to consider:
- The meter range is smaller than the requested bar range
- The graph's time range has +/- infinity as one or both of its values so all dates are desired
- Large swaths of data are missing from the meter data

### Describe alternatives you've considered

While this is more of a possible future feature than an alternative, changing the normalization method for the bar graph as a whole is something that needs to be addressed. Right now, the database normalizes an average for each day and then aggregates the days into bars. Missing days are NOT normalized (#479). This leads to a somewhat inconsistent system, especially when compared to the line graph's method of normalization. 
Adding fully normalized bars, either just for the first/last bar or for all bars, is a possibility. This would require a rewrite of both the database and the display code for bar graphs, and is something that needs to be discussed further before it may be implemented.

### Additional context

This summarizes a discussion I had with @huss about #475 and the way OED displays data for bar graphs in general.

spearec:
PR #811  implemented a version of the original change which was planned. The new code makes the last bar issue less prevalent. However, since all meters/groups need to adhere to the same bar widths, inaccurate bars may still be presented when meters with differing date ranges are displayed. The only way to accurately fix this is to have a per-meter real start and end time, something which is currently not returned by the database.
To make any more changes regarding how OED handles infrequent or missing data requires a longer conversation about whether or not data should be normalized, and if it is to be normalized, how should that be displayed. The final product may end up looking more like the alternative described above.

PR #811, opened 2022-11-02
Bar graph last bar display and line graph zoom slider fixes

### LINKED ISSUE #472: Plotly axis vs meter time endpoints
fast-pt now shows day averaged data in many circumstances. It does not include partial days at the start/end of the meter day and that makes sense. Plotly displays the time as somewhere (currently start of day but proposed to be middle of day) on a line graph. As a result, the x-axis on a line graph only includes the dates within the range returned by fast-pt. When you zoom in on the start/end of the line graph, you cannot get to any points that are beyond what fast-pt provides even though that meter data does exist. A fix might be to make the x-axis range be the min/max dates stored for a meter so you could zoom into these times (but no points would be in the new range until you redraw). If we do this we need to verify that the DB is optimized to quickly find the min/max.

### LINKED ISSUE #475: Last bar might be longer than data available
Once a bar time range is chosen, it is used for all bars (as expected). The last bar often has an end time that is beyond the actual data. This can give the wrong impression of the range of data available and also the height of the bars. I think we should either exclude the partial bar or somehow label it on the graph and/or in a help popup so users know what is going on. Line graphs do not include the partial last day (see #472 ) (compare has issues with this too) but it should all be consistent in some way. This relates to #110 about having one big bar that is larger than the available data.

### LINKED ISSUE #570: Date ranges and what maps display

### LINKED ISSUE #610: remove Plotly grid coordinate shown on map

Maps currently show the Plotly grid coordinate on the map. It has no value to the user and could be confusing.

I think we can leave this if an admin since it can be useful for debugging but remove for everyone else. The y-axis value is shown in the hover text. The x-axis is shown in a popup arrow on the axis. It would be good to get rid of both for a non-admin.

None

### Additional context

None

### LINKED ISSUE #795: line graph redraw using incorrect range when last meter is not included
### Describe the bug
When redraw is used in line graph mode after shortening the viewing window, the time range used is incorrect.
It seems that when the last selected meter is not in the applied range, the range is redrawn incorrect.
### To Reproduce

1. Import meters with different date ranges.
2. Use the selection bar to view a range which does not contain the last selected meter
3. Hit redraw
4. Observe the graph 

### Expected behavior

The viewing window should properly reflect the range shown before hitting redraw

### Screenshots

Full window. Note order of meters.
Screenshot (20)
Window before hitting redraw
Screenshot (21)
Window after hitting redraw
Screenshot (22)
Window after hitting redraw with order of meters changed - i.e working as intended
Screenshot (23)

### Machine info (please complete the following information)

 - OS: Windows running WSL
 - Chrome Version 105.0.5195.127 
 - OED Version 0.8.0, checked out with #788

spearec:
# Description
This PR addresses a number of issues, mainly surrounding the display of the last bar for bar graphs and the line graph date selection slider.
As discussed in #805, this implements an SQL query change for bar graph data which figures out the maximum number of complete bars that can be returned, prioritizing recent data. This is a change from the old code, which started from the beginning of available time, which meant that the date ranges on the last bar were often inaccurate. Since maps use the bar graph calculations, this should also fix map data often being incorrect.
The issue around #795 arose from the improper calculation of the maximum and minimum date ranges on selected meters/groups. This slightly modifies the SQL to return points in sorted order, which allows for quick calculation of date ranges. The new calculations for the ranges also avoid truncating data, so there is no loss of range when data originally displayed at a low resolution (i.e day) is increased to a higher resolution (i.e raw).

Fixes #472
Fixes #475 
Fixes #570 
Fixes #610 
Fixes #795 

Partly Addresses #805 

## Type of change

## Checklist

## Limitations

As discussed in #805, there is still work to be done on displaying the best ranges of data on a bar graph.
Also, as noted in TODOs in the code, some of the calculations in this code can be simplified once meters have defined intervals and ranges.


huss (approved):
Thanks to @spearec for addressing so many  bugs. Checking indicates that all the changes are fine.

I have pushed one commit that I would like you to look over. Basically, the x-axis values were being set by the slider calculation and that seems somewhat off at times. Since we now have the actual desired range from the readings, I changed the code to use those values. It seems to work fine but let me know what you think.

As a note, we should probably revisit the slider value storage at some later time as it has less usage if my change is accepted.

spearec:
Looks good. I did not even notice that the slider was using the stored values, this is certainly more efficient. 
I tested this new change and everything seems to be working as intended. It seems the only other use of the stored slider interval is for use in generating a chart link.

PR #822, opened 2022-11-19
Csv upload page for meters

truongdd03:
# Description

This PR:
- Converts unit and default graphic unit's ids into units on the CSV upload page for meters. It's now possible to import meters from CSV files.
- Refetches meter states when meters are uploaded via CSV files. Hence, users no longer need to refresh to see the changes.
- Brings some meter checks to the database (timezone, reading duplicates, and time sort). The timezone check makes some tests fail since they used meters with invalid timezone. This PR fixes these tests as well.
- Adds migrations for these new checks.

## Type of change

## Checklist

## Limitations


src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -203,21 +203,6 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 			// Set default identifier as name if left blank
 			state.identifier = (!state.identifier || state.identifier.length === 0) ? state.name : state.identifier;
```
huss: I think the checks for area and duplication should be put back. The message from the DB is okay for duplication but the one for area is not clear. That message is "Failed to edit meter with message: "validation failed with instance.area is not exactly one from [subschema 0],[subschema 1]"" The original checks were very clear. I also think it is better to warn here before the routing in these cases because with the DB you only see one of multiple ones possible. Finally, if any change is made then create should mirror these (they do not now since those were unchanged).

src/server/migrations/0.8.0-1.0.0/sql/meter/add_constraints.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ALTER TABLE meters ADD CONSTRAINT default_timezone_meter CHECK (default_timezone_meter IS NULL OR check_timezone(default_timezone_meter::TEXT) = true);
+ALTER TABLE meters ADD CONSTRAINT reading_duplication CHECK (reading_duplication::INTEGER >= 1 AND reading_duplication::INTEGER <= 9);
+ALTER TABLE meters ADD CONSTRAINT time_sort CHECK (time_sort::TEXT = 'increasing' OR time_sort::TEXT = 'decreasing');
```
huss, line 8: I think this also need the area check as done in create_meters_table.sql.

huss:
I have pushed two changes:

1. Change error message so clearer. This was noticed during testing but not something that @truongdd03 should have done.
2. Merged development since I had issues with login that are likely due to my changes in clearing a recent PR.

huss (changes requested):
Thanks to @truongdd03 for carefully dealing with this change for units on the CSV page. I have made a couple of comments to address. I tried to do extensive testing of loading good meters, bad meters, multiple meter in a file and using curl. No issues were found. This can be merged once the small comments are addressed.

src/client/app/components/meters/EditMeterModalComponent.tsx
truongdd03: I got it and brought these checks back.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_constraints.sql
truongdd03: I added in the latest commit.

truongdd03:
Thank you for the review, @huss. I have resolved all the comments. Please let me know if you find something that's needed to be changed. Thanks!

huss:
Thanks to @truongdd03 for the quick edits. Everything looks good.

huss (approved):

PR #825, opened 2022-11-28
change meter/group selection logic and dropdown sorting

### LINKED ISSUE #600: lose selected meter when go to map

spearec:
# Description

This PR changes the logic for determining when a meter/group is selected or deselected and changes the sorting for the dropdown menus. The new sorting has a divider to clearly show the user why incompatible meters are greyed out, as well as putting meters without GPS below those with GPS in the map view while preserving the original order.

image

Fixes #600 

## Type of change

## Checklist


src/client/app/components/ChartDataSelectComponent.tsx
```diff
@@ -225,10 +291,24 @@ export default function ChartDataSelectComponent() {
 			<div style={divBottomPadding}>
 				<MultiSelectComponent
 					options={dataProps.sortedMeters}
-					selectedOptions={dataProps.selectedMeters}
+					selectedOptions={dataProps.displayableSelectedMeters}
 					placeholder={intl.formatMessage(messages.selectMeters)}
-					onValuesChange={(newSelectedMeterOptions: SelectOption[]) =>
-						dispatch(changeSelectedMeters(newSelectedMeterOptions.map(s => s.value)))}
+					// change logic for selection: compute diff between dataProps.selectedMeters and newSelectedMeterOptions\
```
huss: Remove \ at end of comment?

src/client/app/components/ChartDataSelectComponent.tsx
```diff
@@ -212,10 +266,22 @@ export default function ChartDataSelectComponent() {
 			<div style={divBottomPadding}>
 				<MultiSelectComponent
 					options={dataProps.sortedGroups}
-					selectedOptions={dataProps.selectedGroups}
+					selectedOptions={dataProps.displayableSelectedGroups}
 					placeholder={intl.formatMessage(messages.selectGroups)}
-					onValuesChange={(newSelectedGroupOptions: SelectOption[]) =>
-						dispatch(changeSelectedGroups(newSelectedGroupOptions.map(s => s.value)))}
+					onValuesChange={(newSelectedGroupOptions: SelectOption[]) => {
+						const allSelectedGroupIDs: number[] = dataProps.allSelectedGroups.map(s => s.value);
```
huss, line 277: There are more comments in the section below but even with these I think some more would help future developers. Could you describe why each of the filters work and any other choices?

huss (changes requested):
Thanks to @spearec for more good work. I like the separators on the menus and the consistent use. I have a couple of comments on commenting but nothing else. Testing indicates it works as expected.

spearec:
I've added some more comments and changed variable names to be consistent with term usage. This makes the distinction between displayable items (an admin-level setting) and compatible items (determined by graph type/unit selection).

huss (approved):

huss:
Everything looks good. Thanks.

PR #828, opened 2022-12-01
Dealing with long time length readings

### LINKED ISSUE #665: Meter-by-meter logic for serving compressed data

Fictitious data is displayed when the underlying reading data has a lower reading frequency than the frequency that the compressed readings algorithm tries to serve. For example, if the compression algorithm tries to serve data from the now materialized hourly view when the underlying data is 4-hourly data. This compression serves fictitious data because there is not actually any hourly data; instead a constant or linear approximation is applied. As the algorithm is now the only "true" compressed data would be from the daily view.

One idea to resolve this would be to track the reading frequency onto the meter as a new column. This could either be provided from the administrators when they first create the meters or be estimated from the first batch of readings from the data. Then the compression algorithm would determine the appropriate level of compression for serving data for certain time intervals.

OED does not indicate whether the data displayed is compressed or not. We could instead visually indicate the level of compression to the user when they read the graph. This would be less misleading since they should also know the actual reading rate of the meter and understand that the compression is not displaying real data. 

### Additional context
With the addition of PR #660 this only becomes an issue if the meter reads slower than 1 reading per hour. For electricity, this is less of an issue since most meters read data every 15-minutes. As OED generalizes this might become an issue for certain energy sources.

### LINKED ISSUE #806: dealing with low frequency readings

If a meter only acquires data at longer time frames then it is possible OED will graph it at a higher frequency by adding points. An example would be a water meter from the utility bill that only had a data point each month. OED would graph it daily and add 28-31 points for each day between the actual month points. This can happen, in general, if the reading frequency is longer than 1 hour since we have hourly and daily views.

The DevDocs longReadings.md discusses this in detail and proposes a solution. Basically, we use raw/meter reading when possible and then go to hourly and then daily.

Unsure but other ideas are possible

### Additional context

This will address issue #519 that was partly addressed when the minutely readings were removed so it is being closed.

kirtmd:
# Description

Dealing with long time length readings

Fixes #665, Fixes #806.

Changed the way that meter data is returned, previously OED would use daily and hourly readings before going to raw readings which lead to showing readings that didn't exist. With this code, OED starts by checking raw readings, then hourly and daily last to make sure that these issues don't persist. We added a reading frequency column to the meters table to help us do this. The function loops through each meter that is input because we need to know the frequency of each meter specifically to return it's proper reading interval.

## Checklist

## Limitations

Issue #665 is still not completely finished, this covers the meter section of the issue but the group part is still TODO, you can read more about it here: https://github.com/OpenEnergyDashboard/DevDocs/blob/main/longReadings.md


src/server/migrations/0.8.0-1.0.0/sql/meter/add_meter_reading_frequency.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Add unit_id and default_graphic_unit to the meters table.
```
huss: I think this comment was copied from another file and needs to be updated.

src/server/migrations/0.8.0-1.0.0/sql/meter/add_meter_reading_frequency.sql
```diff
@@ -0,0 +1,7 @@
...
+ALTER TABLE meters
+    ADD COLUMN IF NOT EXISTS meter_reading_frequency INTERVAL;
```
huss: There is a default value in the meters creation SQL that should also be here.

src/server/sql/reading/create_reading_views.sql
```diff
@@ -209,84 +210,87 @@ CREATE OR REPLACE FUNCTION meter_line_readings_unit (
...
 	RETURNS TABLE(meter_id INTEGER, reading_rate FLOAT, start_timestamp TIMESTAMP, end_timestamp TIMESTAMP)
-AS $$
+AS $$ 
 DECLARE
 	requested_interval INTERVAL;
 	requested_range TSRANGE;
 	unit_column INTEGER;
-BEGIN
+	frequency INTERVAL;
+	add INTEGER := 1;
```
huss: Since add is the current meter id, should the name reflect that. meter_id is okay but seems close to meter_ids. Open to ideas but what about current_meter_id?

src/server/sql/reading/create_reading_views.sql
```diff
@@ -209,84 +210,87 @@ CREATE OR REPLACE FUNCTION meter_line_readings_unit (
...
-			FROM ((((readings r
-			INNER JOIN unnest(meter_ids) meters(id) ON r.meter_id = meters.id)
-			INNER JOIN meters m ON m.id = meters.id)
-			INNER JOIN units u ON m.unit_id = u.id)
-			INNER JOIN cik c on c.row_index = u.unit_index AND c.column_index = unit_column)
-			WHERE lower(requested_range) <= r.start_timestamp AND r.end_timestamp <= upper(requested_range);
-	 END IF;
+	-- Start with the raw, then hourly and then daily if others will not work
+
+	while add <= cardinality(meter_ids) loop
```
huss: While you nicely put comments below, I think some more, esp. on what the next few lines are doing would help future developers.

src/server/sql/reading/create_reading_views.sql
```diff
@@ -209,84 +210,87 @@ CREATE OR REPLACE FUNCTION meter_line_readings_unit (
...
+					((r.reading / (extract(EPOCH FROM (r.end_timestamp - r.start_timestamp)) / 3600)) * c.slope + c.intercept) 
+				WHEN (u.unit_represent = 'flow'::unit_represent_type OR u.unit_represent = 'raw'::unit_represent_type) THEN
+					-- If it is flow or raw readings then it is already a rate so just convert it but also need to normalize
+					-- to per hour.
+					((r.reading * 3600 / u.sec_in_rate) * c.slope + c.intercept)
+				END AS reading_rate,
+				r.start_timestamp,
+				r.end_timestamp
+				FROM (((readings r
+				--INNER JOIN unnest(current_id) meters(id) ON r.meter_id = meters.id)
```
huss: Should this commented out line be removed?

src/server/sql/reading/create_reading_views.sql
```diff
@@ -209,84 +210,87 @@ CREATE OR REPLACE FUNCTION meter_line_readings_unit (
...
-			INNER JOIN meters m ON m.id = meters.id)
-			INNER JOIN units u ON m.unit_id = u.id)
-			INNER JOIN cik c on c.row_index = u.unit_index AND c.column_index = unit_column)
-			WHERE lower(requested_range) <= r.start_timestamp AND r.end_timestamp <= upper(requested_range);
-	 END IF;
+	-- Start with the raw, then hourly and then daily if others will not work
+
+	while add <= cardinality(meter_ids) loop
+		select meter_reading_frequency into frequency from meters where id = meter_ids[add];
+		IF (((EXTRACT (DAY FROM requested_interval) * 1440) / (EXTRACT (HOUR FROM frequency) * 60 + EXTRACT (MINUTE FROM frequency))) <= 1440)
```
huss: I cannot convince myself that the logic precisely does what is wanted but is close. It seems to assume that the requesteed_interval is an even number of days and I don't think that is the case. I think this since I thought EXTRTACT DAY only returns whole days. I also think we could have an issue with frequency if someone entered it in the general Postgres syntax of '4 days' rather than '96:00:00' since it does not deal with days. A quick test found this did not work with the current code.

To do this in a general way is probably desirable given that there are variety of ways to enter these values into Postgres. I'm hoping there is a way to do this but I have not investigated.

huss (changes requested):
Thanks to @kirtmd for putting in their first PR. I have left some comments to think about and address.

I did modest testing and found this generally works as desired. More substantial testing will be needed after it is finalized and before merging.

I also noticed it has conflicts with development over other changes that need to be resolved. I can help with that if you want.

Also see a comment on this PR (about to do) about other changes needed in OED before this can be deployed.

huss:
Working with this PR and thinking some more lead me to realize that the following changes are needed in OED before this can be deployed:

- Changing the way meters gets data impact groups (that use the new code for each meter in a group) so the group code needs to be updated. It needs to be the case that all meters are getting readings at the same frequency so they can be summed. We have to decide which of the possible meter frequencies to use but the one with the fewest points would be the conservative choice. It also needs to exclude raw. All of this will likely mean changes/new parameters for the line meter function.
- The meter pages need to be changed to allow interaction with the meter frequency of reading.
- The CSV meter import needs to work with the meter frequency of reading (might work okay but need to check).
- Many automated test cases that looks at line readings returned will need to be updated to reflect the new values returned.
- The test data and website data need to be updated to include the correct frequency of reading for each meter. This will also require modifying the meter input code to handle this.

Given that OED v1.0 should happen soon and the changes should be done before this PR is included in an OED release, this PR is on hold for merging to make sure either it happens after v1.0 or all changes are in place before merging.

Please note this does not reflect on the work done for the PR but issues outside of this work that must be accomplished. This PR can still progress with changes and be ready to merge once all the other work is done.

huss:
I have worked on the necessary changes. The important ones to note are:

- The comments on code in the PR are all addressed.
- Each meter correctly calculates the number of points for its readings in DB call. In the past it was for all meters.
- The DB now uses passed parameters to determine when to switch between raw, hourly and daily. It uses 1440 points for now.
- Groups now work with the new code where all meters return the same frequency so they can be added and raw is excluded.
- The preferences page allows setting of a default meter frequency reading. The old environment variable is removed.
- The meter page allows setting of the reading frequency.
- Changes integrated in all graphing pages.
- All meter creation and tests were updated for the changes in this PR.
- Changed so meter create/update returns the id and reading frequency. This is routed back to the client to set state. This means a reload of all meter state is no longer needed.
- A bug was fixed that messed up the meter unit choice on the meters page.
- A bug was fixed where line state was sometimes unset and caused an error.
- The test data for developers and web were updated for the new code. This includes changing the meter loading of test data to be by key:value pairs. The data now sets the reading frequency & area but not area for web since need to figure out what to do. Note groups have not been converted so their area is not set.

Note that a default of 15 minutes is used for reading frequency. A lower value was used since it will return fewer points and be faster if the admin does not set.

With these changes I think this PR is ready to consider for merging into OED 1.0.

spearec (approved):
I've looked over the code and everything looks good. I don't have a perfect understanding of the SQL logic used, but the comments were enough to give me a passable understanding of what is going on.

I wasn't able to test the changes much, but simply graphing the test4DaySin meter in OED is enough to immediately see that the changes seem to fix the issue at least in this case.

The only thing I noted that might be an issue is that reading_line_accuracy is not defined in a migration, but I'm not sure if this is needed depending on when the enum is created in OED.

huss:
> The only thing I noted that might be an issue is that reading_line_accuracy is not defined in a migration, but I'm not sure if this is needed depending on when the enum is created in OED.

Thanks for catching that. A coming commit fixes it.

huss:
Thanks to @spearec for jumping in and doing this review. I am taking it as approval. I should have addressed the one comment. It should be ready to merge.

Issue #698, opened 2022-01-20
improve data tests to do multiple values


huss:
### Is your feature request related to a problem? Please describe.

There are many places that OED tests that a certain value matches the expected value. While this is a good test, it would be better if it was done for multiple values. Examples include:

- src/server/test/routes/compressedReadingsRouteTests.js seems to test a single value.
- src/server/test/db/readingTests.js tests 1 or 2 points
- src/server/test/db/newCompressedReadingsTests.js tests limited points
- src/server/test/db/compareTests.js does a couple of points
- src/server/test/db/barchartReadingsTests.js does a few points

These tests also tend to have readings with 1 hour length.

### Describe the solution you'd like

OED now allows for readings of arbitrary length. Thus, tests should try reading lengths other than 1 hour. The automated test data (see src/server/data/automatedTestingData.js) includes 23 minute data for this reason. The other reason 23 minutes was chosen is that readings cross the hour and day boundaries so are better tests of the compression algorithms. We should be using automated tests to do this type of testing.

Since we can generate multiple points, it would be better to loop over a number of points and test each one in an automated fashion rather than doing a few specialized points.

The graphing is now done with compressed data that can be daily, hourly and original (raw) meter data. We should be testing the values returned from these to be sure they are correct (already done to some extent).

We should more generally test that groups properly add the included meters. The automated data for sine^2 + cosine^2 that gives a constant value is a good example of how this could be done.

### Describe alternatives you've considered

We can continue to test limited number of points and for easy time frames for those points and do the rest manually but that is not what we want.

### Additional context

None.

huss:
The [testing designDocs](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/testing/testing.md) gives more tests to be added to src/server/routes/readings.js. At this time they are line and bar reading tests with more to come. The section "Expected values" has tables for line and bar tests that indicate which have yet to be done. Please communicate with the OED project before starting these tests so duplicate efforts are not made. Note this is a reserved tagged issue since it is normally done by certain students and not for others to complete.

PR #830, opened 2022-12-02
Update to Readings API testing

MikeLimpus:
# Description
This PR overhauls src/server/test/readings.js to enable tests of multiple values using CSV files generated with a spreadsheet. Primarily, it introduces a set of functions to form a testing jig which should accelerate and simplify writing tests for the readings web API. These functions: 

- Insert units, conversions, meters, and optionally groups into the database using the new functions in PR #827 
- Read and parse CSV files to prepare the expected readings that Chai will compare against the actual readings
- Compare readings from the API against a CSV file
- Generate timestamps to simplify the API call query parameters

This jig is intended to allow other developers to create tests rapidly. It may also be used as a basis for other test files to handle multiple data points in testing.

Some tests are written using the jig, but there are many more tests that should be written, as well as corresponding CSV files for these tests. 

The previous tests were also expanded and given their own describe block in the test file.  

Partly Addresses #698

## Type of change

This change correlates with the testing DevDocs update

## Checklist

## Limitations

More tests should be written using the expected values CSVs from the DevDocs. The tests should check varied reading intervals (15 minutes is the only interval tested currently). Varied unit types should be tested. Only line meters are tested right now; bar charts and groups should also be tested. These are all contingent on the creation of new CSV files and should work with the test jig.


MikeLimpus:
Currently failing test script, possibly due to the database being modified by another test. The tests pass when run on their own.

MikeLimpus:
src/server/test/routes/unitReadingsRouteTests.js was using sinon to modify the methods which make up the backend of the readings API without restoring them. This should be resolved now. Recommend a pass is made in the test codebase to ensure that any other use of sinon stubs are restored to ensure this doesn't happen again.

huss:
I updated the PR for a couple of items rather than leaving them to @MikeLimpus due to how long all this took:

1.  Uncommented the flow/raw tests in the code because I added the needed expected CSV files.
2. Merged in development (thanks to @MikeLimpus for doing it recently) to get a very recent update and only list the actual files changed in the PR.

The "successfully creates a user" in usersTest.js was failing. This had nothing directly to do with this PR but began showing up because of changes made. I worked on them and they got mixed in with this PR code so are included. While looking into why, I addressed several items:

1.  A logic issue that almost never arises where an unusual end of the test left the DB in an unusual state and then the next time failed if this was the only test run. Fixed with note in code.
3. Put getConnection() at start of tests as described. This may not be necessary.
4. Fixed bug with missing await in OED production code that was main cause of test failing.
5. Don't ignore duplicate users where create one. map tests had minor issues with the code because this was done. Updated how users are created to test if exist so this error should not occur and will not catch it if it does.
6. Routed error and displayed to user when they have an error in creating a user. This is mostly due to a duplicate user.
7. Changed error notifications to stay for 15 seconds instead of 3. I think it is better to give the user time to see these and they can close sooner if they want. Success notifications are still only 3 seconds.
8. Changed TranslatedString type from a combination of two types to only string. This makes the code easier to deal with. As far as I can tell this does not cause an issue. I left a TODO to simplify the type if this is not an issue that anyone sees.
9. User getByEmail() now returns null if there is no user rather than generating an error. This mimics other places in the code for getting a single item. Updated uses to be compatible and made messages output clearer that this was the reason.

huss:
Thanks to @MikeLimpus for their first PR and doing very nice work in addressing this issue. The code worked fine expected for unexpected issues outside this work. Thanks for fixing the sinon issue. I searched the code and did not find any more uses.

I'm sorry it took so long to finish this but figuring out the issues outside what was the original PR took some time.

This PR is ready to merge.

huss (approved):
Everything is working fine. The other comments address issues that were solved.

Issue #769, opened 2022-04-27
fix JSDoc in code base


huss:
### Is your feature request related to a problem? Please describe.

PR #767 added the ability to run JSDoc linting on the code. When enabled, it finds a lot of issues with the current documentation. We cannot turn this check on until the issues are resolved.

### Describe the solution you'd like

We need to go through all the issues and add the needed documentation. Note anyone is welcome to work on any number of issues and it is not necessary to do them all. See .eslintrc.json for two lines to uncomment to get the checks to happen when you run 'npm run check'.

### Describe alternatives you've considered

We continue to have substandard documentation.

### Additional context

Documentation is awesome :-)

spearec:
I have currently fixed all existing jsdoc errors in my local branch (https://github.com/spearec/OED/tree/jsdoc). After some discussion with @huss, these changes will possibly be folded into some other changes to which ESLint packages OED uses. This will probably happen during the 1.0 release phase, as it is easier to make this transition on a frozen codebase.

In the meantime, please make sure to add jsdoc comments to any new functions you create, as this will minimize the migration process for this issue.

PR #831, opened 2022-12-03
Fix JSDoc linting errors in getGraphColor.ts and two more files plus fix GitHub Actions

Bedolla233:
# Description

partially addresses JSDoc in files:
- getGraphColor.ts
- determineCompatibleUnits.ts
- hasPermissions.ts

Partly addresses #769 

## Type of change

## Checklist

## Limitations
Only fixed JSDoc in a few files.


src/client/app/utils/determineCompatibleUnits.ts
```diff
@@ -115,8 +120,9 @@ export function unitFromPRow(row: number): number {
 
 /**
  * Returns the unit id given the column in Pik.
+ *
  * @param {number} column The column to find the associated unit.
- * @returns {number} The unit id.
+ * @returns {number} Unit id.
```
huss: The modified comment on the equivalent row function has "given the row in Pik units" so I would add the similar text here.

Also, as long as changes are being made, I would remove "units" from Pik so it reads "Pik" instead of "Pik units". This would be hard for you to know that so thanks for making the change. This should be consistend in all comments.

src/client/app/utils/hasPermissions.ts
```diff
@@ -5,10 +5,11 @@
 import { UserRole } from '../types/items';
 
 /**
- * Checks if the user has the permissions of a given role.
- * @param {UserRole} user User role to evaluate
- * @param {UserRole} compareTo User role to compare to
- * @returns {boolean} Whether or not the user has the compareTo role
+ * Checks if the role A has the permission of the role B.
+ *
+ * @param {UserRole} userA variable for role A
```
huss: I like these changes but could we change:
variable for role A/B
to
role for A/B
I suggest this since, in general, we don't put the word "variable" in an @param since most describe variables.

huss (changes requested):
Thanks to @Bedolla233 for a second PR. Overall it looks good. I just made a couple of suggestions. Please let me know if you have thoughts or questions.

src/client/app/utils/hasPermissions.ts
Cao1224: Hello @huss . Sorry for the late reply. Since @Bedolla233 didn't modify it based your comment, I, as his teammate, helped him modify it. Thank you for your patience and comments.

huss: I updated the PR a little. This was mostly due to another PR also changing the same file/comments. The PR is now fine/done from what you did. I am currently seeing if I can get it to pass the GitHub action testing as it started to fail. This has nothing to do with the changes you made.

huss (approved):

huss:
This PR is ready to merge. I thank the developers for their efforts and contribution. I'm sorry it took a while to get this done.

Please note that most of the commits are my many attempts to get GitHub actions to work again as this just happened to be the first PR to fail. This does not reflect in any way on the good work that @Bedolla233 or @Cao1224 did. This really isn't part of this PR but it became part of it during my fixes. The main changes for GitHub Actions that I added are:

- Go to v3 for checkout and update syntax used.
- Update to later node so same as normal OED usage.
- Go to v2 for codeql-actions as needed for v1 about to be eliminated.
- Have the automated test data files go to the tmp directory rather than the associated code directory. Since they should be deleted this should not impact anyone but it helps make the GitHub Actions easier and is better practice.

PR #834, opened 2022-12-04
Multiple export files

### LINKED ISSUE #599: enhance data export

DylanHill54:
# Description

Changed export graph so it exports unique files for each meter of the state graph. it also now passes the correct unit for the graph if you export the graph. For raw export it also now gives unique files for each meter and includes the native unit for each meter in the export. Also altered the logic of the file size check for download to be more accurate to the changed file format when estimating size. Also changed code so that it now uses Meter.ID whenever possible 
@bryanFowles @DylanHill54

See below for changes by @huss.

Fixes #599

## Type of change

## Checklist

## Limitations


src/client/app/components/ExportComponent.tsx
```diff
@@ -9,12 +9,18 @@ import graphExport, { graphRawExport, downloadRawCSV } from '../utils/exportData
 import { ExportDataSet } from '../types/readings';
 import { FormattedMessage } from 'react-intl';
 import { TimeInterval } from '../../../common/TimeInterval';
-import { metersApi } from '../utils/api'
+import { metersApi} from '../utils/api'
```
huss: There are formatting issues in the file. Please use the automatic file formatting in VSC (command palette, the format document) to resolve. Probably should check all files.

src/client/app/components/ExportComponent.tsx
```diff
@@ -9,12 +9,18 @@ import graphExport, { graphRawExport, downloadRawCSV } from '../utils/exportData
...
 import TooltipMarkerComponent from './TooltipMarkerComponent';
+import { State } from '../types/redux/state';
+import { useSelector } from 'react-redux';
+
```
huss, line 21: There are extra blank lines. There should only be one.

src/client/app/components/ExportComponent.tsx
```diff
@@ -27,59 +33,80 @@ export default function ExportComponent(props: ExportProps) {
 	 * Called when Export button is clicked.
 	 * Passes an object containing the selected meter data to a function for export.
 	 */
+	// Meters state
+	const MetersState = useSelector((state: State) => state.meters.byMeterID);
```
huss: Since this is a variable and to be consistent with units, do not capitalize the first word in camelcase.

src/client/app/components/ExportComponent.tsx
```diff
@@ -27,59 +33,80 @@ export default function ExportComponent(props: ExportProps) {
...
+	// Units state
+	const units = useSelector((state: State) => state.units.units);
+
 	const exportReading = () => {
-		const data = props.exportVals.datasets;
+		const data = []
+		//using a loop to push objects one at a time into data to be organized and formated for export
+		for(let i = 0; i < props.exportVals.datasets.length; i++){
```
huss: The current coded seems a bit convoluted and I think it is probably good to fix up. It pushes props.exportVals.datasets[i] into an array, uses it and then uses slice to get it out of the array. As a result, there is only one dataset in the array at any time. Given this, I'm not sure why it should be pushed into another array rather than setting the current index in datasets to a variable and using that. It will mean removing the array index from all the referencing code. I think the code is trying to mirror the old code but that is not important to do here. Note the date/time stamp on each exported file may vary since the readings may span different time ranges but I think that is good. The old code used the first one and that does not seem a good choice.

src/client/app/components/ExportComponent.tsx
```diff
@@ -27,59 +33,80 @@ export default function ExportComponent(props: ExportProps) {
...
-		// Use regex to remove commas and replace spaces/colons/hyphens with underscores
-		const startTimeString = startTime.utc().format('LL_LTS').replace(/,/g, '').replace(/[\s:-]/g, '_');
-		const endTimeString = endTime.utc().format('LL_LTS').replace(/,/g, '').replace(/[\s:-]/g, '_');
-		const chartName = data[0].currentChart;
-		const name = `oedExport_${chartName}_${startTimeString}_to_${endTimeString}.csv`;
-		graphExport(data, name);
 	};
 
 	const exportRawReadings = async () => {
 		if (props.selectedMeters.length === 0)
```
huss: This was not your doing but I'm now noticing. Can you please add {} around the if statement even though it is one line. OED prefers that.

src/client/app/components/ExportComponent.tsx
```diff
@@ -27,59 +33,80 @@ export default function ExportComponent(props: ExportProps) {
...
 			return;
-		const count = await metersApi.lineReadingsCount(props.selectedMeters, props.timeInterval);
-		graphRawExport(count, props.defaultWarningFileSize, props.defaultFileSizeLimit, async () => {
-			const lineReading = await metersApi.rawLineReadings(props.selectedMeters, props.timeInterval);
-			downloadRawCSV(lineReading, props.defaultLanguage);
-		});
+		const data: number[] = []
```
huss: See above about using an array as I think it also applies here but I note you do use the index to get the correct props as I would expect above.

src/client/app/containers/ExportContainer.ts
```diff
@@ -9,6 +9,7 @@ import { ChartTypes } from '../types/redux/graph';
 import { BarReading, LineReading, ExportDataSet } from '../types/readings';
 import { State } from '../types/redux/state';
 
+
```
huss: Please remove the extra line that was added.

src/client/app/containers/ExportContainer.ts
```diff
@@ -47,7 +48,8 @@ function mapStateToProps(state: State) {
 						label,
 						id: state.groups.byGroupID[groupID].id,
 						currentChart: chart,
-						exportVals: dataPoints
+						exportVals: dataPoints,
+						unit:  state.units.units[parseInt(state.graph.selectedUnit.toString())].name
```
huss: This applies to all the similar uses in this file. The code takes state.graph.selectedUnit, converts to a string and then parses back to an int. Given state.graph.selectedUnit is a number, I'm unclear on why this is being done. I also note that it is a constant value so really should not be attached to the group/meter data as done here. I suspect it can be removed and then the selectedUnit can be gotten from state in ExportComponent using React hooks.

src/client/app/components/ExportComponent.tsx
```diff
@@ -9,12 +9,18 @@ import graphExport, { graphRawExport, downloadRawCSV } from '../utils/exportData
...
+import { State } from '../types/redux/state';
+import { useSelector } from 'react-redux';
+
+
+
 
 interface ExportProps {
 	showRawExport: boolean;
 	selectedMeters: number[];
+	units: string;
```
huss: I don't see where this prop is used. It seems the code uses the state to get the units using the new React hooks. Given this, I suspect this can be removed.

src/client/app/containers/ExportContainer.ts
```diff
@@ -145,6 +150,7 @@ function mapStateToProps(state: State) {
 		showRawExport: state.graph.chartToRender === 'line' ? true : false,
 		selectedMeters: state.graph.selectedMeters,
 		selectedGroups: state.graph.selectedGroups,
+		units: state.graph.selectedUnit.toString(),
```
huss: There is only one unit so I'm not sure this should be plural. Other comments may mean it is going away.

src/client/app/components/ExportComponent.tsx
```diff
@@ -9,12 +9,18 @@ import graphExport, { graphRawExport, downloadRawCSV } from '../utils/exportData
...
 import TooltipMarkerComponent from './TooltipMarkerComponent';
+import { State } from '../types/redux/state';
+import { useSelector } from 'react-redux';
+
+
+
 
 interface ExportProps {
 	showRawExport: boolean;
 	selectedMeters: number[];
```
huss: This is a comment about the code that existed before your changes that I am noticing during my review. First, selectedGroups that is returned by ExportContainer is not used here. That was a sign to me that selecedMeters may not be needed. Looking at the code, the datasets contain the meter and group data (why the code loops over that below) and each entry has the id. (I note that raw uses the selectedMeters length which was okay since raw does not export groups.) The other uses seem to allow for getting the values from the datasets. Anyway, the state has these values so why pass via props given Redux hooks. Given all of this, I think the code should be updated to fix the issues and make it better. This is outside your changes so I suspect you will want someone else to do them.

src/client/app/utils/exportData.ts
```diff
@@ -9,23 +9,25 @@ import * as moment from 'moment';
 import { UserRole } from '../types/items';
 import translate from './translate';
 
+
```
huss: Please remove all the extra blank lines.

src/client/app/utils/exportData.ts
```diff
@@ -66,8 +68,8 @@ export default function graphExport(dataSets: ExportDataSet[], name: string) {
  */
 // below comment should be removed when we either remove defaultLanguage or implement it into the following function
 /* eslint-disable @typescript-eslint/no-unused-vars */
-export function downloadRawCSV(items: RawReadings[], defaultLanguage: string) {
-	let csvOutput = 'Label,Readings,Start Timestamp,End Timestamp\n';
+export function downloadRawCSV(items: RawReadings[], defaultLanguage: string, unit: string) {
```
huss: It is interesting that unit is passed here but in the non-raw version it gets it in a similar way to meter with items[0]. This is something that should be looked into.

src/client/app/utils/exportData.ts
```diff
@@ -77,12 +79,13 @@ export function downloadRawCSV(items: RawReadings[], defaultLanguage: string) {
 		// If we switch to the new route, we should remove this warning if we do the formatting here.
 		const startTimestamp = moment(ele.startTimestamp).format('dddd LL LTS').replace(/,/g, ''); // use regex to omit pesky commas
 		const endTimestamp = moment(ele.endTimestamp).format('dddd LL LTS').replace(/,/g, ''); // use regex to omit pesky commas
-		csvOutput += `"${ele.label}",${ele.reading},${startTimestamp},${endTimestamp}\n`; // TODO: add column for units
+		csvOutput += `"${ele.reading},${startTimestamp},${endTimestamp}\n`; // TODO: add column for units
```
huss: The TODO should be removed as the unit is done above.

src/client/app/utils/exportData.ts
```diff
@@ -96,7 +99,7 @@ export function downloadRawCSV(items: RawReadings[], defaultLanguage: string) {
 // NOTE: This function is made with the idea that it will not be called very often
 // Ideally we would have a component that prompts the user and handles all the logic
 export async function graphRawExport(count: number, warningFileSize: number, fileSizeLimit: number, done: () => Promise<void>): Promise<any> {
-	const fileSize = (count * 0.0849 / 1000);
+	const fileSize = (count * 0.0857 / 1000);
```
huss: I have not checked this but it was unexpected to me that the fileSize would get bigger given that we no longer include the id in each row. It may be correct but I want to check.

huss (changes requested):
I thank @bryanFowles and @DylanHill54 for their first PR and addressing this issue. I've left a number of comments. Given some are general (even about previous work on export), you are welcome to discuss them with me to see how best to address them. I'm also happy to answer any questions you may have about the comments.

I did some basic testing but will not do extensive testing until changes are made. I note the following:

1. When a group is exported, the CSV has the word meter associated with it (as in meter identifier). I believe this is because meters and groups are mixed in the datasets. I don't know of an easy solution without separating the data (non-trivial change) so we might want to rename the item so it is clearer. Something to discuss to decide the best text.
2. There is a double quote (") at the start of readings in raw export in the CSV.
3. The issue that a meter's unit is not one that users normally see was known and is still an issue here. We need to come up with a scheme to put in a graphing unit or something appropriate for this. This is not part of the work of those doing this PR but is noted for completeness.

src/client/app/utils/exportData.ts
huss: I checked this out and your estimate was pretty good. I update a little based on more testing.

src/client/app/utils/exportData.ts
huss: With my standardizing how file names are created and other info this is no longer an issue.

huss:
I have made a number of changes that are documented in the commit messages. The overview of significant changes is:

- All the comments about this PR are addressed.
- The unit given for meters is now the graphing one (when possible).
- Converted to React hooks so no container.
- Works correctly when user graphs different rates.
- Can import exported readings. Thus, the format for the CVS changed.
- The logic was simplified and extra steps were removed from older code.
- The route now sends less data.
- The end date given is the end of the last item rather than the start.

src/client/app/translations/data.js
```diff
@@ -74,7 +74,7 @@ const localeData = {
 		"csv.common.param.header.row": "Header Row",
 		"csv.common.param.update": "Update",
 		"csv.download.size.limit": "Sorry you don't have permissions to download due to large number of points.",
-		"csv.download.size.warning.size": "File size will be about",
+		"csv.download.size.warning.size": "File(s) size will be about (usually within 10% for large exports).",
```
spearec: This sounds a little weird, maybe "Total file size" or "File size(s)" instead?

spearec (commented):
PR looks good.
I was unable to get an exported file properly imported into OED. The import worked, but had no data. I don't know if this is an issue with the code or just with my lack of knowledge on how to do this (the documentation isn't great on this). Either way, something to note here.

src/client/app/translations/data.js
huss: Thanks for the careful reading. I agree the wording was clunky. I changed it to "Total size of all files will be about (usually within 10% for large exports)." I hope this is okay with you.

huss:
As for importing and exported file, the help pages are not yet up to date with the 1.0 changes. There are also some quirky behavior that will be address later. I assuming you created the meter when you imported the readings. (If not then let me know as something else is going on.) As with several places when OED automatically creates a meter, it cannot set all the needed values so an admin must do it. I hope these steps will work for you:

1. Import on the CSV readings page with header row checked. Don't bother refreshing here since it does no good.
2. Go to the meter page and refresh. As of now, OED does not add new meters to the state so this is needed.
3. Edit the new meter to set the units. It is a little faster to graph if you also set the default graphic unit.
4. Unfortunately, meter edits do not kick the server/state. It was fixed for units & conversions but I forgot about this. Thus you need to open a shell on the web container and do: updateCikAndViews.

This gave me back the same graphic as what I exported assuming you exported the raw readings. Let me know if this works for you.

spearec (approved):
Following those steps, I was able to properly re-import the meter. I believe the issue was the updateCikAndViews part.

Since this was an issue on my part and not with the code, I have no further issues.

spearec:
Following those steps, I was able to properly re-import the meter. I believe the issue was the updateCikAndViews part.

Since this was an issue on my part and not with the code, I have no further issues.

huss:
review by @spearec is being used to merge this PR.

PR #851, opened 2023-02-03
Fix previousEnd type validation when editing a meter

### LINKED ISSUE #850: previousEndtime fails if not already set
### Describe the bug

The added previousEnd to deal with DST does not work properly if you edit the meter and the value was set to null in the DB.

### To Reproduce

1. Go to meter page as admin.
2. Edit a meter where previousEnd was set to null. The test meters for developers seem to do that.
3. Make any change but leave previousEnd alone.
4. When you save the edits you get an error.

### Expected behavior

The null/empty value for previousEnd should be allowed to be unedited and still allow it to be saved back to the DB.

### Additional context

None.

spearec:
# Description
The meters generated by the testData script have a previousEnd value of undefined in the database. This means that when you go to edit one of the meters, you must set a previousEnd time in order to edit the meter. This PR fixes that, by allowing undefined as a valid type.
This issue only appears to show up when importing data in this way, as adding a new meter via the OED web interface properly initializes the field.

Fixes #850 

## Type of change

## Checklist


huss (approved):
Thanks to @spearec for fixing this annoying issue. The code works as expected.

I just note that the empty previousEnd becomes Unix zero time when you edit a meter without setting this value. This is the DB default value. You don't see that on the admin meter page until you refresh. I think this is fine as both really mean the same thing. I'm just noting so people are aware.

PR #857, opened 2023-02-11
Route messages using response.js

### LINKED ISSUE #793: consistently route return messages

The new routes for meters.js in src/server/routes use success/failure functions to return a string about what happened on add & edit. unit.js is doing it directly without using the functions. conversion.js is doing it either directly or not at all.

We should be consistent. It seems best to use the success/failure functions.

Leave as is which works but isn't optimal.

### Additional context

We could look at other routes to see if we should also do it there.

mosguinz:
# Description

Updated add and edit endpoints in unit.js and conversion.js to use the `success` and `failure` functions.

Fixes #793.

## Type of change

> This PR should not require any major update.

## Checklist

## Limitations

N/A


huss:
I saw this is a draft but I took a quick look. First, thanks for taking on this issue and your first work for OED. Couple of thoughts:

1. You may be planning this but I'll note that it would be good to display the messages returned as is done for meters. See src/client/app/actions/meters.ts in submitEditedMeter() for an example.
2. @mosguinz has a couple of commits in this PR but I don't see their agreement to the OED CLA. Doing that will allow us to accept that work. Thanks and let me know if you have any questions about this.

We look forward to having the final, non-draft version of the PR.

src/server/routes/conversions.js
```diff
@@ -33,6 +34,7 @@ router.get('/', async (req, res) => {
  * Route for POST, edit conversion.
  */
 router.post('/edit', async (req, res) => {
+	console.log("Editing...")
```
chanhenry54: Probably want to remove this since it's mainly just for your own debugging purposes

src/server/routes/conversions.js
```diff
@@ -121,7 +123,7 @@ router.post('/addConversion', async (req, res) => {
 	const validationResult = validate(req.body, validConversion);
 	if (!validationResult.valid) {
 		log.error(`Invalid input for ConversionsAPI. ${validationResult.error}`);
-		res.sendStatus(400);
+        failure(res, 400, `Invalid input: ${validationResult.error}`);
```
chanhenry54: I'd probably do `"Invalid input: " + validationResult.error.toString()` just to stay consistent with the other failure calls, but I'll leave that up to the official reviewer

src/server/routes/conversions.js
```diff
@@ -139,7 +141,7 @@ router.post('/addConversion', async (req, res) => {
 			res.sendStatus(200);
 		} catch (err) {
 			log.error(`Error while inserting new conversion ${err}`, err);
-			res.sendStatus(500);
+            failure(res, 500, `Error while inserting new conversion: ${err}`);
```
chanhenry54: Same as the previous comment

src/server/routes/units.js
```diff
@@ -83,7 +83,7 @@ router.post('/edit', async (req, res) => {
 	const validatorResult = validate(req.body, validUnit);
 	if (!validatorResult.valid) {
 		log.warn(`Got request to edit units with invalid unit data, errors:${validatorResult.errors}`);
-		res.status(400);
+		failure (res, 400, 'validation failed with ' + validatorResult.errors.toString());
```
chanhenry54: Remove the space between failure and (

chanhenry54: Correct me if I'm wrong, but I don't think I see the success/failure functions being imported anywhere from ./response?

src/server/routes/units.js
```diff
@@ -104,9 +104,9 @@ router.post('/edit', async (req, res) => {
 			await unit.update(conn);
 		} catch (err) {
 			log.error('Failed to edit unit', err);
-			res.status(500).json({ message: 'Unable to edit units.', err });
+			failure (res, 500, 'Unable to edit units ' + err.toString());
```
chanhenry54: Remove the space between failure and (

src/server/routes/units.js
```diff
@@ -104,9 +104,9 @@ router.post('/edit', async (req, res) => {
...
 		}
-		res.status(200).json({ message: `Successfully edited units ${req.body.id}` });
+		success (res, 'Successfully edited units ');
```
chanhenry54: Remove the space between success and (. Also, I believe you're missing the `req.body.id`

src/server/routes/units.js
```diff
@@ -185,10 +185,10 @@ router.post('/addUnit', async (req, res) => {
 				);
 				await newUnit.insert(t);
 			});
-			res.sendStatus(200);
+			success(res)
```
chanhenry54: A bit of a nitpick comment for style, but I'd include a semicolon here just remain consistent with the rest of the code

chanhenry54:
For any OED people looking at this PR, I'm a mentor for this group of students who was asked to do a brief code review, just made some comments regarding what I thought would be good to take a look at. I'm not familiar with this project specifically, so feel free to dismiss anything if it's not standard to the project :)

mosguinz:
Thank you both for reviewing this draft PR.

@huss: Just to clarify, would you like the message to be displayed using `window.alert` when `failure` is used in the corresponding files in `/routes`?

https://github.com/OpenEnergyDashboard/OED/blob/f8f54cdefba02a48905a2924a9f15a08b678769e/src/client/app/actions/meters.ts#L89-L91

huss:
@mosguinz that you for checking and being careful. The referenced code has this (part of the full function):

```
			try {
				// posts the edited meterData to the meters API
				await metersApi.edit(editedMeter);
				// Clear meter Id from submitting state array
				dispatch(deleteSubmittedMeter(editedMeter.id));
				// Update the store with our new edits
				dispatch(confirmMeterEdits(editedMeter));
				// Success!
				showSuccessNotification(translate('meter.successfully.edited.meter'));
			} catch (err) {
				// Failure! ):
				// TODO Better way than popup with React but want to stay so user can read/copy.
				window.alert(translate('meter.failed.to.edit.meter') + '"' + err.response.data as string + '"');
				// Clear our changes from to the submitting meters state
				// We must do this in case fetch failed to keep the store in sync with the database
				dispatch(deleteSubmittedMeter(editedMeter.id));
			}

```

So, what you show is correct. I note that the notification functions in src/client/app/utils/notifications.ts now have the autoDismiss parameter so it seems we could get it to stay around and not use the alert. This would make success and failure notices more consistent and in the React style. If we did this then we need to be sure we can copy/paste the text in the error and it is clear to the user how to dismiss it. (along with being sure it does not get removed by something else happening) This does not need to be part of your PR (unless you want to) but something OED should consider doing in the future (maybe creating an issue on this).

mosguinz:
Taking a look at `conversions.ts` and `units.ts`, it looks like they use the `showErrorNotification` function, rather than `window.alert`.

As you mentioned, I think it may be best to tackle that as a separate issue for the sake of containing the scope of this PR. I will be happy to take a look at it.

src/server/routes/units.js
```diff
@@ -8,6 +8,8 @@ const { getConnection } = require('../db');
 const Unit = require('../models/Unit');
 const { removeAdditionalConversionsAndUnits } = require('../services/graph/handleSuffixUnits');
 const validate = require('jsonschema').validate;
+const { success, failure } = require('./response');
+
```
huss, line 12: Small formatting item but could you please remove the new blank line as OED uses a single one between blocks of code.

src/server/routes/conversions.js
```diff
@@ -77,9 +78,9 @@ router.post('/edit', async (req, res) => {
 			await updatedConversion.update(conn);
 		} catch (err) {
 			log.error('Failed to edit conversion', err);
-			res.status(500).json({ message: 'Unable to edit conversions.', err });
+			failure(res, 500, "Unable to edit conversions due to " + err.toString());
 		}
-		res.status(200).json({ message: `Successfully edited conversions ${req.body.sourceId}` });
+		success(res, `Successfully edited conversions ${req.body.sourceId}`);
```
huss: This is not your doing (code already had this) and I can edit if you do not want to. The entire second argument (the success message) to success should be removed. We only have easy access to the ids but an admin does not know what they are. Thus, we should return no string and let the client display the internationalized success message that it already does without further information.

src/server/routes/conversions.js
```diff
@@ -68,7 +69,7 @@ router.post('/edit', async (req, res) => {
 	const validatorResult = validate(req.body, validConversion);
 	if (!validatorResult.valid) {
 		log.warn(`Got request to edit conversions with invalid conversion data, errors:${validatorResult.errors}`);
-		res.status(400);
+		failure(res, 400, "Error(s): " + validatorResult.errors.toString());
```
huss: As is often the case for the related log message, it would be nice to include what was happening in the message returned to the user. For example, this could become: \
Got request to edit conversions with invalid conversion data, errors:${validatorResult.errors.toString() \
This applies to all the new messages in the PR where the needed change only applies to only some.

src/server/routes/units.js
```diff
@@ -104,9 +106,9 @@ router.post('/edit', async (req, res) => {
 			await unit.update(conn);
 		} catch (err) {
 			log.error('Failed to edit unit', err);
-			res.status(500).json({ message: 'Unable to edit units.', err });
+			failure(res, 500, 'Unable to edit units ' + err.toString());
 		}
-		res.status(200).json({ message: `Successfully edited units ${req.body.id}` });
+		success(res, `Successfully edited units ${req.body.id}`);
```
huss: This has a similar issue to conversions in that is returns the id. However, here we can use req.body.identifier instead. Could you make that change (or I can)?

src/server/routes/units.js
```diff
@@ -185,10 +187,10 @@ router.post('/addUnit', async (req, res) => {
 				);
 				await newUnit.insert(t);
 			});
-			res.sendStatus(200);
+			success(res);
 		} catch (err) {
 			log.error(`Error while inserting new unit ${err}`, err);
-			res.sendStatus(500);
+			failure(res, 500, err.toString() + ' with detail ' + err['detail']);
 		}
```
huss: From what I recall, this is the only message that has the err and the err detail. From the messages I could easily generate, the detail is included in the err string. Thus, I propose we remove it unless someone sees something I do not. Whatever is done, it would be nice to be consistent.

huss (changes requested):
That you to @mosguinz, @EdmondLee958 and @leileigoose for this PR and contributing to OED. The changes all do what was expected and is nice work. I have left a few comments that should be addressed. They mostly come from how OED wants to work. Some are a little outside the issue taken on so I can do them if you want (just let me know).

huss:
> Taking a look at `conversions.ts` and `units.ts`, it looks like they use the `showErrorNotification` function, rather than `window.alert`.
> 
> As you mentioned, I think it may be best to tackle that as a separate issue for the sake of containing the scope of this PR. I will be happy to take a look at it.

Just for the record, I put in issue #869 to address this.

mosguinz:
Hi @huss, I've just made the requested changes. Let me know if there's anything I missed!

src/server/routes/units.js
```diff
@@ -104,9 +105,9 @@ router.post('/edit', async (req, res) => {
 			await unit.update(conn);
 		} catch (err) {
 			log.error('Failed to edit unit', err);
-			res.status(500).json({ message: 'Unable to edit units.', err });
+			failure(res, 500, 'Unable to edit units ' + req.body.identifier);
```
huss: I think this line was fine with returning the err. It was the next one so the success uses req.body.identifier instead of id. Sorry for the confusion.

src/server/routes/units.js
```diff
@@ -83,7 +84,7 @@ router.post('/edit', async (req, res) => {
 	const validatorResult = validate(req.body, validUnit);
 	if (!validatorResult.valid) {
 		log.warn(`Got request to edit units with invalid unit data, errors:${validatorResult.errors}`);
-		res.status(400);
+		failure(res, 400, 'Validation failed with ' + validatorResult.errors.toString());
```
huss: Could this message include it is for edit units?

src/server/routes/units.js
```diff
@@ -165,7 +166,7 @@ router.post('/addUnit', async (req, res) => {
 	const validationResult = validate(req.body, validUnit);
 	if (!validationResult.valid) {
 		log.error(`Invalid input for unitsAPI. ${validationResult.error}`);
-		res.sendStatus(400);
+		failure(res, 400, 'Validation failed with  ' + validationResult.error.toString());
```
huss: Could this failure include that it is for adding a unit? I would agree that the original log message right above it could be more specific (not code you did).

src/server/routes/conversions.js
```diff
@@ -121,7 +122,7 @@ router.post('/addConversion', async (req, res) => {
 	const validationResult = validate(req.body, validConversion);
 	if (!validationResult.valid) {
 		log.error(`Invalid input for ConversionsAPI. ${validationResult.error}`);
-		res.sendStatus(400);
+        failure(res, 400, 'Invalid input: ' + validationResult.error.toString());
```
huss: Could this one say it is for adding a converson (and the log message is, again, not the best and I approved it :-)

huss:
Thank you for the quick update. Just a couple of details I commented on above. It is almost there and I appreciate your efforts.

huss:
Thanks for the quick changes. Everything was addressed nicely and the code is good. It should be ready to merge.

huss (approved):

PR #858, opened 2023-02-11
chore: Remove unused dependencies

hyperupcall:
# Description

In `package.json`, there were several dependencies that weren't being used. Over the years, they seem to have been accidentally included in various PRs. This increases `npm install` installation time, and is a security risk.

This fixes that by removing the items from `package.json`. None of the dependencies were `require`'d in any other part of the codebase, so they are fit for removal.

The `babel-plugin-lodash` is a special case as it's recommended for use by `lodash-webpack-plugin`. Since it's not a direct dependency, its status was changed to a `peerDependency`.

## Type of change

- [X] Note merging this changes the node modules

## Checklist

## Limitations

There are a few dependencies that are only used in one or two places, that could be removed since these functions are built into NodeJS these days. But that's for a different PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,15 +15,12 @@
         "@types/history": "~4.7.9",
         "@types/react-router": "~5.1.17",
         "@types/react-router-dom": "~5.3.2",
-        "attr-accept": "~2.2.1",
         "axios": "~0.24.0",
         "babel-core": "~7.0.0-bridge.0",
         "babel-plugin-formatjs": "~10.3.12",
         "bcryptjs": "~2.4.3",
         "body-parser": "~1.18.3",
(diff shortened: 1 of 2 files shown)


huss (approved):
Congratulations to @hyperupcall for their first PR to OED and thanks for addressing this issue. Code review and modest testing found everything works as expected and the changes seem good. I concur that removing unused dependencies is a good idea. 

You mention some other dependencies that could be removed by using standard JS features. If you want to open an issue or provide them to the project that would be appreciated.

In case anyone is confused by this, several of the dependencies removed from pacakge.json are still in package-lock.json because they are used in other dependencies.

hyperupcall:
Thank you, sounds good!

PR #861, opened 2023-02-18
types: Miscellaneous improvements

hyperupcall:
# Description

These changes improve typings in two areas:
1. Correctly specifying the return type of the `failure` function (`||` an operator in TypeScript)
2. Fixing type inference for `express.*` types (lines 8-9) by importing `express`
  2a. This could also be fixed by using the global `Express` type, but this was avoided for consistency

## Type of change

- [] Note merging this changes the node modules
- [] Note merging this changes the database configuration.
- [] This change requires a documentation update

## Checklist

## Limitations

N/A

### DIFF
--- a/src/server/services/csvPipeline/success.js
+++ b/src/server/services/csvPipeline/success.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const express = require('express')
+const express = require('express') /* needed to resolve types in JSDoc comments */
 
 /**
  * Inform the client of a success (200 OK).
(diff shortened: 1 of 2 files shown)


src/server/services/csvPipeline/success.js
```diff
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const express = require('express')
```
huss: I think (guess) it is okay to add a require for the comment. However, if we are going to do that then I would also do it in failure.js as it seems to be the same usage there.

One final thought is to put a comment on the require to indicate it is needed to resolve the type in the comment so it is clear for future developers.

huss (changes requested):
Thanks to @hyperupcall for another PR. This seems fine and I just made what I hope is a simple comment to resolve.

hyperupcall:
made those changes :+1:

huss (approved):

huss:
Thanks to @hyperupcall for the quick changes. Everything looks good and this is ready to merge.

PR #863, opened 2023-02-18
Multiple Fixes: Fixes issues with Modals (Issue #794)

### LINKED ISSUE #794: lost help on pages with cards
### Describe the bug

If you are on one of the new pages that uses modals (unit, meter, conversion) then the help icon works fine when you first load the page. However, once you have a modal popup (edit or create) then it will not work until you leave the page and return. The help on the modal popup works fine.

### To Reproduce

1. To to one of the modal pages.
2. Click the help icon to see it works.
3. Click to edit or create a new item.
4. After the modal pops up you can see the help works. You don't need to actually change anything on the page. Leave the popup in any way.
5. The help icon on the card page will not work.
6. Go to any other page and return and the help icon now works.

### Expected behavior

The help icons should always work.

### Additional context

None

Donny-Guo:
Fixes #794

# Description

(This resolves the tooltips not being shown after displaying a modal. Also fixes an issue with the Menu Modal not allowing tooltips for mobile users on the front page.)

## Type of change

## Checklist

## Limitations

No limitations.

### DIFF
--- a/src/client/app/components/MenuModalComponent.tsx
+++ b/src/client/app/components/MenuModalComponent.tsx
@@ -60,7 +60,7 @@ export default class MenuModalComponent extends React.Component<MenuModalProps, 
 							<p style={labelStyle}>
 								<FormattedMessage id='navigation' />:
 							</p>
-							<HeaderButtonsContainer  showCollapsedMenuButton={false} isModal />
+							<HeaderButtonsContainer  showCollapsedMenuButton={false} isModal={true} />
 						</div>
 						{ this.props.showOptions &&
 							<UIOptionsContainer />
--- a/src/client/app/components/MenuModalComponent.tsx
+++ b/src/client/app/components/MenuModalComponent.tsx
@@ -60,7 +60,7 @@ export default class MenuModalComponent extends React.Component<MenuModalProps, 
 							<p style={labelStyle}>
 								<FormattedMessage id='navigation' />:
 							</p>
-							<HeaderButtonsContainer  showCollapsedMenuButton={false} />
+							<HeaderButtonsContainer  showCollapsedMenuButton={false} isModal />
 						</div>
 						{ this.props.showOptions &&
 							<UIOptionsContainer />
--- a/src/client/app/components/HeaderButtonsComponent.tsx
+++ b/src/client/app/components/HeaderButtonsComponent.tsx
@@ -15,6 +15,7 @@ import { hasPermissions } from '../utils/hasPermissions';
 import { FlipLogOutStateAction } from '../types/redux/unsavedWarning';
 
 interface HeaderButtonsProps {
+	isModal: boolean;
 	showCollapsedMenuButton: boolean;
 	loggedInAsAdmin: boolean;
 	role: UserRole | null;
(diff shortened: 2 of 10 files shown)


src/client/app/components/MenuModalComponent.tsx
```diff
@@ -60,7 +60,7 @@ export default class MenuModalComponent extends React.Component<MenuModalProps,
 							<p style={labelStyle}>
 								<FormattedMessage id='navigation' />:
 							</p>
-							<HeaderButtonsContainer  showCollapsedMenuButton={false} />
+							<HeaderButtonsContainer  showCollapsedMenuButton={false} isModal />
```
huss: This may well be my lack of knowledge about this so just let me know. I'm confused about isModal in two ways. First, it does not seem to be set in this file. Is it assuming a default? Second, it is supplying this value the HeaderButtonsContainer but that does not seem to be part of its props. However, it all seems to work great.

huss:
Thanks to @ConlynPattison and @Donny-Guo for their first PR. OED appreciates your contribution. From the records I see, neither of you have agreed to the [OED CLA](https://openenergydashboard.github.io/developer/cla.html). The project needs that before we can accept your contribution. Please let me know if you have any questions or if you think I missed your CLA. Thanks.

huss (changes requested):
Thank you to @Donny-Guo and @ConlynPattison for this PR. I tried hard to get the failure to occur that was described in the issue but could not. It seems to work perfectly. I did leave one comment that I would like to better understand. I'm mostly requesting changes to hold for the needed CLA (see comment).

huss:
The CLAs are now good. Thanks for the quick work. Now I would just like your insights into the one comment where I was unsure about the code.

ConlynPattison:
@huss thank you for the feedback, and both @Donny-Guo and I have now filled out the OED CLA. Regarding the props for the HeaderButtonsContainer, we could add the isModal property to the file to remain consistent with the showCollapsedMenuButton as a 'ownProps' property. It seems to react to the prop values in the Component file without that addition, but I understand how that could be confusing.

In the MenuModalComponent you've snipped, if I understand your first confusion correctly, it is set to ={true} in that component render call. You could explicitly set the property with isModal={true}, but it is not needed to achieve the same value.

huss:
> @huss thank you for the feedback, and both @Donny-Guo and I have now filled out the OED CLA. Regarding the props for the HeaderButtonsContainer, we could add the isModal property to the file to remain consistent with the showCollapsedMenuButton as a 'ownProps' property. It seems to react to the prop values in the Component file without that addition, but I understand how that could be confusing.

Thanks to @ConlynPattison for the explanation and quick reply. I think consistency and being as transparent as possible are good. Thus, I lean toward adding this if you are willing.

> In the MenuModalComponent you've snipped, if I understand your first confusion correctly, it is set to ={true} in that component render call. You could explicitly set the property w

Again, I lean toward showing the intended value when there might be any doubt. I think the code always wants true so I would show that.

huss:
Thanks for making the suggested change. The code looks good and is ready to merge.

huss (approved):

PR #867, opened 2023-02-19
chore: Replace `es6-promisify` with `util.promisify`

hyperupcall:
# Description

As noted generically in #858, this removed the use of a dependency that can be replaced with a built-in NodeJS function: `util.promisify`. It was added in NodeJS v8.0.0.

## Type of change

## Checklist

## Limitations

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,6 @@
         "csv": "~5.3.2",
         "debug": "~4.3.2",
         "dotenv": "~9.0.0",
-        "es6-promisify": "~5.0.0",
         "express": "~4.17.3",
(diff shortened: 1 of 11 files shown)


huss (approved):
Thanks to @hyperupcall for another PR. My code review found nothing. I also tested that meter & reading CSV upload worked as before along with seeing that login/out still worked. All worked as expected.

There was a warning from CodeQL. This is a standing potential issue and not new. I'm going to leave so others can consider but I don't think it is a problem. Only logged in admins can do this so I think it is okay.

This is good work and ready to merge.

PR #870, opened 2023-02-20
chore: PackageJson improvements

hyperupcall:
# Description

This makes two little improvements to the `package.json`:

1. Removes uppercase from package name (according to [this page](https://github.com/npm/validate-npm-package-name#naming-rules), uppercase is not valid):

Screenshot at 2023-02-20 14-13-30

2. Remove `./node_modules/.bin/...` from invocations. According to the [docs](https://docs.npmjs.com/cli/v9/commands/npm-run-script), this directory is already added to the `PATH`, so it will still work.

## Type of change

## Checklist

## Limitations

N/A


huss (changes requested):
Thanks to @hyperupcall for another PR. I had let the capitalized OED name stay since we are not really creating an NPM package for distribution. However, it seems fine to me to make it conform to the NPM standard since I don't think we use it so it should not impact anyone. Having said that, I did notice that the package-lock.json file has the old name in two places. I think it should be manually edited (to avoid other changes) to match the name in package.json.

Testing indicates no issues.

hyperupcall:
Ah, I didn't think to check there - made those changes :+1:

huss:
The changes are good and it still works :-) Thanks for finalizing this quickly. It is ready to merge.

huss (approved):

Issue #869, opened 2023-02-19
show routed messages to user


huss:
### Is your feature request related to a problem? Please describe.

PR #857 sends back messages from the conversion and unit routes in the desired way (code in src/server/routes/conversions.js or units.js). The API returns what is routed (e.g., src/client/app/utils/api/ConversionsApi.ts). The action that dispatches the code ignores the returned values (e.g., src/client/app/actions/conversions.ts) and just puts up generic messages. It would be nice to display the returned messages. Note it is often only done for failures as success does not return a value.

Examples of code already doing this are:

- src/client/app/actions/meters.ts in submitEditedMeter() for failure.
- src/client/app/components/csv/ReadingsCSVUploadComponent.tsx in handleSubmit() displays the message on success.

### Describe the solution you'd like

The message returned from the route should be passed up the client call chain to an appropriate point to display. Note some code uses a window alert but the showSuccessNotification/showErrorNotification now have a time argument to control how long it stays up so that would probably be better. For some error messages where the user needs lots of time to read or wants to copy, then getting it to stay up until closed would be nice. I also have not checked how easy it is to copy the displayed text on these messages.

### Additional context

There are other places with notices that could be investigated for similar changes.

ArtP1:
I'm working on this

huss:
Given OED now allows admins to easily see log messages, the full messages are now available in the log. OED should limit the content of messages to explain what happened and not, in general, show the internal details of the issue. The message could tell users to see the log or contact the site admin for help.

aduques:
Referencing work done in Issue #1191.

Since I have added a disabled feature that disables the save button if changes are not valid, I'm not sure how the failure messages can be reachable. However, I did notice that the success messages were not implemented everywhere.

Note: While the known ways of the failure messages have been essentially prevented by the implementation of the unsaved warning, failure/success messages should be used everywhere when applicable, so that features can be debugged in the future if they ever fail.

Files that don't have a success/failure message present:
CreateConversionComponent.tsx
EditConversionComponent.tsx
CreateGroupModalComponent.tsx
EditGroupModalComponent.tsx

Could not check completely:
MetersCSVUploadComponent.tsx
ReadingsCSVUploadComponent.tsx

aduques:
[...]

PR #877, opened 2023-03-24
Update conversion error notifications to show error details

EricE147:
# Description

Updated `addConversion`,  `submitEditedConversion`, and `deleteConversion` catch statements in conversions.ts to show error details.

Partly Addresses #869.

1. Add: image
2. Edit: image
3. Delete: image

All tests currently pass: 
OED_Tests

## Type of change

This PR should not require any major update.

## Checklist

## Limitations

1. The time components of `showSuccessNotification` and `showErrorNotification` don't seem to actually change how long the messages stay up as the webpage immediately reloads regardless of how long `autoDismiss` is set for. Another issue may need to be created to fix this. 
2. We were also unable to trigger the error notification naturally and instead forced it using a throw statement so we don't know exactly what the error messages look like.

### DIFF
--- a/src/client/app/actions/conversions.ts
+++ b/src/client/app/actions/conversions.ts
@@ -99,7 +99,7 @@ export function submitEditedConversion(editedConversion: t.ConversionData, shoul
 				showSuccessNotification(translate('conversion.successfully.edited.conversion'));
 			} catch (err) {
 				// Failure! ):
-				showErrorNotification(translate('conversion.failed.to.edit.conversion') + '"' + err.response.data + '"');
+				showErrorNotification(translate('conversion.failed.to.edit.conversion') + ' "' + err.response.data as string + '"');
 			}
 			// Clear conversionData object from submitting state array
 			dispatch(deleteSubmittedConversion(editedConversion));
@@ -121,7 +121,7 @@ export function addConversion(conversion: t.ConversionData): Thunk {
 			dispatch(fetchConversionsDetails());
 			showSuccessNotification(translate('conversion.successfully.create.conversion'));
 		} catch (err) {
-			showErrorNotification(translate('conversion.failed.to.create.conversion') + '"' + err.response.data + '"');
+			showErrorNotification(translate('conversion.failed.to.create.conversion') + ' "' + err.response.data as string + '"');
 		}
 	}
 }
@@ -150,7 +150,7 @@ export function deleteConversion(conversion: t.ConversionData): Thunk {
 				dispatch(confirmDeletedConversion(conversion));
 				showSuccessNotification(translate('conversion.successfully.delete.conversion'));
 			} catch (err) {
-				showErrorNotification(translate('conversion.failed.to.delete.conversion') + '"' + err.response.data + '"');
+				showErrorNotification(translate('conversion.failed.to.delete.conversion') + ' "' + err.response.data as string + '"');
 			}
 			// Inform the store we are done working with the conversion
 			dispatch(deleteSubmittedConversion(conversion));
--- a/src/client/app/actions/conversions.ts
+++ b/src/client/app/actions/conversions.ts
@@ -99,7 +99,7 @@ export function submitEditedConversion(editedConversion: t.ConversionData, shoul
 				showSuccessNotification(translate('conversion.successfully.edited.conversion'));
 			} catch (err) {
 				// Failure! ):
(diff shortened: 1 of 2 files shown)


xJU4Nx (approved):
Waiting for review. Thanks.

salhernandez:
Hi @huss 👋🏼 

I mentored the students that address #869. They found a bug with `showSuccessNotification` and `showErrorNotification` where the `autoDismiss` argument seems to not considered. It didn't matter if we set `autoDismiss` to 2 seconds or 2000 seconds, There may be a fixed autoDismiss value deeper into the call stack. Should the students create a issue report for it?

src/client/app/actions/conversions.ts
```diff
@@ -99,7 +99,7 @@ export function submitEditedConversion(editedConversion: t.ConversionData, shoul
 				showSuccessNotification(translate('conversion.successfully.edited.conversion'));
 			} catch (err) {
 				// Failure! ):
-				showErrorNotification(translate('conversion.failed.to.edit.conversion'));
+				showErrorNotification(translate('conversion.failed.to.edit.conversion') + '"' + err + '"');
```
huss: In other places in OED we do "err.response.data" to see the actual message. I think it may be a little risky if another type of error but I think we should do it here and in the other messages to be consistent.

src/client/app/actions/conversions.ts
```diff
@@ -150,7 +150,7 @@ export function deleteConversion(conversion: t.ConversionData): Thunk {
 				dispatch(confirmDeletedConversion(conversion));
 				showSuccessNotification(translate('conversion.successfully.delete.conversion'));
 			} catch (err) {
-				showErrorNotification(translate('conversion.failed.to.delete.conversion'));
+				showErrorNotification(translate('conversion.failed.to.delete.conversion') + '"' + err + '"');
```
huss: This was not your doing but this message does not show up as expected because src/server/routes/conversions.js near the end of the file for the delete route does not use failure but res.status. This happens in two places. Would you be willing to fix that?

If you do edit that file then you could also fix the addConversion route that has
validationResult.error.toString()
but it should be
validationResult.errors.toString()
as done elsewhere. I missed that when the code was added to OED in the past. It causes an issue so the message is not seen.

huss (changes requested):
Thanks to @EricE147, @ArtP1 & @xJU4Nx for their first PR to OED. Overall, this is good work. I made a couple of comments you can address (or not for the one outside the code you touched) but they are minor.

Just so you know, I tested your code by setting the sourceID to -1 in the route for conversions (src/server/routes/conversions.js):
req.body.sourceId = -1;
This causes the validation to fail and the failure to happen. It can be done in all three of the routes (add, edit, delete) to test each one.
Just for the record, note that I first tried to set the value in src/client/app/actions/conversions.ts but that caused errors in the component since it uses this id to get the string displayed for the source id in 2 of the 3 cases so that is not a good way to do it.

huss:
> I mentored the students that address #869. They found a bug with `showSuccessNotification` and `showErrorNotification` where the `autoDismiss` argument seems to not considered. It didn't matter if we set `autoDismiss` to 2 seconds or 2000 seconds, There may be a fixed autoDismiss value deeper into the call stack. Should the students create a issue report for it?

First, thank you @salhernandez for taking the time to mentor these students. As for the noted this issue, it impacts several parts of the code in OED. Thus, I think putting in an issue is a good idea and I would be pleased to have the students do that. I have been thinking about possible solutions and have an idea. Once the issue is posted, I will comment on why it exists and a first cut at a solution.

EricE147:
@huss we updated the error message to include `response.data` and also fixed the typo in `validationResult.error.toString()`. We tried changing the `res.status` to `failure` but we were still not getting the correct behavior so we left it unchanged.

EricE147:
This is what is shown with `err.response.data`:
image

src/client/app/actions/conversions.ts
huss: I fixed to use success/failure and you did the rest so this is fixed.

huss:
That you for your updates. They all looked great. I'm sorry it took a few days as other thing took up my time.

While looking at the code I noticed inconsistencies in previous code (not what you edited) so I made some changes that are in the commit message. This PR is now ready to merge. Congratulations to  @EricE147, @ArtP1 & @xJU4Nx for their first merged code.

huss (approved):

PR #883, opened 2023-04-12
Adding Maximum and Minimum reading_rate to readings tables

### LINKED ISSUE #594: data range on line graphics
OED went to a system that averages to a daily level to speed up OED (fast-pt). Users can zoom in to see raw data but it can be tricky to know where interesting values are located. To help, OED could add "error" bars to points to represent the range of data for the time interval for each point on the line graph. An easy idea is max/min value. Slightly more sophisticated might be 25/75% quartile labels. Something else would be okay too. The idea is it should be fairly fast (certainly for max/min) if we precalculate when do refresh readings of the daily value table. It would give an idea of the underlying range of values for each point. With the new Plotly zoom and redraw feature you could look at areas of concern with greater resolution. This should be an option that the user can turn on since most will not what to see this.

taylon4:
# Description

This change adds a maximum and minimum reading rate for the daily/hourly readings unit table. This would ideally be used as a visual display on the front end when graphing data.

Fixes #594

## Type of change

## Checklist

## Limitations

Edge cases should be tested with the max/min from the database. Specifically overlapping days - for example June 3rd 12:00 PM to June 4th 2:00 AM


src/server/migrations/0.8.0-1.0.0/sql/readings/create_reading_views.sql
```diff
@@ -2,9 +2,88 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/*
```
huss, line 5: This seems to be a copy of the SQL file. What it should have is any needed changes. So, date_trunc_up, etc. should not be present as those functions did not change. I think you only need to two views you changed.

src/server/sql/reading/create_reading_views.sql
```diff
@@ -127,7 +127,76 @@ daily_readings_unit
 									greatest(r.start_timestamp, gen.interval_start)
 					)
 			))
-		END AS reading_rate,
+		END AS reading_rate, 
+		CASE WHEN u.unit_represent = 'quantity'::unit_represent_type THEN
+    	(max(( --Extract the maximum rate over each day
```
huss: The indenting seems off on this line.

Also, it might be nice to have a comment that explains why the max includes the division (second extract) whereas the sum does not.

These apply to all cases following.

src/server/sql/reading/create_reading_views.sql
```diff
@@ -127,7 +127,76 @@ daily_readings_unit
...
+									-
+									greatest(r.start_timestamp, gen.interval_start)
+					)
+			) / (
+					extract(EPOCH FROM -- The number of seconds that the reading shares with the interval
+									least(r.end_timestamp, gen.interval_start + '1 day'::INTERVAL)
+									-
+									greatest(r.start_timestamp, gen.interval_start)
+					)
+			)) )
```
huss: I think the first ) should be on a separate line and lined up in the appropriate column to match the sum code above.

src/server/sql/reading/create_reading_views.sql
```diff
@@ -127,7 +127,76 @@ daily_readings_unit
...
+			WHEN (u.unit_represent = 'flow'::unit_represent_type OR u.unit_represent = 'raw'::unit_represent_type) THEN
```
huss: Indenting seems off. Maybe all can be checked.

src/server/sql/reading/create_reading_views.sql
```diff
@@ -311,7 +448,6 @@ BEGIN
 END;
 $$ LANGUAGE 'plpgsql';
```
huss, line 508: I don't mind removing the double blank line here but then maybe do above to be consistent?

src/server/sql/reading/create_reading_views.sql
```diff
@@ -237,7 +374,7 @@ BEGIN
 	-- Make sure the time range is withing the reading values.
 	requested_range := shrink_tsrange_to_real_readings(tsrange(start_stamp, end_stamp, '[]'));
 	requested_interval := upper(requested_range) - lower(requested_range);
-	-- unit_column holds the column index into the cik table. This is the unit that was requested for graphing.
+	-- holds the column index into the cik table. This is the unit that was requested for graphing.
```
huss: I'm not clear on why this comment was changed. Can you help me understand?

src/server/migrations/0.8.0-1.0.0/sql/readings/create_reading_views.sql
```diff
@@ -2,9 +2,88 @@
...
+allow for getting a subset of the raw/meter readings quickly.
+ */
+
 /**
 The next two create a view/table that takes the raw/meter readings and averages them for each day or hour.
-This is used by the two line graph functions below to make them faster since the values
+This is used by the line graph function below to make them faster since the values
 are already averaged. There are two types of readings: quantity and flow/raw. The quantity
 readings must be normalized by their time length. The flow/raw readings are already by time
 so they are just averaged. The one table contains both types of readings but are now equivalent
```
huss: While you did not do the first version, I think there should be
DROP MATERIALIZED VIEW IF EXISTS daily_readings_unit;
before the create so the create runs. The view normally should not be there but this would be safer. Same applies to hourly.

huss (changes requested):
Thanks to @taylon4 for their first PR for OED. I have run some basic checks where the readings are all the same length and the results are as expected so the SQL seems correct. I have made a few comments that I would like addressed.

So everyone understands, I plan to comment out the new code and migration change when this PR is approved. It does not reflect in any way on this work. The issue is that OED cannot use the new max/min values until the graphics are changed and adding it to the view will slow down the DB during refreshes and take up space. The hope/plan is to get that done this summer and then the new code will be uncommented for inclusion. It is also hoped to have automated tests including readings that cross time boundaries by that time so all the SQL code can be tested (including this change) to verify it works as expected. Since this code uses the same logic as other places it is not expected to be an issue as manual tests in the past did not find an issue.

src/server/sql/reading/create_reading_views.sql
huss: I addressed this. The part about second extract is unclear (even thought I made the comment) so ignored.

src/server/sql/reading/create_reading_views.sql
huss: addressed,

src/server/sql/reading/create_reading_views.sql
huss: tried to make them all one blank line.

src/server/sql/reading/create_reading_views.sql
huss: decided change was okay.

src/server/migrations/0.8.0-1.0.0/sql/readings/create_reading_views.sql
huss: I realized you don't need to do this since this is a new view and the old one is dropped elsewhere.

src/server/migrations/0.8.0-1.0.0/sql/readings/create_reading_views.sql
huss: I fixed this up but also noticed the migrations were off so I also edited so it has all the correct migrations for this file.

huss (approved):

PR #885, opened 2023-04-17
Allow normalization of displayed data by area

### LINKED ISSUE #849: normalize usage by area

An option to allow each graphic value in OED to be normalized by the associated area.

The [area normalization](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/areaNormalization/areaNormalization.md) design document covers the ideas, discussion and proposed solutions.

This is an enhancement so no alternative at this time.

### Additional context

None.

spearec:
# Description

This PR implements normalization by area as per [the design doc](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/areaNormalization/areaNormalization.md). This allows displayed meter/group data to be shown divided by the area which is assigned to it. This uses the existing area field for meters and groups, and adds a new area unit, allowing for conversions between square feet and square meters.
This also provides database migrations to enforce the new rules for meter/group area, which are `NOT NULL DEFAULT 0 CHECK (area >= 0)`.

For admins, this also adds the feature to automatically calculate a group's area by summing the area of its constituent deep meters. It should be noted that this breaks the rule that a group's usage is the sum of the constituent meter usage (see design doc for more info).

Fixes #849 

## Type of change

## Checklist

## Limitations

PR #884 (group pages) must be cleared before this PR can be merged, as it requires the new state mechanics to work.

While I have done fairly extensive testing of the changes in this PR, it is a large PR that makes multiple database changes, and I did not run a full database migration. I tested each individual change, but there could still be errors revealed with a full migration.

Requires an addition of areaNormalization.html to the help pages. I will also be updating the corresponding designdoc as soon as I can to reflect the finished product.


spearec:
Looking at the output from the build checks, I realized I forgot to note another thing: I edited some of the test files to add area units in order to prevent errors. In doing so, I may have broken some of the tests. It also appears that some of the tests are looking for a null area, but that is no longer a supported feature.
image

src/client/app/utils/getAreaUnitConversion.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/**
+ * Gets the conversion between two area units
+ * @param {AreaUnitType} fromUnit unit to convert FROM
+ * @param {AreaUnitType} toUnit unit to convert TO
+ * @returns {number} conversion multiplier
+ */
+export default function getAreaUnitConversion(fromUnit: AreaUnitType, toUnit: AreaUnitType): number {
+	if(fromUnit === toUnit) {
```
huss: formatting slightly off

src/client/app/containers/LineChartContainer.ts
```diff
@@ -43,76 +44,92 @@ function mapStateToProps(state: State) {
...
-					});
+			let meterArea = state.meters.byMeterID[meterID].area;
+			// we either don't care about area, or we do in which case there needs to be a nonzero area
+			if (!state.graph.areaNormalization || (meterArea > 0 && state.meters.byMeterID[meterID].areaUnit != AreaUnitType.none)) {
+				if(state.graph.areaNormalization) {
+					// convert the meter area into the proper unit, if needed
+					const graphAreaUnit = state.graph.selectedAreaUnit;
+					const meterAreaUnit = state.meters.byMeterID[meterID].areaUnit;
+					if(graphAreaUnit != meterAreaUnit) {
+						meterArea *= getAreaUnitConversion(graphAreaUnit, meterAreaUnit)
```
huss: I think the order of the units are backward in this call. The code is going to normalize by the meterArea so you want this to wind up in the graph unit and not the meter unit. I note that in the group pages where you autocalculate you pass from meter area to group area so it seems the other way. I hope I have this all correct.

Small item is there is no semicolon at the end of the line.

It is the same with groups.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -149,6 +150,44 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
...
+	// or if any other errors are encountered.
+	const handleAutoCalculateArea = () => {
+		if(groupState.deepMeters != undefined && groupState.deepMeters.length > 0) {
+			if(groupState.areaUnit != AreaUnitType.none) {
+				let areaSum = 0;
+				let notifyMsg = '';
+				groupState.deepMeters.forEach(meterID => {
+					const meter = metersState[meterID];
+					if(meter.area > 0) {
+						if(meter.areaUnit != AreaUnitType.none) {
```
huss: formatting off, maybe check all the files?

Also, same thoughts as in meters for this code.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -141,20 +150,88 @@ class PreferencesComponent extends React.Component<PreferencesPropsWithIntl> {
...
+								name='areaUnitType'
+								style={{ marginRight: '10px' }}
+								value={AreaUnitType.meters}
+								onChange={this.handleDefaultAreaUnitChange}
+								checked={this.props.defaultAreaUnit === AreaUnitType.meters}
+							/>
+							<FormattedMessage id='AreaUnitType.meters' />
+						</label>
+					</div>
+					{/* <div className='radio'>
```
huss: I would either get rid of this or add a comment that for units we do not currently support.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -103,6 +105,38 @@ export default function CreateGroupModalComponent(props: CreateGroupModalCompone
...
+			if(state.areaUnit != AreaUnitType.none) {
+				let areaSum = 0;
+				let notifyMsg = '';
+				state.deepMeters.forEach(meterID => {
+					const meter = metersState[meterID];
+					if(meter.area > 0) {
+						if(meter.areaUnit != AreaUnitType.none) {
+							areaSum += meter.area * getAreaUnitConversion(meter.areaUnit, state.areaUnit);
+						} else {
+							notifyMsg += '\n' + meter.identifier + translate('group.area.calculate.error.unit');
```
huss: Is it correct that this really should not happen (area > 0 but no unit for area on a meter)? If so, maybe put a comment.

Also, I've been trying to remember to put quotes around the names/identifiers in case they have spaces, etc. Could those be added here and in groups?

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -103,6 +105,38 @@ export default function CreateGroupModalComponent(props: CreateGroupModalCompone
...
+						}
+					} else {
+						notifyMsg += '\n' + meter.identifier + translate('group.area.calculate.error.zero');
+					}
+				});
+				if(notifyMsg != '') {
+					notifyUser(translate('group.area.calculate.error.header') + notifyMsg);
+				}
+				// the + here converts back into a number
+				setState({...state, ['area']: +areaSum.toFixed(2)});
```
huss: OED does use toFixed in some places but I'm going to advocate for toPrecision as is done for the hover text on graphics. Something similar to 6 places might be good.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -30,11 +30,12 @@ import {
 import { ConversionArray } from '../../types/conversionArray';
 import { GPSPoint, isValidGPSInput } from '../../utils/calibration';
 import { notifyUser, getGPSString, nullToEmptyString, noUnitTranslated } from '../../utils/input';
-import { GroupDefinition } from 'types/redux/groups';
+import { GroupDefinition } from '../../types/redux/groups';
 import ConfirmActionModalComponent from '../ConfirmActionModalComponent'
 import { DataType } from '../../types/Datasources';
 import { groupsApi } from '../../utils/api';
 import { formInputStyle, tableStyle, requiredStyle, tooltipBaseStyle } from '../../styles/modalStyle';
+import getAreaUnitConversion, { AreaUnitType } from '../../utils/getAreaUnitConversion';
```
huss: This type of import works fine. Since there is more than one item that is exported (function & enum) and to make it easier to add future functions, what do you think about removing the default on the function in the imported file and then putting both items inside the braces? I would agree OED has not been consistent on this so I'm very open to leaving it if you think that is better.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -200,16 +239,22 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
...
 				originalGroupState.defaultGraphicUnit != groupState.defaultGraphicUnit ||
 				childMeterChanges ||
-				childGroupChanges
+				childGroupChanges ||
+				originalGroupState.areaUnit != groupState.areaUnit
 			);
 		// Only validate and store if any changes.
 		if (groupHasChanges) {
-			// Check if area is non-negative
+			//Check if area is positive
```
huss: I think a recent merge updated these comments. I would change these three lines to:

// Check if area is non-negative

A subtle point but I think positive does not include 0.

src/client/app/components/groups/GroupsDetailComponent.tsx
```diff
@@ -2,12 +2,13 @@
...
-import { FormattedMessage } from 'react-intl';
+import * as React from 'react';
 import HeaderContainer from '../../containers/HeaderContainer';
 import FooterContainer from '../../containers/FooterContainer';
-import TooltipHelpContainer from '../../containers/TooltipHelpContainer';
 import TooltipMarkerComponent from '../TooltipMarkerComponent';
-import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+import TooltipHelpContainer from '../../containers/TooltipHelpContainer';
+
```
huss: Do we want this blank line? I don't think OED generally puts them between imports.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -172,6 +179,12 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
 			inputOk = false;
 		}
 
+		// A meter area unit must be selected if meter has area
```
huss: This seems to overlap the new check above but have a different message. Are both needed?

src/client/app/types/items.ts
```diff
@@ -16,6 +17,16 @@ export interface SelectOption {
 	style?: React.CSSProperties;
 }
 
+/**
+ * SelectOption but holds strings
+ */
+export interface StringSelectOption {
+	label: string;
+	value: string;
+	isDisabled?: boolean;
```
huss: I could not find where isDisabled & labelIdForTranslate are used. Commenting them out did not cause a build error. Are they needed?

src/client/app/components/ChartDataSelectComponent.tsx
```diff
@@ -243,6 +262,10 @@ export default function ChartDataSelectComponent() {
 			);
 		}
 
+		if(state.graph.selectedAreaUnit == AreaUnitType.none) {
```
huss: Maybe add a comment about why this is here.

src/client/app/components/ExportComponent.tsx
```diff
@@ -52,82 +53,30 @@ export default function ExportComponent() {
...
+					// Make sure it exists in case state is not there yet.
+					if (byMeterID !== undefined) {
+						// The selected rate for scaling
+						let scaling = graphState.lineGraphRate.rate;
+						if(graphState.areaNormalization) {
+							// convert the meter area into the proper unit, if needed
+							const graphAreaUnit = graphState.selectedAreaUnit;
+							const meterAreaUnit = metersState[meterId].areaUnit;
+							if(graphAreaUnit != meterAreaUnit) {
+								scaling *= getAreaUnitConversion(graphAreaUnit, meterAreaUnit)
```
huss: This may have the same concern as noted for line graphs and below for groups.

src/client/app/components/RouteComponent.tsx
```diff
@@ -204,6 +205,11 @@ export default class RouteComponent extends React.Component<RouteProps> {
 									options.toggleBarStacking = true;
 								}
 								break;
+							case 'areaNormalization':
```
huss, line 208: Think about if needs the unit that user selected?

src/client/app/containers/admin/PreferencesContainer.ts
```diff
@@ -12,12 +12,15 @@ import {
 	updateDefaultLanguage,
 	submitPreferencesIfNeeded,
 	updateDefaultWarningFileSize,
-	updateDefaultFileSizeLimit
+	updateDefaultFileSizeLimit,
+	updateDefaultAreaUnit,
```
huss: I know this is picky but could this order be reversed to match the order of unit items below?

src/client/app/containers/BarChartContainer.ts
```diff
@@ -40,44 +41,60 @@ function mapStateToProps(state: State) {
...
-					throw new Error('Unacceptable condition: readingsData.readings is undefined.');
+			let meterArea = state.meters.byMeterID[meterID].area;
+			// we either don't care about area, or we do in which case there needs to be a nonzero area
+			if (!state.graph.areaNormalization || (meterArea > 0 && state.meters.byMeterID[meterID].areaUnit != AreaUnitType.none)) {
+				if(state.graph.areaNormalization) {
+					// convert the meter area into the proper unit, if needed
+					const graphAreaUnit = state.graph.selectedAreaUnit;
+					const meterAreaUnit = state.meters.byMeterID[meterID].areaUnit;
+					if(graphAreaUnit != meterAreaUnit) {
+						meterArea *= getAreaUnitConversion(graphAreaUnit, meterAreaUnit)
```
huss: I think I'll stop marking these but another one that may be reversed.

src/client/app/containers/BarChartContainer.ts
```diff
@@ -40,44 +41,60 @@ function mapStateToProps(state: State) {
...
 			byMeterID[timeInterval.toString()][barDuration.toISOString()] !== undefined) {
-			const readingsData = byMeterID[timeInterval.toString()][barDuration.toISOString()][unitID];
-			if (readingsData !== undefined && !readingsData.isFetching) {
-				const label = state.meters.byMeterID[meterID].identifier;
-				const colorID = meterID;
-				if (readingsData.readings === undefined) {
-					throw new Error('Unacceptable condition: readingsData.readings is undefined.');
+			let meterArea = state.meters.byMeterID[meterID].area;
+			// we either don't care about area, or we do in which case there needs to be a nonzero area
+			if (!state.graph.areaNormalization || (meterArea > 0 && state.meters.byMeterID[meterID].areaUnit != AreaUnitType.none)) {
```
huss, line 46: I think these two if statements can be simplified to:

if (state.graph.areaNormalization && (meterArea > 0 && state.meters.byMeterID[meterID].areaUnit != AreaUnitType.none)) {

The inner parenthesis really are not needed but can be kept if wanted. I think this may apply to other graph type containers.

Having said this, I think the code that begins with readingData needs to be outside this if for this to work. Also, same for groups below and (probably) some other graphic containers.

src/client/app/containers/MapChartContainer.ts
```diff
@@ -167,59 +187,77 @@ function mapStateToProps(state: State) {
...
-									size.push(0);
-								} else {
-									// only display a range of dates for the hover text if there is more than one day in the range
-									timeReading = `${moment.utc(mapReading.startTimestamp).format('ll')}`;
-									if(barDuration.asDays() != 1) {
-										// subtracting one extra day caused by day ending at midnight of the next day.
-										// Going from DB unit timestamp that is UTC so force UTC with moment, as usual.
-										timeReading += ` - ${moment.utc(mapReading.endTimestamp).subtract(1, 'days').format('ll')}`;
+					let groupArea = state.groups.byGroupID[groupID].area;
+					// TODO Consider changing group area to always be defined
```
huss: I think the groups table definition is not null and >= 0 so is this TODO still relevant?

src/client/app/translations/data.js
```diff
@@ -561,6 +594,14 @@ const localeData = {
 		"group": "Groupe",
 		"group.all.meters": "Tous les compteurs",
 		"group.area": "(Need French) Group Area:",
+		"group.area.calculate": "(Need French) Calculate Group Area",
```
huss, line 593: This key shows up twice in French and Spanish.

src/client/app/utils/getAreaUnitConversion.ts
```diff
@@ -0,0 +1,45 @@
...
+	// 'milesfeet': 27880000,
+	// 'kilometersfeet': 10760000,
+	// 'mileskilometers': 2.58999
+};
+
+/**
+ * Gets the conversion between two area units
+ * @param {AreaUnitType} fromUnit unit to convert FROM
+ * @param {AreaUnitType} toUnit unit to convert TO
+ * @returns {number} conversion multiplier
```
huss: I would add "or zero if not found"

src/client/app/utils/graphics.ts
```diff
@@ -14,7 +15,9 @@ import translate from '../utils/translate';
  * @param {LineGraphRate} currentSelectedRate The current selected rate
  * @returns y-axis label and boolean of whether rate needs to be scaled (true if does)
```
huss, line 18: This is missing the new parameters.

src/server/models/Meter.js
```diff
@@ -37,14 +37,15 @@ class Meter {
 	 * @param previousEnd  The last okay reading before crossed out of DST or moment(0) if not, default'1970-01-01 00:00:00'
 	 * @param unitId The foreign key to the unit table. The meter receives data and points to this unit in the graph, default -99
 	 * @param defaultGraphicUnit The foreign key to the unit table represents the preferred unit to display this meter, default -99
-	 */
+	 * @param areaUnit The foreign key to the unit table. Default is 'none'
```
huss: Above for area the default is now 0. (sorry, cannot comment on that line)

src/server/models/Group.js
```diff
@@ -17,15 +18,17 @@ class Group {
 	 * @param note Note about the group
 	 * @param area Area of the group, default null
```
huss: default is now 0.

src/server/models/Group.js
```diff
@@ -17,15 +18,17 @@ class Group {
...
 	 * @param defaultGraphicUnit The foreign key to the unit table represents the preferred unit to display this group.
+	 * @param areaUnit The area unit
```
huss: with default of none.

src/server/models/Preferences.js
```diff
@@ -16,15 +16,18 @@ class Preferences {
 	 * @param {String} defaultTimezone - Option to set the default timezone
 	 * @param {Number} defaultWarningFileSize - Option to set the default warning file size
 	 * @param {Number} defaultFileSizeLimit - Option to set the default file size limit
+	 * @param {Number} defaultAreaNormalization - Option to set the default toggle of area normalization
 	 */
```
huss, line 21: missing defaultAreaUnit

src/server/sql/group/create_groups_tables.sql
```diff
@@ -8,8 +8,9 @@ CREATE TABLE IF NOT EXISTS groups (
 	displayable BOOLEAN,
 	gps POINT DEFAULT NULL,
 	note TEXT,
-	area REAL DEFAULT NULL,
-	default_graphic_unit INTEGER REFERENCES units(id)
+	area REAL NOT NULL DEFAULT 0 CHECK (area >= 0),
+	default_graphic_unit INTEGER REFERENCES units(id),
+	area_unit area_unit_type NOT NULL
```
huss: The migration has:
DEFAULT 'none'.
Should that be added?

src/server/migrations/0.8.0-1.0.0/sql/meter/add_unit_id_default_graphic_unit_area_unit.sql
```diff
@@ -2,7 +2,11 @@
...
 
--- Add unit_id and default_graphic_unit to the meters table.
+-- Add unit_id, default_graphic_unit, and area_unit to the meters table.
+UPDATE meters SET area = 0 WHERE area IS NULL;
 ALTER TABLE meters
     ADD COLUMN IF NOT EXISTS unit_id INTEGER REFERENCES units(id),
-    ADD COLUMN IF NOT EXISTS default_graphic_unit INTEGER REFERENCES units(id);
+    ADD COLUMN IF NOT EXISTS default_graphic_unit INTEGER REFERENCES units(id),
+    ADD COLUMN IF NOT EXISTS area_unit area_unit_type NOT NULL DEFAULT 'none',
+    ALTER COLUMN area SET NOT NULL,
```
huss, line 11: Creating has
CHECK (area >= 0).
Should that be added?

src/server/sql/meter/create_meters_table.sql
```diff
@@ -60,5 +61,6 @@ CREATE TABLE IF NOT EXISTS meters (
     end_timestamp VARCHAR(50) DEFAULT '1970-01-01 00:00:00+00:00',
     previous_end TIMESTAMP DEFAULT '1970-01-01 00:00:00',
     unit_id INTEGER REFERENCES units(id),
-    default_graphic_unit INTEGER REFERENCES units(id)
+    default_graphic_unit INTEGER REFERENCES units(id),
+    area_unit area_unit_type NOT NULL
```
huss: MIgration has DEFAULT 'none'. Should that be added?

src/server/sql/preferences/create_preferences_table.sql
```diff
@@ -11,5 +11,7 @@ CREATE TABLE IF NOT EXISTS preferences (
 	default_language language_type NOT NULL,
 	default_timezone TEXT DEFAULT NULL,
 	default_warning_file_size FLOAT NOT NULL,
-	default_file_size_limit FLOAT NOT NULL
+	default_file_size_limit FLOAT NOT NULL,
+	default_area_normalization BOOLEAN NOT NULL,
```
huss, line 15: There are defaults for the two new area columns in the migration. Should it be added here?

src/server/test/db/meterTests.js
```diff
@@ -83,7 +83,7 @@ mocha.describe('Meters', () => {
 		const conn = testDB.getConnection();
 		const meterPreInsert = new Meter(undefined, 'Meter', null, false, true, Meter.type.MAMAC, 'UTC',
 			gps, 'Identified', 'notes', 33.5, true, true, '05:05:09', '09:00:01', 0, 0, 1, 'increasing', false,
-			25.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 02:12:00', -99, -99);
+			25.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 02:12:00', -99, -99, Unit.areaUnitType.METERS);
 		await meterPreInsert.insert(conn);
```
huss: This test deliberately changes the other unit values so maybe do Unit.feet?

src/server/test/web/csvPipeline/pipe100InputMeter.csv
```diff
@@ -1,3 +1,3 @@
-name,url,enabled,displayable,type,timezone,gps,identifier,note,area,cumulative,reset,reset start,reset end,gap,variation,duplicates,timesort,end only,reading,start time,end time, previous end time, meter unit, graph unit
-pipe100,,FALSE,FALSE,other,,,,,,,,,,,,,,,,,,,,
-pipe100b,123.45.6.0,TRUE,TRUE,obvius,US/Central,"25, 50",pipe100b id,my Note,33,TRUE,TRUE,14:44,15:55,12.34,56.78,7,decreasing,true,89.90,1666-07-08 17:13:19,1777-08-09 05:07:11,1888-09-10 11:12:13,Electric_Utility,kWh
+name,url,enabled,displayable,type,timezone,gps,identifier,note,area,cumulative,reset,reset start,reset end,gap,variation,duplicates,timesort,end only,reading,start time,end time, previous end time, meter unit, graph unit, area unit
```
huss, line 1: I'm just leaving a record that some other input CSV files in these tests set a lot of the meter values. They clearly do not need the area unit but want people to be aware in case they use some as a model. No change needed to this PR.

huss (changes requested):
I want to thank @spearec for taking on another major task and being so careful in your work. This required a lot of thinking on your part and the work shows this. I have a number of comments but, I think, the only one of significance is how the conversion is done for graphing. I have done some basic testing but will hold off on more until that is resolved since it impacts the values graphed. A couple of other thoughts:

- I marked a couple of places where the formatting differs from the standard VSC way. We are trying to stick with that to keep the code consistent. If you can check the files over that would be great.
- I did not load compare data (the new way was broken by recent changes so I need to fix it and I'm trying to get the comments back to you). However, it appeared that you cannot normalize compare data but the option is on that page. Is that correct or did I miss it? If so, should it work (compare is another type of bar graphic where I'm trying to think of why not to include it).

spearec:
Thanks for the comprehensive review. I believe you are correct that the conversion direction is backwards for graphing.

As for compare, since I was unable to get compare data loaded into OED and was not sure what values the compare uses to generate data, I'm not sure I fully implemented the required normalization for compare. I will look into this, as I agree there is no reason for compare not to support normalization.

As I am currently busy finishing up classes for the semester and studying for finals, it may be a little while until I can make these changes.

src/client/app/components/groups/EditGroupModalComponent.tsx
spearec: after looking at what a default function actually is, there's no reason to use one here. I think I just had it because the function I used for reference was a default function.

src/server/sql/preferences/create_preferences_table.sql
spearec: None of the preferences have defaults, because `insert_default_row.sql` handles that, so I don't think it is required here.

src/client/app/components/RouteComponent.tsx
spearec: This is copied from the bar stacking toggle above, so I'm not sure what you mean here.

src/client/app/components/groups/EditGroupModalComponent.tsx
spearec: This was changed in one of the four instances of this check but not the other three.

src/client/app/components/groups/EditGroupModalComponent.tsx
spearec: what is the issue with formatting here?

src/client/app/containers/BarChartContainer.ts
spearec: The reason the statements are done like this is because if area normalization is enabled, then meters that do not satisfy the condition should not be graphed at all, and will fail this check.

src/client/app/utils/getAreaUnitConversion.ts
spearec: Are there still issues with formatting here? I redid this function because we are only converting between two units.

src/client/app/components/groups/CreateGroupModalComponent.tsx
spearec: I believe this cannot happen with just meters added via the web interface, but pre-1.0 meters may have an area but no unit, and maybe imported meters as well.

In short I don't think it should happen but should still be checked anyways, since it will break the code here if a meter has area but no area unit.

huss: You comment was good. I added the quotes.

src/client/app/containers/MultiCompareChartContainer.ts
```diff
@@ -181,4 +182,19 @@ function sortIDs(ids: CompareEntity[], sortingOrder: SortingOrder): CompareEntit
...
+function areaNormalizationValid(state: State, id: number, isGroup: boolean): boolean {
+	if(!state.graph.areaNormalization) {
+		return true;
+	}
+	if(isGroup) {
+		if(state.groups.byGroupID[id].area !== 0 && state.groups.byGroupID[id].areaUnit !== AreaUnitType.none) {
+			return true;
+		}
+	}
+	if(state.meters.byMeterID[id].area !== 0 || state.meters.byMeterID[id].areaUnit !== AreaUnitType.none) {
```
huss: Should this be the same as groups above where need an area AND a unit?

src/client/app/components/RouteComponent.tsx
huss: Sorry. I meant to check this before I finalized the review. This was a note to myself to consider what would happen if the user changed the unit on area normalization. I checked and the unit tell whether it is feet or meters so it works fine.

src/client/app/utils/getAreaUnitConversion.ts
huss: I swept the files for automatic formatting. See my recent commit. They all involved spacing where most were a space between an if and (.  I'm trying to get the code base to use the default, automatic formatting in VSC as it is similar to other systems and means the code is the same when people look at it.

huss (changes requested):
I still have to do a careful check of the values but I suspect all is fine. I have added one place in the updated PR where I have a comment. Could @spearec take a look to see what they think?

huss:
I've mostly finished up testing.  Here are two new thoughts:

1. When you click the custom slider on the bar graphic page it overlaps the Normalize by area checkbox as seen in:
Screenshot 2023-05-01 203648
The current fix is to slide where you want and then click to close to more easily get to the checkbox.
2. Currently you can normalize by area for raw units. They are unique so OED has limited their usage. For example, you cannot do a bar/compare/map graphic with them. I think we should exclude all area normalization for raw units. The best known usage is for temperature and it isn't clear that normalizing that by area makes sense.

@spearec can you address the ones you want and let me know if you want someone else to address any of them. Thanks.

Note: The webData, testData and new automated testing do not yet handle area. I have begun the process but it has more work. I am just noting so other developers are aware that you cannot yet do area without manual edits until this is done. Otherwise they seem to work fine with these changes.

spearec:
I can look into moving the area checkbox down a bit if that is possible, or reordering the sidebar.

As for the compatibility with raw units, that's either very easy or would require a good bit of refactoring. I will have to check, but I think that compatibility could be checked at the same time the code checks for GPS/area unit compatibility.

I also had a thought which relates to this. Currently, OED calculates compatible meters in the select component and also recalculates valid meters when graphing. I was wondering if it makes sense to unify this logic and store whether or not a meter can currently be graphed somewhere in the state. This may not work the way I was thinking, this was just an optimization thought I had.

src/client/app/containers/MultiCompareChartContainer.ts
spearec: Yep, should be an and statement.

I also realized that I've been using `area !== 0` in some cases, and `area > 0` in others. Do you have a preference which one should be used? They both have the same effect because area cannot be negative.

huss: I don't have a reason to prefer one over the other. Do whatever you want/is easier. Thanks for noticing this slight variation in the checks.

huss:
> As for the compatibility with raw units, that's either very easy or would require a good bit of refactoring. I will have to check, but I think that compatibility could be checked at the same time the code checks for GPS/area unit compatibility.

I just looked and in the other cases that avoid raw it seems to be done by the DB not returning any values. In this case, I am now thinking about two ideas (without thinking too much about it):

1. I think a good user solution would be to not show the choice if you have raw units. I think that may just mean seeing if the graphic unit is raw. I want to get OED to hide all choices that do not apply so it will also be hiding the rate meter unless flow (in the future). Thus, this would be compatible with that (assuming others do not have issues with this idea and it goes forward).
2. I think a check can be added to the lineChartContainer for each meter in the loop to stop it from putting the data into what is shown by Plotly or just skip the loop if the graphic unit is raw. This should stop anything from being seen and show an empty graphic.

> I also had a thought which relates to this. Currently, OED calculates compatible meters in the select component and also recalculates valid meters when graphing. I was wondering if it makes sense to unify this logic and store whether or not a meter can currently be graphed somewhere in the state. This may not work the way I was thinking, this was just an optimization thought I had.

A reasonable idea and thanks for thinking about how we can improve the code base. Probably need to think through if edits to meters (or some other items) would update for admins. I assume it does now because it does the calculation right before using that information rather than in state. Please create an issue if you like. OED also needs to decide if this is a post v1.0 change as v1.0 is being settled now and this is a code maintenance rather than feature change.

spearec:
I believe I have successfully disabled area normalization for raw units. The way I implemented this was by both disabling the area unit checkbox when the default graphing unit is raw, and also preventing the user from selecting a raw unit or meter/group with a raw unit when area normalization is enabled.

I also moved the area unit select component above the bar slider, so it no longer overlaps and is also next to the bar stacking checkbox.

huss:
Thanks to @spearec for addressing everything so quickly. I also appreciate thinking about all the cases involving raw and seeing items that I did not mention/think of. I see no issues and this is now ready to merge.

huss (approved):

PR #897, opened 2023-04-25
Dropdown menu for page navigation

### LINKED ISSUE #845: menu for page choices

OED has been adding more pages (conversions, units, ...) where most are related to admin usage. Non-admin users do not use these choices very often. To save page space and to focus user attention on the most important items, the current tabs in the top, right of OED will be converted to a drop-down menu similar to others used for navigation.

Note PR #802 changed the buttons to be in fixed order where the current page is grayed out. This type of behavior will be preserved in the new menu.

The current buttons (for non-admin) look like:

buttons

where there are more choices for admins. This should be converted to the same order as a drop-down menu that has a similar look to the graph type menu:

graphType

As a first implementation, it will sit it the same location as the buttons in the top, right of the screen. It could be moved to the left side if that seems better. The title can be "Select a page" where that is internationalized.

Keep adding buttons that are taking up lots of space :-)

### Additional context

The current code is in src/client/app/components/HeaderButtonsComponent.tsx. src/client/app/components/GraphicRateMenuComponent.tsx has a basic menu with the new look but its logic probably differs. src/client/app/components/ChartSelectComponent.tsx is an older one for selecting the chart.

DylanHNicolay:
# Description

This pull request implements a refactoring of HeaderButtonsComponents to use React Hooks instead of props. Additionally, it adds a dropdown menu to replace the buttons in the header. This change was made because the buttons were starting to wrap, causing the page layout to look worse.

Fixes #845 

## Type of change

## Checklist

- [s] I have removed text in ( ) from the issue request

## Limitations

There is an issue where the help modal is appearing above the dropdown menu since using the css property, display:, will cause an error that prevents the menu from displaying properly. This issue could be related to the warning that pops up whenever the dropdown menu is clicked.


src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -273,4 +273,4 @@ export default function CreateConversionModalComponent(props: CreateConversionMo
 			</Modal>
 		</>
 	);
-}
+}
```
huss: Please put the blank line back. Unsure why this file was edited.

huss (commented):
Thanks to @Nicolax0 for this PR. I appreciate your efforts and success in getting the links to show as desired and getting rid of the console messages. This PR works as desired except for one remaining item which is the help tooltip does not display as desired. I have tried to fix this but have yet to succeed. I'm going to try again later but this PR will need to be merged either way.

huss (approved):
I fixed the last issue and this now works as desired. It is ready to merge.

PR #933, opened 2023-05-26
bug if no meter/group data on line

huss:
# Description

The query failed if there was no meter or group data for lines. This fixes it so it returns nothing so no graphic is shown.

Note bar does not have this issue unless there is no data at all so it seems okay for now.

## Type of change

## Checklist

## Limitations

Note that the DB will automatically update on the next install because this only changed a function. Thus, the box is checked for DB change but it is not as important.


huss:
I accidentally did this on the main repo instead of my fork. We should delete the branch once this is merged.

spearec (approved):
Tested and fixes issue.

One note since there was a similar edit in the file, but line 418 has two keywords which are not capitalized. Since everything else is capitalized it might make sense to change these while the PR is editing this file.

huss:
> One note since there was a similar edit in the file, but line 418 has two keywords which are not capitalized. Since everything else is capitalized it might make sense to change these while the PR is editing this file.

Thanks for catching that. I updated in three places but did not find any others in a quick review.

huss:
Review by @spearec is taken as approving. This is now ready to merge.

Issue #903, opened 2023-05-19
Use reactstrap consistently and update bootstrap


spearec:
### Is your feature request related to a problem? Please describe.

OED currently uses two nearly identical packages, `reactstrap` and `react-bootstrap`. Both of these packages implement Bootstrap in React, which is how OED renders it's UI components such as buttons, modals, dropdowns, etc.

After doing some research, I found that these packages are identical in what they show to the user, and nearly identical in what they offer to developers. As far as I can tell there is nothing that can be done with one that can't be done with the other.

Overall, `react-bootstrap` is a more popular package, but `reactstrap` is simpler and considered easier to use. Since OED was using reactstrap in most cases, I have chosen to move all components using react-bootstrap to reactstrap.

### Describe the solution you'd like

OED should decide between one of these packages. This will allow developers to easily reuse code without worrying which package the component uses, as the syntax for the two packages (color, style, text, onClick, etc) are slightly different.

### Additional context

https://npmcompare.com/compare/react-bootstrap,reactstrap
https://dev.to/cassiolacerda/using-bootstrap-in-reactjs-reactstrap-or-react-bootstrap-gng
https://hackernoon.com/reactstrap-vs-react-bootstrap
https://stackoverflow.com/questions/56061590/what-is-difference-between-reactstrap-and-react-bootstrap

spearec:
Another thing that is relevant to this issue is that OED currently uses Bootstrap 4, and the corresponding react packages. It is probably a good idea to update to Bootstrap 5, as this offers a few more components and is 3 years newer. This may require some migrations, but I think everything OED uses should be backward compatible.

spearec:
Just an update to this - I've been making some slow progress on this. I have completed the migration over to reactstrap, as well as upgrading to reactstrap v9 and corresponding bootstrap v5.

Additionally, I have taken the chance while migrating modals to implement a new, in-form verification system for the create and edit modals. This system gives an admin instant feedback when their selection violates form validation requirements, and tells them what the issue is instantly, without having to click "submit," closing the modal and leaving it unclear as to where the state went.

PR #938, opened 2023-06-09
Fix issues found with recent PRs

spearec:
- Fix issue with not resetting state after finishing Cik/DB update
- fix issues with button theme capitalization

### DIFF
--- a/src/client/app/actions/admin.ts
+++ b/src/client/app/actions/admin.ts
@@ -159,8 +159,8 @@ export function submitPreferencesIfNeeded(): Thunk {
 	};
 }
 
-function updateCikAndDBViews(): t.UpdateCikAndDBViews {
-	return { type: ActionType.UpdateCikAndDBViews };
+function toggleWaitForCikAndDB(): t.ToggleWaitForCikAndDB {
+	return { type: ActionType.ToggleWaitForCikAndDB };
 }
 
 /**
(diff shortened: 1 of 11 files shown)


spearec:
One thing to note in this PR - some of the buttons were changed to use `color` instead of `variant`, as these were using the syntax for the wrong bootstrap package. This is going to be fully fixed as per #903, but the changes here are a temporary patch for 1.0.

huss (approved):
Thanks, again, to @spearec for addressing these issues that were an issue for v1.0. The changes look good and testing shows it addressed the desired issues.

huss:
> One thing to note in this PR - some of the buttons were changed to use `color` instead of `variant`, as these were using the syntax for the wrong bootstrap package. This is going to be fully fixed as per #903, but the changes here are a temporary patch for 1.0.

I see that some use color but the changes listed for this PR do not show any as changes from this PR. Am I missing something?

spearec:
> > One thing to note in this PR - some of the buttons were changed to use `color` instead of `variant`, as these were using the syntax for the wrong bootstrap package. This is going to be fully fixed as per #903, but the changes here are a temporary patch for 1.0.
> 
> I see that some use color but the changes listed for this PR do not show any as changes from this PR. Am I missing something?

Sorry if that was unclear - I meant that the issues with some buttons using `color` and others using `variant` is going to be fixed in my work on #903. This PR just fixes a couple of instances of it, but the main migration to using reactstrap is something that will happen later.

Issue #726, opened 2022-03-02
update React/Redux usage


huss:
### Is your feature request related to a problem? Please describe.

OED has been transitioning to use React hooks. The following pages already use them:

- Unit pages: src/client/app/components/units/ on several pages
- Conversion pages: src/client/app/components/conversions/ on several pages
- Meter pages: src/client/app/components/meters/ on several pages
- Group pages: src/client/app/components/groups/ on several pages
- User pages: src/client/app/components/admin/users/* but they do not deal with state
- Rate menu: src/client/app/components/GraphicRateMenuComponent.tsx
- Export page: src/client/app/components/ExportComponent.tsx
- Menus for graphic data choices: src/client/app/components/ChartDataSelectComponent.tsx
- Map menu (seems done): src/client/app/components/MapChartSelectComponent.tsx

There are other pages that are already being done or planned as part of other work:

- Page choices: src/client/app/components/HeaderButtonsComponent.tsx
- Options on left side of page: src/client/app/components/MultiSelectComponent.tsx, src/client/app/components/groups/DatasourceBoxComponent & Container, src/client/app/components/ChartLinkComponent.tsx & Container, src/client/app/components/LanguageSelectorComponent.tsx & Container, src/client/app/components/UIOptionsComponent.tsx & Container

The items remaining are:

- [x] admin preferences page (issue #886)
- [x] admin CSV upload pages (issue #887)
- [x] line, bar, compare & map graphics pages (issue #888)
- [x] tooltips (issue #889)
- [x] admin user pages (issue #890)
- [x] admin map viewing page (issue #891)
- [x] chart link (issue #892)
- [x] main containers (dashboard, footer, header, route) (issue #893)
- [x] login (issue #894)
- [x] unsaved warning but wonder if that will go away after all pages work with modal (issue awaiting decision)

There should be linked items for each of the needed items but this issue tracks the overall situation and each item should be checked off as the needed code is merged into OED.

## Brief introduction to what needs to be done

When code is transitioned to React hooks, the following can occur:

- The component/container pair has the container eliminated. The component is converted to a function if it was a class.
- The state accessed from the container and passed as props to the component is now accessed via selectState and useState.
- State updates are done via the set function returned by useState.
- useEffect is used, as needed, to cause updates to state/variables needed for rendering.

There are a number of web sources of information on doing this conversion which may be helpful.

### Describe alternatives you've considered

The conversion has already begun so it should just continue. No real alternatives.

[...]

PR #956, opened 2023-06-23
Use jsdoc rules recommended for typescript

spearec:
# Description

Merging #949 caused a few issues with jsdoc, and while resolving those I had a look through the package documentation. There is a specific recommended ruleset for typescript, which is different from what OED was using. This ruleset removes the need for types in jsdoc headers, as they are redundant since typescript is a typed language.
I also took the chance to use eslint overrides to disable jsdoc as a rule for actions, reducers, and containers, instead of disabling jsdoc on a per-file basis. This allows easier control over the rules for these files as a group.

## Type of change

## Checklist

### DIFF
--- a/src/client/app/actions/admin.ts
+++ b/src/client/app/actions/admin.ts
@@ -136,16 +136,16 @@ export function submitPreferences() {
 }
 
 /**
- * @param {State} state The redux state.
- * @returns {boolean} Whether preferences are fetching
+ * @param state The redux state.
+ * @returns Whether preferences are fetching
  */
 function shouldFetchPreferenceData(state: State): boolean {
 	return !state.admin.isFetching;
 }
 
 /**
- * @param {State} state The redux state.
- * @returns {boolean} Whether preferences are submitted
+ * @param state The redux state.
+ * @returns Whether preferences are submitted
  */
 function shouldSubmitPreferenceData(state: State): boolean {
 	return !state.admin.submitted;
@@ -176,8 +176,8 @@ function toggleWaitForCikAndDB(): t.ToggleWaitForCikAndDB {
 }
 
 /**
- * @param {State} state The redux state.
- * @returns {boolean} Whether or not the Cik and views are updating
+ * @param state The redux state.
+ * @returns Whether or not the Cik and views are updating
  */
 function shouldUpdateCikAndDBViews(state: State): boolean {
 	return !state.admin.isUpdatingCikAndDBViews;
@@ -186,8 +186,8 @@ function shouldUpdateCikAndDBViews(state: State): boolean {
 /**
  * Redo Cik and/or refresh reading views.
  * This function is called when some changes in units/conversions affect the Cik table or reading views.
- * @param {boolean} shouldRedoCik Whether to refresh Cik.
- * @param {boolean} shouldRefreshReadingViews Whether to refresh reading views.
+ * @param shouldRedoCik Whether to refresh Cik.
+ * @param shouldRefreshReadingViews Whether to refresh reading views.
  */
 export function updateCikAndDBViewsIfNeeded(shouldRedoCik: boolean, shouldRefreshReadingViews: boolean): Thunk {
 	return async (dispatch: Dispatch, getState: GetState) => {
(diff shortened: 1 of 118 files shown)


src/client/app/actions/admin.ts
```diff
@@ -79,7 +77,6 @@ function markPreferencesSubmitted(defaultMeterReadingFrequency: string): t.MarkP
 }
 
 /* eslint-enable */
```
huss: Should this enable be removed since I think the disable above was?

src/client/app/containers/maps/MapCalibrationChartDisplayContainer.ts
```diff
@@ -104,9 +102,9 @@ function mapStateToProps(state: State) {
 /* eslint-enable */
```
huss: Should this enable be removed since disable above was?

src/client/app/utils/api/ReadingsApi.ts
```diff
@@ -56,11 +56,11 @@ export default class ReadingsApi {
...
-	 * @param {number[]} meterIDs The meter IDs to get readings for
-	 * @param {TimeInterval} timeInterval The range of time to get readings for
-	 * @param {number} barWidthDays The width of the bar in days
-	 * @param {number} unitID The unit id that the reading should be returned in, i.e., the graphic unit
-	 * @returns {Promise<Object<number, Array<{reading: number, start_timestamp: number, end_timestamp: number}>>>} in sorted order
+	 * @param meterIDs The meter IDs to get readings for
+	 * @param timeInterval The range of time to get readings for
+	 * @param barWidthDays The width of the bar in days
+	 * @param unitID The unit id that the reading should be returned in, i.e., the graphic unit
+	 * @returns LineReadings in sorted order
```
huss: Should this be BarReadings?

src/client/app/utils/calculateCompare.ts
```diff
@@ -155,8 +155,8 @@ export interface ComparePeriodLabels {
 
 /**
  * Determines the human-readable names of a comparison period.
- * @param {ComparePeriod} comparePeriod the machine-readable name of the period
- * @returns {{prev: string, current: string}} human-readable names for the compare period
+ * @param comparePeriod the machine-readable name of the period
+ * @returns human-readable names for the compare period
```
huss: Would it make sense at the end to add:
as {{prev: string, current: string}}
so the person knows the format?

src/client/app/utils/calculateCompare.ts
```diff
@@ -174,10 +174,10 @@ export function getComparePeriodLabels(comparePeriod: ComparePeriod): ComparePer
 
 /**
  * Composes a label to summarize compare chart data.
- * @param {number} change the ratio of change between the current and previous period
- * @param {string} name the name of the entity being measured
- * @param {{prev: string, current: string}} labels the names of the periods in question
- * @returns {string} The label summary
+ * @param change the ratio of change between the current and previous period
+ * @param name the name of the entity being measured
+ * @param labels the names of the periods in question
```
huss: Same as ... if make change above.

src/client/app/utils/graphics.ts
```diff
@@ -11,12 +11,11 @@ import { AreaUnitType } from './getAreaUnitConversion';
...
- * @param {UnitData} selectUnitState The unit state for the selected unit for graphing
- * @param {LineGraphRate} currentSelectedRate The current selected rate
- * @param {boolean} areaNormalization Whether area normalization is enabled
- * @param {AreaUnitType} selectedAreaUnit The currently selected area unit to normalize to
- * @returns {{unitLabel: string, needsRateScaling: boolean}} y-axis label and boolean of whether rate needs to be scaled (true if does)
+ * @param selectUnitState The unit state for the selected unit for graphing
+ * @param currentSelectedRate The current selected rate
+ * @param areaNormalization Whether area normalization is enabled
+ * @param selectedAreaUnit The currently selected area unit to normalize to
+ * @returns y-axis label and boolean of whether rate needs to be scaled (true if does)
```
huss, line 18: As other comments, return the JSON structure if we are doing that.

huss (changes requested):
Thanks to @spearec for quickly updating the files based on the recent update to eslint. I know removing all those recent edits was painful so I am esp. grateful.

I've made a couple of comments but overall this looks very good. Note I added a comment in issue #726 so the TODOs in .eslintrc.json will be addressed.

spearec:
> As other comments, return the JSON structure if we are doing that. 

In this case, the JSON structure is directly noted in the function header instead of an interface, so the mouseover text of the function has the structure.

huss (approved):

PR #957, opened 2023-06-23
New readings test data added

dzheng13341:
# Description

3 test data from testing.md added to readings.js:

- should have daily points for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as BTU chained with reverse conversion
- should have daily points for 15 minute reading intervals and raw units with +-inf start/end time & C as Widget with intercept & chained
- should have daily points for 15 minute reading intervals and raw units with +-inf start/end time & C as Widget with intercept & chained & reverse conversions

Tests are successful.

## Type of change

## Checklist

## Limitations

File testing.md should be updated if pull request is accepted.
For conversion c8 ['F', 'C', true, 0.555556, -17.77778, 'Fahrenheit → Celsius'], gave rounding errors so it is better to use 
['F', 'C', true, 1/1.8, -32/1.8, 'Fahrenheit → Celsius'] for closer values.


huss:
I updated the code for changes I thought we should use. This was not the result of the work by @dzheng13341 but what I noticed during the review of previous work.

huss (approved):
Thanks to @dzheng13341 for their first PR to OED. The work was very good and no issues were found. I have updated the testing design document to mark these three tests as complete.

PR #958, opened 2023-06-23
Update reactstrap and edit/create modal validation

### LINKED ISSUE #903: Use reactstrap consistently and update bootstrap

### LINKED ISSUE #907: meter units should have displayable of none

Meter units are not visible on graphic menus so they are effectively displayable of none. Currently the admin can set it to another value during creation/editing. It would be nice to stop changes once the unit is of type meter. If there already is a displayable other than none then the admin should be warned and the value changed to none.

See above.

This is only confusing and the change does not change how OED works. Thus, the current situation could stay so this is a lower priority.

### Additional context

None

spearec:
# Description

The main scope of this PR is to remove the uses of `react-bootstrap` and replace them with `reactstrap`.
The `react-bootstrap` package was used in the header buttons component and in edit/create modal components.

In the process of updating these components, I changed their UI significantly. The header buttons are now a navbar, which supports multiple compact dropdowns and menus.
Edit/create modals now include in-form validation and are displayed in two columns on larger screens. This allows OED to give instant feedback to admins, as well as prevent them from trying to submit an invalid meter in most cases. In the process of adding validation, I also prevented meter units from being displayable. The two-column layout allows for more information to be displayed at once and will revert to the old one-column layout on small screens.

Also fixes the issue that the admin and local language preferences were the same, so an admin could not change their display language without potentially affecting the site default. In doing so, a new state section (state.options) and options dropdown in the navbar were created, which can be used to add similar user-specific settings in the future.

Fixes #903 
Fixes #907
Partially Addresses #726 (Header & Language containers removed)

## Type of change

## Checklist


huss:
Thanks to @spearec for taking this on. Overall, this looks good. I have not yet reviewed the code but did try using it. Here are some thoughts:

- The message for the sec in rate says greater than the min in unit create/edit (1 in this case) but 1 is allowed. Either the message or value should be changed to include 1.
- There are merge conflicts (sigh).
- As we have discussed (just documenting), we might reduce the number of dropdown choices in the navbar if horizontal space becomes an issue.

Once you feel this is ready then formally submit and it will have all the code reviewed.

spearec:
With the latest commits, this PR should be ready. I have no done much testing yet, so I would not deem it ready to merge until it gets tested more.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -185,121 +184,143 @@ export default function CreateConversionModalComponent(props: CreateConversionMo
...
+							<Col>
+								{/* Slope input*/}
+								<FormGroup>
+									<Label for='slope'>{translate('conversion.slope')}</Label>
+									<Input
+										id='slope'
+										name='slope'
+										type='number'
+										defaultValue={state.slope}
+										onChange={e => handleNumberChange(e)}
```
huss: Intercept is required but not slope. I think slope should also be required. On create too.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -148,104 +148,111 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+							</Col>
+							<Col>
+								{/* Intercept input */}
+								<FormGroup>
+									<Label for='intercept'>{translate('conversion.intercept')}</Label>
+									<Input
+										id='intercept'
+										name="intercept"
+										type="number"
+										defaultValue={state.intercept}
```
huss: I think intercept should be required here. If you delete it becomes zero which is okay but not as clear.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -108,12 +108,27 @@ export default function CreateGroupModalComponent(props: CreateGroupModalCompone
...
+		Area must be positive or zero
+		If area is nonzero, area unit must be set
+		Group must have at least one child (checked on submit)
+	*/
+	const [validGroup, setValidGroup] = useState(false);
+	useEffect(() => {
+		setValidGroup(
+			state.name !== '' &&
+			(state.area === 0 || (state.area > 0 && state.areaUnit !== AreaUnitType.none)) &&
+			(state.deepMeters.length > 0)
```
huss, line 123: Could we add back the part of the comment that said:
// Must have children. From a practical standpoint, this means there are no deep children.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -282,7 +280,7 @@ export default function CreateGroupModalComponent(props: CreateGroupModalCompone
 			});
 		}
 		// If any of these change then it needs to be updated.
-		// metersState normally does not change but can so include.
+		// metersState normally does not change but can so include.')}</Label>
```
huss: Unsure aoubt the ')}... at the end of the line.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -404,266 +408,292 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
...
+								<Input
+									id='name'
+									name='name'
+									type='text'
+									autoComplete='on'
+									value={groupState.name}
+									disabled/>
+							</FormGroup>
+							{/* default graphic unit display for non-admin */}
+							<FormGroup>
```
huss, line 470: This is a really small point. There is still a dropdown arrow that does nothing for graphic unit when non-admin. Other items don't have so I wondered if it is easy to make this one go away.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -120,6 +114,44 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
...
+		Name cannot be blank
+		Area must be positive or zero
+		If area is nonzero, area unit must be set
+		Reading Gap must be greater than zero
+		Reading Variation must be greater than zero
+		Reading Duplication must be between 1 and 9
+		Reading frequency cannot be blank
+		Unit and Default Graphic Unit must be set (can be to no unit)
+		Meter type must be set
+		If displayable is true and unitId is set to -99, warn admin
```
huss, line 128: All the other tests are here but not the last one. It is done in the form. Should it be here? Same for edit.

src/client/app/translations/data.js
```diff
@@ -645,6 +647,9 @@ const localeData = {
 		"groups.select": "Sélectionnez des Groupes",
 		"has.no.data": "(Need French) has no current data",
 		"has.used": "a utilisé",
+		"header.pages": "(Need French) Pages",
+		"header.options": "Options",
```
huss: Should have (Need French).

src/client/app/translations/data.js
```diff
@@ -1032,7 +1028,11 @@ const localeData = {
 		"DisplayableType.none": "(Need Spanish) none",
 		"DisplayableType.all": "(Need Spanish) all",
 		"DisplayableType.admin": "(Need Spanish) admin",
-		"duplication.invalid": "(Need Spanish) The reading duplication must be an integer between 1 and 9 but you gave ",
+		"error.bounds": "(Need Spanish) Must be between {min} and {max}.",
+		"error.displayable": "(Need Spanish) Displayable will be set to false.",
+		"error.greater": "(Need French) Must be greater than {min}.",
```
huss: French -> Spanish

src/client/app/translations/data.js
```diff
@@ -582,7 +580,11 @@ const localeData = {
 		"DisplayableType.none": "(Need French) none",
 		"DisplayableType.all": "(Need French) all",
 		"DisplayableType.admin": "(Need French) admin",
-		"duplication.invalid": "(Need French) The reading duplication must be an integer between 1 and 9 but you gave ",
+		"error.bounds": "(Need French) Must be between {min} and {max}.",
+		"error.displayable": "(Need French) Displayable will be set to false.",
+		"error.greater": "(Need French) Must be greater than {min}.",
+		"error.negative": "(Need French) Cannot be negative.",
```
huss, line 594: You did not do this but while checking I realized error.gps is missing after this. Would you be willing to add it? Also missing from Spanish.

huss (commented):
Thanks to @spearec  for more very good work. It has taken too long to get this review done so I'm posting my comments from reading the code.  I still need to carefully test it running. There are a few thoughts but nothing significant.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -148,104 +148,111 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+								})}
+							</Input>
+						</FormGroup>
+						<Row xs='1' lg='2'>
+							<Col>
+								{/* Slope input */}
+								<FormGroup>
+									<Label for='slope'>{translate('conversion.slope')}</Label>
+									<Input
+										id='slope'
```
huss, line 212: slope does not have "name='slope'" and I found that was the reason that changing only the slope did not update on save. It was because the handleNumberChange receives an empty string for target.name. The same issue was found for note.

src/client/app/components/meters/CreateMeterModalComponent.tsx
spearec: The admin is warned via validation on the displayable input component. This error doesn't prevent the admin from submitting, it just warns them that displayable will be set to false if they do submit.

src/client/app/components/groups/EditGroupModalComponent.tsx
spearec: The reason is because the input has to be a select type. Without pulling the entire units state, I'm not sure how to get the identifier of the currently selected graphic unit. There is graphicunitsstate, but that stores a set of UnitData, and I'm not sure how to easily filter that for the identifier I want.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -331,364 +317,445 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
...
+							</FormFeedback>
+						</FormGroup></Col>
+					</Row>
+					<Row xs='1' lg='2'>
+						{/* meter unit input */}
+						<Col><FormGroup>
+							<Label for='unitId'> {translate('meter.unitName')}</Label>
+							<Input
+								id='unitId'
+								name="unitId"
```
huss: A few of the strings in this file use double quotes instead of single quotes. Not sure why the linting did not catch this. This also appears in several of the edited files.

huss (changes requested):
Testing only found one small issue and double quotes. This seems to work very well. Thanks to @spearec for the careful work and very nice changes.

src/client/app/components/LanguageSelectorComponent.tsx
```diff
@@ -3,97 +3,52 @@
...
+						Français
+					</DropdownItem>
+					<DropdownItem
+						onClick={() => dispatch(updateSelectedLanguage(LanguageTypes.es))}
+						disabled={selectedLanguage === LanguageTypes.es}>
+						Español
+					</DropdownItem>
+					<DropdownItem divider />
+					<DropdownItem
+						href={'https://openenergydashboard.github.io/help/' + version + '/language.html'}>
```
huss: Another thought. From src/client/app/components/TooltipHelpComponent.tsx you can change the BASE_URL to get the help pages from another location. How about changing that name to BASE_HELP_URL and exporting it. Then this value could be used here and in HeaderButtonsComponent.tsx so it would follow the same idea. This makes testing new help versions easier as it can be somewhere else on all pages. We may also allow the admin to change the help location in the future to allow for local deployment if desired, e.g., slow internet access. What do you think @spearec?

spearec: That is a good idea, I meant to make note of this but never wrote it down anywhere.

spearec: Adding to this - if we do add support for admins to host their own help pages, then I think it makes sense to make that URL  either a .env variable or an admin preferences setting, instead of compiled into the code directly. Should an issue be opened that addresses this?

huss: Feel free to add an issue if you want. I was waiting until after the admin page changes were complete to consider doing this. We don't yet have a request for this but it was an idea.

src/client/app/components/groups/EditGroupModalComponent.tsx
huss: The TODO you left is fine.

huss:
Thanks to @spearec for addressing all the comments and even fixing some other issues they found. Everything looks good. I changed the wording on one message but that was minor. This is ready to merge and will be included in v1.0.0.

huss (approved):

PR #959, opened 2023-06-24
Testing

sparakala21:
# Description

I added 2 tests for:
should have daily points for 15 minute reading intervals and raw units with +-inf start/end time & C as F with intercept

should have daily points for 15 minute reading intervals and raw units with +-inf start/end time & C as F with intercept reverse condition


huss (approved):
Thanks to @sparakala21 for their first PR to OED. The code looks good and runs as expected. I've updated the testing design doc to reflect the two tests completed. I auto formatted in VSC that changed two minor formats.

PR #975, opened 2023-07-31
Error bar

### LINKED ISSUE #594: data range on line graphics

Elias0127:
# Description

We grabbed the min and max data from the database and added it to the line graph as error bars. Note that we did not write the SQL query (src/server/sql/reading/create_reading_views.sql) that was responsible for creating the min and max data in the database. The code was already written, and we only uncommented it.

This is a joint work
Co-authored - by JesseVarGar

Fixes #594

## Type of change

## Checklist

## Limitations

Adding min and max to the database caused a mismatch with the tests and resulted in them failing.


src/client/app/actions/lineReadings.ts
```diff
@@ -159,4 +158,4 @@ export function fetchNeededLineReadings(timeInterval: TimeInterval, unitID: numb
 
 		return Promise.all(promises);
 	};
-}
+}
```
spearec: Generally, OED uses newlines at EOF and between functions. There are a few cases of removed newlines, which I assume are due to a differing vscode configuration. These newlines should be kept to match the rest of the codebase.

src/client/app/components/ErrorBarComponent.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { State } from '../types/redux/state';
+import { toggleShowMinMax } from '../actions/graph';
+import translate from '../utils/translate';
+import TooltipMarkerComponent from './TooltipMarkerComponent';
+
+/**
+ * React Component rendering an Error Bar checkbox for toggle operation.
+ * @returns Error Bar checkbox with tooltip and label
+ */
+const ErrorBarComponent = () => {
```
spearec: All other components in OED use `export default function thisComponent()`, so although this is functionally similar, this should be kept consistent with the codebase.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -113,6 +129,7 @@ function mapStateToProps(state: State) {
 							color: getGraphColor(colorID, DataType.Meter)
 						}
 					});
+
```
spearec: As noted before, there's a few cases of out-of-place newlines either added or removed.

src/client/app/translations/data.js
```diff
@@ -149,6 +149,7 @@ const localeData = {
 		"edit.unit": "Edit Unit",
 		"email": "Email",
 		"enable": "Enable",
+		"error.bar": "Error Bar",
```
spearec: I would recommend this includes a verb, such as "Show error bars". We try to have the UI be as descriptive as possible. See the checkboxes for area normalization and bar stacking for examples.

src/client/app/translations/data.js
```diff
@@ -233,6 +234,7 @@ const localeData = {
...
 		"help.home.compare.interval.tip": "Selects the time interval (Day, Week or 4 Weeks) to compare for current to previous.  Please see {link} for further details and information.",
 		"help.home.compare.sort.tip": "Allows user to select the order of multiple comparison graphs to be Alphabetical (by name), Ascending (greatest to least reduction in usage) and Descending (least to greatest reduction in usage).  Please see {link} for further details and information.",
+		"help.home.error.bar": "Toggles error bar with min and max value.Please visit {link} for further details and information.",
```
spearec: Error bars should be plural and there is no space after the period.

src/client/app/types/redux/graph.ts
```diff
@@ -56,6 +56,12 @@ export interface ChangeChartToRenderAction {
 export interface ToggleAreaNormalizationAction {
 	type: ActionType.ToggleAreaNormalization;
 }
+export interface ToggleShowMinMaxAction {
+	type: ActionType.ToggleShowMinMax;
+}
+export interface ToggleShowMinMaxAction {
```
spearec, line 60: Duplicate entry for the interface definition.

src/server/migrations/1.0.0-1.1.0/index.js
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const database = require('../../models/database');
+const sqlFile = database.sqlFile;
+
+module.exports = {
```
spearec, line 8: Migration was added, but also needs to be added to `registerMigration.js`

src/client/app/containers/LineChartContainer.ts
```diff
@@ -18,6 +18,8 @@ function mapStateToProps(state: State) {
 	const timeInterval = state.graph.timeInterval;
 	const unitID = state.graph.selectedUnit;
 	const datasets: any[] = [];
+	const yMinData: number[] = [];
```
spearec: Since these arrays are per-meter, these need to be defined inside of the for loop by meterID below (see xData and yData arrays). When defined here, these result in a failure to add error bars when more than one meter is graphed.
image

src/client/app/containers/LineChartContainer.ts
```diff
@@ -81,7 +82,16 @@ function mapStateToProps(state: State) {
 						xData.push(timeReading.format('YYYY-MM-DD HH:mm:ss'));
 						const readingValue = reading.reading * scaling;
 						yData.push(readingValue);
-						hoverText.push(`<b> ${timeReading.format('ddd, ll LTS')} </b> <br> ${label}: ${readingValue.toPrecision(6)} ${unitLabel}`);
+						const minValue = reading.min * scaling;
+						yMinData.push(minValue);
+						const maxValue = reading.max * scaling;
+						yMaxData.push(maxValue);
+						if (state.graph.showMinMax) {
```
spearec: As an optimization, the filling of these arrays should be inside of the if statement. If showMinMax is false, these arrays are not used, so it is wasteful to fill them with data.

spearec (changes requested):
Overall, great work from @Elias0127 and @JesseVarGar. This is a good implementation of the work that was done on the backend. I've got a few small code styling comments, and there are also two issues with the implementation.

As noted below, the min/max points array is kept across meters, which leads to problems when displaying bars on multiple meters.

Additionally, there should not be error bars for meter-resolution data. Currently, the code tries to display error bars on this data, which leads to errors in the console as well as weird plotly output.
image

src/client/app/actions/lineReadings.ts
```diff
@@ -130,7 +130,6 @@ function fetchGroupLineReadings(groupIDs: number[], timeInterval: TimeInterval,
 		dispatch(receiveGroupLineReadings(groupIDs, timeInterval, unitID, groupLineReadings));
 	};
 }
```
huss: Could the blank line between function be put back.

src/client/app/components/UIOptionsComponent.tsx
```diff
@@ -88,8 +89,11 @@ class UIOptionsComponent extends React.Component<UIOptionsPropsWithIntl, UIOptio
 				<ChartSelectComponent />
 				<ChartDataSelectComponent />
 				<GraphicRateMenuComponent />
-				<AreaUnitSelectComponent/>
-
+				<AreaUnitSelectComponent />
+				{/* Controls error bar, specifically for the line chart.  */}
```
huss: Really small point but there are two spaces before the */. Could you please make it one.

src/client/app/translations/data.js
```diff
@@ -602,6 +604,7 @@ const localeData = {
 		"edit.unit": "(need French) Edit Unit",
 		"email": "E-mail",
 		"enable": "Activer",
+		"error.bar": "(Need French) Error Bar",
```
huss: This text differs from en. It should be as in en. Note comments on fr apply to es too.

src/client/app/translations/data.js
```diff
@@ -686,6 +689,7 @@ const localeData = {
...
 		"help.home.compare.interval.tip": "(Need French) Selects the time interval (Day, Week or 4 Weeks) to compare for current to previous.  Please see {link} for further details and information.",
 		"help.home.compare.sort.tip": "(Need French) Allows user to select the order of multiple comparison graphs to be Alphabetical (by name), Ascending (greatest to least reduction in usage) and Descending (least to greatest reduction in usage).  Please see {link} for further details and information.",
+		"help.home.error.bar": "(Need French) Toggles error bar with min and max value.Please visit {link} for further details and information.",
```
huss: Text does not have a space after the . and also please check for other differences with en.

While it could go either way, I think you should make it "error bars" per the comment from @spearec (was marked resolved).

src/client/app/utils/exportData.ts
```diff
@@ -24,7 +36,14 @@ function convertToCSV(readings: LineReading[], meter: string, unitLabel: string,
 		// somewhat universal way of formatting.
 		const startTimeStamp = moment.utc(reading.startTimestamp).format('YYYY-MM-DD HH:mm:ss');
 		const endTimeStamp = moment.utc(reading.endTimestamp).format('YYYY-MM-DD HH:mm:ss');
-		csvOutput += `${value},${startTimeStamp},${endTimeStamp}\n`;
+		csvOutput += `${value},${startTimeStamp},${endTimeStamp}`;
+		// Populate the min and max columns only for LineReading types.
```
huss: I might change this comment to be: "Include min and max in export if appropraite". The comment above specifies the conditions where this should be the case. The currrent comment leaves out that you must be showing error bars. I think it might be best to not duplicate the conditions so they don't get out of sync.

src/server/models/Reading.js
```diff
@@ -240,16 +240,15 @@ class Reading {
...
 		 */
 		const allMeterLineReadings = await conn.func('meter_line_readings_unit',
 			[meterIDs, graphicUnitId, fromTimestamp || '-infinity', toTimestamp || 'infinity', 'auto', maxRawPoints, maxHourlyPoints]
 		);
-
 		const readingsByMeterID = mapToObject(meterIDs, () => []);
 		for (const row of allMeterLineReadings) {
 			readingsByMeterID[row.meter_id].push(
-				{ reading_rate: row.reading_rate, start_timestamp: row.start_timestamp, end_timestamp: row.end_timestamp }
+				{reading_rate: row.reading_rate, min_rate: row.min_rate, max_rate: row.max_rate, start_timestamp: row.start_timestamp, end_timestamp: row.end_timestamp }
```
huss: I think the space after { should be put back.

huss:
@Elias0127 From the records I have, you have yet to sign the [CLA ](https://openenergydashboard.github.io/developer/cla.html)for OED. OED needs a signed CLA to accept this work. If you think my records are in error then please let me know.

huss:
Before this PR I pushed a commit to get the tests working. I want to note that the fix did not include testing the new error bar values returned. I have opened issue #984 so the tests will be enhanced in the future. For now, the PR will be manually tested.

huss:
There is a slightly annoying behavior where adding error bars to some meters shifts the slider bars for the time range. What is interesting is the actual range displayed is not altered but the slider can now be zoomed out to earlier and later dates. If you graph test15MinSin over all time with and without error bars you can see the effect. I tried to debug this and found that simply commenting out the include of the error_y but leaving all the calculations removes the shift. This was consistent with finding that the graphic values seemed the same for the values with and without the error bars except for the error ranges. I have looked at the data in the debugger and it only seems to change the y-axis values so why the x-axis is effected is unclear. It isn't a big issue and the graph is correct. I wanted to document this and also see if someone else has any ideas.

huss:
Per agreement with the developers of this PR, I just added min/max to the chart link. It should work correctly now and reproduce the desired graphic. Others should test that it is correct and code  is okay.

Elias0127:
> @Elias0127 From the records I have, you have yet to sign the [CLA ](https://openenergydashboard.github.io/developer/cla.html)for OED. OED needs a signed CLA to accept this work. If you think my records are in error then please let me know.

huss:
Given the comments by @spearec, I have removed min/max for any meter that is raw data. Other meters still show min/max. Again, others should check this out to agree that this is a good fix.

src/client/app/containers/LineChartContainer.ts
```diff
@@ -81,7 +83,25 @@ function mapStateToProps(state: State) {
...
+							// We want to show min/max. Note if the data is raw for this meter then all the min/max values are null.
+							// In this case we still push the min/max but plotly will not show them. This is a little extra work
+							// but makes the code cleaner.
+							const minValue = reading.min * scaling;
+							yMinData.push(minValue);
+							const maxValue = reading.max * scaling;
+							yMaxData.push(maxValue);
+							if (reading.min != null) {
+								// There are min/max values so put in hover.
+								hoverText.push(`${hoverStart} <br> min: ${minValue.toPrecision(6)} <br> max: ${maxValue.toPrecision(6)}`);
```
huss: The strings min and max should be internationalized.

huss (changes requested):
Congratulation to Elias and Jesse on their first PR to OED. Thanks for through work. My testing did not find any issues. I made a few comments that should be addressed. I also note:

- I made a few commits to the PR (see comments to PR) that others should look at and verify to be correct.
- I edited the description to remove the [ ] around the issue number so GitHub properly links.

huss:
>     * I made a few commits to the PR (see comments to PR) that others should look at and verify to be correct.

Thanks to @spearec for catching my mistake. I missed all the zero value at the bottom of the graphic. I believe their fix is correct.

src/client/app/reducers/graph.ts
```diff
@@ -93,6 +94,12 @@ export default function graph(state = defaultState, action: GraphAction) {
 				...state,
 				areaNormalization: !state.areaNormalization
 			};
+		case ActionType.ToggleShowMinMax:
+			return {
+				...state,
+				showMinMax: !state.showMinMax
+			};
+
```
spearec: extra newline here

src/client/app/translations/data.js
```diff
@@ -686,6 +689,7 @@ const localeData = {
...
 		"help.home.compare.interval.tip": "(Need French) Selects the time interval (Day, Week or 4 Weeks) to compare for current to previous.  Please see {link} for further details and information.",
 		"help.home.compare.sort.tip": "(Need French) Allows user to select the order of multiple comparison graphs to be Alphabetical (by name), Ascending (greatest to least reduction in usage) and Descending (least to greatest reduction in usage).  Please see {link} for further details and information.",
+		"help.home.error.bar": "(Need French) Toggles error bar with min and max value. Please visit {link} for further details and information.",
```
spearec: This should match the english text (Toggle error bars)

src/client/app/translations/data.js
```diff
@@ -1139,6 +1144,7 @@ const localeData = {
...
 		"help.home.compare.interval.tip": "Selecciona el intervalo de tiempo (Día, Semana o Cuatro semanas) para comparar el actual con el anterior.  Por favor, vea {link} para más detalles e información.",
 		"help.home.compare.sort.tip": "Permite al usuario seleccionar el orden de los gráficos de comparación múltiple para que sean alfabéticos (por nombre), ascendentes (de mayor a menor reducción de uso) y descendentes (de menor a mayor reducción de uso). Por favor, vea {link} para más detalles e información.",
+		"help.home.error.bar": "(Need Spanish) Toggles error bar with min and max value.Please visit {link} for further details and information.",
```
spearec: This should match the english text (Toggle error bars)

src/client/app/containers/LineChartContainer.ts
```diff
@@ -81,7 +83,20 @@ function mapStateToProps(state: State) {
...
+						const hoverStart = `<b> ${timeReading.format('ddd, ll LTS')} </b> <br> ${label}: ${readingValue.toPrecision(6)} ${unitLabel}`;
+						if (state.graph.showMinMax && reading.max != null) {
+							// We want to show min/max. Note if the data is raw for this meter then all the min/max values are null.
+							// In this case we still push the min/max but plotly will not show them. This is a little extra work
+							// but makes the code cleaner.
+							const minValue = reading.min * scaling;
+							yMinData.push(minValue);
+							const maxValue = reading.max * scaling;
+							yMaxData.push(maxValue);
+							hoverText.push(`${hoverStart} <br> min: ${minValue.toPrecision(6)} <br> max: ${maxValue.toPrecision(6)}`);
```
spearec: As Steve said, min and max need to be internationalized. This can be easily done using the `translate('string')` function inline.

spearec (changes requested):
Most of the above comments have been addressed. I just have a couple of minor things to note.

src/client/app/containers/LineChartContainer.ts
Elias0127: Sorry for the delay. It’s done now.

huss (approved):

huss:
I have reviewed all the requested changes for both myself and @spearec. Everything looks good and this PR is ready to merge. I appreciate the efforts of @Elias0127 and @JesseVarGar and their understanding for the time it took us to do the final reviews.

PR #981, opened 2023-08-03
properly update moment locale based on selectedLanguage

spearec:
# Description

Updates moment locale based on selectedLanguage instead of defaultLanguage. This was an oversight in the reactstrap PR.

## Type of change

## Checklist


src/client/app/actions/admin.ts
```diff
@@ -98,6 +96,9 @@ function fetchPreferences(): Thunk {
 				}
 				if (preferences.defaultLanguage !== state.options.selectedLanguage) {
```
huss, line 97: The logic seems to work fine but I think having a comment(s) about what the cases are and what/why the value specified is set would be good.

huss (approved):
It seems to work fine. I just made a minor suggestion for commenting.

Issue #940, opened 2023-06-14
update entry of group, units & conversion data


huss:
### Is your feature request related to a problem? Please describe.

PR #828 included a change in how the code specifies meter data to become key:value pairs when using insertMeters (see src/server/data/automatedTestingData.js). Similar style should be used for groups, units & conversions. This impacts inserting the standard units and test data.

### Describe the solution you'd like

The code is modified as described. It can be done separately for each item if preferred.

### Describe alternatives you've considered

The code works but it is harder to keep track of the values with the current code.

### Additional context

The sooner we do this then the less code that needs to be updated for inserting values.

huss:
PR #983 did conversions & units so only groups remains.

PR #983, opened 2023-08-03
Testing format changes

dzheng13341:
# Description

- Changed insertUnits() and insertConversions() so when adding inputs into test, it now matches with meterData instead of being arrays in an array, and also makes it easier to see/understand and change test data.
- Subsequently changed all the current tests to match the new formatting.
- Tests returned successful.

Partly addresses issue #940 because it does not do groups.
## Type of change

## Checklist


spearec (approved):
Changes look good. Both `npm run test` and `npm run testData` work fine, and I believe those are the only scripts impacted by these changes.
This certainly improves readability of this part of the code.

src/server/util/insertData.js
```diff
@@ -20,22 +20,22 @@ const fs = require('fs').promises;
 async function insertUnits(unitsToInsert, update = false, conn) {
 	await Promise.all(unitsToInsert.map(
```
huss, line 21: The meter code checks that required keys are provided. It would be nice to do that here. I think you must give: name, unitRepresent, typeOfUnit, unitIndex, displayable, preferredDisplay due to what is set in the Unit constructor and what is required by the DB not to be null.

src/server/util/insertData.js
```diff
@@ -75,10 +175,10 @@ async function insertStandardUnits(conn) {
 async function insertConversions(conversionsToInsert, conn) {
 	await Promise.all(conversionsToInsert.map(
 		async conversionData => {
```
huss: Similarly to units for required values: sourceId, destinationId, bidirectional, slope, intercept. The last two are not required by database but you would have issues if not provided.

Also, could you rename the sourceId/destinationId to be sourceName/DestinationName since they are the names and not the Id being passed.

huss (changes requested):
Thanks to @dzheng13341 for their first PR. The work is good. I made a couple of comments that I would like addressed. I note that the developer also checked that a cold start would work since these functions are used to initialize OED. 

These are also used by the website setup so I will take care of updating that once this PR clears. That is not done within the production code of OED.

src/server/util/insertData.js
dzheng13341: I have created checks for name, unitRepresent, typeOfUnit, displayable, and preferredDisplay. None of the tests provided a unitIndex so it seems it is not a required key

src/server/models/Conversion.js
```diff
@@ -7,16 +7,16 @@ const sqlFile = database.sqlFile;
 
 class Conversion {
 	/**
-	 * @param {*} sourceId The unit id of the source.
-	 * @param {*} destinationId The unit id of the destination.
+	 * @param {*} sourceName The unit id of the source.
```
huss: It seems some files unrelated to the changes in input of conversions had the name of variables changed. In this file OED is using the id and not the name. The variable names should not have been changed and should Id not name.

src/server/routes/conversions.js
```diff
@@ -13,7 +13,7 @@ const router = express.Router();
 
 function formatConversionForResponse(item) {
 	return {
-		sourceId: item.sourceId, destinationId: item.destinationId, bidirectional: item.bidirectional, slope: item.slope, intercept: item.intercept, note: item.note
+		sourceName: item.sourceName, destinationName: item.destinationName, bidirectional: item.bidirectional, slope: item.slope, intercept: item.intercept, note: item.note
```
huss: This is another file where the variable names should have remained as Id.

src/server/services/graph/createConversionArrays.js
```diff
@@ -71,10 +71,10 @@ async function createCikArray(graph, conn) {
 	await assignIndex(destinations, conn);
 	for (const source of sources) {
 		for (const destination of destinations) {
-			const sourceId = source.id;
-			const destinationId = destination.id;
+			const sourceName = source.id;
```
huss: This is another file where the variable names should have remained as Id.

src/server/services/graph/createConversionGraph.js
```diff
@@ -21,9 +21,9 @@ async function createConversionGraph(conn) {
 
 	const conversions = await Conversion.getAll(conn);
 	for (let i = 0; i < conversions.length; ++i) {
-		graph.addLink(conversions[i].sourceId, conversions[i].destinationId);
+		graph.addLink(conversions[i].sourceName, conversions[i].destinationName);
```
huss: This is another file where the variable names should have remained as Id.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -9,17 +9,17 @@ const { getAllPaths } = require('./createConversionGraph');
 
 /**
  * Adds the new unit and conversions to the database and the conversion graph.
- * @param {*} sourceId The source unit's id.
- * @param {*} destinationId The destination unit's id.
+ * @param {*} sourceName The source unit's id.
```
huss: This is another file where the variable names should have remained as Id.

src/server/services/graph/pathConversion.js
```diff
@@ -88,9 +88,9 @@ async function pathConversion(path, conn) {
 	let intercept = 0;
 	let suffix = '';
 	for (let i = 0; i < path.length - 1; ++i) {
-		const sourceId = path[i].id;
-		const destinationId = path[i + 1].id;
-		const [newSlope, newIntercept, newSuffix] = await conversionValues(sourceId, destinationId, conn);
+		const sourceName = path[i].id;
```
huss: This is another file where the variable names should have remained as Id.

src/server/sql/conversion/insert_new_conversion.sql
```diff
@@ -3,4 +3,4 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 INSERT INTO conversions(source_id, destination_id, bidirectional, slope, intercept, note)
-VALUES (${sourceId}, ${destinationId}, ${bidirectional}, ${slope}, ${intercept}, ${note});
+VALUES (${sourceName}, ${destinationName}, ${bidirectional}, ${slope}, ${intercept}, ${note});
```
huss: This is another file where the variable names should have remained as Id.

src/server/sql/conversion/update_conversion.sql
```diff
@@ -7,4 +7,4 @@ UPDATE conversions
         slope = ${slope},
         intercept = ${intercept},
         note = ${note}
-	WHERE source_id = ${sourceId} AND destination_id = ${destinationId};
+	WHERE source_id = ${sourceName} AND destination_id = ${destinationName};
```
huss: This is another file where the variable names should have remained as Id.

src/server/test/db/conversionTests.js
```diff
@@ -12,8 +12,8 @@ const Unit = require('../../models/Unit');
  * @param {*} actual The actual conversion.
  */
 function expectConversionToBeEquivalent(expected, actual) {
-	expect(actual).to.have.property('sourceId', expected.sourceId);
-	expect(actual).to.have.property('destinationId', expected.destinationId);
+	expect(actual).to.have.property('sourceName', expected.sourceName);
```
huss: This is another file where the variable names should have remained as Id.

src/server/util/insertData.js
```diff
@@ -20,26 +20,51 @@ const fs = require('fs').promises;
...
+					dbUnit.unitRepresent = unitData.unitRepresent;
+					dbUnit.secInRate = unitData.secInRate;
+					dbUnit.typeOfUnit = unitData.typeOfUnit;
+					dbUnit.suffix = unitData.suffix;
+					dbUnit.displayable = unitData.displayable;
+					dbUnit.preferredDisplay = unitData.preferredDisplay;
+					dbUnit.note = unitData.note;
+					// Should update even though exists.
+					await dbUnit.update(conn);
+					// Otherwise do not update.
```
huss: I'm unclear on why this comment moved. The line above just did the update. I think it was correct where it was.

src/server/util/insertData.js
```diff
@@ -20,26 +20,51 @@ const fs = require('fs').promises;
...
+			let ok = true;
+			for (let i = 0; i < unitsToInsert.length; ++i) {
+				// Meter key/value pairs for the current meter.
+				const uData = unitsToInsert[i];
+		
+				// Check that needed keys are there.
+				const requiredKeys = ['name', 'unitRepresent', 'typeOfUnit', 'displayable', 'preferredDisplay'];
+				ok = true;
+				requiredKeys.forEach(key => {
+					if ((key == 'name') && (typeof uData[key] !== 'string')){
```
huss: A few thoughts on these new checks:

1. The formatting is off. I suggest you use the standard Visual Studio Code formatting if you are using that IED.
2. These checks differ from the way meter does it that was given as an example. I can understand that checking the type of the value could be valuable. However, we have not been doing that and only doing in some cases seems inconsistent. It might be easiest to remove them as there are lots of ways that the data could be incorrect. If you want to keep this then some structure of requiredKeys that has the name and type that is then checked the same way for all keys seems better. You should also update meters to be consistent.
3. The first and second tests seem to do the same thing but are separated out. I'm unclear why. This will go away if you replace with the way meter did it or if you do consistent checks for all keys.

src/server/util/insertData.js
```diff
@@ -75,10 +200,35 @@ async function insertStandardUnits(conn) {
...
-				await new Conversion(sourceId, destinationId, conversionData[2], conversionData[3], conversionData[4], conversionData[5]).insert(conn);
+			let ok = true;
+			for (let i = 0; i < conversionsToInsert.length; ++i) {
+				// Meter key/value pairs for the current meter.
+				const cData = conversionsToInsert[i];
+		
+				// Check that needed keys are there.
+				const requiredKeys = ['sourceName', 'destinationName', 'bidirectional', 'slope', 'intercept'];
+				ok = true;
+				requiredKeys.forEach(key => {
```
huss: See comment on units as they apply here.

huss (changes requested):
I have a few comments and thoughts on the recent changes. Could you please address them before I do a careful check on the code? Thanks.

src/server/util/insertData.js
dzheng13341: Understood. The checks differ from the way meters does it because some nonstring keys return values such as 0 or false which will cause the program to think the key is missing. I could only think of the basic workaround which is to check the type of value, if there is any suggestions on how to do these checks better or change the checks meters does to fit units and conversions please let me know.

huss: While trying to figure out how to check the keys in a better way I ran into issues I found in the original code. I believe the issue was not seen because the same conversions were being inserted many times but the code skipped the insert after the first time. Below is my revised code to insert conversions that I believe is correct. The same basic logic should be used for units. The improved log message should be used in meters but it is important not to use a map function in that case so only change the message. You are welcome to make these changes or I can do them if you prefer. 

```js
async function insertConversions(conversionsToInsert, conn) {
	await Promise.all(await Promise.all(conversionsToInsert.map(
		async (conversionData, index) => {
			// Check that needed keys are there.
			const requiredKeys = ['sourceName', 'destinationName', 'bidirectional', 'slope', 'intercept'];
			ok = true;
			requiredKeys.forEach(key => {
				if (!conversionData.hasOwnProperty(key)) {
					console.log(`********key "${key}" is required but missing so conversion number ${index} not processed with values:`, conversionData);
					// Don't insert
					ok = false;
				}
			})
			if (ok) {
				const sourceName = (await Unit.getByName(conversionData.sourceName, conn)).id;
				const destinationName = (await Unit.getByName(conversionData.destinationName, conn)).id;
				if (await Conversion.getBySourceDestination(sourceName, destinationName, conn) === null) {
					await new Conversion(sourceName, destinationName, conversionData.bidirectional, conversionData.slope, conversionData.intercept, conversionData.note).insert(conn);
				}
			}
		}
	)))
}
```

dzheng13341: Sorry for the delay. I have updated insertConversions to match theses changes.

src/server/util/insertData.js
huss: I made this change.

huss (approved):

Issue #947, opened 2023-06-20
cookie-parser usage


huss:
### Is your feature request related to a problem? Please describe.

While working on PR #942, it was unclear if the package was actually used by OED. The only use is in src/server/app.js and I could not find any place that cookie was used elsewhere. OED does not seem to set any cookies. I commented it out and did not find any issues. Someone should see if it is really used (and hopefully with more knowledge of this than I have).

### Describe the solution you'd like

Remove package if not used.

### Describe alternatives you've considered

It stays in OED.

### Additional context

None

salhernandez:
@huss I would like for this issue to be assigned to me so that the students I'm mentoring for CodeDay Labs can work on it.

huss:
> @huss I would like for this issue to be assigned to me so that the students I'm mentoring for CodeDay Labs can work on it.

Done. I look forward to your contributions. Please let me know if you need any help.

spearec:
@salhernandez I'm not sure if it will be related, but #958 adds a new user-specific language setting. Currently, this is set to the site default every time OED is loaded, but it could be useful to store this value in a cookie, so the user's preference is saved when they reload the site. 
Even if OED chooses to remove this package for now, it would be good to have documentation on how to add it back in if we decide we want to start using cookies.

salhernandez:
@spearec Thanks for the info!

PR #985, opened 2023-08-07
Solved #947 by updating the usage of cookie-parser dependencies

moll0928:
# Description

Fixes #[issue]

## Type of change

## Checklist

## Limitations

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "bcryptjs": "~2.4.3",
     "body-parser": "~1.20.2",
     "bootstrap": "~5.2.3",
-    "cookie-parser": "~1.4.6",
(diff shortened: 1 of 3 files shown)


moll0928:
I solved the problem #947 by updating the cookie parser dependencies. The original version of the dependency is 1.4.6, which causes the problem. First, I changed the dependency to the lower version 1.4.4, and after the testing, I found out that the website worked smoothly. So I directly deleted it, and did the test again. The whole website works fine. Thus this dependency is redundant and can be deleted.

spearec:
It looks like you've updated the pull request file in the codebase, instead of the pull request message itself on git. If you could fix this and update the PR instead we can move ahead with approving this PR.

PR #989, opened 2023-08-09
Enable reading validation via admin input

### LINKED ISSUE #511: set checks in pipeline
This relates to the merging of PR #482. The conditionSet is undefined in insertMamacData, insertMetasysData & insertObviusData. We need to decide what to put in for these so we actually do testing on inputted data. This may require new admin input to decide.

carlsonrob:
# Description
Currently, readings are not validated upon upload. This PR enables an admin to define default test values, which are then executed when readings are uploaded. The admin panel, meter modal (create/edit), meter object and database have been modified to accommodate these values. The conditionSet object was previously undefined, but can now be determined using these values and enable readings to be validated.
@TrongQuocLe @codehomie1

Fixes #511

## Type of change

## Checklist

## Limitations
Written test cases need to be created.


src/server/models/Meter.js
```diff
@@ -236,7 +251,7 @@ class Meter {
 		readingDuplication = this.readingDuplication, timeSort = this.timeSort, endOnlyTime = this.endOnlyTime,
 		reading = this.reading, startTimestamp = this.startTimestamp, endTimestamp = this.endTimestamp,
 		previousEnd = this.previousEnd, unitId = this.unitId, defaultGraphicUnit = this.defaultGraphicUnit, areaUnit = this.areaUnit,
-		readingFrequency = this.readingFrequency) {
+		readingFrequency = this.readingFrequency, minVal = this.minVal, maxVal = this.maxVal, minDate = this.minDate, maxDate = this.maxDate, maxError = this.maxError, disableChecks = this.disableChecks) {
```
huss: OED wants lines to be 150 characters or less. Please split this line.

src/server/sql/preferences/insert_default_row.sql
```diff
@@ -15,9 +15,10 @@ IF NOT EXISTS(SELECT *
 	THEN
 	INSERT INTO preferences (display_title, default_chart_to_render, default_bar_stacking,
 	default_language, default_timezone, default_warning_file_size, default_file_size_limit,
-	default_area_normalization, default_area_unit, default_meter_reading_frequency) 
-	VALUES ('', 'line', FALSE, 'en', NULL, 5, 25, FALSE, 'meters', '00:15:00');
-
+	default_area_normalization, default_area_unit, default_meter_reading_frequency, 
+	default_meter_minimum_value, default_meter_maximum_value, default_meter_minimum_date, 
+	default_meter_maximum_date, default_meter_reading_gap, default_meter_maximum_errors) 
+	VALUES ('', 'line', FALSE, 'en', NULL, 5, 25, FALSE, 'meters', '00:15:00', -9007199254740991, 9007199254740991, '1970-01-01 00:00:00+00:00', '6970-01-01 00:00:00+00:00', 0, 75);
```
huss: OED wants lines to be 150 characters or less. Please split this line.

src/client/app/actions/admin.ts
```diff
@@ -104,13 +128,45 @@ function fetchPreferences(): Thunk {
...
+		state.admin.defaultMeterMinimumValue >= MIN_VAL &&
+		state.admin.defaultMeterMinimumValue <= state.admin.defaultMeterMaximumValue &&
+		state.admin.defaultMeterMinimumValue <= MAX_VAL &&
+		state.admin.defaultMeterMinimumDate !== '' &&
+		state.admin.defaultMeterMaximumDate !== '' &&
+		moment(state.admin.defaultMeterMinimumDate) >= MIN_DATE_MOMENT &&
+		moment(state.admin.defaultMeterMaximumDate) <= MAX_DATE_MOMENT &&
+		moment(state.admin.defaultMeterMinimumDate) <= moment(state.admin.defaultMeterMaximumDate) &&
+		(state.admin.defaultMeterMaximumErrors >= 0 && state.admin.defaultMeterMaximumErrors <= 75)
+	) return true;
```
huss: OED likes to have  braces around if statement blocks. Also, in this case, I would use an else statement for the false case.

src/client/app/actions/admin.ts
```diff
@@ -104,13 +128,45 @@ function fetchPreferences(): Thunk {
 	};
 }
 
+/* Because there is no warn for invalid data in admin pannel, I use function validPreferences(state: State)
```
huss: Could you make the comment about doing the tests here into a TODO so that it is done as other modal pages are when this page is changed? Also, pannel -> panel.

src/client/app/actions/admin.ts
```diff
@@ -104,13 +128,45 @@ function fetchPreferences(): Thunk {
...
+	return false;
+}
 /**
  * Submits preferences stored in the state to the API to be stored in the database
  */
 export function submitPreferences() {
 	return async (dispatch: Dispatch, getState: GetState) => {
 		const state = getState();
 		try {
+			if (!validPreferences(state)) throw new Error('invalid input');
```
huss: if should have braces.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -379,6 +470,37 @@ class PreferencesComponent extends React.Component<PreferencesPropsWithIntl> {
...
+		this.updateUnsavedChanges();
+	}
+
+	private handleDefaultMeterReadingGapChange(e: { target: HTMLInputElement; }) {
+		this.props.updateDefaultMeterReadingGap(parseFloat(e.target.value));
+		this.updateUnsavedChanges();
+	}
+
+	private handleDefaultMeterMaximumErrorsChange(e: { target: HTMLInputElement; }) {
+		this.props.updateDefaultMeterMaximumErrors(parseFloat(e.target.value));
```
huss: Should this parse to an int? I know JS thinks about number but it would be a stronger test as the code will not really do a FP value.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -74,7 +75,13 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
...
 		previousEnd: '',
 		areaUnit: AreaUnitType.none,
-		readingFrequency: adminState.defaultMeterReadingFrequency
+		readingFrequency: adminState.defaultMeterReadingFrequency,
+		minVal: adminState.defaultMeterMinimumValue,
+		maxVal: adminState.defaultMeterMaximumValue,
+		minDate: adminState.defaultMeterMinimumDate,
+		maxDate: adminState.defaultMeterMaximumDate,
+		maxError: adminState.defaultMeterMaximumErrors,
+		disableChecks: false
```
huss: I think this is a reasonable default. However, would it make sense to allow the admin input this value on the preferences page since they might not want tests on meters at their site? This would save changing for each new meter and also get automatically created meters to do the right thing if we use this value as is suggested elsewhere.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -126,8 +133,20 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
...
+		Minimum Date cannot be after Maximum Date
+		Minimum Date and Maximum Value must be between valid input
+		Maximum No of Error must be between 0 and 75
 	*/
 	const [validMeter, setValidMeter] = useState(false);
+	const MIN_VAL = Number.MIN_SAFE_INTEGER;
+	const MAX_VAL = Number.MAX_SAFE_INTEGER;
+	const MIN_DATE_MOMENT = moment(0).utc();
+	const MAX_DATE_MOMENT = moment(0).utc().add(5000, 'years');
+	const MIN_DATE_STRING = moment(0).utc().format('YYYY-MM-DD HH:mm:ssZ');
```
huss: I know this does not change what the code does but I think it might be best to make the string values use the variables you just said above such as MIN_VAL. It will help if somebody changes the value in the future.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -138,7 +157,16 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
...
+			state.meterType !== '' &&
+			state.minVal >= MIN_VAL &&
+			state.minVal <= state.maxVal &&
+			state.maxVal <= MAX_VAL &&
+			state.minDate !== '' &&
+			state.maxDate !== '' &&
+			moment(state.minDate) >= MIN_DATE_MOMENT &&
+			moment(state.maxDate) <= MAX_DATE_MOMENT &&
+			moment(state.minDate) <= moment(state.maxDate) &&
+			(state.maxError >=0  && state.maxError <= 75)
```
huss: You use the value 75 in a number of places so I wonder if it makes sense to create a const at the top for this value. Maybe export to use in other places?

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -743,7 +824,54 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
...
+							<Label for='minDate'>{translate('meter.minDate')}</Label>
+							<Input
+								id='minDate'
+								name='minDate'
+								type='text'
+								autoComplete='on'
+								onChange={e => handleStringChange(e)}
+								placeholder='YYYY-MM-DD HH:MM:SS'
+								required value={state.minDate}
+								invalid={state.minDate === '' || moment(state.minDate) < MIN_DATE_MOMENT || moment(state.minDate) > moment(state.maxDate)}/>
```
huss: I'm surprised that this seems to work. moment has its own comparison functions so you should use something such as isBefore (https://momentjs.com/docs/#/query/is-same-or-before/).

I don't think we have done this anywhere else in the code but with the new type of checks it just hit me we might be  able to use the moment function isValid()  to verify that it is an okay date.  Right now I think it fails later in the process and is not as informative for the user as the new test on input. 

Also applies to max.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -122,18 +128,50 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+		Minimum Date cannot be after Maximum Date
+		Minimum Date and Maximum Value must be between valid input
+		Maximum No of Error must be between 0 and 75
 	*/
 	const [validMeter, setValidMeter] = useState(false);
+	const MIN_VAL = Number.MIN_SAFE_INTEGER;
+	const MAX_VAL = Number.MAX_SAFE_INTEGER;
+	const MIN_DATE_MOMENT = moment(0).utc();
+	const MAX_DATE_MOMENT = moment(0).utc().add(5000, 'years');
+	const MIN_DATE_STRING = moment(0).utc().format('YYYY-MM-DD HH:mm:ssZ');
```
huss: A number of the comments for create also apply here and I will not repeat them.

src/client/app/translations/data.js
```diff
@@ -123,6 +123,12 @@ const localeData = {
 		"default.warning.file.size": "Default Warning File Size",
 		"default.file.size.limit": "Default File Size Limit",
 		"default.time.zone": "Default Time Zone",
+		"default.meter.minimum.value": "Default meter minimum reading value check",
```
huss, line 126: The new translation keys are not in other languages.

src/client/app/types/redux/admin.ts
```diff
@@ -24,7 +24,13 @@ export type AdminAction =
...
 	| ToggleWaitForCikAndDB
 	| UpdateDefaultAreaUnitAction
-	| UpdateDefaultMeterReadingFrequencyAction;
+	| UpdateDefaultMeterReadingFrequencyAction
+	| UpdateDefaultMeterMinimumValueAction
+	| UpdateDefaultMeterMaximumValueAction
+	| UpdateDefaultMeterMinimumDateAction
+	| UpdateDefaultMeterMaximumDateAction
+	| UpdateDefaultMeterMaximumErrorsAction
+	| UpdateDefaultMeterReadingGapAction;
```
huss: A really small point but could MaximumErrors come last in the order as it is in other files/places.

src/server/services/eGauge/readEgaugeData.js
```diff
@@ -18,7 +18,7 @@ async function readEgaugeData(meter, conn) {
 	await requestor.setRegisterId();
 	const meterReadings = await requestor.getMeterReadings();
 	await requestor.logout()
-
+
```
huss: Please remove the tab at the start of the line.

src/server/services/obvius/loadLogfileToReadings.js
```diff
@@ -52,7 +52,13 @@ async function loadLogfileToReadings(serialNumber, ipAddress, logfile, conn) {
 				undefined, // unit
 				undefined, // default graphic unit
 				undefined, // area unit
-				undefined // reading frequency
+				undefined, // reading frequency
+				undefined, // minVal
```
huss: I wonder if we should use the site defaults here rather than undefined. They may often be the same due to the default values but they could differ. Note  would apply to the new reading gap in preferences too.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -627,7 +627,7 @@ async function processData(rows, meterID, timeSort = TimeSortTypesJS.increasing,
 		prevEndTimestampTz = endTimestampTz;
 	}
 	// Validate data if conditions given
-	if (conditionSet !== undefined && !validateReadings(result, conditionSet)) {
+	if (conditionSet !== undefined && !conditionSet['disableChecks'] && !validateReadings(result, conditionSet)) {
```
huss: The processData.js uses appendMsgTotal() to add messages to what is returned to the user. I think it would be good to return these messages in the same way so that the caller receives the messages and not only goes to the log file. I know this is a non-trivial change so let us know if you prefer not to do it.

The error messages in validateReadings.js has the meter id. Could we pass the meter name as an argument so we could print that instead of the ID as is done in other messages? 

A couple of thoughts about the messages such as:
```js
log.warn(`ERROR WHEN CHECKING READING TIME FROM METER ${reading.meterID}: Time ${reading.startTimestamp} is earlier than lower bound ${minDate}`);
```
 First, OED does not use uppercase in messages. Second, I think this should be a log.error because it causes the readings to be rejected.

I think line 45 has startTimestamp when it should be endTimestamp so the code would be:
```js
log.warn(`ERROR WHEN CHECKING READING TIME FROM METER ${reading.meterID}: Time ${reading.endTimestamp} is later than upper bound ${maxDate}`);
```

I cannot put these comments in the other file so I'm putting it here.

src/server/services/readMamacMeters.js
```diff
@@ -96,7 +96,13 @@ async function getMeterInfo(url, ip, csvLine) {
 				unitId, // unit
 				unitId, // default graphic unit
 				undefined, // area unit
-				undefined // reading frequency
+				undefined, // reading frequency
+				undefined, // minVal
```
huss: As suggested elsewhere, use the site defaults if that is done.

src/server/sql/meter/create_meters_table.sql
```diff
@@ -64,5 +64,11 @@ CREATE TABLE IF NOT EXISTS meters (
     unit_id INTEGER REFERENCES units(id),
     default_graphic_unit INTEGER REFERENCES units(id),
     area_unit area_unit_type NOT NULL DEFAULT 'none',
-    reading_frequency INTERVAL NOT NULL DEFAULT '00:15:00'
+    reading_frequency INTERVAL NOT NULL DEFAULT '00:15:00',
+    min_val DOUBLE PRECISION NOT NULL DEFAULT -9007199254740991 CHECK (min_val::DOUBLE PRECISION >= -9007199254740991),
```
huss: The readings table uses FLOAT for the reading so maybe it would be best to use that here instead of double precision. Note all SQL changes need to be done similarly for the migrations.

src/server/sql/preferences/create_preferences_table.sql
```diff
@@ -14,5 +14,11 @@ CREATE TABLE IF NOT EXISTS preferences (
 	default_file_size_limit FLOAT NOT NULL,
 	default_area_normalization BOOLEAN NOT NULL,
 	default_area_unit area_unit_type NOT NULL,
-	default_meter_reading_frequency INTERVAL NOT NULL
+	default_meter_reading_frequency INTERVAL NOT NULL,
+	default_meter_minimum_value DOUBLE PRECISION NOT NULL,
```
huss: Change to float if done in meter.

huss:
To do testing, I merged in the latest development and also added the database migration that was needed. Since I had to do it I just pushed the changes myself. You can check if the migration seems correct to you and any changes in the PR SQL needs to stay in sync with this.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -627,7 +627,7 @@ async function processData(rows, meterID, timeSort = TimeSortTypesJS.increasing,
...
 		errMsg = `<h2>For meter ' + meterName + ': error when validating data so all reading are rejected</h2>`;
```
huss: This is printing meterName. I think the code should be:
```js
`<h2>For meter ${meterName}: error when validating data so all reading are rejected</h2>`
```

huss (changes requested):
I want to thank @carlsonrob, @TrongQuocLe and @codehomie1 for their first contribution to OED. Overall, this is very good work and I found very few issues during testing. I have made a number of comments that should be addressed. Please let me know if you have any thoughts or questions.

huss:
I made a  few minor changes to the code. I don't think there are any issues but I'll give others one day to see if there are any issues before I merge this code.

huss (approved):
I want to thank all the developers of this code for being willing to implement additional features after the pull request was created, being so careful in finding a few things I missed and the speed with which they did all their work. The changes address all the comments and testing has found this to work well. This code is now ready to merge. Congratulations on your first contribution to OED.

PR #1011, opened 2023-09-17
Fix failure if no map file chosen or no name

### LINKED ISSUE #686: failure if no map file chosen or no name on creation
### Describe the bug

If you go to the map creation page and enter a name and angle but do not select a map then when you Save and continue you get a blank page. You cannot easily get off this page.

Similarly, if you create a map but do not give it a name, it will fail. However, this will allow you to get through the entire calibration step but gives a failure message when you try to save to the database at the end of the calibration.

### To Reproduce

See above.

### Expected behavior

OED should give an error popup (as it does for a missing or invalid angle) during the map creation step to force the user to input all the needed information.

### Additional context

None.

nathang15:
# Description

Add alert messages when user fails to upload a map image or input a map name, forcing the user to enter the information before being able to continue.

Fixes #686 

## Type of change

## Checklist

## Limitations


src/client/app/translations/data.js
```diff
@@ -297,6 +297,8 @@ const localeData = {
 		"map.is.not.calibrated": "Calibration Needed",
 		"map.is.not.displayable": "Display Disabled",
 		"map.load.complete": "Map load complete from",
+		"map.bad.load": "Map image needed",
```
huss: This should be placed alphabetically with the other map.bad items. There also needs to be translations in the other two languages in this file as with other keys.

Also could the string become "Map image file needed".

src/client/app/components/maps/MapCalibrationInitiateComponent.tsx
```diff
@@ -90,13 +98,19 @@ class MapCalibrationInitiateComponent extends React.Component<MapInitiatePropsWi
 			</form>
 		);
 	}
-
 	private async confirmUpload(event: any) {
```
huss, line 98: Please add back the blank line before this function.

src/client/app/components/maps/MapCalibrationInitiateComponent.tsx
```diff
@@ -90,13 +98,19 @@ class MapCalibrationInitiateComponent extends React.Component<MapInitiatePropsWi
...
-			this.props.updateMapMode(CalibrationModeTypes.calibrate);
+			if (this.fileInput.current.files.length === 0) {
+				this.notifyBadMapLoad();
+			}
+			else if (this.state.mapName.trim() === '')	{
+				this.notifyBadName();
+			}
+			else {
+				await this.handleInput(event);
+				this.props.updateMapMode(CalibrationModeTypes.calibrate);
```
huss, line 109: I notice the notifyLoadComplete() was removed on a successful check situation. I have wondered about it and my testing indicates that it does not really tell you if a bad map is uploaded. Thus, I think it is okay to remove it but the function that was being called and related code should be removed if we no longer do this.

huss (changes requested):
Thanks to @nathang15 for their first contribution to OED. Overall, this works very well and addresses the issue. I have made a few suggestions that ought to be done before this pull request is merged.

nathang15:
I have updated the changes requested.

nathang15 (commented):
I have addressed the requested changes.

huss:
Thanks to @nathang15 for the quick changes. Everything looks good. I made a small change to the Spanish translations to parallel the other ones and be slightly better grammatically (at least I hope). This is ready to merge. Congratulations your first approved submission to OED.

huss (approved):

PR #1012, opened 2023-09-18
Hide Rate dropdown menu when not applicable

### LINKED ISSUE #899: hide rate menu when not applicable

OED has introduced a rate menu on the line graphic page. It does not apply for these line choices:

- When the units is of raw type.
- When displaying kW on the y-axis. This can happen because you are doing the quantity kWh or a direct graphing of kW.

The rate menu should automatically disappear when either of these line choices are active and reappear when not.

The current situation where the menu is present but changing the choice makes no difference.

### Additional context

The Normalize by Area has logic to make it disappear/reappear for selections that do not support that option. Looking at that code may be useful. PR #885 added this and [this commit](https://github.com/OpenEnergyDashboard/OED/pull/885/commits/92d57abd6d7751cbc8843764491ef5a38308508e) should have the related code.

nathang15:
# Description

Set the Rate dropdown menu in Line Graph to be hidden away when either kW appears on y axis or raw unit type is selected.  

Fixes #899 

## Type of change

## Checklist

## Limitations

### DIFF
--- a/src/client/app/components/GraphicRateMenuComponent.tsx
+++ b/src/client/app/components/GraphicRateMenuComponent.tsx
@@ -30,14 +30,19 @@ export default function GraphicRateMenuComponent() {
 	// Unit data by Id
 	const selectedUnitData = unitDataById[graphState.selectedUnit];
 
+	// Should the rate drop down menu be rendered.
+	let shouldRender = true;
 	// Compare the value of name to be 'kW' or 'kWh' or unitRepresent type to be 'raw' and update the visibility of the Rate menu
 	if (selectedUnitData) {
 		const { name, unitRepresent } = selectedUnitData;
-		if(unitRepresent === UnitRepresentType.raw || name === 'kW' || name === 'kWh'){
-			return null;
+		if (unitRepresent === UnitRepresentType.raw || name === 'kW' || name === 'kWh') {
+			shouldRender = false;
 		}
 	}
-
+	// Also don't show if not the line graphic.
+	if (graphState.chartToRender !== 'line'){
+		shouldRender = false;
+	}
 	// Array of select options created from the rates
 	const rateOptions: SelectOption[] = [];
 
@@ -58,7 +63,7 @@ export default function GraphicRateMenuComponent() {
 	return (
 		<div>
 			{
-				graphState.chartToRender == 'line' &&
+				shouldRender &&
 				<div>
 					<p style={labelStyle}>
 						<FormattedMessage id='rate' />:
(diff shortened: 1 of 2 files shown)


nathang15:
I want to ask that if this is the right way to address the issue? Overall, the code fixes the issue but I'm not sure if this is the approach that OED wants.

huss:
Thanks to @nathang15  for another pull request. Overall, the idea worked well. I pushed a commit that includes the exclusion for graphics other than line from the rate menu since I think it makes them more consistent within the page and with other pages OED has. Please let me know if you think this change is good or something else should be done.

I did notice one issue while testing the code and I don't think it is an issue you introduced but something that became apparent due to this change. When you hide the rate menu while still on the line graphic page and then allow it to reappear, the help for rate no longer pops up. If you leave this graphic page for another one and come back then it will work again. For example, select the unit kWh and then unselect it (click x) and the rate menu help will not work. I suspect it has to do with the fact that it was changed while still on the page but I have not tried to track it down carefully. If you want to look at this then great. If you want someone else to look at it then please let me know.

nathang15:
Thank you for the updated changes. I think the new code makes it easier for other developers to understand. 

For the issue, may I ask that is the rate menu help refers to the tooltip? If so, I can take a look at the problem and fix it. Do I need to create a new issue for this or fix the newly occurred problem directly on this issue?

Thank you.

huss:
> For the issue, may I ask that is the rate menu help refers to the tooltip? If so, I can take a look at the problem and fix it. Do I need to create a new issue for this or fix the newly occurred problem directly on this issue?

Yes, you cannot click the tooltip icon and get a pop up at times. I don't think we need a new issue but it would be nice to fix this before we merge this PR. You are welcome to look at this if you wish. Thanks.

nathang15:
I have tried locating where the issue occurs for a while, but I still cannot find where the exact problem happens. I have tried different approaches to identifies where exactly the problem happens but kind of unsuccessful. 

Therefore, I resort to a "cheat" solution, which is to add a onClick event to rebuild the Tooltip. Essentially, after the Rate Menu is toggled between hidden and visible, the user would need to click twice for the tooltip to appear (which does not appear before the fix after the toggle event).

I believe that the underlying cause is because of some kind of bug in the React Tooltip package because the code for the tooltip component and its trigger does not look out of the ordinary.

Thank you.

nathang15 (commented):
I have tried locating where the issue occurs for a while, but I still cannot find where the exact problem happens. I have tried different approaches to identifies where exactly the problem happens but kind of unsuccessful.

Therefore, I resort to a "cheat" solution, which is to add a onClick event to rebuild the Tooltip. Essentially, after the Rate Menu is toggled between hidden and visible, the user would need to click twice for the tooltip to appear (which does not appear before the fix after the toggle event).

I believe that the underlying cause is because of some kind of bug in the React Tooltip package because the code for the tooltip component and its trigger does not look out of the ordinary.

Thank you.

huss:
The solution works reasonably well and only happens when you click the icon. I noted that you have to click the icon twice to get the help pop up and get the code to fire up when the issue occurs. I played with it a little and did not find a quick solution so I'm leaving this with a TODO when more global changes are made to tooltips.

huss (approved):
Thanks to @nathang15 for another contribution to OED. The code works well and is ready to merge.

PR #1039, opened 2023-10-11
Add radar graph feature

### LINKED ISSUE #208: radar/clock graph
Some sites support a radar/clock graphic that shows usage in polar coordinates where time is the angle and the length is the usage. We should add this ability.

nathang15:
# Description

Finish up the work on Radar Graph. These includes fixing the display, add date range picker, allow groups to work on radar.

The code for Date Range Picker is taken from 3D graph branch of @ChrisMart21.

As the commits show, @Elias0127 and @JesseVarGar did the initial work on this feature and @nathang15 finished it up.

Fixes #208 

## Type of change

## Checklist


nathang15:
This is a linting error. Weird thing is that my local device does not show any errors regarding linting.

huss:
I will try to look at this soon. As for the linting errors, if you open a shell on the OED web container and run "npm run check" you should see the same linting errors. It looks as if your VSC is using spaces instead of tabs for indenting. You can change that by clicking the indentation on the bottom ribbon in VSC. There you can set to use tabs and can convert the current spaces to tabs. Let me know if this does not work.

nathang15:
I believe the indentation is now fixed.

huss (commented):
Thanks to @nathang15 for this PR. I need to review the code but wanted to comment on a few things I noticed when initially trying it. Overall it looks good.

1. The line and radar Redux state is very similar (line also has max/min). Given the current plan is to use the line values for radar, I don't think we need the radar state. This is what is done for maps where it uses the bar state. Unlike that case, you do need to potentially fetch the new date ranges selected if not already in Redux state.
2. The time set from the date range picker for the final day is 23:59:59. This can be seen from the Redux state. This potentially causes the last point in the day to be dropped. For example, if you select 1 day for 15 min test data you get 95 instead of 96 points. The last point is at 11:37:30 instead of 11:52:30. OED uses the end of a day as 00:00:00 of the next day. This can be seen in the bar state/values on the graph.
3. The dates seem off when multiple meters are graphed with different date/time values. This figure shows the 15 min and 4 day sin where the dates are repeated. I suspect the label algorithm chooses different labels in these two cases. Somehow they needs to be the same. I think other cases also show issues.

Screenshot 2023-10-14 094157

huss (approved):
Thanks to @JesseVarGar & @Elias0127 for the initial work on radar and @nathang15 for additional work. I have review the code, made a few changes and found that it is fine and works as expected. This should be ready to merge.

PR #1040, opened 2023-10-13
Fixes to suffix type when creating or editing units.

### LINKED ISSUE #1019: Units of type suffix should have a suffix
### Describe the bug

Currently a unit can be of type suffix but not have a suffix included and the inverse is also possible for both creating and editing a unit. It does not make sense to have such a combination and an admin should be forced to include both or neither. If a suffix is added but the unit is not of type suffix then the admin should be told the type will be changed to suffix. If the unit is of type suffix but there is no suffix unit then the suffix input should become red and an entry required before saving.

### Screenshots

Examples for create a unit are shown but edit behaves similarly.

Screenshot 2023-09-25 073325

Screenshot 2023-09-25 073307

### Additional context

Currently the admin help warns about doing this so hopefully it will not happen but this fix would mean the documentation can be updated to say it cannot happen.

Jake-Bodin:
# Description

Changes to use effect, changing submit, suffix, and typeofUnit, in createUnit and editUnit files to complete the Issue #1019 
Also added a formatted message to data.js

Fixes #1019 

## Type of change

## Checklist

- [ X] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations


src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -69,11 +69,13 @@ export default function CreateUnitModalComponent() {
...
 		Name cannot be blank
 		Sec in Rate must be greater than zero
+		If type of unit is suffix their must be a suffix
 	*/
 	const [validUnit, setValidUnit] = useState(false);
 	useEffect(() => {
-		setValidUnit(state.name !== '' && state.secInRate > 0);
-	}, [state.name, state.secInRate]);
+		setValidUnit(state.name !== '' && state.secInRate > 0 &&
+			(state.typeOfUnit !== UnitType.suffix || (state.typeOfUnit === UnitType.suffix && state.suffix !== '')));
```
huss: The first check to make sure it is unequal suffix does not seem to be needed. It also does not follow the other checks in this file. I changed it and did not see any issue. Am I missing something?

This also applies to edit.

huss (changes requested):
Thanks to @Jake-Bodin for another pull request. It worked well and no issues were encountered. I make one comment on the code that should be addressed.

src/client/app/components/unit/CreateUnitModalComponent.tsx
Jake-Bodin: It's redundant, I believe i did that early on because I was looking at this line in the CreateMeter file
             (state.area === 0 || (state.area > 0 && state.areaUnit !== AreaUnitType.none
Which is unnecessary in this case.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -69,11 +69,13 @@ export default function CreateUnitModalComponent() {
...
 		Name cannot be blank
 		Sec in Rate must be greater than zero
+		If type of unit is suffix their must be a suffix
 	*/
 	const [validUnit, setValidUnit] = useState(false);
 	useEffect(() => {
-		setValidUnit(state.name !== '' && state.secInRate > 0);
-	}, [state.name, state.secInRate]);
+		setValidUnit(state.name !== '' && state.secInRate > 0 &&
+			(state.typeOfUnit === UnitType.suffix && state.suffix !== ''));
```
huss: I think the logic is off here. I believe the correct logic is
```
	setValidUnit(state.name !== '' && state.secInRate > 0 &&
			(state.typeOfUnit !== UnitType.suffix || state.suffix !== ''));
```
I found the current logic did not allow saving if you removed the suffix and then changed the type of unit to non-suffix. Let me know if you agree and, if so, change in both files.

Jake-Bodin: Your correct I see the same issue and after using your logic it seems to be working fine I apologize.

huss:
Thanks to @Jake-Bodin for making these changes. The PR is fine and ready to merge. Note it will be held for a little while until the pending Vite PR is (hopefully) merged.

huss (approved):

huss:
The merge after review only adds in approved code from development so this PR is still fine and queue to be merged.

PR #1056, opened 2023-10-31
Timezones loaded from moment instead of database

### LINKED ISSUE #681: enhance loading meter timezone
### Describe the bug

Each time the meters page loads, the timezone of each meter is fetched separately, taking a lot of requests. This makes the rate limit in PR #676 high, which is very risky. Optimization is needed to reduce the number of requests

### To Reproduce

1. Click the "Meters" button
2. Look at the Docker console window, you will see the timezones requests starting with "GET /api/timezones"

### Expected behavior

The number of timezones requests is reduced.

Jake-Bodin:

Fixes #681 

## Type of change

- [ X] Note merging this changes the database configuration.

## Checklist

- [X ] I have removed text in ( ) from the issue request
- [X ] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

### DIFF
--- a/src/client/app/components/TimeZoneSelect.tsx
+++ b/src/client/app/components/TimeZoneSelect.tsx
@@ -22,7 +22,6 @@ const TimeZoneSelect: React.FC<TimeZoneSelectProps> = ({ current, handleClick })
 			const zone = moment.tz(zoneName);
 			const abbrev = zone.format('z');
 			return { value: zoneName, label: `${zoneName} (${abbrev})` };
-
 		});
 	};
 	const resetTimeZone = [{ value: null, label: translate('timezone.no') }];
--- a/src/client/app/components/TimeZoneSelect.tsx
+++ b/src/client/app/components/TimeZoneSelect.tsx
@@ -4,9 +4,9 @@
 
 import * as React from 'react';
 import Select from 'react-select';
-import axios from 'axios';
-import { TimeZones, TimeZoneOption } from 'types/timezone';
+import { TimeZoneOption } from 'types/timezone';
 import translate from '../utils/translate';
+import * as moment from 'moment-timezone';
 
 interface TimeZoneSelectProps {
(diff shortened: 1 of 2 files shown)


src/client/app/components/TimeZoneSelect.tsx
```diff
@@ -4,42 +4,35 @@
...
+		});
+	};
+	const resetTimeZone = [{ value: null, label: translate('timezone.no') }];
+	const options = [...resetTimeZone, ...getTimeZones()];
 
 	const handleChange = (selectedOption: TimeZoneOption) => {
 		handleClick(selectedOption.value);
 	};
 
 	return (options !== null ?
```
huss: The need to check for null options was a result of the way the data was requested from the database. With it now being gotten from moment this cannot happen. I removed the conditional and used:
	return <Select isClearable={false} value={options.filter(({ value }) => value === current)} options={options} onChange={handleChange} />;
It seems to work fine. I think this change should be moade if @Jake-Bodin does not find any issues with is.

huss (changes requested):
Thanks to @Jake-Bodin for addressing this issue. Testing found it worked fine. I have one small comment to address to clean up code.

src/client/app/components/TimeZoneSelect.tsx
Jake-Bodin: I don't see any changes made from my original code, I don't feel I know enough to know if the new timezones are properly effecting the meters but i don't see any changes or problems.

huss: Sorry this was not clear. The comment had the desired code. I made this change and push a commit to put it into the PR. This PR should now be ready to merge.

huss (approved):

PR #1057, opened 2023-10-31
shouldFetchVersion function checks for version

### LINKED ISSUE #1048: only get version once

OED displays the version at the bottom of each page. Each time you change pages (to/from graphic page or any other page) it does an api call to get the version. Since the version cannot change except after an upgrade, the version will remain the same so this api call is unneeded.

If there is a version in Redux state then the api call should not be made. The code in src/client/app/actions/version.ts already has a shouldFetchVersion() function so it can be augmented to check for the Redux state.

OED works fine but this is an inefficiency and not in line with the new way other state is only fetched once.

### Additional context

None

Jake-Bodin:
# Description

Fixes #1048 

## Type of change

## Checklist

- [ X] I have removed text in ( ) from the issue request
- [X ] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

### DIFF
--- a/src/client/app/actions/version.ts
+++ b/src/client/app/actions/version.ts
@@ -27,7 +27,7 @@ export function receiveVersion(data: string): t.ReceiveVersion {
  * @returns Whether preferences are fetching
  */
 function shouldFetchVersion(state: State): boolean {
-	return !state.version.isFetching;
+	return !state.version.isFetching && !state.version.version;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/client/app/actions/version.ts
```diff
@@ -27,7 +27,7 @@ export function receiveVersion(data: string): t.ReceiveVersion {
  * @returns Whether preferences are fetching
```
huss: I think this comment should be update because it also checks if it already exists in state.

huss (changes requested):
Thanks to @Jake-Bodin for addressing another issue. Testing found it works as desired. There is one small comment to address to update the code comments.

huss (approved):

PR #879, opened 2023-04-03
Migrate from Webpack to Vite

hyperupcall:
# Description

## New Edits

- I have interactively rebased the previous commits and split them up (with extra comments in the Git description) to make things hopefully easier to follow
- As noted below, performance will improve much better than indicated by the benchmarks, due to upgrading from Vite 4.2 to 4.3!
- There are no current issues with these changed (from my testing), so this PR has been undrafted :rocket: 

## Description

This migrates the build system from Webpack to Vite. The web ecosystem as a whole is moving away from Webpack to newer, faster, and more modern solutions like Vite.

Vite is different than Webpack. In general, Webpack is very lenient with what it accepts. For example, when ECMAScript Modules and CommonJS modules are mixed in a weird/incorrect way, Webpack tries to find _some_ way to correctly resolve the imports, construct a module graph, and output the bundle.

On the other hand, Vite will fail fast if there is something wrong with the imports. Internally, vite uses the ESBuild bundler for the development server, and then the Rollup bundler when bundling for production. Since there are two bundlers, `import`s and `requires` must "work"/"be compatible" with both bundlers.

Much of the changes in this PR change the `import`s to be more precise and to make both Rollup and ESBuild happy. There are still some issues though, as mentioned further down.

The webpack npm scripts are replaced with Vite-specific ones: `npm run vite:dev` and `npm run vite:build` 

Fixes #871

Some statistics about bundle time and size are shown below for comparison:

## Webpack vs Vite Stats

Note that the Vite numbers are actually significantly better since [updating to](547113e277bc7836e7b30828bf06d949ffab5da5) Vite 4.3. Vite 4.3 performs 40-76% better in benchmarks compared to version 4.2!

### Docker-Compose

Webpack:
  - command: `docker-compose up`
  - time: 1 minute 19 seconds
  
Vite
  - command: `docker-compose up`
  - time: 48 seconds

### Bundling

Webpack:
  - command: `npm run webpack:build`
  - time: 58 seconds
  - bundle size: 5.79 MiB (bundle.js)

Vite:
  - command: `npm run vite:build`
  - time: 17 seconds
  - bundle size: ~4.6 MiB (bundle.js)

## Current Issues

[...]

PR #1059, opened 2023-11-03
restructure: Export Redux store from separate file

hyperupcall:
# Description

Defining the Redux store in the same file as the app will create problems with Vite. There is some issue with circular imports not working properly with HMR (hot module replacement). Upstream tracks this issue at vitejs/vite#3033; a workaround is to define and export the Redux store from a different file. This commit does exactly that to circumvent the circular import issue.

Related to #879

## Type of change

- [ ] Note merging this changes the database configuration.
- [ ] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

N/A

PR #1060, opened 2023-11-03
fix: Use static plotly.js imports to make Vite happy

hyperupcall:
# Description

Vite will not play nice with these dynamic requires for some reason. To make Vite happy, convert the imports to static imports. Also change them to ESM-imports for consistency.

Related to #879

## Type of change

## Checklist

## Limitations

N/A

### DIFF
--- a/src/client/app/types/locales.ts
+++ b/src/client/app/types/locales.ts
@@ -1,8 +1,12 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// @ts-expect-error Suppress TypeScript declaration file not found error
+import esLocale from 'plotly.js/lib/locales/es'
+// @ts-expect-error Suppress TypeScript declaration file not found error
+import frLocale from 'plotly.js/lib/locales/fr'
 
-export default class Locales{
-	static es = require('plotly.js/lib/locales/es');
-	static fr = require('plotly.js/lib/locales/fr');
+export default class Locales {
+	static es = esLocale;
+	static fr = frLocale;
 }


huss:
@hyperupcall Thanks for working on this. It all looked fine but testing found the language is not changing in Poltly. This is what I get with the PR when I set the language to Spanish:
Screenshot 2023-11-07 122206
and this is from the development branch:
Screenshot 2023-11-07 121746
Notice Jan becomes Ene as expected. I tried this with PR #1059 and it works the same as development so this change seems the likely source. I have not had time to look into the reason. If you can look at this or have any ideas then I welcome that.

hyperupcall:
Looking at the [source code local file](https://github.com/plotly/plotly.js/blob/master/lib/locales/es.js), it seems they use `module.exports` to create a "default ~~import~~ export".

It seems Webpack complies with the JavaScript Specification and CommonJS specification in the following way:

- When using `import`, Webpack will bundle the code with the semantics of ECMAScript Modules. That is, the default import is specified with `export default`, so the bundled code "ignores" the `module.exports` because that's not a part of ESModules
  - Usually in these situations, the imported file contains a `require`, and crashes the program. Since there are no `requires` in `locales/es.js`, that is why there is a silent failure
- When using `require`, Webpack will bundle the code with the semantics of CommonJS. That is, it will "look for" `module.exports` to find the default import (and `exports.default`, depending on the configuration)

But, Vite should do this too. And it should be okay to `require` a CommonJS file inline. It could be possible the error was something to do with tree-shacking, but I don't remember. One thing to note is that `plotly.js` has a [webpack field](https://github.com/plotly/plotly.js/blob/33dc2de5be0cdb7eb9ce51ad00283ec54b36f729/package.json#L7) in it's package.json. That could explain the discrepancy: Vite and Webpack are reading different files (Webpack, `packageJson.webpack` and Vite, `packageJson.main`).

hyperupcall:
I think we should close this issue bcause it works with Webpack, and we're using Webpack right now. I originally made the change because Vite ~`v3.3` previously errored and was unhappy. Now, Vite has version `v4.0` (and is even faster). It may be possible that this issue has been fixed or resolved in Vite. In any case, I think it might be better to resolve this once Vite is integrated, so we have a chance to test this with Vite. What do you think?

huss:
> I think we should close this issue bcause it works with Webpack, and we're using Webpack right now. I originally made the change because Vite ~`v3.3` previously errored and was unhappy. Now, Vite has version `v4.0` (and is even faster). It may be possible that this issue has been fixed or resolved in Vite. In any case, I think it might be better to resolve this once Vite is integrated, so we have a chance to test this with Vite. What do you think?

I'm okay with closing this pull request. I'm hoping we can get Vite incorporated into OED in the foreseeable future. Thanks for looking after this.

hyperupcall:
> I'm okay with closing this pull request. I'm hoping we can get Vite incorporated into OED in the foreseeable future. Thanks for looking after this.

Me too! Since #1086 was merged, I can focus my efforts on that.

PR #1065, opened 2023-11-07
Delete units

### LINKED ISSUE #1020: Delete a unit

It would be good if a unit could be deleted as can be a conversion.

If a unit is not used in a conversion, meter, or group then it should be safe to remove. This could be added to the edit unit modal page. We should check for any other conditions where it would be unsafe to delete a unit.

Currently states can be hidden so no one can use them which has a similar effect to deleting but does not remove them.

### Additional context

As part of this effort, it would be good to remove suffix units created by OED if the suffix is illuminated. Currently, they are made hidden.

### LINKED ISSUE #1042: meter unit change

Changing the unit (not the default graphic unit) associated with a meter can have a ripple effect if it is used in a group. It is desirable to let the admin know this is the case and stop the change.

The meter edit page will sweep Redux state to see if the meter is included in any group (via deep meters). If so, the admin will be told via a popup and the change of meter unit will not be allowed by resetting the menu back to the original value.

OED could put in the complex logic to see if the change is okay and fix up some issues. This is done in groups now so not redoing in meters seems much easier.  This seems fine since changing a meter unit should not be common after adding to a group.

### Additional context

If an admin wants to change the meter unit then they will need to:

1. Remove the meter from all groups that include it.
2. Change the meter unit.
3. Add the meter back to the group(s) if that is permissible.

sparakala21:
# Description
I added the handleDelete function in EditUnitModalComponent.tsx so that admins can delete units when it is necessary.
I also added a way to ensure safety when an admin changes a meters intrinsic properties. 

Fixes #1020 and Fixes #1042.

## Type of change

## Checklist

- [ ] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [ ] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.


src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -132,7 +132,7 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 		Group must have at least one child (i.e has deep child meters)
 	*/
 	const [validGroup, setValidGroup] = useState(false);
-	useEffect(() => {
+	useEffect(() => {deleteConfirmText
```
huss: I'm unclear what "deleteConfirmText" is for here. Could you explain?

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 				props.meter.maxError != state.maxError ||
 				props.meter.disableChecks != state.disableChecks
 			);
+		let error_message = '';
+		for (const value of Object.values(groups)) {
+			for (let i = 0; i < value.deepMeters.length; i++) {
+				console.log(value.deepMeters[i], props.meter.id)
+				if(value.deepMeters[i]== props.meter.id)
+				{
```
huss: OED follows the default VSC formatting where the { is on the same line as the if. There are also spacing issues. Would you please do a "Format Document" to clean this up.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 				props.meter.maxError != state.maxError ||
 				props.meter.disableChecks != state.disableChecks
 			);
+		let error_message = '';
```
huss: Could some comments be added to describe the basic functioning of this code?

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+		for (const value of Object.values(groups)) {
+			for (let i = 0; i < value.deepMeters.length; i++) {
+				console.log(value.deepMeters[i], props.meter.id)
```
huss: Was this console.log for debugging? In general, we don't have them in the final code.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 				props.meter.maxError != state.maxError ||
 				props.meter.disableChecks != state.disableChecks
 			);
+		let error_message = '';
+		for (const value of Object.values(groups)) {
```
huss: I think this check is run every time. Is that correct? For example, if there is no change should it run? Another is if you only change the name then does it matter if it is in another group? I suspect this should be restricted to certain cases. Sorry if the issue and our conversations did not cover this and let me know if you think otherwise.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+			for (let i = 0; i < value.deepMeters.length; i++) {
+				console.log(value.deepMeters[i], props.meter.id)
+				if(value.deepMeters[i]== props.meter.id)
+				{
+					error_message += translate('group') + value.name + ' uses ' + translate('meter') + ' "' + metersByID[value.deepMeters[i]].name + '" \n';
+				}
+			}
+		}
+		state.unitId = props.meter.unitId;
```
huss: First, I'm not sure why the state is always being reset. Second, directly setting state can cause issues. I think using setState() is the way to go here.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -22,6 +22,7 @@ import { Dispatch } from 'types/redux/actions';
 import { useSelector } from 'react-redux';
 import { State } from 'types/redux/state';
```
huss: Please remove new blank line.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -69,6 +70,45 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 		setState({ ...state, [e.target.name]: Number(e.target.value) });
 	}
 
+
```
huss: Extra blank line.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -69,6 +70,45 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+
+	// Log the Redux state to the console
```
huss: See formatting comment in edit meter. Other comments on alert, what done when issue found and commenting apply here.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -69,6 +70,45 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+			if(conversions[i].sourceId==state.id){
+				error_message += translate('conversion') + conversions[i].note + 'uses ' + translate('unit') + ' "' + state.id+ '"as a source unit\n';
+			}
+
+			if(conversions[i].destinationId==state.id){
+				error_message += translate('conversion') + conversions[i].sourceId + '→' + conversions[i].destinationId + ' uses '
+				+ translate('unit') + ' "' + state.id+ '" as a destination unit\n';
+			}
+		}
+		//finish meter graphicalUnitId and UnitId
```
huss: Are these TODO items? Not sure about these two comments.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -69,6 +70,45 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+		//finish conversion check both source or destination unit
+		if(error_message)
+		{
+			error_message = translate('unit.failed.to.delete.unit') + '\n' +  error_message;
+			alert(error_message);
+		}
+
+
+		//
```
huss: Unsure about these comments.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+
+		if (error_message) {
+			error_message = translate('meter.is.not.editable') + '\n' + error_message;
+			alert(error_message);
```
huss: I know alert is used elsewhere but we are trying to convert to Reaxct friendly alerts. Please look at src/client/app/utils/notifications.ts for success and error message functions. I think they would work here.

Also, I'm unclear that this stops the changes. After the alert it does not seem to have logic to stop the following code to update the meter.

src/client/app/translations/data.js
```diff
@@ -334,7 +334,7 @@ const localeData = {
 		"meter.is.enabled": "Updates Enabled",
 		"meter.is.not.displayable": "Display Disabled",
 		"meter.is.not.enabled": "Updates Disabled",
-		"meter.previousEnd": "Previous End Time Stamp:",
+				"meter.previousEnd": "Previous End Time Stamp:",
```
huss: extra spacing at start of line

src/client/app/translations/data.js
```diff
@@ -785,6 +786,7 @@ const localeData = {
 		"meter.is.enabled": "Mises à Jour Activées",
 		"meter.is.not.displayable": "Affichage Désactivé",
 		"meter.is.not.enabled": "Mises à Jour Désactivées",
+		"meter.is.not.editable": "(need French) This meter cannot be changed",
```
huss: I don't see this key under en.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 				props.meter.maxError != state.maxError ||
 				props.meter.disableChecks != state.disableChecks
 			);
+		let error_message = '';
+		for (const value of Object.values(groups)) {
+			for (let i = 0; i < value.deepMeters.length; i++) {
+				console.log(value.deepMeters[i], props.meter.id)
+				if(value.deepMeters[i]== props.meter.id)
+				{
+					error_message += translate('group') + value.name + ' uses ' + translate('meter') + ' "' + metersByID[value.deepMeters[i]].name + '" \n';
```
huss: The word 'uses' is not translated. This is true edit unit file too.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -240,6 +243,22 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+				}
+			}
+		}
+		state.unitId = props.meter.unitId;
+
+		if (error_message) {
+			error_message = translate('meter.is.not.editable') + '\n' + error_message;
```
huss: Maybe use += to be consistent with code above?

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -69,6 +70,45 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+	// Log the Redux state to the console
+	const meterState = useSelector((state: State) => state.meters.byMeterID);
+	const conversions = useSelector((state: State) => state.conversions.conversions)
+
+	const handleDeleteUnit = () => {
+		// Access the Redux store state
+		let error_message = '';
+		for (const value of Object.values(meterState)) {
+			if(value.unitId==state.id){
+				error_message += translate('meter') + value.name + ' uses ' + translate('unit') + ' "' + state.id+ '" \n';
```
huss: Using the state.id instead of the identifier. Applies below too.

huss (changes requested):
Thanks to @sparakala21 for more work on OED and this PR. I have made a number of comments. Given this may cause changes in the code, I am going to wait to do testing until they are resolved.

src/client/app/components/meters/EditMeterModalComponent.tsx
sparakala21: i changed the implementation so that it only runs when it needs to. What do you think?

src/client/app/components/groups/EditGroupModalComponent.tsx
sparakala21: ill be honest i dont think that was me

src/client/app/components/meters/EditMeterModalComponent.tsx
sparakala21: ok i made the change but now it displays both alerts. Im still figuring out how to fix that.

huss:
@sparakala21 I could not easily separate out the changes I had tried so I decided to push then out to the PR. There are some comments about open items. Feel free to change my code as needed and to address the items left. Let me know if I can help and thanks for looking into this.

huss:
I just wanted to document that this PR is waiting on PR #1113 since it changes how Redux state is accessed. After that, this work will be updated and all concerns addressed.

src/client/app/components/groups/EditGroupModalComponent.tsx
huss: This is gone with the change to RTK so no longer an issue.

src/client/app/components/meters/EditMeterModalComponent.tsx
huss: fixed

src/client/app/components/meters/EditMeterModalComponent.tsx
huss: removed

src/client/app/components/meters/EditMeterModalComponent.tsx
huss: I also restricted it more.

src/client/app/components/unit/EditUnitModalComponent.tsx
huss: done

src/client/app/components/meters/EditMeterModalComponent.tsx
huss: Fixed

huss:
I updated the PR to merge in #1113 and edited so uses the Redux Toolkit. I also made all the other needed changes to it works as desired. It should be ready to merge.

huss (approved):

PR #1109, opened 2024-01-07
Convert pik to cik

### LINKED ISSUE #901: convert from pik array to cik state

For a variety of reasons, convert from a pik array to cik state on client-side. See the [pikState.md](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/pikState.md) design document.

See design document.

We would have to change the proposed way of doing baselining on client-side.

### Additional context

NA

qnhn22:
# Description

I have created a new state called 'Cik' for client-side usage. The code has been refactored to eliminate the need for the 'pik' array, replacing it with the newly introduced 'Cik' state. In order to make 'Cik' accessible, I have developed a new API that fetches 'Cik' data, which includes updating the model, creating an SQL file for retrieving information from the database, and implementing a new route.  
  
Additionally, since the 'pik' is no longer required, the unit index can be removed from the unit table and refactor the code to eliminate any usage of the unit index. (database changed)

Fixes #901

## Type of change

## Checklist

## Limitations


src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -355,7 +356,7 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 		// pik is needed since the compatible units is not correct until pik is available.
 		// metersState normally does not change but can so include.
 		// globalGroupsState can change if another group is created/edited and this can change ones displayed in menus.
-	}, [ConversionArray.pikAvailable(), metersState, globalGroupsState, groupState.defaultGraphicUnit, groupState.deepMeters]);
+	}, [metersState, globalGroupsState, groupState.defaultGraphicUnit, groupState.deepMeters]);
```
huss: Other places have ciksState when ConversionArray.pikAvailable() removed. Should it be here?

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -122,6 +121,7 @@ export default function CreateMeterModalComponent(props: CreateMeterModalCompone
 
 	// Dropdowns
 	const [dropdownsState, setDropdownsState] = useState(dropdownsStateDefaults);
+	const globalCiksState = useSelector((state: State) => state.ciks.ciks);
```
huss: In editMeterModalComponent this is ciksState. Should the name be consistent?

src/client/app/types/redux/ciks.ts
```diff
@@ -0,0 +1,39 @@
...
+	intercept: number;
+}
+
+export interface CiksState {
+	hasBeenFetchedOne: boolean;
+	isFetching: boolean;
+	ciks: CikData[];
+}
+
+
```
huss: Please remove all but one last blank line.

src/client/app/types/conversionArray.ts
```diff
@@ -1,28 +1,28 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
-* License, v. 2.0. If a copy of the MPL was not distributed with this
-* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// /* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: Given pik is now gone from the code, maybe this file should be removed rather than commented out? There is a record in git if it is ever needed.

src/server/models/Cik.js
```diff
@@ -20,6 +33,25 @@ class Cik {
 		return conn.none(sqlFile('cik/create_cik_table.sql'));
 	}
 
+	/**
+	 * Create a new Cik object from row's data.
+	 * @param {*} row The row from which Cik will be created.
+	 * @returns {Promise.<Array.<Cik>>}
```
huss: This seems to return an object and not a promise. Is it a promise?

src/server/routes/ciks.js
```diff
@@ -0,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+  * License, v. 2.0. If a copy of the MPL was not distributed with this
+  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const express = require('express');
+const { log } = require('../log');
+const { getConnection } = require('../db');
+const Cik = require('../models/Cik');
+const { success, failure } = require('./response');
```
huss: I don't see where these two imports or the next one is used.

src/server/routes/conversions.js
```diff
@@ -1,6 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
-  * License, v. 2.0. If a copy of the MPL was not distributed with this
-  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+	* License, v. 2.0. If a copy of the MPL was not distributed with this
```
huss: I think OED normally uses a single space so the * linrd up on all lines.

huss (changes requested):
Thanks to @qnhn22 for working on this issue. Testing indicates it works as expected. I have a few modest comments to address. Also, I noticed that a search reveals places where "pik" and "conversionArray" still appear (or are part of a word) in the code or comments. I looked at several of them and it is not clear to me that they are needed. Do you want to check them and remove as possible or is that something I need to do?

src/client/app/components/groups/EditGroupModalComponent.tsx
qnhn22: Actually no. Because the Cik data is fetched in the initialization, ciksState shouldn't be here. Thank you for your suggestion.

huss: A couple of thoughts on this:

1. It is done in createGroupModalComponent which is very similar so I'm unclear on why not here.
2. The assumption where is that the initialization must be completed before the admin gets to this modal. I agree that is quite likely but is not guaranteed. I believe it happened with pik during a manual refresh of the group page. I'm not positive it can happen but I think it has so it seems safer to include the condition, esp. as this is very similar to pik that needed it.

qnhn22: I just thought that the initialization step run first, then it would make Cik available before accessing the modals or dropdown which use the Cik data. But as you suggested, I will include it as a condition because it would be safer. Thank you.

huss:
I just pushed changes to this PR. They do the final removal for the unit index from code and DB. It also merged in development to be sure there are no issues. I noted in the description that this changes the DB configuration and announced on Discord. With @qnhn22 changes and these, this PR is hopefully ready to merge. This completes the design doc steps until "Integrating cik".

huss (approved):

Issue #1093, opened 2023-11-29
Alerts need to be switched


sparakala21:
### Is your feature request related to a problem? Please describe.

Currently, our application relies on generic alerts for error and success notifications. While these have served their purpose, it's become evident that we need a more robust and consistent solution, especially when dealing with React components. The current approach lacks the flexibility and reliability required for a modern React application.

### Describe the solution you'd like
Introducing showErrorNotification and showSuccessNotification functions which will enhance the user experience and provide better control over the presentation of notifications.

### Additional context

check out PR #1065 for some ideas

huss:
Let me clarify that in some places we do Window.alert() that is not React friendly. These should be switched to the new methods in src/client/app/utils/notifications.ts that use a React way. Care should be taken to set the time (autoDismiss) if the desire is that the notice stay up for a very long time (or indefinitely in some cases). It is also important to choose the right one (success or failure). Anyone doing this that has questions is welcome to reach out to me.

huss:
After PR #1113, the only ones left are Window.confirm. It would be good to find a confirmation dialog that is compatible with the toast notifications that are used.

MatthewTran22:
> After PR [#1113](https://github.com/OpenEnergyDashboard/OED/pull/1113), the only ones left are Window.confirm. It would be good to find a confirmation dialog that is compatible with the toast notifications that are used.

Could I work on this issue?

shaikhibrahim2000:
I would like to work on this

huss:
This is available for someone to work on. It might not be too hard if a solution can be readily found. Assuming it is a new package, it would be good to check with the OED team before finalizing the choice.

koki-star:
Hi @huss my teammate and I will be working on this issue together. I couldn’t assign myself on GitHub, so could you please assign me to it?

[...]

Issue #893, opened 2023-04-19
React hooks for main containers (dashboard, footer, header, route)


huss:
### Is your feature request related to a problem? Please describe.

Please see issue #726 for information.

### Describe the solution you'd like

This transitions a number of components where the code located in files src/client/app/components/DashboardComponent.tsx, src/client/app/containers/DashboardContainer.ts, src/client/app/components/FooterComponent.tsx, src/client/app/containers/FooterContainer.ts, src/client/app/components/HeaderComponent.tsx, src/client/app/containers/HeaderContainer.ts, src/client/app/components/RouteComponent.tsx & src/client/app/containers/RouteContainer.ts need to be updated for React hooks. The look of each item will stay the same.

### Additional context

It is hoped that other pages do not need to be changed as part of this but these are used across the OED React code base.

spearec:
#958 converts HeaderComponent to use hooks

huss:
Fixed by PR #1113 so closing.

Issue #888, opened 2023-04-19
React hooks for line, bar, compare & map graphics pages


huss:
### Is your feature request related to a problem? Please describe.

Please see issue #726 for information.

### Describe the solution you'd like

The graphic pages located in files src/client/app/containers/LineChartContainer.ts, src/client/app/containers/BarChartContainer.ts, src/client/app/containers/CompareChartContainer.ts & src/client/app/containers/MapChartContainer.ts need to be updated for React hooks. The look of the page will stay the same.

Note that each of the Containers interacts with react-plotly using the old style mapStateToProps. Information on a way to just use a Component is available. Some useful links may be [react-plotly README](https://github.com/plotly/react-plotly.js/blob/master/README.md) and React [Plotly information page](https://plotly.com/javascript/react/). Many other help pages are available.

### Additional context

src/client/app/components/DashboardComponent.tsx is part of another issue but changes may be needed.

ChrisMart21:
#1113 Addresses All other than compare which still has a mapStateToProps container.

huss:
Replaced with issue #1195 for last part.

Issue #886, opened 2023-04-19
React hooks for admin preferences page


huss:
### Is your feature request related to a problem? Please describe.

Please see issue #726 for information.

### Describe the solution you'd like

The admin page located in files src/client/app/components/admin/PreferencesComponent.tsx & src/client/app/containers/admin/PreferencesContainer.ts need to be updated for React hooks. The look of the page will stay the same.

### Additional context

When the admin user pages are updated, the "Users" button on this page will be eliminated. This may or may not have already happened.

src/client/app/components/admin/AdminComponent.tsx may require some minor changes.

huss:
Fixed by PR #1113 so closing.

PR #1113, opened 2024-01-15
Redux Toolkit Migration

### LINKED ISSUE #817: routecomponent's authentication functions do not work as expected
### Describe the bug

Due to how asynchronous functions are handled, the current method in which checkAuth, requireAuth, and requireRole are written has a fundamental issue. The token validation is an asynchronous function, but the functions themselves must return a value synchronously. This means a value is returned before the async block is finished. because of this, users are directed to the page they try to access regardless of the token authentication. This is the root cause of issue #816, although I have found a workaround for that issue.

Currently, apart from #816, this is a silent issue. Some error is encountered during the process which causes the token to be removed and then the function to be re-run, which will properly redirect the user (see screenshots). For now, this works as intended. Users are able to access a page only if they have access, and invalid tokens are deleted and then the user is redirected. The problem is that the code only works because of further errors popping up downstream.

For example, consider requireAuth:
```typescript
public requireAuth(component: JSX.Element) {
// Redirect route to home page if the auth token does not exist
if (!hasToken() || !this.props.loggedInAsAdmin) {
	return <Redirect to='/' />;
}
// Verify that the auth token is valid.
// Needs to be async because of the network request
(async () => {
	if (!(await verificationApi.checkTokenValid())) {
		showErrorNotification(translate('invalid.token.login.admin'));
		// We should delete the token when we know that it is expired. Ensures that we don't not leave any unwanted tokens around.
		deleteToken();
		// This ensures that if there is no token then there is no stale profile in the redux store.
		this.props.clearCurrentUser();
		// Route to login page since the auth token is not valid
		return <Redirect to='/' />;
	}
	return component;
})();
return component;
}
```

When the function is evaluated, if the user has a token, the last return statement is immediately returned. Only after the await function finishes does the rest of the code evaluate. This means that the function may attempt to return a redirect, but this second return value is never processed.

### To Reproduce

1. Log in as an admin user
2. Wait for the token to become invalid
3. Access an admin page
4. Inspect the web console

### Expected behavior

There should be no improper API calls nor other react errors.

### Screenshots

image

ChrisMart21:
# Description

This pull request primarily focuses on updating the project to modern Redux patterns by migrating to the recommended '[Redux-Toolkit](https://redux-toolkit.js.org/)' while addressing several other minor issues along the way.

Fixes:
- #817 
  - Reworked the snippets in question to use react-routerv6's outlets + redux state.
- #92 
  - Data Fetching now managed via RTK Query. Fetching/caching simplified and more robust.
- #886 
  - Preferences removed the container, and added hooks + selectors
- #893 
  - All components now are functional. Dropped the container+component HOC pattern.
- #466 
  - Graphs now scale / are responsive 

Partially addresses 

- #726 / #888 
  -  Continues the ongoing push towards functional components.
  - All graph types other than compare
  - main containers
  - admin preferences/userpages
  
- #1093 
  - Continue ongoing switch towards notifications. 

## Type of change

## Checklist

## Limitations
- This pull request introduces numerous changes across the front end that are yet to be thoroughly tested. While not entirely feature-complete, it represents a significant step towards migrating to RTK. A thorough evaluation is necessary to ensure these changes have not impaired the functionality of the app.
  - Close evaluation and iterations have been ongoing since initial posting of PR. It is reaching its final stages, and is believed to be stable. Further testing and evaluation from the community is always welcome, and encouraged.

- This PR has a version bump to react-router-v6. This change introduces several newer hooks that pair well with the ongoing migration towards functional components. This did however introduce some backwards compatibility issues with the current implementation of the unsaved warning. This component has to be reworked for compatibility with react-router-v6, but also Redux RTK.

 - There was issues/difficulty migrating the entirety of 'Maps'. Maps currently is still using legacy redux and either needs to be reworked, or migrated properly by someone more knowledgeable on this section of the code base.
   - While maps utilizes some newer RTK elements, it is still heavily reliant on legacy redux patterns.  
   - Further work is necessary on this section of the migration.

- Redux-Toolkit introduces some new devmode checks which conflict with some of the legacy Redux code. For example, the 'serializableCheck' has to currently be suppressed due to the use of TimeInverval in Redux State. TimeInterval is a class which is non-serialiazable which needs to be addressed. Additionally 'immutableCheck' is also currently has conflicts with the maps implementation. For the time being, these checks are disabled to prevent the app from malfunctioning, however the goal is to address these issues, and have all dev mode checks enabled. 
  - Progress has been made on this front and will likely be resolved in tandem with the Maps rework.


src/client/app/redux/api/userApi.ts
```diff
@@ -0,0 +1,53 @@
...
+				url: 'api/users/delete',
+				method: 'POST',
+				body: { email }
+			}),
+			invalidatesTags: ['Users']
+		})
+	})
+})
+
+// public async editUsers(users: User[]) {
```
huss: Since these capabilities are above, is this commented out code needed?

src/client/app/redux/api/userApi.ts
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import { NewUser, User } from '../../types/items';
+// import { authApi } from './authApi';
```
huss: Is auth done here?

src/client/app/components/admin/UsersDetailComponent.tsx
```diff
@@ -2,78 +2,79 @@
...
-	const submitUnsavedChangesFunction = (successCallback: () => void, failureCallback: () => void) => {
-		// This function is called to submit the unsaved changes
-		props.submitUserEdits().then(successCallback, failureCallback);
+	const editUser = (e: React.ChangeEvent<HTMLInputElement>, targetUser: User) => {
+		// copy user, and update role
+		const updatedUser: User = { ...targetUser, role: e.target.value as UserRole }
+		// make new list from existing local user state
+		const updatedList = localUsersChanges.map(user => (user.email === targetUser.email) ? updatedUser : user)
+		setLocalUsersChanges(updatedList)
+		// editUser(user.email, target.value as UserRole);
```
huss: I think this is done in submit so commented out code can be removed.

src/client/app/redux/actions/admin.ts
```diff
@@ -0,0 +1,192 @@
...
+import { showErrorNotification, showSuccessNotification } from '../../utils/notifications';
+import { Dispatch, GetState, Thunk } from '../../types/redux/actions';
+import { State } from '../../types/redux/state';
+import { conversionArrayApi, preferencesApi } from '../../utils/api';
+import translate from '../../utils/translate';
+import * as moment from 'moment';
+// import { updateSelectedLanguage } from './options';
+import { graphSlice } from '../slices/graphSlice';
+import { adminSlice } from '../slices/adminSlice';
+// TODO Marked For Deletion after RTK migration solidified
```
huss: Still needed?

src/client/app/redux/actions/admin.ts
```diff
@@ -0,0 +1,192 @@
...
+			dispatch2(graphSlice.actions.changeChartToRender(state.admin.defaultChartToRender));
+			if (preferences.defaultBarStacking !== state.graph.barStacking) {
+				dispatch2(graphSlice.actions.changeBarStacking());
+			}
+			if (preferences.defaultAreaNormalization !== state.graph.areaNormalization) {
+				dispatch2(graphSlice.actions.toggleAreaNormalization());
+			}
+			if (preferences.defaultLanguage !== state.options.selectedLanguage) {
+				// if the site default differs from the selected language, update the selected language and the locale
+				// dispatch2(updateSelectedLanguage(preferences.defaultLanguage));
```
huss: Is this now done in the component so commented out code no longer needed? Import can go if this goes.

src/client/app/redux/actions/admin.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Dispatches a fetch for admin preferences and sets the state based upon the result
+ */
+function fetchPreferences(): Thunk {
+	return async (dispatch: Dispatch, getState: GetState) => {
+		dispatch(adminSlice.actions.requestPreferences());
+		const preferences = await preferencesApi.getPreferences();
+		dispatch(adminSlice.actions.receivePreferences(preferences));
+		moment.locale(getState().admin.defaultLanguage);
+		//  TODO reference only DELETE ME
```
huss: Remove?

src/client/app/redux/actions/admin.ts
```diff
@@ -0,0 +1,192 @@
...
+export function fetchPreferencesIfNeeded(): Thunk {
+	return (dispatch: Dispatch, getState: GetState) => {
+		if (shouldFetchPreferenceData(getState())) {
+			return dispatch(fetchPreferences());
+		}
+		return Promise.resolve();
+	};
+}
+
+export function submitPreferencesIfNeeded(): Thunk {
```
huss: I don't see where this function is used. Can you help me understand? (seems shouldSubmitPreferenceData is only used here). This may apply to other, similar functions.

src/client/app/components/conversion/ConversionViewComponent.tsx
```diff
@@ -36,7 +37,7 @@ export default function ConversionViewComponent(props: ConversionViewComponentPr
 	const handleClose = () => {
 		setShowEditModal(false);
 	}
-
+	React.useEffect(() => undefined, [props.conversion])
```
huss: Could a comment be added about what this is doing? I think it updates this modal if the conversion changes.

src/client/app/components/conversion/ConversionsDetailComponent.tsx
```diff
@@ -25,23 +20,22 @@ import { Dispatch } from 'types/redux/actions';
...
-	const conversionsState = useSelector((state: State) => state.conversions.conversions);
+	const { data: conversionsState = [], isFetching: conversionsFetching } = conversionsApi.useGetConversionsDetailsQuery();
+	// Units DataById
+	const { unitDataById = {}, isFetching: unitsFetching } = unitsApi.useGetUnitsDetailsQuery(undefined, {
+		selectFromResult: ({ data, ...result }) => ({
+			...result,
+			unitDataById: data && unitsAdapter.getSelectors().selectEntities(data)
+		})
+	})
+	// const x = useAppSelector(state => conversionsApi.endpoints.refresh.select()(state))
```
huss: status?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -37,8 +35,9 @@ export default class MetersCSVUploadComponent extends React.Component<MetersCSVU
 			window.alert(error.response.data as string);
 		}
 		// Refetch meters details.
-		store.getState().meters.hasBeenFetchedOnce = false;
-		fetchMetersDetails();
+		// Removed with rtk migration
```
huss: The meters are not being updated as desired.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -324,10 +324,10 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 						childGroups: thisGroupState.childGroups, gps: gps, displayable: thisGroupState.displayable,
 						note: thisGroupState.note, area: thisGroupState.area, defaultGraphicUnit: thisGroupState.defaultGraphicUnit, areaUnit: thisGroupState.areaUnit
 					}
+					console.log(submitState, 'removeme')
 					// This saves group to the DB and then refreshes the window if the last group being updated and
 					// changes were made to the children. This avoid a reload on name change, etc.
-					dispatch(submitGroupEdits(submitState, (i === groupsChanged.length ? true : false) && (childMeterChanges || childGroupChanges)));
-					i++;
+					submitGroupEdits(submitState)
```
huss: Chris?: How does this make sure that all state is okay without reload. Also, since there is an invalidate in src/client/app/redux/api/groupsApi.ts does this do it more than needed?

Maybe look at editUnitMutation for passing parameters.

src/client/app/components/AppLayout.tsx
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @param props Optional Children prop to render instead of using Outlet
+ * @returns The OED Application Layout. The current route as the outlet Wrapped in the header, and footer components
+ */
+export default function AppLayout(props: LayoutProps) {
+	return (
+		<>
+			<ToastContainer transition={Slide} />
+			<HeaderComponent />
+			{props.children ?? <Outlet />}
```
huss: Maybe comment to describe

src/client/app/components/MapChartComponent.tsx
```diff
@@ -0,0 +1,387 @@
...
+			sizex: 500,
+			sizey: 500,
+			xanchor: 'left',
+			yanchor: 'bottom',
+			sizing: 'contain',
+			opacity: 1
+		}]
+	};
+
+	/***
```
huss: Can this be removed?

src/client/app/redux/api/groupsApi.ts
```diff
@@ -0,0 +1,102 @@
...
+							id: childrenInfo.groupId,
+							changes: {
+								childMeters: childrenInfo.childMeters,
+								childGroups: childrenInfo.childGroups
+							}
+						}));
+						dispatch(groupsApi.util.updateQueryData('getGroups', undefined, groupDataById => { groupsAdapter.updateMany(groupDataById, updates) }))
+					}
+				} catch (e) {
+					console.log(e)
```
huss: This should be changed. log in standatd OED way and/or notify user.

src/client/app/components/router/GraphLinkComponent.tsx
```diff
@@ -0,0 +1,113 @@
...
+					dispatchQueue.push(graphSlice.actions.updateComparePeriod({ comparePeriod: validateComparePeriod(value), currentTime: moment() }))
+					break;
+				case 'compareSortingOrder':
+					dispatchQueue.push(graphSlice.actions.changeCompareSortingOrder(validateSortingOrder(value)))
+					break;
+				case 'optionsVisibility':
+					dispatchQueue.push(appStateSlice.actions.setOptionsVisibility(value === 'true' ? true : false))
+					break;
+				case 'mapID':
+					// 'TODO, Verify Behavior & FIXME! MapLink not working as expected
```
huss: Resolve. Maybe part of map issues.

huss (commented):
@ChrisMart21 This is not a formal review but some comments for us to work on as I continue to do the formal review. Many will be deleted and are informal.

src/client/app/redux/api/userApi.ts
ChrisMart21: Was used as translation reference. This & import at the top needs to get deleted. Will mark as resolved when deleted,

src/client/app/redux/api/userApi.ts
ChrisMart21: Auth is handled on startup, or login. If authenticated, these request headers\s will have a token for additional verification if needed. That import isn't necessary here.

src/client/app/components/admin/UsersDetailComponent.tsx
ChrisMart21: Agreed.

src/client/app/redux/actions/admin.ts
ChrisMart21: This Entire File could be purged. 
It is not used anywhere and was used as reference for translation.

src/client/app/redux/actions/admin.ts
ChrisMart21: This Entire File could be purged.
It is not used anywhere and was used as reference for translation.

ChrisMart21: I added a comment to files that need to get purged.

// TODO Marked For Deletion after RTK migration solidified

src/client/app/components/conversion/ConversionsDetailComponent.tsx
ChrisMart21: refreshing now occurs as an onQueryStarted Property of the relevant mutations. Comment can be deleted.

src/client/app/components/groups/EditGroupModalComponent.tsx
ChrisMart21: This and most other admin pages are written such that the page will reset if the cached values it relies on gets modified. In this case the groups cache is updated, therefore resets the current page/ modal. 

It seems however that this mutation is missing the onQueryStarted Logic. The window reload will be replaced with the appropriate cache invalidation(s).

src/client/app/components/meters/MetersDetailComponent.tsx
```diff
@@ -4,129 +4,71 @@
...
-							possibleGraphicUnits={possibleGraphicUnits}
-						/>
+						<CreateMeterModalComponent />
 					</div>
 				}
-				{metersStateLoaded && unitsStateLoaded &&
+				{
 					<div className="card-container">
 						{/* Create a MeterViewComponent for each MeterData in Meters State after sorting by identifier */}
+						{/*  Optional Chaining to prevent from crashing upon startup race conditions*/}
```
huss: Can you help me understand why this is optional chaining?

ChrisMart21: This comment should be removed. It seems that this comment is still lingering around after some of the logic was shifted elsewhere.. This optional chaining still occurs in many of the selectors. 
One issue I was encountering during was that selectors were firing before the initial data was finished loading. This would cause fatal undefined index access issues. To prevent this I utilized optional chaining to short-circuit some of the methods. Doing this caused the methods to gracefully finish and return no data. Not sure if this is ideal, but works for now.

Similarly, the comment above no longer applies either. The sorting of meters is now handled at the store level by entity adapter. This comment should be edited . Data is no longer sorted during the render.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -103,15 +103,16 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 	const shouldUpdateUnit = (): boolean => {
 		// true if inputted values are okay and there are changes.
 		let inputOk = true;
+		const meterDataByID = selectMeterDataById(store.getState())
```
huss: Curious why this one uses store directly and not useAppSelector as in somw other places.

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -180,13 +203,15 @@ function mapStateToProps(state: State, ownProps: CompareChartContainerProps): an
 			yaxis: {
 				title: unitLabel,
 				showgrid: true,
-				gridcolor: '#ddd'
+				gridcolor: '#ddd',
+				fixedrange: true
```
huss, line 209: Want to check why added fixedrange?

src/client/app/containers/MapChartContainer.ts
```diff
@@ -339,14 +349,10 @@ function mapStateToProps(state: State) {
...
-	const props: any = {
+	const props = {
 		data,
-		layout,
-		config: {
-			locales: Locales // makes locales available for use
-		}
-	};
-	props.config.locale = state.options.selectedLanguage;
+		layout
```
huss, line 350: I think removing the locales is why plotly is no longer using the chose language to label the axis items.

src/client/app/containers/CompareChartContainer.ts
ChrisMart21: I believe to avoid zooming when dragging across the graph. If this is desired behavior it can be omitted. Let me know.

src/client/app/components/unit/EditUnitModalComponent.tsx
ChrisMart21: Good catch! This is bad practice!! It is higly advised against to import store.ts anywhere other than index. Honeslty not sure how this one made it through... Definitely needs to be addressed

src/client/app/components/AppLayout.tsx
ChrisMart21: Sure. Done!

src/client/app/components/BarChartComponent.tsx
```diff
@@ -0,0 +1,257 @@
...
+		if (dataset.x.length > 1) {
+			enoughData = true
+		}
+	})
+	if (datasets.length === 0) {
+		return <h1>
+			{`${translate('select.meter.group')}`}
+		</h1>
+	} else if (!enoughData) {
+		// This normal so plot.
```
huss: First, this does not seem a normal plot as it has no data. Second, I think we may want to change the translate key since not just 3D. Let me know and I can do if you want.

src/client/app/components/BarChartComponent.tsx
```diff
@@ -0,0 +1,257 @@
...
+	} else {
+		return (
+			<Plot
+				data={datasets as Plotly.Data[]}
+				style={{ width: '100%', height: '100%' }}
+				onRelayout={handleRelayout}
+				useResizeHandler={true}
+				config={{
+					displayModeBar: false,
+					responsive: true
```
huss: locales is gone and does not seem to show language aware labels.

src/client/app/components/BarChartComponent.tsx
```diff
@@ -0,0 +1,257 @@
...
+					bargroupgap: 0.1, // Gap between different meter's readings under the same timestamp
+					showlegend: true,
+					legend: { x: 0, y: 1.1, orientation: 'h' },
+					yaxis: {
+						title: unitLabel, showgrid: true,
+						gridcolor: '#ddd', fixedrange: true
+					},
+					xaxis: {
+						rangeslider: { visible: true },
+						// rangeselector: { visible: true },
```
huss: Needed?

src/client/app/components/ChartDataSelectComponentSave.tsx
```diff
@@ -0,0 +1,751 @@
+// /* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: Can this go?

src/client/app/components/ChartSelectComponent.tsx
```diff
@@ -3,85 +3,62 @@
...
-						<FormattedMessage id='radar' />
-					</DropdownItem>
-					<DropdownItem
-						onClick={() => dispatch({ type: 'CHANGE_CHART_TO_RENDER', chartType: ChartTypes.threeD })}
-					>
-						<FormattedMessage id='3D' />
-					</DropdownItem>
+					{
+						// Make items for dropdown from enum
+						Object.values(ChartTypes)
```
huss, line 45: I like the new code but the items should be alphabetical.

src/client/app/containers/MapChartContainer.ts
ChrisMart21: Indeed, Corrected the issue.

src/client/app/components/BarChartComponent.tsx
ChrisMart21: Yes please fix the translations to whatever you find suitable. Also, yes the comment is in the wrong location.

src/client/app/components/BarChartComponent.tsx
ChrisMart21: Agreed & fixed

src/client/app/components/BarChartComponent.tsx
ChrisMart21: Not sure. Let me know. I'm still testing things & potential ui controls. I removed the redraw and restore, which i think should come back in some way. Similarly, I added a range slider to bar ,but not entirely sure if its here to stay. We need to solidify the UI design In the near future.

src/client/app/components/ChartDataSelectComponentSave.tsx
ChrisMart21: yes! fixed

src/client/app/components/ChartSelectComponent.tsx
ChrisMart21: I added a .sort() to the map method. Let me know if this works for you. Otherwise we can rollback.

src/client/app/components/ChartDataSelectComponentSave.tsx
ChrisMart21: Yes, Fixed!

src/client/app/components/DateRangeComponent.tsx
```diff
@@ -2,63 +2,52 @@
...
-import TooltipMarkerComponent from './TooltipMarkerComponent';
 import translate from '../utils/translate';
-import { State } from '../types/redux/state';
-import { ChartTypes } from '../types/redux/graph';
-import { Dispatch } from '../types/redux/actions';
-import { changeGraphZoomIfNeeded } from '../actions/graph';
+import TooltipMarkerComponent from './TooltipMarkerComponent';
+import { selectSelectedLanguage } from '../redux/slices/appStateSlice';
+
+// Potential Fixes, for now omitted
```
huss: Could you explain what the fix is?

ChrisMart21: There's an issue with the date range picker overflowing & getting overshadowed by the plotly graph when the screen size is not wide enough.

 By default the date picker does not honor the width of the parents container. Enabling this behavior with those few lines of CSS fixes this issue, but I'm not too keen on how it looks visually.

When a chart is displaying, shrink the width of the window some (with and without the css) to see what I mean.

src/client/app/components/ExportComponent.tsx
```diff
@@ -2,213 +2,186 @@
...
-								graphExport(sortedReadings, meterIdentifier, unitLabel, unitIdentifier, chartName, scaling, MeterOrGroup.meter, errorBarState);
-							}
-						}
+					// Convert the meter area into the proper unit if normalizing by area or use 1 if not so won't change reading values.
+					const areaScaling = graphState.areaNormalization ?
+						meterArea * getAreaUnitConversion(meterDataById[meterId].areaUnit, graphState.selectedAreaUnit) : 1;
+					// Divide areaScaling into the rate so have complete scaling factor for readings.
+					const scaling = rateScaling / areaScaling;
+					// Make sure they are there and not being fetched.
+					if (readingsData) {
```
huss: Should this have as lineMeterIsFetching?

src/client/app/components/conversion/ConversionViewComponent.tsx
ChrisMart21: This should be deleted.

src/client/app/components/GraphicRateMenuComponent.tsx
```diff
@@ -22,13 +23,10 @@ export default function GraphicRateMenuComponent() {
 	const dispatch = useDispatch();
```
huss: Most places use useAppDispatch but a few use the old useDispatch. Wondering why?

ChrisMart21: The behavior 'should' be identical & not cause issues, however they all really should be using 'useAppDispatch'.  Must've been overlooked during translation.

src/client/app/components/ExportComponent.tsx
ChrisMart21: Line 66 should cover this, but the other cases should also be carefully evaluated. I suppose a check that there is actually meaningful data like readings.length ... etc could be checked here as well. 

I didn't significantly alter the logic in this or most components,  except for how the data is retrieved. 
Upon careful inspection of this code, it appears that many of these checks can be deferred to some already existing selectors, and possibly re-written for easier maintenance which seams like a near-future endeavor.
A refactor should be in order for this and probably many other files.

src/client/app/components/HeaderButtonsComponent.tsx
```diff
@@ -77,15 +76,16 @@ export default function HeaderButtonsComponent() {
 	// Local state for rendering.
 	const [state, setState] = useState(defaultState);
 	// Information on the current user.
-	const currentUser = useSelector((state: State) => state.currentUser.profile);
+	const { profile: currentUser } = useAppSelector(selectCurrentUser);
 	// Tracks unsaved changes.
-	const unsavedChangesState = useSelector((state: State) => state.unsavedWarning.hasUnsavedChanges);
+	// TODO Re-implement AFTER RTK Migration
```
huss, line 72: Is unsaved still used and status of this comment? Does it need to be done later per TODO below?

src/client/app/components/HeaderComponent.tsx
```diff
@@ -3,36 +3,24 @@
...
-		paddingBottom: '5px'
-	};
-	const largeTitleStyle = {
-		display: 'inline-block'
-	};
-	const smallTitleStyle = {
-		display: 'inline-block',
-		marginTop: '10px'
-	};
+	const siteTitle = useAppSelector(state => state.admin.displayTitle);
```
huss: Another palce where admin state is directly accessed.

src/client/app/components/HistoryComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * @returns Renders a history component with previous and next buttons.
+ */
+export default function HistoryComponent() {
+	const dispatch = useAppDispatch();
+	const backStack = useAppSelector(selectPrevHistory)
+	const forwardStack = useAppSelector(selectForwardHistory)
+
+	return (
+		<div style={{ width: '80%', display: 'flex', flexDirection: 'row', justifyContent: 'space-between' }}>
+			<svg width={20} height={20} viewBox="0 0 10 10" fill="none" xmlns="http://www.w3.org/2000/svg"
```
huss: xmlns="http://www.w3.org/2000/svg" does not seem necessary when I removed it. Should we have this URL to w3?

src/client/app/components/LineChartComponent.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { selectLineUnitLabel } from '../redux/selectors/plotlyDataSelectors';
+import { graphSlice } from '../redux/slices/graphSlice';
+import { LineReadings } from '../types/readings';
+import translate from '../utils/translate';
+import LogoSpinner from './LogoSpinner';
+import { selectSelectedLanguage } from '../redux/slices/appStateSlice';
+import locales from '../types/locales';
+
+// Stable reference for when there is not data.
+const stableEmptyReadings: LineReadings = {}
```
huss: Is it okay to move this inside the function? It only seems to be used here but maybe there is an efficienty issue I'm not thinking of.

ChrisMart21: It is intentionally left outside. Since it is declared outside it will retain a stable object reference. This is there to prevent unnecessary re-rendering when initial data is in flight. Putting it inside will create a new reference each render which defeats its purpose. These stable references could be moved to a different location , or maybe wrapped on a selector if desired. This pattern will show up in a handful of places so we need to decide on this sooner than later.

src/client/app/components/ReadingsPerDaySelectComponent.tsx
```diff
@@ -2,58 +2,45 @@
...
-			return actualReadingInterval
+			actualReadingInterval = endTS.diff(startTS) / 3600000;
 		}
 
-		// Return normal interval
-		return readingInterval;
-	})
+	}
+
+	// Return normal interval
```
huss: First, could this commented out code be removed? Also, I'm unclear on why the new code returns the same value as the old code in this case. Finally, should there be a return above and not be implicit?

src/client/app/redux/actions/conversions.ts
```diff
@@ -1,75 +1,40 @@
...
-import translate from '../utils/translate';
-import * as t from '../types/redux/conversions';
-import { conversionsApi } from '../utils/api';
+/* eslint-disable jsdoc/check-param-names */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+// @ts-nocheck
+/* eslint-disable jsdoc/require-param */
+
+// TODO Marked For Deletion after RTK migration solidified
```
huss: Not sure used and checking if you are okay with removing this now?

src/client/app/redux/actions/graph.ts
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as moment from 'moment';
+import { TimeInterval } from '../../../../common/TimeInterval';
+import * as t from '../../types/redux/graph';
+import { ComparePeriod, SortingOrder } from '../../utils/calculateCompare';
+
+export interface LinkOptions {
```
huss: I don't see this used. Ok to remove?

src/client/app/redux/actions/groups.ts
```diff
@@ -0,0 +1,193 @@
+// /* This Source Code Form is subject to the terms of the Mozilla Public
+//  * License, v. 2.0. If a copy of the MPL was not distributed with this
+//  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// TODO Marked For Deletion after RTK migration solidified
```
huss: Ok to remove?

src/client/app/redux/actions/meters.ts
```diff
@@ -1,65 +1,41 @@
...
+import { metersApi } from '../../utils/api';
+import { showErrorNotification, showSuccessNotification } from '../../utils/notifications';
+import translate from '../../utils/translate';
+import { updateCikAndDBViewsIfNeeded } from './admin';
 
-export function receiveMetersDetails(data: t.MeterData[]): t.ReceiveMetersDetailsAction {
-	return { type: ActionType.ReceiveMetersDetails, data };
-}
 
 export function fetchMetersDetails(): Thunk {
```
huss: This function is still in a comment in src/client/app/components/csv/ReadingsCSVUploadComponent.tsx but maybe it can be cleaned up and removed. The rest seem to be unused so can delete per TODO above.

src/client/app/redux/actions/units.ts
```diff
@@ -1,61 +1,41 @@
...
+	* License, v. 2.0. If a copy of the MPL was not distributed with this
+	* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { ActionType, Thunk, Dispatch, GetState } from '../types/redux/actions';
-import { showSuccessNotification, showErrorNotification } from '../utils/notifications';
-import translate from '../utils/translate';
-import * as t from '../types/redux/units';
-import { unitsApi } from '../utils/api';
-import { updateCikAndDBViewsIfNeeded } from './admin';
+// TODO Marked For Deletion after RTK migration solidified
```
huss: I don't think this is used. Ok to remove?

src/client/app/redux/api/authApi.ts
```diff
@@ -0,0 +1,60 @@
...
+				url: 'api/login',
+				method: 'POST',
+				body: loginArgs
+			}),
+			// When this mutation is successful, cache stores with these tags are marked as invalid
+			// next time the corresponding endpoint is queried, cache will be ignored and overwritten by a fresh query.
+			// in this case, a user logged in which means that some info for ADMIN meters groups etc.
+			// invalidate forces a refetch to any subscribed components or the next query.
+			invalidatesTags: ['MeterData', 'GroupData']
+			// Listeners for this query (ExtraReducers):
```
huss: What is this commented out code for?

src/client/app/redux/api/baseApi.ts
```diff
@@ -0,0 +1,42 @@
...
+		'GroupData',
+		'GroupChildrenData',
+		'Preferences',
+		'Users',
+		'ConversionDetails',
+		'Units'
+	],
+	// Initially no defined endpoints, Use rtk query's injectEndpoints
+	endpoints: () => ({})
+	// Defaults to 60 seconds or 1 minute
```
huss: Keep commented out code?

src/client/app/components/groups/EditGroupModalComponent.tsx
huss: Now that I'm looking at the API, I'm still unsure if this updates too much. The old code determined if other groups were impacted. The new API seems to always invalidate the GroupData. Not sure and other issues with groups.

src/client/app/redux/api/metersApi.ts
```diff
@@ -0,0 +1,114 @@
...
+		}),
+		rawLineReadings: builder.query<RawReadings[], { meterID: number, timeInterval: TimeInterval }>({
+			query: ({ meterID, timeInterval }) => `api/readings/line/raw/meter/${meterID}?timeInterval=${timeInterval.toString()}`
+		})
+	})
+})
+
+
+export const selectMeterDataResult = metersApi.endpoints.getMeters.select()
+
```
huss, line 80: Is it standard practice to export even if only used in this file? Just trying to understand.

src/client/app/redux/api/readingsApi.ts
```diff
@@ -0,0 +1,156 @@
...
+
+export const readingsApi = baseApi.injectEndpoints({
+	endpoints: builder => ({
+		threeD: builder.query<ThreeDReading, ThreeDReadingApiArgs>({
+			// ThreeD requests only single meters at a time which plays well with default cache behavior
+			// No other properties are necessary for this endpoint
+			// Refer to the line endpoint for an example of an endpoint with custom cache behavior
+			query: ({ id, timeInterval, graphicUnitId, readingInterval, meterOrGroup }) => ({
+				// destructure args that are passed into the callback, and generate/ implicitly return the API url for the request.
+				url: `api/unitReadings/threeD/${meterOrGroup}/${id}?`,
```
huss: Should there be a ? at the end (below does not have)?

huss: Should there be a ? at the

src/client/app/redux/actions/conversions.ts
ChrisMart21: Yep, Should go.

src/client/app/redux/actions/graph.ts
ChrisMart21: yes, this can/should go

src/client/app/redux/api/authApi.ts
ChrisMart21: RTK Slices have an optional property called extra reducers, which can listen to any action defined outside of itself. Sometimes its difficult to know which slices are listening to things like rtkQuery actions. In this case its current user slice listens for the match fulfilled action. This was an early on idea- that I'm not certain this will work cleanly. It wasto document the listeners for the endpoint, but should probably go.

src/client/app/redux/api/metersApi.ts
ChrisMart21: Not sure if there's a standard practice here, or if there are performance implications. Just think its useful to have selectors accessible wherever if need.

src/client/app/redux/api/readingsApi.ts
ChrisMart21: Interesting that it works either way. I'm assuming the fetchBaseQuery is smart enough to sort itself out. I guess it should be omitted since it handles the url generation internally when given a url + params object.

src/client/app/redux/componentHooks.ts
```diff
@@ -0,0 +1,38 @@
...
+// Overloads to support TS key completions
+type TranslateFunction = {
+	(messageID: LocaleDataKey): string;
+	(messageID: string): string;
+}
+
+// usage
+// const translate = useTranslate()
+// translate('myKey')
+export const useTranslate = () => {
```
huss, line 27: I'm trying to figure out how this relates to src/client/app/utils/translate.ts and the fact that some places use one or the other.

src/client/app/redux/selectors/adminSelectors.ts
```diff
@@ -0,0 +1,309 @@
...
+	unitData => {
+		let possibleMeterUnits = new Set<UnitData>();
+		// The meter unit can be any unit of type meter.
+		unitData.forEach(unit => {
+			if (unit.typeOfUnit == UnitType.meter) {
+				possibleMeterUnits.add(unit);
+			}
+		});
+		// Put in alphabetical order.
+		possibleMeterUnits = new Set(_.sortBy(Array.from(possibleMeterUnits), unit => unit.identifier.toLowerCase(), 'asc'));
```
huss, line 51: This is not a change but I noticed in src/client/app/redux/api/unitsApi.ts that the sorting includes an accent term and localCompare. There are other places where the way we sort differs. Do you think they should be consistent? If so, we should see if a function(s) can be used to do this in an easier way.

src/client/app/redux/selectors/adminSelectors.ts
```diff
@@ -0,0 +1,309 @@
...
+		(_state, meterDetails: MeterData) => meterDetails.unitId,
+		(_state, meterDetails: MeterData) => meterDetails.defaultGraphicUnit
+	],
+	(possibleGraphicUnits, possibleMeterUnits, unitId, defaultGraphicUnit) => {
+		// Graphic units compatible with currently selected unit
+		const compatibleGraphicUnits = new Set<UnitData>();
+		// Graphic units incompatible with currently selected unit
+		const incompatibleGraphicUnits = new Set<UnitData>();
+		// If unit is not 'no unit'
+		if (unitId != -99) {
```
huss, line 97: This is used for both edit and create. For create there used to be a separate test for -999. I'm noting since this may need to change to get the other code to allow no unit. Same below.

src/client/app/redux/selectors/adminSelectors.ts
```diff
@@ -0,0 +1,309 @@
...
+			enabled: false,
+			displayable: false,
+			meterType: '',
+			url: '',
+			timeZone: '',
+			// String type conflicts with MeterDataType GPSPoint
+			gps: '',
+			// Defaults of -999 (not to be confused with -99 which is no unit)
+			// Purely for allowing the default select to be "select a ..."
+			unitId: -99,
```
huss: -99 vs -999 here too.

src/client/app/redux/selectors/threeDSelectors.ts
```diff
@@ -0,0 +1,47 @@
...
+// Memoized Selectors
+export const selectThreeDComponentInfo = createSelector(
+	selectThreeDMeterOrGroupID,
+	selectThreeDMeterOrGroup,
+	selectMeterDataById,
+	selectGroupDataById,
+	(id, meterOrGroup, meterDataById, groupDataById) => {
+		//Default Values
+		let meterOrGroupName = 'Unselected Meter or Group'
+		let isAreaCompatible = true;
```
huss, line 27: My reading of the code indicates this must be set by selectNameFromEntity so this value does not matter. Wanted to verify that is correct. If so, I think we should set to empty string (or comment).

src/client/app/redux/selectors/uiSelectors.ts
```diff
@@ -0,0 +1,545 @@
...
+		// Get the new min and max times, in seconds, from the slider box
+		const newMinXTimestamp = Math.floor(minTimeStamp + (secondsPerPixel * sliderMinX));
+		const newMaxXTimestamp = Math.floor(minTimeStamp + (secondsPerPixel * sliderMaxX));
+		// The newMin/MaxTimestamp is equivalent to a Unix time in milliseconds. Thus, it will
+		// shift with timezone. It isn't clear if we want it in local or UTC. It depends on what
+		// plotly does. Here it is assumed that local is what is desired. This seems to work
+		// and not shift the graphs x-axis so using.
+		return new TimeInterval(moment(newMinXTimestamp), moment(newMaxXTimestamp)).toString();
+	} else {
+		return 'all'
```
huss: This used to throw an error. I can see that all would graph the full range but is it covering up an error? Should we log here or is it normal?

ChrisMart21: I think this approach needs to be reworked. We need to tap into the plotly component callbacks to retrieve and manage this value.
 I haven't got around to implementing this and kept this in as a reminder/to cover up errors for the time being.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -0,0 +1,380 @@
...
+			if (cleared) {
+				const clearedMeterOrGroups = meta.removedValues;
+				// A Select has been cleared(all values removed with clear)
+				// use the first index of cleared items to check for meter or group
+				isAMeter = clearedMeterOrGroups[0].meterOrGroup === MeterOrGroup.meters
+				// if a meter clear meters, else clear groups
+				isAMeter ? current.selectedMeters = [] : current.selectedGroups = []
+
+			}
+			if (valueRemoved) {
```
huss: Are these mutually exclusive? If so, should it be else if for these?

src/client/app/types/locales.ts
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-export default class Locales{
+export default class locales{
```
huss: Normally class names are capitalized. Is this to avoid the same name from Plotly or something else? Maybe we should refactor name if this is the case.

src/client/app/types/redux/state.ts
```diff
@@ -16,15 +15,15 @@ import { CurrentUserState } from './currentUser';
...
-import { OptionsState } from './options';
+import { AppState } from 'redux/slices/appStateSlice';
 
 export interface State {
+	appState: AppState;
 	meters: MetersState;
 	readings: {
 		line: LineReadingsState;
 		bar: BarReadingsState;
 		compare: CompareReadingsState;
```
huss, line 26: Just curious why 3D was removed from here.

src/client/app/utils/hasPermissions.ts
```diff
@@ -21,5 +21,7 @@ export function hasPermissions(user: UserRole, compareTo: UserRole): boolean {
  * @returns Whether or not user is an admin
  */
 export function isRoleAdmin(user: UserRole): boolean {
+	// TODO Already Converted to a selector
```
huss: The only place this is used is src/client/app/components/HeaderButtonsComponent.tsx. The comment seems out of date since that function does not exist. I think we should purge the last use and get rid of this but want to check with you on the best way to do that.

src/server/test/web/readingsData/expected_bar_ri_15_mu_kWh_gu_kWh_st_-inf_et_inf_bd_75.csv
```diff
@@ -0,0 +1,2 @@
+reading,start time,end time
```
huss: Not sure why this is listed as a change but same as current development so should be fine. Just noting.

src/server/test/web/readingsBarMeterQuantity.js
```diff
@@ -94,9 +94,23 @@ mocha.describe('readings API', () => {
                     // Check that the API reading is equal to what it is expected to equal
                     expectReadingToEqualExpected(res, expected);
                 });
-
-                // Add B5 here
-
+                mocha.it('B5: 75 day bars for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as kWh', async () => {
```
huss: Another change that already in development so noting.

src/client/app/components/groups/GroupsDetailComponent.tsx
```diff
@@ -3,77 +3,29 @@
...
-		// You need the basic group state loaded since going to modify.
-		if (groupsStateLoaded) {
-			// Get all groups' meter and group immediate children into state. Since all modals done for all groups
-			// we get them all here.
-			dispatch(fetchAllGroupChildrenIfNeeded());
-		}
-		// In case the group state was not yet loaded you need to do this again.
-	}, [groupsStateLoaded, groupsAllChildrenLoaded]);
+	// page may contain admin info so verify admin status while admin is authenticated.
+	authApi.useTokenPollQuery(undefined, { skip: !isAdmin, pollingInterval: authPollInterval })
```
huss: I'm wondering why tis is not in admin, conversion & unit (and maps at some point) pages? Maybe others.

src/client/app/redux/api/metersApi.ts
huss: Maye see https://stackoverflow.com/questions/76705375/does-exporting-a-module-and-never-importing-it-anywhere-affect-performance

src/client/app/components/groups/GroupsDetailComponent.tsx
ChrisMart21: Admin outlet has this polling effect so any of it's children also inherit the behavior.

Meter and group pages are kinda different because they modify behaviors based on auth status so they live outside the adminOutlet and explicitly need the polling added here

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: fixed

src/client/app/components/MapChartComponent.tsx
huss: done

src/client/app/components/router/GraphLinkComponent.tsx
huss: gone, map link elsewhere

src/client/app/containers/CompareChartContainer.ts
huss: Just noting that we don't want to zoom on compare so this is fine.

src/client/app/components/BarChartComponent.tsx
huss: done

src/client/app/components/ChartSelectComponent.tsx
huss: looks good

src/client/app/redux/actions/groups.ts
huss: done

src/client/app/redux/actions/meters.ts
huss: done

src/client/app/redux/actions/units.ts
huss: done

src/client/app/redux/api/metersApi.ts
huss: Maybe see https://stackoverflow.com/questions/76705375/does-exporting-a-module-and-never-importing-it-anywhere-affect-performance. I'm not sure which way to go on this.

src/client/app/redux/api/readingsApi.ts
huss: Removed.

src/client/app/components/groups/EditGroupModalComponent.tsx
huss: It is true that this invalidates the entire group redux state. However, it is possible that the edit of one group impacts another that includes it. OED could check if the full update is needed but that would require changing how group state is acquired. Given this isn't done very often and only by admins, OED will not make this change now (or maybe ever). This same philosophy will be used for all such changes/invalidations for admin work.

src/client/app/redux/api/groupsApi.ts
huss: Changed to error popup.

src/client/app/components/DateRangeComponent.tsx
huss: I added a small comment to document.

src/client/app/components/GraphicRateMenuComponent.tsx
huss: I replaced all uses with useAppDispatch.

src/client/app/components/LineChartComponent.tsx
huss: Clarified comment

src/client/app/types/locales.ts
ChrisMart21: Reverted back to UpperCase.

src/client/app/components/HeaderComponent.tsx
ChrisMart21: Corrected.

src/client/app/redux/selectors/uiSelectors.ts
ChrisMart21: This has been pushed back until after this PR. Graphs are going to all be reworked to track all necessary attrributes will live in Redux( including the range Slider interval) Resolving for now.

src/client/app/redux/api/authApi.ts
ChrisMart21: removing for now since not standard.

src/client/app/redux/api/baseApi.ts
ChrisMart21: Deleted, can change if needed.

src/client/app/redux/api/metersApi.ts
ChrisMart21: I think this is a non-critical issue. I will deffer this to some later PR if desired.

src/client/app/components/BarChartComponent.tsx
ChrisMart21: Fixed For now.

src/client/app/components/HeaderButtonsComponent.tsx
ChrisMart21: Unsaved warning needs an overall update and is planned to get re-implemented post-rtk migration.

src/client/app/redux/selectors/threeDSelectors.ts
ChrisMart21: This is correct. Not is an empty string instead.

src/client/app/redux/slices/graphSlice.ts
ChrisMart21: yes, only one event should happpen per reducer pass. elifs were used in the later half of this reducer, but didn't seem to make it up here. Fixed

src/client/app/utils/hasPermissions.ts
ChrisMart21: Yeah, these are essentially selectors. This file is legacy, and will be deleted.

src/client/app/types/redux/state.ts
ChrisMart21: This is old state, which is only used in maps. 3d was the first to implement RTKQ, so was removed here a while ago. But this all should go soon, so not of much importance. Resolving convo.

src/client/app/components/ReadingsPerDaySelectComponent.tsx
ChrisMart21: Commented out code, removed.

ChrisMart21: What do you mean by old code. vs new code. I'm unsure there's a difference between the two iterations in this context. Could you expand on this?

src/client/app/redux/selectors/adminSelectors.ts
huss: I think this is resolved. Could @ChrisMart21 verify.

src/client/app/redux/selectors/adminSelectors.ts
huss: Changes have fixed.

src/client/app/redux/componentHooks.ts
huss: Need to resolve store more generally so will address in future.

src/client/app/redux/selectors/adminSelectors.ts
huss: Will get consistent in future.

src/client/app/redux/selectors/adminSelectors.ts
ChrisMart21: Is resolved with recent create meter modal updates.

huss:
I want to express my deep thanks to @ChrisMart21 for initiating and completing the migration of OED to the Redux Toolkit and Query system. This not only gets OED up to the latest Redux usage but adds a number of new features such as history of graphics. This is a substantial step in OED keeping current with technology and reducing our technical debt. Many thanks.

huss (approved):
Thanks to @ChrisMart21 for another substantial contribution. All the comments have been addressed and testing finds the code works as expected. It is ready to merge.

PR #1153, opened 2024-03-12
Add compare tests

alfarj83:
# Description

Added functions to parse and test the compare meter data.

## Type of change

Test data implementation

- [X] This change requires a documentation update

## Checklist

- [X] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

The function is still a work in progress (i.e. not completely able to run), and for now the test case just has values hardcoded into it.

### DIFF
--- a/src/server/test/web/readingsCompareMeter.js
+++ b/src/server/test/web/readingsCompareMeter.js
@@ -13,7 +13,10 @@ const { prepareTest,
     expectCompareToEqualExpected,
     getUnitId,
     ETERNITY,
-    METER_ID } = require('../../util/readingsUtils');
+    METER_ID,
+    unitDatakWh,
+    conversionDatakWh,
+    meterDatakWh } = require('../../util/readingsUtils');
 
 mocha.describe('readings API', () => {
     mocha.describe('readings test, test if data returned by API is as expected', () => {
@@ -21,18 +24,19 @@ mocha.describe('readings API', () => {
             mocha.describe('for meters', () => {
                 // Test 15 minutes over all time for flow unit.
                 mocha.it('Expected and actual data should line up', async () => {
-                    await prepareTest(unitData, conversionData, meterData);
+                    await prepareTest(unitDatakWh, conversionDatakWh, meterDatakWh);
                     // Get the unit ID since the DB could use any value.
                     const unitId = await getUnitId('kWh');
                     const expected = '';
-                    //for compare, need the unitID, currentStart, currentEnd, shift
+                    // for compare, need the unitID, currentStart, currentEnd, shift
                     const res = await chai.request(app).get(`/api/compareReadings/meters/${METER_ID}`)
-                        .query({ 
-                            currStart: "2022-10-30T00:00:00.000Z",
-                            currEnd: "2022-10-31T17:00:00.000Z",
+                        .query({
+                            curr_start: '2022-10-30 00:00:00',
+                            curr_end: '2022-10-31 17:00:00',
                             shift: 'P1D',
                             graphicUnitID: unitId
                         });
+                    console.log('res: ' , res.body);
                     expectCompareToEqualExpected(res, expected);
                 });
             });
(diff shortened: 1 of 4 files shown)


src/server/test/web/readingsCompareMeter.js
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: I think to be consistent, this file should be named readingsCompareMeterQuantity.js.

src/server/test/web/readingsCompareMeter.js
```diff
@@ -0,0 +1,43 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+    This file tests the readings retrieval API compare chart meters.
+    See: https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/testing/testing.md for information.
+*/
+
+const { chai, mocha, app } = require('../common');
+const Unit = require('../../models/Unit');
```
huss: This import is not used so it can be removed.

src/server/util/readingsUtils.js
```diff
@@ -98,6 +98,21 @@ function expectMaxMinToEqualExpected(res, expected, id = METER_ID) {
     }
 }
 
+/**
+ * Compares readings from api call against the expected readings csv
```
huss: I realize other ones don't do this but as we add more expected fuctions it would be good to have the description be more precise. Also, this does not use a CSV so remove that. Thus, could this become: 
Compares readings from compare api call against the expected readings

src/server/test/web/readingsCompareMeter.js
```diff
@@ -0,0 +1,43 @@
...
+mocha.describe('readings API', () => {
+    mocha.describe('readings test, test if data returned by API is as expected', () => {
+        mocha.describe('for compare charts', () => {
+            mocha.describe('for meters', () => {
+                // Test 15 minutes over all time for flow unit.
+                mocha.it('Expected and actual data should line up', async () => {
+                    await prepareTest(unitDatakWh, conversionDatakWh, meterDatakWh);
+                    // Get the unit ID since the DB could use any value.
+                    const unitId = await getUnitId('kWh');
+                    const expected = [ [ '7962.23097109771', '8764.06090894387', '2022-10-31 17:00:00', 'P1D' ] ];
```
huss: Unlike other expected values, compare only has the set 4 values for comparison. Thus, I'm not convinced having an array of arrays makes sense in this case. I think just having the single array would be best so here it becomes:
const expected = [ '7962.23097109771', '8764.06090894387', '2022-10-31 17:00:00', 'P1D' ];
The ``expectCompareToEqualExpected`` needs to be updated to match.

huss (changes requested):
Thanks to @alfarj83 for another contribution to OED. I've made a few comments to address but overall it looks good.

src/server/test/web/readingsCompareMeter.js
alfarj83: Will do!

src/server/test/web/readingsCompareMeterQuantity.js
```diff
@@ -0,0 +1,41 @@
...
+mocha.describe('readings API', () => {
+    mocha.describe('readings test, test if data returned by API is as expected', () => {
+        mocha.describe('for compare charts', () => {
+            mocha.describe('for meters', () => {
+                // Test 15 minutes over all time for flow unit.
+                mocha.it('Expected and actual data should line up', async () => {
+                    await prepareTest(unitDatakWh, conversionDatakWh, meterDatakWh);
+                    // Get the unit ID since the DB could use any value.
+                    const unitId = await getUnitId('kWh');
+                    const expected = [ '7962.23097109771', '8764.06090894387', '2022-10-31 17:00:00', 'P1D' ];
```
huss: Sorry for not noticing this earlier but the compare API only returns 2 values (as checked in expectCompareToEqualExpected()). Thus, the last two values in the expected array should not be there.

alfarj83: Alright fixed!

huss:
Thanks to @alfarj83 for addressing all the comments. This works as expected and is ready to merge. Note I will update the testing design doc to reflect this and add more desired compare tests (hopefully) in the near future.

huss (approved):

PR #1234, opened 2024-05-14
Fixes Issue #1221 - confirm unit delete

### LINKED ISSUE #1221: confirm unit delete

Other pages such as conversion do a final popup check to verify that a delete is actually wanted. The unit page checks that the unit is safe to delete but does not do a confirm with the admin.

After clicking delete, the admin should have to confirm the delete after it is confirmed to be safe.  This is what conversion shows:

image

None

### Additional context

None

mikedivine:
# Description

Added Confirmation Modal to "Delete Unit".  Tested manually and looks to be functioning properly.  

Fixes #1221

## Type of change

## Checklist

## Limitations

None


src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -71,7 +75,34 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+		// Hide the edit modal
+		handleClose();
+		// Show the warning modal
+		setShowDeleteConfirmationModal(true);
+	};
+
+	/* End Confirm Delete Modal */
+
 	const handleDeleteUnit = () => {
+
```
huss: Please remove this blank line.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -396,9 +438,10 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 							placeholder='Note'
 							onChange={e => handleStringChange(e)} />
 					</FormGroup>
-				</Container></ModalBody>
+				</Container>
+				</ModalBody>
```
huss: The open tags are both on one line but this change puts the end tags on two lines. I think together is okay since it will reduces indenting in this case. If you want to keep this change then the open tags need to be the same and the file needs reformatting (normally from VSC auto format).

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -71,7 +75,34 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 		setState({ ...state, [e.target.name]: Number(e.target.value) });
 	};
 
+	/* Confirm Delete Modal */
+	// Separate from state comment to keep everything related to the warning confirmation modal together
+	const [showDeleteConfirmationModal, setShowDeleteConfirmationModal] = useState(false);
+	const deleteConfirmationMessage = translate('unit.delete.unit') + ' [' + props.header + '] ?';
```
huss: You are following current usage but I noticed this now. I tried repacing props.header with props.unit.identifier and found it worked fine. Thus, props.header is duplicating the original, other props state. As such, I think it should be removed from the props. I see this also in EditConversionModalComponent so the same change there if you are willing (or I can do).

huss (changes requested):
Thanks to @mikedivine for their first code contribution to OED. Overall, this looks good and works as expected. I made a few minor comments to address.

src/client/app/components/unit/EditUnitModalComponent.tsx
mikedivine: done.

src/client/app/components/unit/EditUnitModalComponent.tsx
mikedivine: I put them back on one line.

src/client/app/components/unit/EditUnitModalComponent.tsx
mikedivine: Ah, I was using the conversion code as a template, but there is no identifier in the conversion interface as they use two units and arrows to identify it.  So I have updated the prop in conversion to be called conversionIdentifier and pass it in so it is more telling of what it is.  I got rid of the heading prop for units and just use values.identifier which receives the unit props.  Let me know if this all works

huss (approved):
Thanks to @mikedivine for quickly addressing all the comments and carefully reviewing that props state to fix it in a good way. Everything looks good and runs as expected. Congratulations on your first accepted work into OED.

PR #1237, opened 2024-05-18
Fixes Issue #1209 - don't allow cumulative reset without cumulative on meter

### LINKED ISSUE #1209: don't allow cumulative reset without cumulative on meter

The meter edit and create pages allow the admin to select cumulative reset without also choosing cumulative. This should not be allowed to avoid mistakes.

Either don't allow checking the cumulative reset if cumulative is not checked or determine after clicked, warn the admin and unselect.

Leave as is

### Additional context

One example of how this is done is in the group create page where the default graphic unit is set to no unit if the selected child meter/group is not compatible with it.

mikedivine:
# Description

This disables the Cumulative Reset option and the Cumulative Reset Start and Stop times if Cumulative is selected as 'no'.
The Cumulative Reset Start and Stop times are also disabled if Cumulative Reset is marked as 'no'.

Fixes #1209

## Type of change

## Checklist

## Limitations

Needs Reviewing

### DIFF
--- a/src/client/app/components/meters/CreateMeterModalComponent.tsx
+++ b/src/client/app/components/meters/CreateMeterModalComponent.tsx
@@ -33,7 +33,7 @@ import TooltipMarkerComponent from '../TooltipMarkerComponent';
  * @returns Meter create element
  */
 export default function CreateMeterModalComponent() {
-	// Tracks wheter a unit/ default unit has been selected.
+	// Tracks whether a unit/ default unit has been selected.
 	// RTKQ Mutation to submit add meter
 	const [submitAddMeter] = metersApi.endpoints.addMeter.useMutation();
 	// Memo'd memoized selector
@@ -60,7 +60,6 @@ export default function CreateMeterModalComponent() {
 			setMeterDetails(details => ({ ...details, defaultGraphicUnit: -999 }));
 		}
 	}, [meterDetails.unitId]);
-
 
 	const handleShow = () => setShowModal(true);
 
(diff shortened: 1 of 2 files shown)


src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -428,7 +433,9 @@ export default function CreateMeterModalComponent() {
 							<Input id='cumulativeResetStart' name='cumulativeResetStart' type='text' autoComplete='off'
 								onChange={e => handleStringChange(e)}
 								value={meterDetails.cumulativeResetStart}
-								placeholder='HH:MM:SS' />
+								placeholder='HH:MM:SS'
+								disabled={meterDetails.cumulativeReset === false || meterDetails.cumulative === false ? true : false}
```
huss: A small point but can
```meterDetails.cumulative === false ? true : false```
become
```meterDetails.cumulative === false```
This also mirrors the cumulativeReset check.

This also applies below and in the two similar cases for edit.

File: src/client/app/components/meters/CreateMeterModalComponent.tsx [hunk 2]
```diff
@@ -412,13 +411,19 @@ export default function CreateMeterModalComponent() {
...
+							{meterDetails.cumulative === true ? (
+								<Input id='cumulativeReset' name='cumulativeReset' type='select'
+									value={meterDetails.cumulativeReset.toString()}
+									onChange={e => handleBooleanChange(e)}>
+									{Object.keys(TrueFalseType).map(key => {
+										return (<option value={key} key={key}>{translate(`TrueFalseType.${key}`)}</option>);
+									})}
+								</Input>
+							) : (
+								<Input id='cumulativeReset' name='cumulativeReset' type='select' disabled>
```
huss, line 429: If you set cumulative to yes, then cumulativeReset to yes and finally cumulative to no, then cumulativeReset shows unavailable but it stores to the DB as true (yes). This means that it is likely doing the reset at times. I think that if you change cumulative to no then it should modify the Redux state to make cumulativeReset be no so the DB gets the desired value. The same applies to edit.

huss (changes requested):
Thanks to @mikedivine for another contribution. Overall, the code looks good and works correctly. I made a couple of small comments to address.

src/client/app/components/meters/CreateMeterModalComponent.tsx
mikedivine: done.

src/client/app/components/meters/CreateMeterModalComponent.tsx
mikedivine: I added a useEffect hook to check if cumulative changes and then if cumulative is false, it updates cumulativeReset to false. I hope this is an acceptable solution.

huss:
Thanks to @mikedivine for the quick updates. Everything looks good. Thanks for fixing another issue.

huss (approved):

PR #1263, opened 2024-06-11
Fixes #1208

### LINKED ISSUE #1208: meter reading duplication should be drop-down menu

On the CSV page, the reading duplication is a drop-down menu of 1-9. This forces the value to be integral and that should be the case. The meter create and edit pages allow for arbitrary input of values. When not an allowed value, the save does not actually happen to the DB and no error is shown. It is in Redux state but if the page is refreshed it will go back to its old value.

Convert meter pages to use drop-down. 

Leave as is with issues involved.

### Additional context

None

davanmur:
# Description

This changes the Duplicate input box when creating and editing a meter to a dropdown with the option to select 1-9, as per issue #1208.

Fixes #1208 

## Type of change

## Checklist

## Limitations

None.


src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -495,13 +495,25 @@ export default function CreateMeterModalComponent() {
 						{/* readingDuplication input */}
 						<Col><FormGroup>
 							<Label for='readingDuplication'>{translate('meter.readingDuplication')}</Label>
-							<Input id='readingDuplication' name='readingDuplication' type='number'
+							{/* <Input id='readingDuplication' name='readingDuplication' type='number'
```
huss: Can you please remove the commented out code here and in edit.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -495,13 +495,25 @@ export default function CreateMeterModalComponent() {
...
+								<option> 2 </option>
+								<option> 3 </option>
+								<option> 4 </option>
+								<option> 5 </option>
+								<option> 6 </option>
+								<option> 7 </option>
+								<option> 8 </option>
+								<option> 9 </option>
+							</Input>
 							<FormFeedback>
```
huss: Is this feedback still needed given the only allowed values are 1-9? Same for edit.

huss (changes requested):
Thanks to @davanmur for their first contribution to OED. This works as expected. I made a couple of small comments to address. Also, could you fill out the PR template so it has the desired information (see https://openenergydashboard.org/developer/pr.html) and links to the issue in GitHub.

huss (commented):
Thanks to @davanmur for the update. The code is now fine. Could you please address the request to update the PR description and title (see previous comment). Then the PR should be ready to merge. Thanks.

huss (approved):

PR #1284, opened 2024-06-25
CSV Changes with Redux Toolkit and Formatting Updates

### LINKED ISSUE #1198: convert CSV upload page to React Toolkit

PR #1113 did a lot of the migration to RTK. The CSV upload page still needs to be done.

Convert to RTK as done for other pages.

Leave as is

### Additional context

It might be good to reconsider the look of these pages. For example, should these be two separate pages for meters and readings now that we can easily add to the menu choice. The look could also be made consistent with the input on newer pages. If this is not done as part of this issue then a new issue should be made.

This issues supersedes issue #887.

mikedivine:
# Description

Removed CSV container and split into separate CSV pages.  Uses Create Meter Modal in Readings and seems to be working with Redux.

@rfrost30 has helped contribute on this PR.

Fixes #1198

## Type of change

## Checklist

- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement]

## Limitations

None


src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -26,315 +33,477 @@ function range(lower: number, upper: number): number[] {
...
+													/>
+												</div>
+												<div className='ps-2'>
+													{translate('csv.readings.param.relaxed.parsing')}
+												</div>
+											</div>
+										</Label>
+									</Col>
+								</Row>
+								<Row>
```
huss: I think we should change the refresh behavior from the past (you correctly mirrored the old code). I think this should only have a "Refresh Readings" box so the two become one. The rationale is the graphing code needs both now and the refresh is moderately fast so doing both should not be too bad. The function in src/server/services/refreshAllReadingViews.js could be used to do this rather than two calls.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -26,315 +33,477 @@ function range(lower: number, upper: number): number[] {
...
-	}
+	// dispatch(fetchMetersDetails());
+	// TODO Using an alert is not the best. At some point this should be integrated
+	// with react.
+	// There should be a message (not void) but that is not the type so overriding.
+	const handleSubmit = async (e: React.MouseEvent<HTMLFormElement>) => {
+		e.preventDefault();
+		if (selectedFile) {
+			try {
+				const msg = await submitReadings(selectedFile);
```
huss: This is continuing to use the old style routes outside the RTK (new in src/client/app/redux/api). I think OED should probably go to the new system to be consistent and to allow other actions such as invalidating the current readings so you don't have to refresh the web browser after an upload. This could be done now or an issue opened if people agree.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
-					</FormGroup>
-					<Button color='secondary' type='submit'>
-						<FormattedMessage id='csv.submit.button' />
-					</Button>
-				</Form>
-			</div>
-		);
-	}
+	return (
+		<Container>
```
huss: This begins shifted down the page so it is not near the title. Is it easy to get to start at the top of the page?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+			<TooltipHelpComponent page='help.csv.header' />
+			<Form onSubmit={handleSubmit}>
+				<Row className="justify-content-md-center">
```
huss, line 97: I see the parallel with CSV Readings with Row/Col. Could you explain if these are also helping with the page format? I just want to understand.

src/client/app/components/TooltipHelpComponent.tsx
```diff
@@ -47,6 +47,8 @@ export default function TooltipHelpComponent(props: TooltipHelpProps) {
 		'help.admin.unitview': { link: `${helpUrl}/adminUnitViewing.html` },
 		'help.admin.user': { link: `${helpUrl}/adminUser.html` },
 		'help.csv.header': { link: `${helpUrl}/adminDataAcquisition.html` },
```
huss: I see this tag is used for the page label but I don't see it being used as a tooltip/translation. Also, data.ts has it in two languages but not French. Should this tag be removed? If so, the documentation needs to be checked that it gives what is needed on the linked pages (someone else can do the docs).

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+import { uploadCSVApi } from '../../utils/api';
+import translate from '../../utils/translate';
 
-	constructor(props: MetersCsvUploadPropWithCacheDispatch) {
-		super(props);
-		this.handleSubmit = this.handleSubmit.bind(this);
-		this.handleSetMeterName = this.handleSetMeterName.bind(this);
-		this.fileInput = React.createRef<HTMLInputElement>();
-	}
+interface MetersCSVUploadComponentProps {}
```
huss: This probably is needed but I wonder if there is another way than an empty interface.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+
+const MetersCSVUploadComponent: React.FC<MetersCSVUploadComponentProps> = () => {
+	const dispatch = useDispatch();
```
huss: Most of the code base now uses useAppDispatch and wondewr if this can too. This may relate to updating for RTK.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+	const[ meterData, setMeterData] = React.useState({
```
huss: I'm not taking a position on this but many places use a defaultState/defaultValues (name not consistent) to define and initialize state. The CSV Readings does it too. Should OED prefer one? Should the code try to be consistent?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
-			// A failed axios request should result in an error.
-			showErrorNotification(error.response.data as string);
 		}
-		// Refetch meters details by invalidating its api cache.
-		this.props.resetApiCache();
+		resetApiCache();
+	};
 
-	}
+	const handleFileChange = (file : File | null) => {
```
huss: OED tries to use the default VSC formatting (command: format document). I think this should be applied to all files to conform (here no space before the :).

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+
+const MetersCSVUploadComponent: React.FC<MetersCSVUploadComponentProps> = () => {
+	const dispatch = useDispatch();
+	const[ meterData, setMeterData] = React.useState({
+		meterName: '',
+		gzip: false,
+		headerRow: false,
+		update: false
+	});
+	const [ file, setFile ] = React.useState<File | null>(null);
```
huss: I'm curious about using state vs a local variable when no React/Redux is used (such as useEffect). I don't know which OED has used but I think it is variables. Something to discuss/decide.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+	const fileInput = React.useRef<HTMLInputElement>(null);
+	const resetApiCache = () => {
+		dispatch(baseApi.util.invalidateTags(['MeterData']));
```
huss: I suspect this would not be needed here if the RTK was used for routing.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
-			const current = this.fileInput.current as HTMLInputElement;
-			const { files } = current;
-			if (files && (files as FileList).length !== 0) {
-				await this.props.submitCSV(files[0]);
-				// TODO Using an alert is not the best. At some point this should be integrated
-				// with react.
-				showSuccessNotification('<h1>SUCCESS</h1>The meter upload was a success.');
+	const handleSubmit = async(e: React.MouseEvent<HTMLFormElement>) => {
+		e.preventDefault();
+		if(file) {
```
huss: Some places in the newer OED forms it can check if the required elements are done and not enable the submit until that is done. Is that possible with a file picker?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -26,315 +33,477 @@ function range(lower: number, upper: number): number[] {
 	return arr;
 }
```
huss: It seemed the message with info on upload status was lost due to a web page refresh. Need to look into.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+				// A failed axios request should result in an error.
+				showErrorNotification(error.response.data as string);
 			}
-		} catch (error) {
-			// A failed axios request should result in an error.
-			showErrorNotification(error.response.data as string);
 		}
-		// Refetch meters details by invalidating its api cache.
-		this.props.resetApiCache();
+		resetApiCache();
```
huss: Could there be a comment if this is needed. I would also like to know why it is here.

src/client/app/translations/data.ts
```diff
@@ -580,6 +584,7 @@ const LocaleTranslationData = {
 		"create.user": "Créer un utilisateur",
 		"csv": "CSV",
 		"csv.file": "Fichier CSV",
+		"csv.clear.button": "Forme claire",
 		"csv.common.param.gzip": "GzipComment",
```
huss: I'm just noticing now but is the French string for gzip correct?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+								<TooltipMarkerComponent page='help.csv.header' helpTextId={tooltipStyle.tooltipReadings} />
+							</div>
+						</h2>
+					</Col>
+				</Row>
+				<Row className='justify-content-md-center'>
+					<Col md='auto'>
+						<FormFileUploaderComponent
+							formText='csv.upload.meters'
+							onFileChange={handleFileChange}
```
huss: Do people think there should be a second string saying meter upload when in the title above?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
+							reference={fileInput}
+							required
+						/>
+						<FormGroup>
+							<Container>
+								<Row>
+									<Col>
+										<Label for='gzip'>
+											<div style={checkBox}>
+												<div>
```
huss: Is this div needed? If it is for spacing maybe use the checkBox style to add that.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -3,14 +3,21 @@
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import * as React from 'react';
```
huss, line 6: In all the files it would be nice to use the VSC command "organize imports" that will sort them in a standard way.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -26,315 +33,477 @@ function range(lower: number, upper: number): number[] {
...
+								</Input>
+							</Label>
+						</FormGroup>
+						<FormGroup>
+							<Label for='duplications'>
+								<div className='pb-1'>
+									{translate('meter.readingDuplication')}
+								</div>
+								<Input type='select' id='duplications' name='duplications' value={selectedMeter?.readingDuplication || ''} onChange={handleChange}>
+									{range(1, 10).map(i => (
```
huss: I like the way this is done. The recent change to the meter create/edit pages lists each one. Are you willing to roll this to those pages? Maybe put the range functions into a general util directory.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -26,315 +33,477 @@ function range(lower: number, upper: number): number[] {
...
-			showErrorNotification(error.response.data as string);
-		}
-	}
-
-	private handleSetMeterName(e: React.ChangeEvent<HTMLInputElement>) {
-		const target = e.target;
-		this.props.setMeterName(MODE.readings, target.value);
-	}
+/**
+ * Defines the CSV Readings page card view
```
huss: Is this a card view? Maybe put this type of comment in the CSV meters file.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -26,315 +33,477 @@ function range(lower: number, upper: number): number[] {
...
-									<FormattedMessage id='csv.readings.param.cumulative' />
-								</Label>
-								<Col sm={12}>
-									<Input type='select' name='cumulative' onChange={this.handleSetCumulative}>
-										<option value={BooleanMeterTypes.meter}> {translate('BooleanMeterTypes.meter')} </option>
-										<option value={BooleanMeterTypes.true}> {translate('BooleanMeterTypes.true')} </option>
-										<option value={BooleanMeterTypes.false}> {translate('BooleanMeterTypes.false')} </option>
-									</Input>
+						<br />
+						<CreateMeterModalComponent onCreateMeter={handleCreateMeter}/>
```
huss: I tried being logged in with the CSV role and was able to create a new meter. However, only an admin should be able to do that and if the CSV user goes to the meter page they cannot. I'm not sure what is up but it is an issue.

Also, if a CSV user cannot create a meter then it might be good to revisit the logic above about a newly created meter that is not visible to the CSV user.

src/client/app/components/HeaderButtonsComponent.tsx
```diff
@@ -174,8 +174,15 @@ export default function HeaderButtonsComponent() {
 								style={state.csvViewableLinkStyle}
 								disabled={state.shouldCSVButtonDisabled}
```
huss: This is not working as I would expect. It is supposed to gray out the current page. Given this, I think there needs to be state for each CSV page and not just one. NOte the route seems wrong above and that should be part of this change.

src/client/app/translations/data.ts
```diff
@@ -86,35 +86,38 @@ const LocaleTranslationData = {
 		"create.user": "Create a User",
 		"create.unit": "Create a Unit",
 		"csv": "CSV",
-		"csv.file": "CSV File",
+		"csvMeters": "CSV Meters",
```
huss, line 89: Many of the changes in English are not reflected in the other languages.

huss (commented):
Thanks to @mikedivine and anyone else involved in this PR. Overall, it works as desired. I am releasing this as comments to get my thoughts to the team. I suspect we will want to discuss at an upcoming meeting. I still need to do some more careful reading/testing but maybe that can wait until these comments are decided. As always, I appreciate your thoughts and I'm here to clarify, esp. given my thoughts were released quickly.

src/client/app/components/HeaderButtonsComponent.tsx
mikedivine: I did not see this first go around.  This is now updated and seems to work properly.

src/client/app/components/TooltipHelpComponent.tsx
mikedivine: Yes, I believe this can be removed as a help link but the documentation can stay because it has useful information outside of using the site to upload data.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: If a CSV user can create meters by uploading csv data then I find it odd that they cannot manually do so. Seems like a contradiction, but I definitely can implement where the create meter button is disabled if they are not an admin if that is preferred.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Not a card view, oopsie, removed that portion of the comment

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: done, thanks for the tip

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Yeah, that is a good point.  I would be inclined to say it is unnecessary with the title above.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: This range function was written in the original combined CSV components, says by lindavin. But yes, I can add this in utilities and use this function in the create and edit pages.

mikedivine: I have now added this to utilities and is now used in create and edit Meters components.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Done. I have now changed it to only use one parameter of Refresh Readings and the Server API now uses the function in src/server/services/refreshAllReadingViews.js to do this.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: no longer needed, was when we had to pass props from a container

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Yes, correct, I believe we will have to create a seperate issue for this since all the CSV processing is done on the server side currently, so I don't see an easy way to use the RTK

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: This has now been set to not allow CSV users to create a meter manually and still only see visible meters to upload readings for.

src/client/app/translations/data.ts
mikedivine: Ok, I believe these should all be addressed now.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Fixed to use AppDispatch

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Yes, it just helps with alignments and spacing, etc.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Looks like this is resolved now.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Yes, default state and values now used.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: hmmm, not sure I understand, let's discuss in meeting.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: I removed it.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: you are correct, not needed, will remove.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: done

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: now implemented

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: added a comment in the code:
	// This is needed until a React CSV API is created or integrated with the Redux meter API

src/client/app/translations/data.ts
```diff
@@ -230,7 +235,8 @@ const LocaleTranslationData = {
 		"help.admin.unitedit": "This page allows admins to edit units. Please visit {link} for further details and information.",
 		"help.admin.unitview": "This page shows information on units. Please visit {link} for further details and information.",
 		"help.admin.user": "This page allows admins to view and edit users. Please visit {link} for further details and information.",
-		"help.csv.header": "This page allows certain users to upload meters and readings via a CSV file. Please visit {link} for further details and information.",
+		"help.csv.meters": "This page allows certain users to upload meters via a CSV file. Please visit {link} for further details and information.",
```
huss: I think this should be "allows admins to" since we decided to limit to admins.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -3,338 +3,490 @@
...
+									<FormGroup>
+										<Label for='cumulativeReset'>
+											<div className='pb-1'>
+												{translate('csv.readings.param.cumulative.reset')}
+											</div>
+											<Input
+												type='select'
+												id='cumulativeReset'
+												name='cumulativeReset'
+												value={getBooleanMeterType(selectedMeter?.cumulative)}
```
huss: I think this should be cumulativeReset.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -3,338 +3,490 @@
...
+						<FormGroup>
+							<Label for='timeSort'>
+								<div className='pb-1'>
+									{translate('meter.timeSort')}
+								</div>
+								<Input
+									type='select'
+									id='timeSort'
+									name='timeSort'
+									value={selectedMeter?.timeSort || ''}
```
huss: I think all the input with state similar to this have an issue. The handle updates the readingsData state but the value shown is the meter value. I think this relates to another concern I had that when you change a meter you only set the meterIdentifier in the readingsData state and not all the related values. You can fix as you want but here is a thought:
- Initially (as is done) the readingsData gets the default values. This means in most cases it will be the meter value. This, by definition, must change when the meter is selected (see next).
- When a meter is selected (which includes creation), the readingsData state is updated with the new meter's values. This does mean any selected items are reset.
- The value on item change is set to the readingsData value and not the meter. This avoids modifying the meter values but keeps the menus showing the current value.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,192 @@
...
-			</div>
-		);
-	}
+	const checkBox = {
+		display: 'flex'
+	};
 
-}
-export default connect(null, mapDispatchToProps)(MetersCSVUploadComponent);
+	return (
```
huss, line 93: I know we all discussed this but noting that only an admin will be allowed to create a meter.

src/client/app/utils/range.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Get array of numbers from a low and a high number
```
huss: I liked the JSDoc in the original version in ReadingsCSVUploadComponent.tsx. It made it clear whether lower/higher was inclusive or exclusive. May not be needed given another comment.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 import * as _ from 'lodash';
```
huss: I think most files have a blank line after the MPL notice so can it be added back.

src/client/app/utils/range.ts
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: I became curious if there was a standard way to do this. I found lots of solutions but I think these two seem the best choices:
1. lodash has a range function that does exactly what this function does.
2. [...Array(5).keys()]
Given we already use lodash, I think the first choice is the way to go. Just make sure to do the new import way to only import the needed function. This removes this file.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,192 @@
```
huss, line 4: This is a general comment so just sticking here. Given the time since development was merged with this work, I think it would be good to merge it again to be sure it is good with all the changes. (I just noted there are merge conflicts so this really needs to happen.)

src/client/app/translations/data.ts
```diff
@@ -579,35 +585,38 @@ const LocaleTranslationData = {
...
-		"csv.readings.param.lengthGap": "Écart de longueur",
-		"csv.readings.param.length.variation": "Variation de longueur",
-		"csv.readings.param.meter.name": "Nom du compteur",
-		"csv.readings.param.refresh.hourlyReadings": "Actualiser les relevés horaires",
-		"csv.readings.param.refresh.readings": "Actualiser les lectures quotidiennes",
+		"csv.readings.param.length.gap": "Écart de longueur:",
+		"csv.readings.param.length.variation": "Variation de longueur:",
+		"csv.readings.param.meter.name": "Nom du compteur:",
+		"csv.readings.param.meter.identifier": "Identifiant du compteur:",
+		"csv.readings.param.refresh.hourly.readings": "Actualiser les relevés horaires",
```
huss: Is this translation used?

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -32,7 +32,7 @@ function validateReadings(arrayToValidate, conditionSet, meterName = undefined)
  * @param {Moment} maxDate inclusive latest acceptable date (won't be rejected)
  * @param {number} maxError maximum number of errors to be reported, ignore the rest
```
huss, line 33: Not your doing but since you changed the name of the last parameter could you add it to the JDDoc? Same for ones below.

src/server/services/csvPipeline/uploadReadings.js
```diff
@@ -20,72 +17,21 @@ const Preferences = require('../../models/Preferences');
  * @returns 
  */
 async function uploadReadings(req, res, filepath, conn) {
-	const { meterName, createMeter, headerRow, update, honorDst, relaxedParsing } = req.body; // extract query parameters
+	const { meterIdentifier, headerRow, update, honorDst, relaxedParsing } = req.body; // extract query parameters
```
huss: I have a feeling that the changes in the PR to meterIdentifier impact what is sent by curl. Is that correct? If so, we need to be careful that the documentation is updated for what is actually sent.

src/server/services/csvPipeline/uploadReadings.js
```diff
@@ -20,72 +17,21 @@ const Preferences = require('../../models/Preferences');
...
-				meterCreated = true;
-				log.info('Creating meter ' + tempMeter.name);
-				return await Meter.getByName(tempMeter.name, conn); // Get meter from DB after insert because some defaults are set within the DB.
-			}
 		});
-	if (!meterCreated && createMeter === BooleanTypesJS.true) {
-		log.warn('The create meter was set but the meter already existed for meter ' + meter.name);
-	}
-
+	console.log('The meter details from getByIdentifier: ' + meter);
```
huss: Should this output be left?

src/server/models/Meter.js
```diff
@@ -166,6 +166,17 @@ class Meter {
 		return Meter.mapRow(row);
 	}
 
+	/**
```
huss, line 169: Bad formatting.

src/server/sql/meter/get_meter_by_identifier.sql
```diff
@@ -0,0 +1,5 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss, line 1: This needs a DB migration. Let me know if you need help with this.

src/client/app/components/FormFileUploaderComponent.tsx
```diff
@@ -3,14 +3,15 @@
...
-import { FormattedMessage } from 'react-intl';
-import { Col, Input, FormGroup, FormText, Label } from 'reactstrap';
+import { Col, Input, FormGroup, Label } from 'reactstrap';
+import translate from '../utils/translate';
 
 interface FileUploader {
-	reference: React.RefObject<HTMLInputElement>;
+	reference?: React.RefObject<HTMLInputElement>;
 	required: boolean;
 	formText: string;
```
huss: This used to be used on this page but it seems removed. Should it stay? If not, all the uses of this interface should be updated.

src/client/app/translations/data.ts
```diff
@@ -86,35 +86,40 @@ const LocaleTranslationData = {
...
+		"csv.readings.param.endOnly": "End Only times:",
 		"csv.readings.param.honor.dst": "Honor Daylight Savings Time",
-		"csv.readings.param.lengthGap": "Length Gap",
-		"csv.readings.param.length.variation": "Length Variation",
-		"csv.readings.param.meter.name": "Meter name",
-		"csv.readings.param.refresh.hourlyReadings": "Refresh Hourly Readings",
-		"csv.readings.param.refresh.readings": "Refresh Daily Readings",
+		"csv.readings.param.length.gap": "Length Gap:",
+		"csv.readings.param.length.variation": "Length Variation:",
+		"csv.readings.param.meter.name": "Meter name:",
```
huss: Is this key needed anymore?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -3,338 +3,490 @@
...
-		this.props.selectCumulativeReset(target.value as BooleanMeterTypes);
-	}
-
-	private handleSetCumulativeResetStart(e: React.ChangeEvent<HTMLInputElement>) {
-		const target = e.target;
-		this.props.setCumulativeResetStart(target.value);
-	}
+	const handleFileChange = (file: File | null) => {
+		setSelectedFile(file);
+		if (!file) return;
```
huss: Please use { } on all if statements. Applies the CSV meter too. I might consider and if/else for all these cases since I think it is less prone to errors.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -3,338 +3,490 @@
...
+									<FormGroup>
+										<Label for='cumulativeResetStart'>
+											<div className='pb-1'>
+												{translate('csv.readings.param.cumulative.reset.start')}
+											</div>
+											<Input
+												id='cumulativeResetStart'
+												name='cumulativeResetStart'
+												onChange={handleChange}
+												required
```
huss: I'm unsure what this required does since I don't see any warning if nothing is entered.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -3,338 +3,490 @@
...
+												</div>
+											</div>
+										</Label>
+									</Col>
+								</Row>
+							</Container>
+						</FormGroup>
+						<div className='d-flex flex-row-reverse'>
+							<div className='p-3'>
+								<Button color='primary' type='submit' disabled={!isValidCSV}>
```
huss: I like the disabled check. I also see the warning if you don't select a meter and submit. Would it make sense to not allow submit on no meter selected?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: Maybe add a comment, even if similar to the one above about the React CSV API.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,192 @@
...
-		try {
-			e.preventDefault();
-			const current = this.fileInput.current as HTMLInputElement;
-			const { files } = current;
-			if (files && (files as FileList).length !== 0) {
-				await this.props.submitCSV(files[0]);
-				// TODO Using an alert is not the best. At some point this should be integrated
-				// with react.
-				showSuccessNotification('<h1>SUCCESS</h1>The meter upload was a success.');
+	// This is needed until a React CSV API is created or integrated with the Redux meter API
```
huss: Would the team like to put in an issue about the needed CSV updates for React API? If not, I can do it.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: I'm still unclear so could someone help me understand?

src/client/app/components/TooltipHelpComponent.tsx
huss: I see it gone from data.ts but it is still in this file. Is it needed?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: Is this resolved?

huss (changes requested):
Thanks for all the updates. I have made some comments to address. I'm also delaying final testing until this is done and development is merged. This is looking good and I think it should all be done soon.

huss:
I just pushed a commit to fix the pipeline test for identifier use and no create meter. It also fixes meter upload to consistently use identifier.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Sure, we will do this.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Will do.

src/client/app/components/meters/EditMeterModalComponent.tsx
mikedivine: done

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: These are ReactStrap elements I saw used on the other components and they just make sure everything is aligned in rows and columns. This could be done with div elements if you prefer or another method if you have a suggestion, just let me know if you have a preference.

src/server/models/Meter.js
mikedivine: oops, fixed

src/client/app/components/TooltipHelpComponent.tsx
mikedivine: Oh, I see, removed.

src/client/app/translations/data.ts
mikedivine: no, removing

src/client/app/translations/data.ts
mikedivine: Should just be Gzip like the others, fixed

src/client/app/translations/data.ts
mikedivine: Got it, fixed

src/client/app/translations/data.ts
mikedivine: no, removing, fixed

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Got it, fixed

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: yes, now removed

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Got it, the required on the file input doesn't allow to submit, but I've added to also disable if meter is not selected.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: yes, fixed

src/server/services/csvPipeline/uploadReadings.js
mikedivine: no, removed

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: this has been resolved.

src/client/app/utils/range.ts
mikedivine: got it, moving forward to use lodash

src/client/app/utils/range.ts
mikedivine: Got it, will do.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: I rewrote the api to update redux with the meters adding, although you may need to look and see if my method is ideal for you or what your thoughts would be.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Yes, this is resolved now

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: I have updated the frontend CSV api to now send new meters using dispatch to update the store. If it's to update meters, it invalidates all meters as discussed with you previous. This is how the current redux api does it, but we weren't sure why exactly. If it changes in the future, we can adjust the api accordingly.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Rewrote this api to have readings invalidated once new readings are added.  An issue can still be added so in the future it is written to add the data without invalidating.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Also they can no longer access the CSV Meters page to upload meters

src/client/app/components/FormFileUploaderComponent.tsx
mikedivine: Are you talking about formText? Looks like this component is only used on uploadReadings and uploadMeters, I will remove it from the component

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Understood, this makes a lot of sense and I will change it to use the method you described.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: OK, I got rid of the Row/Col except for outside to center everything inside a column and then rows and columns for the checkboxes.

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -4,53 +4,97 @@
...
-			update: uploadPreferences.update ? BooleanTypes.true : BooleanTypes.false
-		};
-		for (const [preference, value] of Object.entries(uploadPreferencesForm)) {
-			formData.append(preference, value.toString());
+	try {
+		const response = await backend.doPostRequest<MetersUploadResponse>('/api/csv/meters', formData);
+		const { message, meters } = response;
+		const meterNames = meters.map(meter => meter.name).join(', ');
+		// If new meters were added to DB, have redux add them to getMeters state
+		/// If meters were updated to the DB, invalidate meters
```
huss: extra /

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -4,53 +4,97 @@
...
+		message = await backend.doPostRequest<string>('/api/csv/readings', formData);
+	} catch (error) {
+		return { success: false, message: error.response.data };
+	}
+	dispatch(baseApi.util.invalidateTags(['Readings']));
+	return { success: true, message: message };
+};
+
+export const submitMeters = async (uploadPreferences: MetersCSVUploadPreferencesItem,
+	metersFile: File, dispatch: any): Promise<ApiResponse> => {
```
huss: I'm curious if you think that dispatch should be passed or allocated here. In the calling code the dispatch is not used except to pass to this function. Same for submitReadings.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -4,107 +4,191 @@
...
-			const current = this.fileInput.current as HTMLInputElement;
-			const { files } = current;
-			if (files && (files as FileList).length !== 0) {
-				await this.props.submitCSV(files[0]);
-				// TODO Using an alert is not the best. At some point this should be integrated
-				// with react.
-				showSuccessNotification('<h1>SUCCESS</h1>The meter upload was a success.');
+	const handleFileChange = (file: File | null) => {
+		setSelectedFile(file);
+		if (!file) {
```
huss: I don't think there is anything wrong with this but I'm trying to figure out when it can happen. I tried Chrome and Firefox and I could not select a file that did not exist nor unset once set. Do you know if it is possible for the file not to exist? Applies to readings too.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
+		}
+	};
 
-	private handleSetLengthGap(e: React.ChangeEvent<HTMLInputElement>) {
-		const target = e.target;
-		this.props.setLengthGap(target.value);
-	}
+	const handleSelectedMeterChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		const selectedMeterIdentifier = e.target.value;
+		const foundMeter = visibleMeters.find(meter => meter.identifier === selectedMeterIdentifier) || null;
```
huss: Since the meter came from a drop down menu, I'm trying to figure out how it would not be in visibleMeters?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
+			[name]: checked
+		}));
+	};
 
-	private handleSetHonorDst() {
-		this.props.toggleHonorDst();
-	}
+	const handleNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		let aNumber = Number(e.target.value);
+		if (aNumber < 0) {
```
huss: This may go away but I have a couple of thoughts:
1. The value is set to a different one that the user input without notice.
2. This uses the same name as other places but has modified functionality.
3. Given the min is 0 on the form, I cannot find a way to enter a negative value so is this needed?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
+	authApi.useTokenPollQuery(undefined, { skip: !isAdmin, pollingInterval: authPollInterval });
+	// We only want displayable meters if non-admins because they still have
+	// non-displayable in state.
+	const { visibleMeters } = useAppSelector(selectVisibleMeterAndGroupData);
+	// This is the state for the form data for readings
+	const [readingsData, setReadingsData] = React.useState<ReadingsCSVUploadPreferencesItem>(ReadingsCSVUploadDefaults);
+	// This is the state for the file to be uploaded
+	const [selectedFile, setSelectedFile] = React.useState<File | null>(null);
+	// This is the state for the chosen meter to upload readings for
+	const [selectedMeter, setSelectedMeter] = React.useState<MeterData>(defaultValues);
```
huss: Can you help me understand why selectedMeter exists? From my quick check, it seems readingsData is also set with the updated meter info. Thus, is this a duplicated state?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
-				window.location.reload();
+	// If a new meter was created then select it as the meter to be used
+	React.useEffect(() => {
+		if (createdMeterIdentifier) {
+			const createdMeter = visibleMeters.find(meter => meter.identifier === createdMeterIdentifier) || null;
+			if (createdMeter) {
+				setSelectedMeter(createdMeter);
+				setReadingsData(prevState => ({
+					...prevState,
+					meterIdentifier: createdMeter.identifier
```
huss: A lot of other values are set below on meter change. Curious why not here?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
-					<FormFileUploaderComponent formText='csv.upload.readings' reference={this.fileInput} required labelStyle={titleStyle} />
-					<FormGroup>
-						<Label style={titleStyle}>
-							<FormattedMessage id='csv.readings.section.cumulative.data' />
 						</Label>
-						<Col sm={8}>
+						<FormGroup>
+							{isAdmin && <><CreateMeterModalComponent onCreateMeter={handleCreateMeter} /></>}
+						</FormGroup>
+						<FormGroup>
```
huss: I'm really sorry to be bringing this up so late in the process. It is fine to either address or make an issue instead. I was looking at the meter create page and noticed it looks different than this page in terms of order, layout and checks. I was thinking the order should be similar to the meter create/edit page and the checks are the same (so you cannot enter the cumulative start/end if reset not chosen, etc.) I think making similar columns for the items would also be nice. I'm open to ideas on the exact layout (here there are section headers) but being consistent seems good.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
-
+import TooltipHelpComponent from '../TooltipHelpComponent';
+import TooltipMarkerComponent from '../TooltipMarkerComponent';
+import CreateMeterModalComponent from '../meters/CreateMeterModalComponent';
 /**
- * Returns a range of values between the specified lower and upper bounds.
- * @param lower The lower bound, which will be included in the range.
- * @param upper The upper bound, which will be excluded from the range.
- * @returns An array of values between starting from the lower bound and up to and excluding the upper bound.
+ * Defines the CSV Readings page
```
huss, line 26: Blank line nice.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,513 @@
...
+import { submitReadings } from '../../utils/api/UploadCSVApi';
+import { ReadingsCSVUploadDefaults, convertBoolean } from '../../utils/csvUploadDefaults';
+import { showErrorNotification, showSuccessNotification } from '../../utils/notifications';
 import translate from '../../utils/translate';
 import FormFileUploaderComponent from '../FormFileUploaderComponent';
-
+import TooltipHelpComponent from '../TooltipHelpComponent';
+import TooltipMarkerComponent from '../TooltipMarkerComponent';
+import CreateMeterModalComponent from '../meters/CreateMeterModalComponent';
 /**
```
huss, line 25: Blank line between is nice.

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -4,53 +4,97 @@
...
-		formData.append('csvfile', readingsFile); // It is important for the server that the file is attached last.
-		return await this.backend.doPostRequest<void>('/api/csv/readings', formData);
+	let message = '';
+	try {
+		message = await backend.doPostRequest<string>('/api/csv/readings', formData);
+	} catch (error) {
+		return { success: false, message: error.response.data };
+	}
+	dispatch(baseApi.util.invalidateTags(['Readings']));
+	return { success: true, message: message };
```
huss: If I understand the logic, I think it is much clearer to put all the success in the try part. Then it catches all errors (not just the post) and it makes clear what goes with what. Using a return to avoid code should, in my opinion, only be used when necessary and where the logic would be difficult.

src/server/services/csvPipeline/uploadReadings.js
```diff
@@ -126,43 +71,43 @@ async function uploadReadings(req, res, filepath, conn) {
 		readingRepetition = parseInt(duplications, 10);
 	}
 
-	if (timeSort === undefined || timeSort === TimeSortTypesJS.meter) {
+	if (timeSort === undefined) {
```
huss, line 88: I think this works since the help page says "If a parameter is not supplied in a direct file upload then it defaults to the meter value.". However, I wanted to ask if you have tested this.

src/server/test/web/csvPipelineTest.js
```diff
@@ -518,23 +557,40 @@ for (let fileKey in testCases) {
...
 				let inputPath = `${__dirname}/csvPipeline/${inputFile}`;
 				let inputBuffer = fs.readFileSync(inputPath);
 				let evalString = `${testCases[fileKey]['chaiRequest'][index]}.attach('csvfile', inputBuffer, '${inputPath}')`;
 				// TODO It would be nice if this was not an eval. Tried a function with closure but could not get it to work as did not find chai.
 				const res = await eval(evalString);
 				expect(res).to.have.status(testCases[fileKey]['responseCode'][index]);
-				expect(res).to.be.html;
 				// OED returns a string with messages that we check it is what was expected.
-				expect(res.text).to.equal(testCases[fileKey]['responseString'][index]);
+				try {
```
huss: A few of thoughts on this:
1. The check for html was removed but a json check did not replace it.
2. The fact that the test needed to be modified is an indication that you changed the expected response. These responses have been published and sites have a reason to expect they will not change unless absolutely needed. They were set up as html for a reason and now this one is not html. I'm hoping the old behavior can easily be put back. Maybe you can not put a success code in the response as JSON and look at the response code. It looks as if that changes then the added item to via the JSON string is not needed.
3. Usually you have no need for a try/catch like this in a test. I think something unusual/undesirable is happening here.
I have not spent a lot of time on this so maybe I've missed something. We might need to discuss.

src/server/routes/csv.js
```diff
@@ -124,10 +123,13 @@ router.post('/meters', validateMetersCsvUploadParams, async (req, res) => {
 		}
 
 		const conn = getConnection();
-		await uploadMeters(req, res, csvFilepath, conn);
-		success(req, res, 'Successfully inserted the meters.');
+		const newlyCreatedMeters = await uploadMeters(req, res, csvFilepath, conn);
+		// Send the newly created meters back to the frontend
+		res.status(200).send({ message: '<h1>SUCCESS</h1>Successfully inserted the meters.', meters: newlyCreatedMeters });
```
huss: See a longer discussion in src/server/test/web/csvPipelineTest.js. I think this returned message changed too. It also changed how it is sent.

src/server/sql/meter/get_meter_by_identifier.sql
huss: My mistake. Not needed.

huss:
I'm noting that I pushed two commits.

huss (changes requested):
Thanks, again, to @mikedivine for continued work on this. Overall, this is looking good. I'm sorry to have more comments but please check them out. Let me know if I can help in any way.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Added.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Comment same as above, not a different line right?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: You are correct, think I was overdoing defensive programming.
I don't know if there any browsers that would let you put a filename in for a file that doesn't exist.
Removed Null and now doesn't check if File is Null.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Was coding defensively, but I will get rid of the if statement and use Type Assertion. Is that what you meant?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Yes, it should be, will add now.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: You are correct. This is not needed. I will remove this.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Before this I was able to enter a negative number by typing it in instead of using the arrows. I can put a warning when a user attempts to enter a negative number and I'll change the function name to handlePositiveNumberChange.  This could be done differently, so I'll check in with you before moving forward on this.

src/server/services/csvPipeline/uploadReadings.js
mikedivine: I don't believe it will change much for what is sent back because it just looks like:
  <h1>FAILURE</h1> <h2>It looks like the insert of the readings had issues with some or all of the readings where the processing of the readings returned these warning(s)/error(s):</h2>
  For reading #1 on meter pipe3 in pipeline: previous reading has value 0 start time 1970-01-01T00:00:00+00:00 end time 1970-01-01T00:00:00+00:00 and current reading has value 24 start time 2021-06-01T00:00:00+00:00 end time 2021-06-02T00:00:00+00:00 with timeSort increasing; duplications 1; cumulative true; cumulativeReset false; cumulativeResetStart 00:00:00; cumulativeResetEnd 23:59:59.999999; lengthGap 0; lengthVariation 0; onlyEndTime false
So "pipe3" could be referring to meterIdentifier or meterName, but the input of meterIdentifier is definitely a change so documentation needs to be updated with pertinent info.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: I will work on this.

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: Removed.

src/server/routes/csv.js
mikedivine: Removed meterData being sent back as per discussion.

src/server/test/web/csvPipelineTest.js
mikedivine: Yes, reversing this back to previous per our discussion.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: I matched the matching input fields in the same order as is on Create/Edit Meter.  I moved the file picker above the options and I moved the CSV file options directly under the file picker as it made the most sense to me.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: Is this referring to the file reference or this: const [selectedFile, setSelectedFile] = React.useState<File | null>(null);

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: I have moved it inside the try catch

src/server/services/csvPipeline/uploadReadings.js
mikedivine: Yes, I tested this using the curl command and didn't include parameters and it still worked.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
mikedivine: It's working this way.

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: This is not a component, that's why it is passed as a parameter

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: We spoke, I will utilize what the other components use.

src/client/app/utils/api/UploadCSVApi.ts
huss: Needed since not a component.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,521 @@
...
+								</div>
+							</Col>
+						</Row>
+						<div className='pb-3'></div>
+						<Row xs='1' lg='2'>
+							<Col>
+								<FormGroup>
+									<Label for='cumulative'>
+										<div className='pb-1'>
+											{translate('csv.readings.param.cumulative')}
```
huss: I noticed some of the titles differed from meter. I realized that there are a meter and csv equivalent keys in data.ts. I think it should be changed to only use a single set of keys. Ideally, the name would not be associated with one of these items (such as cumulative vs. meter.cumulative).

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,521 @@
...
+										</div>
+									</Label>
+									<Input
+										type='select'
+										id='cumulative'
+										name='cumulative'
+										value={readingsData.cumulative}
+										onChange={handleChange}
+									>
+										<option value={BooleanTypes.true}> {translate('BooleanMeterTypes.true')} </option>
```
huss: Create meter looks over the enum to create the menu. While it is only two values, I think it is good practice. If you agree, could it mirror meter?

This applies to a couple of items.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -2,351 +2,521 @@
...
+									<Input
+										type="number"
+										id='lengthGap'
+										name='lengthGap'
+										min='0'
+										value={readingsData.lengthGap}
+										onChange={handleNumberChange}
+										invalid={readingsData.lengthGap < 0}
+									/>
+									<FormFeedback>
```
huss: I'm okay with the feedback but I was unable to enter a negative value so I'm uncertain if the user can see these.

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -4,53 +4,75 @@
...
-	ReadingsCSVUploadPreferencesItem, MetersCSVUploadPreferencesItem, BooleanTypes, CSVUploadPreferencesForm, ReadingsCSVUploadPreferencesForm
+	BooleanTypes,
+	CSVUploadPreferencesForm,
+	MetersCSVUploadPreferencesItem,
+	ReadingsCSVUploadPreferencesForm,
+	ReadingsCSVUploadPreferencesItem
 } from '../../types/csvUploadForm';
+import ApiBackend from './ApiBackend';
+
+interface apiResponse {
```
huss: I'm curious as to why this was changed from starting capitalized. A search found lots of interfaces in OED and I did not easily see one this way. What do you think?

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -4,53 +4,75 @@
...
-			formData.append(preference, value.toString());
-		}
-		formData.append('csvfile', metersFile); // It is important for the server that the file is attached last.
-		await this.backend.doPostRequest<void>('/api/csv/meters', formData);
+	try {
+		const response = await backend.doPostRequest<string>('/api/csv/meters', formData);
+		// Meter Data was sent to the DB, invalidate meters for now
+		dispatch(baseApi.util.invalidateTags(['MeterData']));
+		// meters were invalidated so all meter changes will now reflect in Redux state, now return
+		return { status: true, message: response };
```
huss: Given this is used as a boolean where true is success, I'm curious about changing the name. Could you help me understand?

src/server/services/csvPipeline/uploadReadings.js
huss: Probably fine but can I get a confirmation on this?

huss (changes requested):
Thanks for another update. Testing did not find issues. I have a few more comments that I hope are easy to address. Other than that it seems ready.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Yes, will mirror this style.

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: No valid reason, think I was editing some stuff and must have thought it made more sense in the moment, but I agree 'success' makes more sense.  Will change it back.

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: mistake, I agree it should be capitalized. Will change back.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: I will go through these and match them and remove unused translations.

src/client/app/utils/api/UploadCSVApi.ts
```diff
@@ -4,53 +4,75 @@
...
-export default class UploadCSVApi {
-	private readonly backend: ApiBackend;
+export const submitReadings = async (uploadPreferences: ReadingsCSVUploadPreferencesItem, readingsFile: File,
+	dispatch: Dispatch): Promise<ApiResponse> => {
+	const backend = new ApiBackend();
+	const formData = new FormData();
+	// The Boolean values in state must be converted to the submitted values of yes and no.
+	const uploadPreferencesForm: ReadingsCSVUploadPreferencesForm = {
+		...uploadPreferences,
+		gzip: uploadPreferences.gzip ? TrueFalseType.true : TrueFalseType.false,
```
huss: When I try to upload a file I get an enum error. I did not try to track down where it is coming from. Whatever changes are made, we also need to be sure that the curl commands work.

I also notice that BooleanTypes was removed from the client but is in many places in the server code. I supect this related.

huss (changes requested):
I was not sure if this was ready for review but I wanted to move it forward. I've made one comment to address and there are a couple of old ones to finalize. If you want to discuss the new one then let me know as it may not be trivial. Thanks for all your efforts.

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: Yes, I removed BooleanTypes as Meter Components were using TrueFalseType which mirrored the BooleanTypes.  Curl method worked for me to add the meters while the Web Readings upload also worked but I did get the type error on the web CSV Meters upload attempt.  I'll try to figure what is happening.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: Yes they don't allow a negative number, so we will remove the form feedback for these

src/server/services/csvPipeline/uploadReadings.js
mikedivine: This has been tested, not sure if you will see this..

mikedivine (commented):
Oh I see I have to submit review for you to see comments

mikedivine (commented):
Oh I see I have to submit review for you to see comments

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
mikedivine: completed

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: I am no longer able to reproduce this error. Let me know if you are still getting it.

src/server/services/pipeline-in-progress/validateReadings.js
mikedivine: How would I add something to the JSdoc?

huss: You can put
`` * @param {string} meterName The name of the meter``
at the end of the param list. It will show up as part of the JSDoc in VSC (or if manually created which we don't do).

src/client/app/utils/api/UploadCSVApi.ts
huss: This is the error I see:

image

I still see the error and I think I have debugged it. You converted the client CSVUploads to be true/false. That maps true/false to yes/no which is good for the display. The issue is that the server still expected yes/no. See src/server/services/csvPipeline/validateCsvUploadParams.js for BooleanTypesJS which is used as the allowed values for cumulative, etc. I believe I did this a while back to show yes/no to the user but it was not the best way to do it. Now the enum loop shows that value but sends true/false. This is better since internally the value is a boolean. I verified this in two ways:

1. I put a console.log in src/server/services/csvPipeline/validateCsvUploadParams.js at the start of validateReadingsCsvUploadParams to show the values sent were true/false.
2. I changed `cumulative: new EnumParam('cumulative', [BooleanTypesJS.true, BooleanTypesJS.false]),` to `cumulative: Boolean,` and the error went away (there are other errors for the other params that have a similar issue.

Thus, I think this type of change will resolve the issue and improve the code from what it was before your changes. Please change for all the types needed, if you agree, and let me know if you think otherwise.

As long as I was deep in the code, I also wanted to double check about what happens if a direct (e.g., curl) request is made. In src/server/services/csvPipeline/uploadReadings.js on line 86 is checks if cumulative is not provided and then uses the meter value (assuming it exists). So, you can still get the meter value by not providing the value. This is what is already documented in https://openenergydashboard.org/helpV1_0_0/adminReadingsImport/ so no changes are needed. I'll add that what you said before about checking it was correct (I'm just paranoid :-).

huss: I'm unresolving since I still see this. I was doing final testing when I saw this.

huss:
I just merged in development since recent PRs caused a merge conflict and I had meant to do this PR earlier. I'll get it to work. After that there is one item unresolved that remains.

@mikedivine Can you look over the last comment and also verify my large merge did not cause an issue. My basic testing did not find any issues outside the last comment.

huss:
I'm going to work with @mikedivine on the failed test to resolve meter name vs identifier.

src/client/app/utils/api/UploadCSVApi.ts
mikedivine: Got it, I will work on fixing this.

mikedivine: I have fixed this issue as far as I can tell. I created a utility function called normalizeBoolean which took any 'yes' 'no' values and converted to boolean.

src/server/services/csvPipeline/validateCsvUploadParams.js
```diff
@@ -31,27 +19,30 @@ MeterTimeSortTypesJS = Object.freeze({
...
- */
-BooleanMeterTypesJS = Object.freeze({
-	true: 'yes',
-	false: 'no',
-	// meter means to use value stored on meter or the default if not.
-	meter: 'meter value or default'
-});
+// This is to allow curl users to still send 'yes' and 'no' for parameters sent,
+// The frontend will just use standard boolean values to send to the API
+BooleanCheckArray = ['yes', 'no', 'true', 'false', true, false];
```
huss: Should this be a const?

huss (commented):
In my attempt to keep up with the PRs, I reviewed your changes up to now. They all seem good. I made one small comment.

src/server/services/csvPipeline/validateCsvUploadParams.js
mikedivine: yes, fixed

huss (approved):
I want to thank @mikedivine for a lot of work and effort to meet all the changing requests. I have reviewed the recent changes and retested. I found everything works as desired. This is ready to merge. Congratulations.

PR #1288, opened 2024-06-26
Spanish edits

juanjoseguva:
# Description

Proofreading of all Spanish translations in data.ts. Some messages had not been translated at all and others were missing from one or the other language. 

Fixes #[issue]

No associated issue

## Type of change

## Checklist

## Limitations

French translations are unproofed and many remain untranslated. Those lines are marked by "\u{26A1}" at the end of the message, which is the high voltage symbol in unicode.


src/client/app/translations/data.ts
```diff
@@ -27,7 +27,7 @@ const LocaleTranslationData = {
 		"AreaUnitType.none": "no unit",
 		"ascending": "Ascending",
 		"as.meter.unit": "as meter unit",
-		"as.meter.defaultgraphicunit": "as meter unit",
+		"as.meter.defaultgraphicunit": "as meter unit",//Ask
```
huss: I'm uncertain about this comment so you could you please explain?

src/client/app/translations/data.ts
```diff
@@ -328,8 +328,8 @@ const LocaleTranslationData = {
 		"meter.cumulativeReset": "Cumulative Reset:",
 		"meter.cumulativeResetEnd": "Cumulative Reset End:",
 		"meter.cumulativeResetStart": "Cumulative Reset Start:",
-		"meter.enabled": "Updates:",
-		"meter.endOnlyTime": "End Only Time:",
+		"meter.enabled": "Meter enabled",
```
huss: First, the others don't include "Meter". Second, I think the original with "Updates:" is better. Removing the colon was not desired.

src/client/app/translations/data.ts
```diff
@@ -1003,39 +1003,39 @@ const LocaleTranslationData = {
...
 		"AreaUnitType.meters": "metros cuadrados",
 		"AreaUnitType.none": "sin unidad",
-		"ascending": "Ascendente",
-		"as.meter.unit": "as meter unit\u{26A1}",
-		"as.meter.defaultgraphicunit": "as meter unit\u{26A1}",
+		"ascending": "Ascendiente",
+		"as.meter.unit": "como unidad de medidor",
+		"as.meter.defaultgraphicunit": "como unidad de medidor",
 		"bar": "Barra",
 		"bar.days.enter": "Enter in days and then hit enter\u{26A1}",
```
huss: This was just added. Could you translate it?

src/client/app/translations/data.ts
```diff
@@ -1071,91 +1072,92 @@ const LocaleTranslationData = {
...
+		"default.meter.maximum.date": "Revisión de la fecha de lectura máxima del medidor predeterminado",
+		"default.meter.reading.gap": "Deistancia predeterminada entre lecturas del medidor",
+		"default.meter.maximum.errors": "Número máximo de errores en la lectura del medidor",
+		"default.meter.disable.checks": "Desactivar revisiones de medidor predeterminado",
+		"delete.group": "Borrar grupo",
 		"delete.map": "Borrar mapa",
 		"delete.user": "Borrar usario",
 		"descending": "Descendente",
-		"discard.changes": "Descartar los Cambios",
+		"discard.changes": "Descartar los cambios",
```
huss, line 1133: In English the words are all capitalized. Should they be here? This applies in several of the changes.

src/client/app/translations/data.ts
```diff
@@ -1270,96 +1273,96 @@ const LocaleTranslationData = {
...
-		"map.is.not.displayable": "Pantalla desactivada",
-		"map.load.complete": "Carga de mapas completa de",
-		"map.modified.date": "Ultima modificación",
-		"map.name": "Nombre de mapa",
-		"map.new.angle": " Enumere el ángulo con respecto al norte verdadero (0 a 360)",
+		"map.is.deleted": "Mapa borrado de la base de datos",
+		"map.is.displayable": "Visualización activada",
+		"map.is.not.calibrated": "Necesita calibración",
+		"map.is.not.displayable": "Visualización desactivada",
+		"map.load.complete": "Carga de mapas completada de",
```
huss: Extra space at end of line.

src/client/app/translations/data.ts
```diff
@@ -1270,96 +1273,96 @@ const LocaleTranslationData = {
...
 		"name": "Nombre:",
 		"navigation": "Navegación",
-		"need.more.points": "Nececitar más puntos",
+		"need.more.points": "Nececita más puntos",
 		"no": "no",
 		"note": "Nota:",
-		"no.data.in.range": "No Data In Date Range\u{26A1}",
-		"oed": "Panel de Energía Abierta",
+		"no.data.in.range": "No hay datos para el rango de fechas",
+		"oed": "Open Energy Dashboard",
```
huss: I'm wondering if this should stay in English or be translated. It is a name but it also has a meaning as words. What do you think?

src/client/app/translations/data.ts
```diff
@@ -1368,114 +1371,115 @@ const LocaleTranslationData = {
...
-		"submit.new.user": "Entregar un nuevo usario",
+		"submit.changes": "Ingresar los cambios",
+		"submit.new.user": "Ingresar un nuevo usario",
 		"the.unit.of.meter": "La unidad del medidor",
 		"this.four.weeks": "Estas cuatro semanas",
-		"timezone.no": "no zona horaria",
+		"TimeSortTypes.decreasing": "decreciente",
+		"TimeSortTypes.increasing": "creciente",
+		"TimeSortTypes.meter": "valor del medidor o predeterminado",
+		"true": "Verdad",
```
huss: By the way OED does the ordering, this should not have moved in the file. It turns out the English one is in the wrong place. Could you put this back and fix the English placement?

src/client/app/translations/data.ts
```diff
@@ -1071,91 +1072,92 @@ const LocaleTranslationData = {
...
-		"csv.tab.meters": "Medidor",
+		"csv.readings.section.time.gaps": "Longitud de tiempos",
+		"csv.submit.button": "Entregar datos CSV",
+		"csv.tab.meters": "Medidores",
 		"csv.tab.readings": "Lecturas",
-		"csv.upload.meters": "Cargar medidores de CSV",
-		"csv.upload.readings": "Cargar lecturas de CSV",
+		"csv.upload.meters": "Subir medidores CSV",
+		"csv.upload.readings": "Subir lecturas CSV",
 		"custom.value": "Custom value\u{26A1}",
```
huss: A new key. Could you translate?

huss (changes requested):
Thanks to @juanjoseguva for doing this work. I appreciate you catching some issues and doing all the translations. I did not carefully check each one but they seem good. I put in a few comments to consider but overall this is very good.

huss:
I forgot to mention that I merged development to get rid of the conflict and also just remembered to push a small change for something I noticed that was not part of the PR.

src/client/app/translations/data.ts
juanjoseguva: I noticed the message is the same as the line above it in all languages, so I was unsure if "as.meter.defaultgraphicunit" should be deleted in favor of the generalized use of "as.meter.unit" or if the message associated with as.meter.defaultgraphicunit should be "as meter default graphic unit" instead of "as meter unit"

src/client/app/translations/data.ts
juanjoseguva: No, Spanish conventionally uses much less capitalization than in English. In fact, despite being proper nouns, the names of languages like "English" and "Spanish" are always inglés and español, never capitalized. The same is true of days of the week and months. For book and movie titles, only the first letter is capitalized. See https://es.wikipedia.org/wiki/Harry_Potter_y_la_c%C3%A1mara_secreta as an example or https://www.thoughtco.com/capitalization-in-spanish-3080293 for more information.

src/client/app/translations/data.ts
juanjoseguva: I go back and forth on this myself, as well as what the best translation would be. Torn between "Panel" and "Tablero". For either one you have a funny acronym, TEA or PEA

src/client/app/translations/data.ts
huss: I see and good catch. The as.meter.defaultgraphicunit should be "as meter default graphic unit". Could you update all the languages?

src/client/app/translations/data.ts
huss: Sounds. OED capitalized for emphasis but I think following standard Spanish is good so leave as is.

src/client/app/translations/data.ts
huss: I decided to see how it is used to see if that would help. And....I cannot see any use of the key "oed" (or 'oed'). If it was to be used then probably src/client/index.html is the place. I tired to get that title to show but could not even by disabling JS. I think we can delete this key.

huss (approved):
Thanks to @juanjoseguva for addressing all the comments, including ones outside the original work. Review found no issues so this is ready to merge. Congratulations on your first accepted work.

huss:
I made a mistake and missed that src/client/app/components/LogoComponent.tsx uses the "oed" translation key. I caught this in my final checks. I pushed a change to put it back. It is only used for the hover text over the icon so not too important.

PR #1300, opened 2024-07-09
Gitingnore

davanmur:
# Description

Updates gitignore to include src/server/tmp.


huss (approved):
Thanks to @davanmur for doing this. It is something that has annoyed me but I did not fix it for the project. Review and testing found it works as expected.

PR #1305, opened 2024-07-12
Fixing 3D axis labels

### LINKED ISSUE #1194: standardize how graphics get language aware times

Different parts of the code seem to use different methods to get date aware labeling. Two for hover text are:

- src/client/app/components/ThreeDComponent.tsx uses moment (``dateTS.format(moment.localeData().longDateFormat('L').replace(/YYYY/g, 'YY'))``).
- src/client/app/redux/selectors/lineChartSelectors.ts uses ``timeReading.format('ddd, ll LTS')``

There may be other ways in other places.

OED should decide the best way and consistently use that (maybe by centralizing the code).

Leave as is but the format may vary.

### Additional context

None

ianLewis8:
# Description

In order to standardize language aware times, the only one I noticed that was unique from the others was the 3D component.

The main reason for this was that the tick marks on the graph would overlap with the graph title and lines due to the axis labels being too long.

In order to fix this, I added a 'tickangle' to the y-axis of the graph that allowed the ticks to display horizontally and thus have more space. I also changed the format statement of this axis to use the same as other charts which was dateTS.format('YYYY-MM-DD HH:mm:ss').

Note that the issue describes another scenario where timeReading.format('ddd, ll LTS') is used. However, this is the standard when adjusting the vertical axis which is supposed to be different from the one I changed.

After adding these changes, the y-axis would display times for a small number of days. I fixed this by adding a few if-else statements that specify the number of ticks and tick interval for a few limited cases and then set to default for all other cases.

Fixes #1194 

## Type of change

## Checklist

## Limitations

No limitations to be noted.


huss (commented):
Thanks to @ianLewis8 for this work. Overall, it works well and I like it. I have one small question. When you zoom in to limited days, the axis shows the time of day. With 3D another axis is showing/graphing the time across the day so the date axis should only be the date in an ideal case. Would it be easy to only show the date and year on this axis?

ianLewis8:
Yes I can do that

ianLewis8:
I managed to resolve the issue with the yaxis showing time of day on limited number of days. Let me know if it looks good to you.

huss:
Thanks to @ianLewis8 for figuring this out. I know it was tricky and I appreciate your efforts. Everything looks good.

huss (approved):
All good.

Issue #890, opened 2023-04-19
modal for admin user pages


huss:
### Is your feature request related to a problem? Please describe.

The user page is linked from the admin panel and does not have the look of the newer pages with modals.

### Describe the solution you'd like

The look of these pages will change substantially to align with the new modal card look in OED. The top of the page will have a create button to create a user that is similar to the current create page but as a modal popup. A card for each user will have the email & role listed along with a button to edit the user. The modal that pops up will allow changes to the role and, unlike the current page, should allow for password change. There should also be a button to delete the user similar to the conversion page.

Another change is the "Users" button under "Manage:" on the admin page will be removed. In the dropdown menu in the top, right of the OED page, an additional option will be added that is labeled "Users" (only if an admin). This will take the admin to the user card page. The user card page should have an edit/create button with a cancel/discard as on other admin modal card pages along with similar logic/layout. The delete choice should confirm as on other pages.

A couple of other changes:

- The translation in src/client/app/translations/data.ts for "Manage" is likely to be unused after these changes and can be removed in all languages.
- There should be no default user role and the admin should be required to choose one.
- The current page does not shift to the top as other pages. It is assumed this will be fix automatically when made similar to other pages.

### Additional context

None

huss:
This issue was updated to only include the modal look as the other part was done earlier.

huss:
This was fixed with PR #1306 and #1324 so closing.

PR #1306, opened 2024-07-14
User Changes: Cards and Modals for Users Components. Also added notes to users.

mikedivine:
# Description

This PR changes the look of the user components and aligns it with the new modal card look in OED. User Notes have been added to the DB. Create button to create a user, a card for each user, edit button opens a modal that allows changes to the email, password, role, and notes.  Button to delete the user and confirmation modal.

Option is added that is labeled "Users" (only if an admin). This takes the admin to the user card page.

Partly Addresses #890 - Users Components completed only

## Type of change

## Checklist

## Limitations

The admin component changes will be a separate PR.


src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+										type="email"
+										value={userDetails.email}
+										onChange={e => handleStringChange(e)}
+										invalid={!userDetails.email}
+										required
+									/>
+								</FormGroup>
+							</Col>
+						</Row>
+						{!userDetails.passwordMatch && (
```
huss: I'm unclear on what this is for/doing. Below is FormFeedback which is used on other pages to give form feedback. This seems to try to also do that. When I commented it out it did not seem to make a difference. Should it be removed?

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+							<Col>
+								<FormGroup>
+									<Label for="confirmPassword">Confirm Password</Label>
+									<Input
+										id="confirmPassword"
+										name="confirmPassword"
+										type="password"
+										value={userDetails.confirmPassword}
+										onChange={e => handleStringChange(e)}
+										invalid={userDetails.confirmPassword !== userDetails.password && userDetails.confirmPassword !== ''}
```
huss: I'm wondering if OED should warn and/or require a password to have a minimum length of say 8. I think at least doing a warning might be good. What do others think?

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss, line 1: This is just a general note that many of the comments on this file apply to the edit user modal too.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+			<Modal isOpen={showModal} toggle={handleCloseModal} size="lg">
+				<ModalHeader>
+					<FormattedMessage id="create.user" />
+				</ModalHeader>
+				<ModalBody>
+					<Container>
+						<Row>
+							<Col>
+								<FormGroup>
+									<Label for="email">Email</Label>
```
huss: First, this and all the labels need to be translated.

Second, I've been thinking about the use of email here, even before this PR. OED may have originally had other plans but it does not send email to users. (Sites are not required to set up email service on their servers.) I also tried and had no issues in giving a non-email input and then using it. Given this, I'm thinking it would be better to change the title and any messages that users see to "user name". What do others think? It might be nice to change all the internal names away from email but I think it is okay to leave them for now if people want to skip this step (not requires DB migration too).

src/client/app/components/admin/users/CreateUserModalComponent.tsx
huss: I'm going to suggest that the check at the end for an empty password be removed. The reason is that if no confirmation password is entered but the rest of the form is valid then the save is disabled but no error messages are shown. I do see that it leads to an immediate error after entering the password but I found that okay.

Also, the way the invalid is done here differs from edit. I wonder if it should be standardized.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+	const handleShowModal = () => setShowModal(true);
+	const handleCloseModal = () => {
+		setShowModal(false);
+		resetForm();
+	};
+
+	// Handlers for each type of input change
+	const handleStringChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		setUserDetails(prevDetails => ({
+			...prevDetails, ...userDetails, [e.target.name]: e.target.value
```
huss: I'd like to understand why this needs both preDetails and userDetails? (not done in edit)

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+ * @returns CreateUserModal component
+ */
+export default function CreateUserModal() {
+
+	// user default values
+	const userDefaults = {
+		email: '',
+		password: '',
+		confirmPassword: '',
+		role: '',
```
huss: It is a little curious that the role is a string here but an enum when prepared to submit and in the form menu. I thought a little about changing this. I suspect one might need an invalid role. If it was ever sent to the server (the form should stop that) then it would be rejected by the DB because its enum for roles does not have this. Do people think it sould be changed or just leave with a comment?

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+		note: '',
+		passwordMatch: true
```
huss: I have mixed feeling about having passwordMatch in the state as it can be directly calculated from other state. I might be inclined to create a function instead. However, this might just go away given a comment about its use below.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -0,0 +1,283 @@
...
+							<Col>
+								<FormGroup>
+									<Label for="confirmPassword"><FormattedMessage id="password.confirm" /></Label>
+									<Input
+										id="confirmPassword"
+										name="confirmPassword"
+										type="password"
+										value={userDetails.confirmPassword}
+										onChange={e => handleStringChange(e)}
+										invalid={!userDetails.passwordMatch}
```
huss, line 260: See create on comments about this and using a function instead.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+			<Button color="secondary" onClick={handleShowModal}>
+				<FormattedMessage id="create.user" />
+			</Button>
+			<Modal isOpen={showModal} toggle={handleCloseModal} size="lg">
+				<ModalHeader>
+					<FormattedMessage id="create.user" />
+				</ModalHeader>
+				<ModalBody>
+					<Container>
+						<Row>
```
huss: Row is defined differently in edit which, from a quick check, seems to follow other modals more (at least for unit).

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -0,0 +1,283 @@
...
+	// user edit form state
+	const [userDetails, setUserDetails] = useState({
+		...props.user,
+		...userDefaults
+	});
+
+	// get current logged in user
+	const currentLoggedInUser = useAppSelector(selectCurrentUserProfile) as User;
+
+	// set user details to passed in user
```
huss: This not only does it when first created but for any changes. Is there a case where the props change so the useEffect is needed?

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -0,0 +1,283 @@
...
+		setUserDetails({ ...userDetails, disableDelete: false });
+		if (currentLoggedInUser) {
+			if (props.user.email === currentLoggedInUser.email) {
+				setUserDetails(prevDetails => ({
+					...prevDetails, disableDelete: true }));
+			}
+		}
+	}, [currentLoggedInUser, props.user]);
+
+	useEffect(() => {
```
huss, line 49: The logic is not mirroring create user which does not use a useEffect. I think it would be nice for them to be parallel unless there is a reason not to.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+		const userRole: UserRole = UserRole[userDetails.role as keyof typeof UserRole];
+		const newUser: User = { email: userDetails.email, role: userRole, password: userDetails.password, note: userDetails.note };
+		createUser(newUser)
+			.unwrap()
+			.then(() => {
+				showSuccessNotification(translate('users.successfully.create.user'));
+				handleCloseModal();
+			})
+			.catch(() => {
+				showErrorNotification(translate('users.failed.to.create.user'));
```
huss: Edit user displays the error received. Could this?

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -0,0 +1,283 @@
...
+		// Show the warning modal
+		setShowDeleteConfirmationModal(true);
+	};
+
+	const handleDeleteUser = () => {
+		// Closes the warning modal
+		// Do not call the handler function because we do not want to open the parent modal
+		setShowDeleteConfirmationModal(false);
+
+		// Delete the conversion using the state object, it should only require the source and destination ids set
```
huss: Update comment for user not conversion state.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -0,0 +1,283 @@
...
+										invalid={!userDetails.passwordMatch}
+									/>
+									<FormFeedback><FormattedMessage id="user.password.mismatch" /></FormFeedback>
+								</FormGroup>
+							</Col>
+						</Row>
+						<Row>
+							<Col>
+								<FormGroup>
+									<Label for="note">Note</Label>
```
huss: translate

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,201 @@
...
+			});
```
huss, line 102: The edit has a way to reset passwords and it does it here. I think that is a good idea for security to also do with create.

src/client/app/components/admin/users/UserViewComponent.tsx
```diff
@@ -0,0 +1,56 @@
...
+	return (
+		<div className="card">
+			<div className="identifier-container">
+				{props.user.email}
+			</div>
+			<div className="item-container p-2">
+				<b><FormattedMessage id="role" />: </b>
+				{props.user.role}
+			</div>
+			<div className="edit-btn">
```
huss: I think it would be good to show the note (limited to (I think) 30 characters) as on other cards.

src/client/app/components/admin/users/UsersDetailComponent.tsx
```diff
@@ -0,0 +1,60 @@
...
+import TooltipMarkerComponent from '../../TooltipMarkerComponent';
+import CreateUserModalComponent from './CreateUserModalComponent';
+import UserViewComponent from './UserViewComponent';
+
+/**
+ * Component which shows user details
+ * @returns User Detail element
+ */
+export default function UserDetailComponent() {
+	const { data: users = stableEmptyUsers } = userApi.useGetUsersQuery(undefined);
```
huss: I removed the undefined and it worked fine. Do people think you should have that on a void method?

src/client/app/components/admin/users/UsersDetailComponent.tsx
```diff
@@ -0,0 +1,60 @@
...
+	const [localUsers, setLocalUsers] = React.useState<User[]>([]);
+
+	React.useEffect(() => { setLocalUsers(users); }, [users]);
```
huss: Can you help me understand why there is a copy of the users state in this component?

src/client/app/translations/data.ts
```diff
@@ -979,15 +984,16 @@ const LocaleTranslationData = {
...
-		"users.successfully.delete.user": "Successfully deleted user.\u{26A1}",
-		"users.successfully.edit.users": "Successfully edited users.\u{26A1}",
-		"uses": "uses\u{26A1}",
+		"users.failed.to.create.user": "Échec de la création de l'utilisateur: ",
+		"users.failed.to.delete.user": "Échec de la suppression de l'utilisateur: ",
+		"users.failed.to.edit.user": "Échec de la modification de l'utilisateur: ",
+		"users.successfully.create.user": "L'utilisateur a été créé avec succès: ",
+		"users.successfully.delete.user": "L'utilisateur a été supprimé avec succès: ",
+		"users.successfully.edit.user": "L'utilisateur a été modifié avec succès: ",
+		"uses":"uses\u{26A1}",
```
huss: Please put space back

src/client/app/translations/data.ts
```diff
@@ -1475,13 +1483,16 @@ const LocaleTranslationData = {
...
 		"users": "Usuarios",
-		"users.failed.to.create.user": "No se pudo crear un usuario.",
-		"users.failed.to.delete.user": "No se pudo borrar el usuario.",
-		"users.failed.to.edit.users": "No se pudo editar los usuarios.",
-		"users.successfully.create.user": "Usuario creado con éxito.",
-		"users.successfully.delete.user": "Usuario borrado con éxito.",
-		"users.successfully.edit.users": "Usuarios editados con éxito.",
-		"uses":"usos",
+		"user.delete.confirm": "Borrar usuario: ",
+		"user.confirm.delete": "Supprimer l'utilisateur: ",
```
huss: This seems to be similar to the one above and not used anywhere.

src/client/app/translations/data.ts
```diff
@@ -1157,7 +1164,8 @@ const LocaleTranslationData = {
 		"edit.group": "Editar grupo",
 		"edit.meter": "Details/Editar medidor\u{26A1}",
 		"edit.unit": "Editar unidad",
-		"email": "Correo electrónico",
+		"edit.user": "Editar Usuario",
+		"email": "Correo Electronico",
```
huss: The change to email translation should be put back to how it was since I think that was correct.

src/server/models/User.js
```diff
@@ -12,12 +12,14 @@ class User {
 	 * @param email This user's email
 	 * @param passwordHash The user's passwordHash
 	 * @param role The user's role
+	 * @param note The user note
 	 */
-	constructor(id, email, passwordHash, role) {
+	constructor(id, email, passwordHash, role, note) {
```
huss: There are several test files with "new User" that are missing the note. I think a clean way to deal with this addition is to have ``note = ''`` so it will default to an empty string.

src/server/models/User.js
```diff
@@ -49,40 +51,69 @@ class User {
...
 	 * Returns a promise to update a user's password
-	 * @param email the email of the user whose password is to be updated
+	 * @param id the id of the user whose password is to be updated
 	 * @param passwordHash the new password's hash
 	 * @param conn is the connection to use.
 	 * @returns {Promise.<array.<User>>}
 	 */
-	static async updateUserPassword(email, passwordHash, conn) {
-		return conn.none(sqlFile('user/update_user_password.sql'), { email: email, password_hash: passwordHash });
+	static async updateUserPassword(id, passwordHash, conn) {
```
huss, line 86: src/server/services/user/editUser.js has ``await User.updateUserPassword(email, passwordHash, conn);`` that is still using the user email. It doesn't look hard to use the id from the user DB request above this.

src/server/routes/users.js
```diff
@@ -105,22 +105,25 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
 			role: {
 				type: 'string',
 				enum: Object.values(User.role)
+			},
```
huss, line 108: Should note be a required param?

src/server/routes/users.js
```diff
@@ -135,47 +138,98 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
...
-							type: 'string'
-						},
-						role: {
-							type: 'string',
-							enum: Object.values(User.role)
-						}
-
+			user: {
+				type: 'object',
+				required: ['id', 'email', 'role'],
```
huss: require note?

src/server/routes/users.js
```diff
@@ -135,47 +138,98 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
...
 					}
 				}
 			}
 		}
 	};
+
 	if (!validate(req.body, validParams).valid) {
 		res.status(400).json({ message: 'Invalid params' });
 	} else {
+
```
huss: OED does prefer not to have blank lines in places such as this.

src/server/routes/users.js
```diff
@@ -135,47 +138,98 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
...
-			} else {
-				const roleUpdates = users.map(async user => await User.updateUserRole(user.email, user.role, conn));
-				await Promise.all(roleUpdates);
-				res.sendStatus(200);
+			const { user } = req.body;
+			const userBeforeChanges = await User.getByID(user.id,conn);
+			const numberOfAdmins = (await User.getNumberOfAdmins(conn)).count;
+			
+			// This protects the database so that there will always be at least one admin
+			if (numberOfAdmins < 2 && userBeforeChanges.role === 'admin' && user.role !== 'admin') {
```
huss: Not a big deal but the database calls and check for # admims could be moved inside this if so it only happens if a role change is being made.

src/server/routes/users.js
```diff
@@ -135,47 +138,98 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
...
+				log.error(errorMessage);
+				return res.status(400).json({
+					message: errorMessage,
+				});
+			}
+			
+			// set up Asynchronous database queries
+			const userUpdates = [];
+
+			// not the only admin, so proceed
```
huss: I'm okay with separating out every change but OED normally does an entire update so it is easier and one call. Now, here you cannot do the password if not provided with a new one so that could be a separate call. Up to you if you want to change this.

src/server/services/user/createUser.js
```diff
@@ -37,7 +37,7 @@ const { getConnection } = require('../../db');
 	}
 
 	const passwordHash = bcrypt.hashSync(password, 10);
-	const admin = new User(undefined, email, passwordHash, User.role.ADMIN);
+	const admin = new User(undefined, email, passwordHash, User.role.ADMIN, undefined);
```
huss: In maps, it saves the calibration time with ``modifiedDate: moment().toISOString()``. OED also puts a message about automatically created units, etc. in the note. How about we put: `` `User ${email} created via createUser on ${moment().toISOString()}` `` as the note.

src/server/sql/user/insert_new_user.sql
```diff
@@ -3,6 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 DO $$ BEGIN
```
huss: OED does not usually do a "DO/END" in these files unless it is a function or handles an exception. Is there a special reason it is here?

huss (changes requested):
Thanks to @mikedivine, @Nireves333, @hazelcsumb & @juangutierrez for this significant change. I've made some comments to address. Please let me know if anything is not clear.

I pushed a commit with small changes (see commit message) that were easy for me to do. Let me know if you think any have an issue.

src/server/sql/user/insert_new_user.sql
mikedivine: Not sure, this was here before I touched the code.  I will remove.

--------------------------------------------------------------------
Steven Huss-Lederman, 4 years ago via PR #507   (October 6th, 2020 9:17 PM) 

make createUser more resilient, fix install bug

This tried to make createUser closer to the new createDB.
The SQL avoids an error if the user exists.
createUser now stops on errors since none should happen.
A new, optional parameer to terminateReadline allows code to supply
an exit status value where default is old value of 0.
The install script for createDB retries fixes error where it stopped on
last attempt even if it worked. It also stops if createUser fails.

src/server/services/user/createUser.js
mikedivine: Sounds good, this now has `User ${email} created via createUser on ${moment().toISOString()}` as the note.

src/server/routes/users.js
mikedivine: Got it.

src/server/models/User.js
mikedivine: Sounds good, now note is defaulted to an empty string

src/server/models/User.js
mikedivine: Ah, The editUser.js code was indented incorrectly, so I never updated that code except to fix the indentation issue. I'll look at this script and update it.

src/server/routes/users.js
mikedivine: I didn't make it required as I figured you could have no note, but I'll add it as required since it seems we are going to be using en ampty string as the default now.

src/server/routes/users.js
mikedivine: Sounds good, now required.

src/server/routes/users.js
mikedivine: Sounds good. I have now moved these calls inside the if statement

src/server/routes/users.js
mikedivine: Got it, it is now combined.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Yes, now matches editUser

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Sounds good, will reset passwords on success or failure.

mikedivine: Although, since this is a new user modal, I'll just reset all fields.

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: These are all using translate now.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Yes, now removed

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: removed

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: working on this. I'm planning to add an invalid role.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Now updated to match other modals.
I assume you meant Row xs='1' lg='2'

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: You are correct, it does not, now removed.

src/client/app/components/admin/users/UsersDetailComponent.tsx
mikedivine: Yes, it does seem to work fine without it. Now removed.

src/client/app/translations/data.ts
mikedivine: removed

src/client/app/components/admin/users/UsersDetailComponent.tsx
mikedivine: You are right, should not be. Removed

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Working on changing over to username instead of email.

src/client/app/translations/data.ts
mikedivine: Email is gone now :|

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: That is a good point, I don't believe props should change, so changing to just a standard useState

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: done

src/client/app/translations/data.ts
mikedivine: done.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Completed changing to username and all internal from email to username.

src/client/app/components/admin/users/UserViewComponent.tsx
mikedivine: I have added the card to show the notes up to 50 chars. I tried 30, 50 looked better to me, but I can change to 30 if preferred.

src/server/models/User.js
mikedivine: This is now updated to use ID.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: I agree that this would be good. I would thinking requiring would be good along with implementing more stringent password requirements in the future would be good in my opinion. I will work on requiring the password be at least 8 chars long.

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: Understood, will have CreateUser mirror this useEffect

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Yes, I have decided to remove it.

mikedivine: I am unsure this should go away.  I feel like it has to be checked if either password and confirmPassword changes to ensure they are matching. Otherwise if you only checked after confirmPassword, the user could go back to password and change it and there would be no check. useEffect seems the simplest method to handle this. Let me know if I am overthinking or missing something.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Sounds good, I will remove the check at the end for an empty password.  Also I will make the invalid password check on editUser to mirror this method

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: I felt checking both password fields with useEffect to see if they match seems to make the most sense to me. I am unsure of how to make it as effective another way, but I am open to your thoughts.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: I have implemented this change

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: I have added an invalid role.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,245 @@
...
+				handleCloseModal();
+			})
+			.catch(error => {
+				showErrorNotification(translate('users.failed.to.create.user') + userDetails.username + ' ' + error.data.message);
+				resetPasswordFields();
+			});
+	};
+
+	// allow enter key to submit form
+	const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
```
huss: First, this ability is only on the user pages and not any others. If we are going to keep this then it needs an issue for all pages to do this. Having said that, I have mixed feeling about having enter automatically submit the form. This and other forms have multiple buttons at the bottom so this means OED is picking submit as the default. I could see cancel as an option too. So, I'm leaning toward no default action on enter even though that can be faster. I think it is better to err on not accidentally making changes than make it easier/faster. I have a vague recollection that this may have been discussed when the modals were done and this was the decision.

Second, if this does stay then the console.log needs to be a popup and translated.

src/server/sql/user/create_users_table.sql
```diff
@@ -4,7 +4,8 @@
 
 CREATE TABLE IF NOT EXISTS users(
   id SERIAL PRIMARY KEY NOT NULL,
-  email VARCHAR(254) UNIQUE,
+  username VARCHAR(254) UNIQUE,
   password_hash CHAR(60) NOT NULL,
-  role user_type NOT NULL
+  role user_type NOT NULL,
```
huss, line 9: This change needs a migration. Let me know if you need help with this.

src/client/app/types/items.ts
```diff
@@ -104,3 +105,15 @@ export enum TrueFalseType {
...
 	false = 'no'
 }
+
+// user default values
+export const userDefaults = {
+	username: '',
+	password: '',
+	confirmPassword: '',
+	note: '',
+	role: UserRole['Select Role'],
```
huss: This should be set up so it is translated when shown to the user.

src/client/app/translations/data.ts
```diff
@@ -1390,7 +1399,7 @@ const LocaleTranslationData = {
 		"remove": "Eliminar",
 		"restore": "Restaurar",
 		"return.dashboard": "Regresar al panel principal",
-		"role": "Rol",
+		"role": "Role: ",
```
huss: I don't think this is the Spanish word for this. Maybe add the untranslated symbol.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -0,0 +1,315 @@
...
+							<Col>
+								<FormGroup>
+									<Label for="password">
+										{translate('password')}
+									</Label>
+									<Input
+										id="password"
+										name="password"
+										type="password"
+										placeholder="Only enter password to update password"
```
huss: This string should be translated.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
huss: Given the added useEffect and additional checks I think it is okay to leave as is.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,245 @@
...
+							helpTextId={tooltipStyle.tooltipUsersView}
+						/>
+					</div>
+				</ModalHeader>
+				<ModalBody>
+					<Container onKeyDown={handleKeyDown}>
+						<Row xs='1' lg='2'>
+							<Col>
+								<FormGroup>
+									<Label for="username">
```
huss: I'm not sure why it passed the checks but all of these should be single quotes. I searched and we are not consistent about this but it is the OED way.

src/scripts/installOED.sh
```diff
@@ -145,7 +145,7 @@ else
 	# Create a user
 	set -e
 	if [ "$production" == "no" ] && [ ! "$OED_PRODUCTION" == "yes" ]; then
-		npm run createUser -- test@example.com password
+		npm run createUser -- testUser password
```
huss: First, given you had to edit this in two places it probably should have been a variable from the beginning. Second, while I like having shorter username without the special symbols I have mixed feeling about changing it. A lot of people expect the old email and this may throw them off. I'm inclined to either leave it alone or insert a second admin user with the new name. What do you thinkg?

src/server/routes/login.js
```diff
@@ -15,16 +15,16 @@ const router = express.Router();
...
 	const validParams = {
 		type: 'object',
 		maxProperties: 2,
-		required: ['email', 'password'],
+		required: ['username', 'password'],
 		properties: {
-			email: {
+			username: {
 				type: 'string',
 				minLength: 3,
```
huss, line 29: Do you have any thoughts on the minLength of 3 for both username and password? This does not match the OED requirement for creation of either. I'm inclined to say it is required and put no limit here since it will mismatch if wrong. Thoughts?

Also, should we add maxLength on the password to stop bad actors or is it not really needed?

src/server/routes/obvius.js
```diff
@@ -112,17 +112,17 @@ function obviusLog(req, res, next){
...
  */
 function verifyObviusUser(req, res, next){
-	// First we ensure that the password and email parameters are provided.
+	// First we ensure that the password and username parameters are provided.
 	if (!req.param('password')) {
 		failure(req, res, 'password parameter is required.');
 		return;
-	} else if (!req.param('email')){
-		failure(req, res, 'email parameter is required.');
+	} else if (!req.param('username')){
```
huss: This could break scripts that send Obvius data to OED. I don't recall (and did not quickly find) if Obvius sends the email key. We need to check this out to figure out what is allowed. We also need to decide if we should allow either for backward compatibility.

src/server/test/web/obvius/README.md
```diff
@@ -5,7 +5,7 @@ Here are some sample files for testing Obvius routes.
 For instance to upload the config file (with meter info):
 ```bash
 curl localhost:3000/api/obvius -X POST \
-	-F 'email=test@example.com' \
+	-F 'username=testUser' \
```
huss: See another comment on this change.

src/server/test/web/csvPipelineTest.js
```diff
@@ -21,8 +21,8 @@ const UPLOAD_METERS_ROUTE = '/api/csv/meters';
 const UPLOAD_READINGS_ROUTE = '/api/csv/readings';
 
 // The default fields and routes for the requests
-const CHAI_READINGS_REQUEST = `chai.request(app).post('${UPLOAD_READINGS_ROUTE}').field('email', '${testUser.email}').field('password', '${testUser.password}')`;
-const CHAI_METERS_REQUEST = `chai.request(app).post('${UPLOAD_METERS_ROUTE}').field('email', '${testUser.email}').field('password', '${testUser.password}')`;
+const CHAI_READINGS_REQUEST = `chai.request(app).post('${UPLOAD_READINGS_ROUTE}').field('username', '${testUser.username}').field('password', '${testUser.password}')`;
```
huss, line 24: All these changes to username for test indicate a breaking change. We need to discuss if this is okay. Right now I'm included to either put it back or accept the old and new way. Lets discuss before making any changes.

src/client/app/components/admin/users/UserViewComponent.tsx
huss: I think it should be 30 so consistent across pages.

src/client/app/components/admin/users/EditUserModalComponent.tsx
huss: Okay per other comment.

src/client/app/components/admin/users/UserViewComponent.tsx
```diff
@@ -0,0 +1,64 @@
...
+				<b>
+					{translate('role')}
+				</b>
+				{props.user.role}
+			</div>
+			<div className="item-container p-2">
+				<b>
+					{translate('note')}
+				</b>
+				{props.user.note ? (props.user.note.length > 50 ? props.user.note.substring(0, 50) + '...' : props.user.note) : ''}
```
huss: Frist, other pages just do the slice so maybe do that here too (if don't add the extra ...). Second, if going to add ... then it needs to be on all pages and we need to verify that 30 characters still stays on one line on all cards.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,245 @@
...
+		setUserDetails(prevDetails => ({
+			...prevDetails,
+			passwordMatch: (userDetails.password === userDetails.confirmPassword),
+			passwordLength: userDetails.password.length > 7
+		}));
+	}, [userDetails.password, userDetails.confirmPassword]);
+
+	// check if form is valid
+	const isFormValid = () => {
+		return !!userDetails.username &&
```
huss: Should this use the state variables rather than check again? Similar comment for edit.

src/server/sql/user/create_users_table.sql
```diff
@@ -4,7 +4,8 @@
 
 CREATE TABLE IF NOT EXISTS users(
   id SERIAL PRIMARY KEY NOT NULL,
-  email VARCHAR(254) UNIQUE,
+  username VARCHAR(254) UNIQUE,
```
huss, line 7: This change needs a migration. Let me know if you need help with this.

huss (changes requested):
Thanks once again to @mikedivine for addressing all the previous comments. I esp. appreciate the large number of changes due to email -> username. I think this is very close where I have some comments to think about/get input.

src/client/app/components/admin/users/UserViewComponent.tsx
mikedivine: Got it, changing to 30.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
Nireves333: Will do.

src/server/routes/login.js
Nireves333: Having a minimum and maximum length could be a good idea for security reasons. 
I think matching them here and in the creation modal would be a good idea.

src/client/app/components/admin/users/UserViewComponent.tsx
mikedivine: Got it, I'll use slice and not include the '...'

src/client/app/components/admin/users/CreateUserModalComponent.tsx
Nireves333: I'll take care of it

src/client/app/types/items.ts
Nireves333: What do you mean by this?

Nireves333: We made this a default value enum, how would we go about doing this?

src/client/app/translations/data.ts
Nireves333: I'll fix this

src/client/app/components/admin/users/CreateUserModalComponent.tsx
Nireves333: Completed. Removed the enter key functionality.

src/client/app/translations/data.ts
Nireves333: Completed

src/client/app/components/admin/users/CreateUserModalComponent.tsx
Nireves333: Completed

src/client/app/components/admin/users/EditUserModalComponent.tsx
mikedivine: Will do, I will add in translations/data.ts

src/server/sql/user/create_users_table.sql
mikedivine: Ok I added the changes, I beleive just renaming the email column to username should be all that is needed.  Let me know if I missed something.

src/scripts/installOED.sh
mikedivine: Up to you, I just thought since we were moving away from email, it would be better to use a username. I am good with whatever you prefer. I'll change it back and leave it how it was for now. Let me know and I can have it add both if you'd prefer.

src/server/routes/login.js
mikedivine: I will remove the minLength as requested as your explanation makes sense to me.  I will add 128 as max password, to prevent bad actors mucking with it.

mikedivine: let me know if you think 128 is not large enough.

src/server/test/web/obvius/README.md
mikedivine: Changed back for now

src/server/test/web/csvPipelineTest.js
mikedivine: Ok, sounds good, we can discuss at the meeting.

src/server/routes/obvius.js
mikedivine: Yeah, it most likely does as I found the function obviusUsernameAndPasswordAuthMiddleware in authenticator.js and it has this:
const user = await verifyCredentials(req.body.username, req.body.password, true);

We can discuss or let me know if you prefer to allow for backward compatibility.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Got it, now using the state variables that check on useEffect

src/client/app/types/items.ts
mikedivine: I think I figured out a way and I'll implement it.

src/server/routes/obvius.js
mikedivine: altered to be backwards compatible, left note.

src/server/test/web/csvPipelineTest.js
mikedivine: I updated the obvius meter login with backwards compatibility with email and left a note. Will need documentation changes for curl commands with CSV uploads to use username instead of email moving forward.

src/scripts/installOED.sh
```diff
@@ -160,7 +160,7 @@ else
 			fi
 		else
 			# There was no createdb error so assume database ready for use so stop process
-			printf "%s\n" "User creation had no errors so default user 'test@example.com' with password 'password' should exist"
+			printf "%s\n" "User creation had no errors so default user 'testUser' with password 'password' should exist"
```
huss: This outputs testUser and not the actual name. I think it would be good to create a variable at the top of the script and use it here and in the npm command above.

src/scripts/installOED.sh
huss: Lets put in user called test with the same password to start to transition. Then leave a TODO to remove the old one in the future.

src/server/routes/login.js
huss: I think we should put back minLength but set it to 8 on the password. Username can be any min length.

src/server/test/web/obvius/README.md
```diff
@@ -5,7 +5,7 @@ Here are some sample files for testing Obvius routes.
 For instance to upload the config file (with meter info):
 ```bash
 curl localhost:3000/api/obvius -X POST \
-	-F 'email=test@example.com' \
+	-F 'username=test@example.com' \
```
huss: Please change this to <obvius user> instead of test@example.com (both places). This indicates it needs to be replaced and can be any user with obvius role (including admin).

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,234 @@
...
+			...prevDetails,
+			passwordMatch: (userDetails.password === userDetails.confirmPassword),
+			passwordLength: userDetails.password.length > 7
+		}));
+	}, [userDetails.password, userDetails.confirmPassword]);
+
+	// check if form is valid
+	const isFormValid = () => {
+		return userDetails.username &&
+        userDetails.passwordMatch &&
```
huss: The spacing should be tabs in a couple of places.

src/client/app/components/admin/users/UsersDetailComponent.tsx
```diff
@@ -0,0 +1,60 @@
...
+				<div className='edit-btn'>
+					<CreateUserModalComponent />
+				</div>
+				<div className='card-container'>
+					<Row className='justify-content-center'>
+						{// display users and sort by username alphabetically
+							[...users]
+								.sort((a, b) => a.username.localeCompare(b.username))
+								.map(user => (
+									<Col key={user.username} className="d-flex justify-content-center mb-3">
```
huss: Could you help me understand what the key is for?

src/scripts/installOED.sh
mikedivine: Will do.

src/scripts/installOED.sh
mikedivine: Will do.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -0,0 +1,234 @@
...
+									</Label>
+									<Input
+										id='role'
+										name='role'
+										type='select'
+										value={userDetails.role}
+										onChange={handleRoleChange}
+										invalid={userDetails.role === UserRole['user.role.select']}
+									>
+										{Object.entries(UserRole).map(([role, value]) => (
```
huss: This is done differently than the meter modal where the need to select is gone once you make a selection (cannot choose it again). I think that is a little nicer. Please either fix of put in a TODO.

src/server/routes/obvius.js
```diff
@@ -112,17 +112,23 @@ function obviusLog(req, res, next){
  * Verifies an Obvius request via username and password.
  */
 function verifyObviusUser(req, res, next){
-	// First we ensure that the password and email parameters are provided.
-	if (!req.param('password')) {
+	// First we ensure that the password and username parameters are provided.
+	const password = req.param('password');
+	// allowing for backwards compatibility if previous eGuage meters are using the 'email' parameter instead of
```
huss: First, indenting is off. Second, could you please add a TODO marker at start of line so it will stand out more and work with TODO systems.

src/server/services/csvPipeline/validateCsvUploadParams.js
```diff
@@ -92,12 +92,16 @@ const DEFAULTS = {
...
 // These are the common upload params shared between meters and readings upload.
-// Note: Even though, the 'email' and 'password' properties are not used to validate the upload, 
+// Note: Even though, the 'username' and 'password' properties are not used to validate the upload, 
 // they may be attached to the request body when the CSV Pipeline handles user authentication 
 // (i.e. when the user performs a curl request to the pipeline). Thus, we list these properties 
 // here so that they do not falsely trigger the 'additionalProperties' User Error.
 const COMMON_PROPERTIES = {
 	meterName: new StringParam('meterName', undefined, undefined),
+	username: new StringParam('username', undefined, undefined),
+	// allowing for backwards compatibility to allow for curl users to use the 'email' parameter instead of
```
huss: First, add TODO per other comment. Also, what happens if someone passes both the username and email? For example, which one is used?

src/server/routes/login.js
mikedivine: OK I will add back minLength to 8 on password and 3 on username.

src/server/routes/obvius.js
```diff
@@ -112,17 +112,23 @@ function obviusLog(req, res, next){
...
+		// the 'username' parameter to login. Developers need to decide in the future if we should deprecate email
+		// or continue to allow this backwards compatibility
+	const username = req.param('username') || req.param('email');
```
huss, line 120: As in another file, does this fail if they supply both? We need to have a consistent, determined result that we can doucment.

src/server/routes/obvius.js
```diff
@@ -112,17 +112,23 @@ function obviusLog(req, res, next){
...
+	} else if (!username) {
+		failure(req, res, 'username parameter is required.');
 		return;
 	} else { // Authenticate Obvius user.
-		req.body.email = req.param('email');
-		req.body.password = req.param('password');
-		obviusEmailAndPasswordAuthMiddleware('Obvius pipeline')(req, res, next);
+		req.body.username = username;
+		req.body.password = password;
+		obviusUsernameAndPasswordAuthMiddleware('Obvius pipeline')(req, res, next);
```
huss, line 131: I'm not really knowledgeable here but I want to verify you did not break this code. The original set req.body.password and that is what src/server/routes/authenticator.js uses. Is the fact it is in req.param('password') the same? We have to be very careful with any authentication changes.

src/server/models/eGauge/eGaugeRequestor.js
```diff
@@ -19,7 +19,11 @@ class EgaugeRequestor {
 		// Information expected by the eGauge API and needed by OED.
 		const url = new URL(`https://${meter.url}`);
 		this.apiUrl = `${url.origin}/api`; // can we add an ending slash??
-		this.username = url.searchParams.get('username');
+		// allowing for backwards compatibility if previous eGuage meters are using the 'email' parameter instead of
```
huss: A couple of thoughts. First, TODO per other comments. Second, issue if provide both to settle.

src/server/test/web/csvPipelineTest.js
```diff
@@ -472,6 +476,13 @@ const testCases = {
...
+	},
+	pipe1: {
```
huss: A test key should never be reused. It is a unique identifier. This should be a new test number - maybe pipe130? The same applies for the other test changed.

src/server/routes/csv.js
```diff
@@ -67,9 +67,13 @@ router.use(function (req, res, next) {
 					(await isTokenAuthorized(token, csvRole)) ? cb(null, true) : cb(new Error('Invalid token (either unauthorized or logged out'));
 				} else {
 					// If no token is found, then the request is mostly like a curl request. We require an
-					// email and password to be supplied for curl requests.
-					const { email, password } = request.body;
-					const verifiedUser = await verifyCredentials(email, password, true);
+					// username and password to be supplied for curl requests.
+					const { username, email, password } = request.body;
+					// allowing for backwards compatibility if previous eGuage meters are using the 'email' parameter instead of
```
huss: Same thoughts as other places it was changed.

huss (changes requested):
Thanks to the team for the updates. This is looking good overall and works as expected. I've made a few comments to address including on some previous ones. I'm hoping the changes are not very much.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: OK, this now works in a similar manner.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
mikedivine: Got it, will resolve this

src/client/app/components/admin/users/UsersDetailComponent.tsx
mikedivine: React requires key on root element inside the array, I'll remove from UserViewComponent since it isn't needed there.

src/server/routes/obvius.js
mikedivine: Yes, will do.

src/server/test/web/csvPipelineTest.js
mikedivine: Got it. I will change these.

src/server/routes/obvius.js
mikedivine: This takes the username param if it exists, otherwise it takes the email param. It only fails if neither is supplied. I tested with both in using curl and it works as intended.

src/server/services/csvPipeline/validateCsvUploadParams.js
mikedivine: This takes the username param if it exists, otherwise it takes the email param. It only fails if neither is supplied. I tested with both in using curl and it works as intended.

src/server/test/web/obvius/README.md
mikedivine: OK, I changed to test on both spots.

src/server/models/eGauge/eGaugeRequestor.js
mikedivine: works the same as the others, takes username if it exists, otherwise takes email. If neither, then fails

src/server/routes/csv.js
mikedivine: Same as others, added TODO, username parameter is taken if it is given, otherise email parameter is taken as username. If both are supplied, username takes priority. If niether is given, then fails.

src/server/routes/obvius.js
mikedivine: I do not believe so. This: req.body.username = req.param('username'); I believe is equivalent to this: const username = req.param('username') || req.param('email'); req.body.username = username;

mikedivine: I see some tests in the readme.md but not sure how to run them to test if everything is working properly

huss:
I just pushed two items via commits:

1. Added code to more consistently enforce min/max on user info.
2. Merged development to remove conflict.

I did this because I accidentally merged the new admin settings page so I wanted to quickly get this merged so people have access to the user page. If anyone sees any issue then let me know.

huss:
The new mins on user name means that if someone had one of less than 3 it would not longer work. Since they used to be emails this seems safe.

Though unlikely, a now invalid password could be present. If this is the case then it would need to be reset.

huss (approved):
Thanks to @mikedivine for these last changes. I appreciate your willingness to consider changes and for making the all even when that mean new items. I've tested the code and reviewed. It all seems fine.

PR #1318, opened 2024-07-24
Issue 1182 date range & 1165 dates used by calendar

### LINKED ISSUE #1165: dates used by calendar

The last day is included on the line, map & radar graphic but not map & bar (maybe others). For example, select a single day with the calendar on the line graphic.  You will see that one day graphed (assuming it has data).  Same behavior on map &  radar. If you go to bar or map then you see nothing. In a more general sense, the last day is not included in bar or map but is on line and radar. This seems inconsistent and confusing to the user.

It seems intuitive to the user that the last listed day in the date range is shown. I think bar and map should do this.

None good if leave the current behavior.

### Additional context

None.

### LINKED ISSUE #1182: date range does not include last day
### Describe the bug

When you select a date range, the last day is not included in the range. I’m not sure this is intuitive for the user. For example, you select one day (6/1/21-6/1/21) and you get nothing or 3 days (6/1/21-6/3/21) gives two days. I think we should reconsider this behavior.

### Expected behavior

I think OED should include all the dates in the date range for all graphics. Note OED considers midnight of a day (00:00:00) to be the start and this may relate to this issue.

### Screenshots

This show the current way for 3 days showing 2 on a day bar graphic:

Screenshot 2024-04-03 102005

### Additional context

None

williamw04:
# Description

In graphs, such as line, bar and radar, final dates within the date range are not displayed. This is issue since intuitively users would expect data from the entire range of dates including the last date. This was unintentionally caused by OED's decision to only display full days. This behavior stems from the decision that displaying partial dates was undesirable for graphs such as bar and 3D. Given an end date, OED would cut the rest of the date i.e. given a date at 11:00:00 OED would truncate this date to 00:00:00 of the current date. 

To address this issue, the date is first taken from the redux state then manipulated pushing the end date forward by a millisecond. This would roll the end date to the beginning of the next day providing the correct range when fetching data. However, we want to display the proper date range in the OED web application so that action should be reversed. 

This was implemented in the dateRangeToTimeInterval method in which the end date was pushed forward and timeIntervalToDateRange when that action was reversed. 

In the DateRangeComponent, when a date interval was selected, handleChange would call dispatch to update the TimeInterval using the dateRangeToTimeInterval method. The dateRangeToTimeInterval method reformats the date range to the international standard for time before being converted to a moment object. The dispatch action changes the state, and redux & react would realize that the changed is directly related to timeinterval which causes the site to be repainted. This was result in value being updated, in which timeIntervalToDateRange(queryTimeInterval) would be called and would reverse the added millisecond.

Fixes #1182 & Fixes #1165

## Type of change

## Checklist

- [ X ] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [ X ] I have removed text in ( ) from the issue request
- [ X ] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

The case when the user selects a partial date by using the slider the last date might not be displayed. This is caused when there are no points of data available before the end time on the end date. This is due to the unit/frequency of the data points returned by the server. I.e when you move a slider to 10:51:00 and the server is returns data at noon 12:00:00 of every date, the data point of 12:00:00 of the end date is out of range and not graphed. As a result the last point it graphs it 12:00:00 of the previous date. 

Another issue found, is bar graph is unable to display data for a singular date.


src/client/app/components/BarChartComponent.tsx
```diff
@@ -39,6 +39,8 @@ export default function BarChartComponent() {
 		})
 	});
```
huss: I think the blank line and console log should be removed.

src/client/app/redux/selectors/chartQuerySelectors.ts
```diff
@@ -45,6 +45,7 @@ export const selectCommonQueryArgs = createSelector(
 	selectQueryTimeInterval,
 	selectSelectedUnit,
 	(selectedMeters, selectedGroups, queryTimeInterval, selectedUnit) => {
+
```
huss: Please remove the new blank line.

src/client/app/utils/dateRangeCompatibility.ts
```diff
@@ -46,7 +52,14 @@ export function dateRangeToTimeInterval(dateRange: Value): TimeInterval {
 	}
 	if (start && end) {
 		start = moment(toUTC(start));
-		end = moment(toUTC(end));
+		/* Adds a millisecond to the end time.
+		For the case in which the end time is the last moment of the day (date = date.endOf('day)),
+		adding a millisecond pushes the end time to the very beginning of the next day.
+		i.e the end date is 7.22.2020 at 23:59:59. Due to how OED works, the end date will be truncated to
+		7.20.2020 at 00:00:00. This is due to OED decision to graph only entire days causing the lost of the last date.
```
huss: I think 7.20 -> 7.22. Also, if you want to use the standard representation that the code uses then it is 2020-07-22 23:59:59 (as example).

huss:
I'm leaving a note that when you use the slider to select the dates and the graphic shows whole days then the date range picker may be one day earlier at start and/or one day later at end. This is because any partial days are not used (except 3D) but they are shown on the date range picker. This is not a bug but the way OED works. The documentation should be updated for this.

huss (changes requested):
Thanks to @williamw04 for finding a clean solution to issue #1182. I have a few minor comments to address. Also, I tried using one day per issue #1165 and found that bar still does not show any data. I suspect the reason is separate from the change so far. If you agree either fix the issue or remove that issue from this PR.

williamw04:
Implemented changes and addressed issue #1165

huss (approved):

PR #1323, opened 2024-07-30
Pr1319SelectorDraft

ChrisMart21:
- selectHelpUrl

# Description

same as #1319, but with change suggestions . Aims to reuse existing selector selectHelpUrl.


huss (approved):
This works as expected and is a better design than PR #1319 (which is now closed). Given the new site is now live, I moved this from draft to open and am approving. If @ChrisMart21 agrees then please merge. However, see my comment on announcing to developers due to needing to change help location.

huss:
The help location is new in the next release so there is no migration issue on the value. However, current developers will have the old location in the admin panel. They need to update to remove the help/ at the end so it is https://openenergydashboard.github.io/. Any newly created version and DB should be fine. An announcement about this needs to be made (I can do it if @ChrisMart21 does not but let me know).

ChrisMart21:
@huss Please make the announcement for this merge. Thanks

PR #1325, opened 2024-08-02
Issue#1163 resolved, Issue#1292 resolved, and Issue#1293 partially resolved

### LINKED ISSUE #1163: values wrong on map graphic based on bar choice
### Describe the bug

The dates shown on the map graphic seem to show the correct start date but the end date is based on the bar interval and not the map interval. Furthermore, it seems to normalize the value by the number of bar days not map interval days.

### To Reproduce

1. Select the Sin Amp 1 kWh meter.
2. On map note the values shown or day, week, 4 week.
3. On bar interval select day from what should have defaulted to 4 week.
4. On map the dates are only one day and the values are off.

### Expected behavior

Bar and map intervals should be independent and work properly.

### LINKED ISSUE #1292: dropdown for all interval choices

PR #1249 modified the bar interval to be a dropdown menu. Other pages do not do this.

The compare and map graphic pages have similar choices (day, week, 4 week but not custom). They should become dropdown menus similar to bar.

The compare graphic does not have a label/title as does the other ones so this should be added as internationalized "Compare Interval".

Leave UI inconsistent.

### Additional context

Another PR discusses adding custom interval option and this may be easy to do at this time (if someone wants).

### LINKED ISSUE #1293: custom interval for compare and map

Bar allows for a custom interval. Compare and map do not. For consistency they should all allow custom.

Add the custom option in the same way as bar.

Note to simplify the code and to make all graphics consistent, the interval on any page will change the interval on all pages. This is how, for example, the date ranges work.

Don't allow.

### Additional context

Issue #1292 is changing the interval choice to a dropdown and the custom should be done the same as the bar dropdown. It may be easy to do both at the same time.

This issue supersedes #1164.

It probably relates to the issue #1163.

danielshid:
# Description

I fixed the issue with map displaying the wrong data based on bar by implementing it so bar and map are dependent on the same range. I also implemented dropdown menus for bar, map, and compare intervals, as well as the custom option for map. Map is complete; however, custom cannot currently be implemented for compare.

Fixes #1163, Fixes #1292, and Partly Addresses #1293

## Type of change

## Checklist

## Limitations

N/A


huss:
This PR has merge conflicts. It would help me to have them resolved so I can test the final result and run the tests. If you need help then let me know. Thanks.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,288 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as moment from 'moment';
```
huss, line 5: I'm curious about the choice to remove the three separate control components for bar, compare & map to create this one combined one. It is nice to separate out components rather than have conditionals in one when that is possible/appropriate. What I'm thinking about is if it makes sense to have the original three with a new component for the interval/width that would be included into the other three. That would centralize this one component but leave the other three. See below on ideas to make all three more similar. What do you think?

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,288 @@
...
+	return (
+		<div>
+			{chartType === ChartTypes.bar && (
+				<div className='checkbox'>
+					<input type='checkbox' style={{ marginRight: '10px' }} onChange={handleChangeBarStacking} checked={barStacking} id='barStacking' />
+					<label htmlFor='barStacking'>{translate('bar.stacking')}</label>
+					<TooltipMarkerComponent page='home' helpTextId='help.home.bar.stacking.tip' />
+				</div>
+			)}
+			{(chartType === ChartTypes.bar || chartType === ChartTypes.map) && (
```
huss: Would it make sense to fuse this with compare so there is only one? It might involve these changes:

- Use input in both cases so the same.
- Create an enum to specify the allowed options in the input. It could be set at the start based on the graph type where custom is not included if compare. Loop over enum as is done in other places to create the input values. This would also make it very easy to switch once the TODO on custom for compare is done.
- Use a standardized onChange function. I don't know if they could all use the same state or if an if is needed to separate out compare as is now done. I have not looked at the details to see why it is different. If only for the custom value and it cannot be chosen then it may be okay to combine into one.

What do you think?

huss:
I think this only partly addresses issue #1293 since it does not do compare. I think the description should be modified to reflect this so it will not automatically close when this PR is merged. Also, adding a comment to the issue that map is complete would be good.

huss (changes requested):
Thanks to @danielshid for addressing all these issues. Testing found it works as expected. I made a couple of comment involving code layout to think about. Let me know if you have thoughts or want to discuss. I also made a comment about one linked issue.

src/client/app/components/MapControlsComponent.tsx
```diff
@@ -1,53 +1,19 @@
...
-				<ButtonGroup style={zIndexFix}>
-					<Button outline={barDurationDays !== 1} onClick={() => handleDurationChange(1)}> {translate('day')} </Button>
-					<Button outline={barDurationDays !== 7} onClick={() => handleDurationChange(7)}> {translate('week')} </Button>
-					<Button outline={barDurationDays !== 28} onClick={() => handleDurationChange(28)}> {translate('4.weeks')} </Button>
-				</ButtonGroup>
-				<TooltipMarkerComponent page='home' helpTextId='help.home.map.interval.tip' />
-			</div>
-		</div>
-	);
+    return (
```
huss: These should be tabs for indenting.

src/client/app/components/CompareControlsComponent.tsx
```diff
@@ -1,89 +1,59 @@
...
-	const comparePeriod = useAppSelector(selectComparePeriod);
-	const compareSortingOrder = useAppSelector(selectSortingOrder);
-	const [compareSortingDropdownOpen, setCompareSortingDropdownOpen] = React.useState<boolean>(false);
-	const handleCompareButton = (comparePeriod: ComparePeriod) => {
-		dispatch(graphSlice.actions.updateComparePeriod({ comparePeriod, currentTime: moment() }));
-	};
-	const handleSortingButton = (sortingOrder: SortingOrder) => {
-		dispatch(graphSlice.actions.changeCompareSortingOrder(sortingOrder));
-	};
+    const dispatch = useAppDispatch();
```
huss: tabs for indenting

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,169 @@
...
+					} />
+				</p>
+				<Input
+					id='durationDays'
+					name='durationDays'
+					type='select'
+					value={chartType === ChartTypes.compare ? comparePeriod?.toString() : days}
+					onChange={e => chartType === ChartTypes.compare ? handleComparePeriodChange(e.target.value) : handleDaysChange(e.target.value)}
+				>
+					{chartType !== ChartTypes.compare && (
```
huss: I think this would be cleaner if the same three options were added and then custom is added if it is not compare. Let me know if you want help with the way to do this.

huss (changes requested):
Thanks to @danielshid for the updated code. Review found it works as desired. I've made a few small comments to address but this is very close to done.

Also, I merged development to test. You will need to update your DB per the msg to Discord developers.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,169 @@
...
+				</p>
+				<Input
+					id='durationDays'
+					name='durationDays'
+					type='select'
+					value={chartType === ChartTypes.compare ? comparePeriod?.toString() : days}
+					onChange={e => handleDaysChange(e.target.value)}
+				>
+					{options.map(option => (
+						<option value={option.value} key={option.label}>
```
huss: For four weeks, the old version had "4 Weeks" on the menu (see https://openenergydashboard.org/helpV1_0_0/compareGraphic/). I think that is better and matches the other ones. I'm not sure the best way but maybe treat the items as keys and translate via data.ts. Also, these options for compare are not translating so that would address that too.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,169 @@
...
+	};
+
+	const updateDurationChange = (value: number) => {
+		// Update if okay value. May not be okay if this came from user entry in custom form.
+		if (daysValid(value)) {
+			dispatch(graphSlice.actions.updateDuration(moment.duration(value, 'days')));
+		}
+	};
+
+	// Define the initial set of options for the dropdown when the chart type is not 'compare'.
```
huss: This is definately better than the previous code. However, the custom option is showing on the compare graphic (but it does nothing). What I'm thinking is to add the custom option here and then, below, where you reset the menu for compare it would not be added. I'm open to other ideas but that may get the two set of choices without much effort.

Now, for the future, I'm noting that when we add compare to the custom choices, it would be best to fuse the options so there is only one enum that serves them all.

huss (changes requested):
Thanks to @danielshid for addressing the previous comments. I've made a couple more that should be fairly easy to address. Let me know if you have questions or cannot address them.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,181 @@
...
+	// Defines the set of options for the dropdown
+	const options = chartType === ChartTypes.compare
+		? Object.entries(ComparePeriod).map(([key, value]) => (
+			<option value={value} key={key}>
+				{translate(comparePeriodTranslations[key as keyof typeof comparePeriodTranslations])}
+			</option>
+		))
+		: (
+			<>
+				<option value='1'>{translate('day')}</option>
```
huss: Why not use comparePeriodTranslations? Another comment might naturally lead to this.

src/client/app/translations/data.ts
```diff
@@ -29,7 +29,7 @@ const LocaleTranslationData = {
 		"as.meter.unit": "as meter unit",
 		"as.meter.defaultgraphicunit": "as meter default graphic unit",
 		"bar": "Bar",
-		"bar.days.enter": "Enter in days and then hit enter",
+		"days.enter": "Enter in days and then hit enter",
```
huss: Old name in other languages.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,181 @@
...
+			<div style={divTopBottomPadding}>
+				<p style={labelStyle}>
+					{translate(
+						chartType === ChartTypes.bar ? 'bar.interval' :
+							chartType === ChartTypes.map ? 'map.interval' :
+								'compare.period'
+					)}:
+					<TooltipMarkerComponent page='home' helpTextId={
+						chartType === ChartTypes.bar ? 'help.home.bar.days.tip' :
+							chartType === ChartTypes.map ? 'help.home.map.days.tip' :
```
huss: The map help is not linking to a page. I think it is because the help system expects help.home.map.interval.tip.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,181 @@
...
+	};
+
+	const comparePeriodTranslations: Record<keyof typeof ComparePeriod, string> = {
+		Day: 'day',
+		Week: 'week',
+		FourWeeks: '4.weeks'
+	};
+
+	// Defines the set of options for the dropdown
+	const options = chartType === ChartTypes.compare
```
huss: I was thinking that the menu would be set up with the three standard option and the custom added after the loop when appropriate. Here is code from an upcoming change to units that shows the basic idea:
```js
{Object.entries(LineGraphRates).map(
	([rateKey, rateValue]) => (
		<option value={rateValue * 3600} key={rateKey}>
			{translate(rateKey)}
		</option>
	)
)}
<option value={CUSTOM_INPUT}>
	{translate('custom.value')}
</option>
```
What do you think?

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,181 @@
...
+	const [daysCustom, setDaysCustom] = React.useState<number>(duration.asDays());
+	// True if custom duration input for bar or map is active.
+	const [showCustomDuration, setShowCustomDuration] = React.useState<boolean>(false);
+
+	// Keeps react-level state, and redux state in sync.
+	// Two different layers in state may differ especially when externally updated (chart link, history buttons.)
+	React.useEffect(() => {
+		// Assume value is valid since it is coming from state.
+		// Do not allow bad values in state.
+		const isCustom = !['1', '7', '28'].includes(duration.asDays().toString());
```
huss: I think the old bar code and the upcoming unit code has custom remain even if the value entered is a standard value. This code flips to the standard menu choice. My thinking is that if you are doing custom input then you want that and can repeatedly change the custom value (to include standard values) without selecting custom again. If anyone feels otherwise then speak up. If not, then this should change to be consistent. DM me to see the code used in unit to use similar logic.

huss (changes requested):
Thanks to @danielshid for the updated code. While they addressed the comments at the time, I've made a few more based on other code coming and some things I noticed. I appreciate you looking into them.

src/client/app/components/IntervalControlsComponent.tsx
```diff
@@ -0,0 +1,181 @@
...
+	const [showCustomDuration, setShowCustomDuration] = React.useState<boolean>(false);
+	// Define a flag to track if custom input is actively being used
+	const [isCustomInput, setIsCustomInput] = React.useState<boolean>(false);
+
+	// Keeps react-level state, and redux state in sync.
+	// Two different layers in state may differ especially when externally updated (chart link, history buttons.)
+	React.useEffect(() => {
+		// If user is in custom input mode, don't reset to standard options
+		if (!isCustomInput) {
+			const isCustom = !['1', '7', '28'].includes(duration.asDays().toString());
```
huss: To fully integrate looping over the enum for duration, what if this line is replaced with:
```js
			const durationValues = Object.values(ComparePeriod) as string[];
			const isCustom = !(durationValues.includes(duration.asDays().toString()));
```
What do you think?

huss (commented):
Thanks to @danielshid for addressing the previous comments. Review and testing found it works as desired. I have one some comment on a possible change to the code that I wanted your input on. If you like it then you are welcome to make the change.

src/client/app/components/IntervalControlsComponent.tsx
danielshid: That works.

huss (approved):

Issue #1161, opened 2024-03-21
Remove most plotly icons from radar & map graphic


huss:
### Is your feature request related to a problem? Please describe.

PR #1113 removed most Plotly icons. The radar graphic still has three of the icons. On map they appear as the mouse is moved over the graphic. They should be made consistent with the other graphics.

### Describe the solution you'd like

See above.

### Describe alternatives you've considered

On Radar, one icon appears to do nothing and one lists the Plotly version so those don't hurt much. The camera is actually desired. On map, it is as before with all the icons. Thus, OED can leave as is until there is a fix.

### Additional context

Issue #1160 proposes to add the graphic capture icon to all graphics and this should be consistent with that one.

williamw04:
I'll attempt this problem

huss:
Radar is done. Map will be made consistent with others once the map PR is merged. Then OED can see if the icons only appear when hover over the graphic. If not, I don't think that is a big deal so I'm closing this.

Issue #1160, opened 2024-03-21
Plotly capture of graphic image


huss:
### Is your feature request related to a problem? Please describe.

Before PR #1113, OED included Plotly action icons at the top, right of the graphics. Many were problematic and users were advised not to use them except to capture the graphic as an image. If possible, it would be nice to add only the Plotly icon to capture the graphic image and not any others. If the one with the Plotly version and link is left then that would be okay but not required.

### Describe the solution you'd like

Add icon as described.

### Describe alternatives you've considered

Users can often get screen capture but that is often a less scalable image and requires knowledge of how to do it.

### Additional context

The previous controls can get seen at https://openenergydashboard.github.io/help/v1.0.0/lineGraphic.html where the camera should be the icon to get a graphic image. Issue #1161 addresses differences in the icons for radar.

williamw04:
I'd like to attempt this

huss:
This is solved by the linked PR so it is being closed.

PR #1314, opened 2024-07-18
Maps Update: Cards and Modals for Maps Components

juanjoseguva:
# Description

Updating Maps page to display as cards with modals, matching the other pages. 

UPDATE: 
 - This PR has been updated to also migrate over to RTKQuery for data fetching, and slices, for state management.
 - Fixes: #1197, #1145, #667 ,#1192, #1146 immutability check and serializable checks are both enabled
 
 2509 Update:
- Fixes: #1147

## Type of change

- [ ] Note merging this changes the database configuration.
- [x] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

Saving functionality should be moved into the modals, and the save changes button under the cards should be deleted.

UPDATE: The migration maintains much of the prior functionality but, for simplicity, some of the maps logging was temporarily omitted. Logging as a whole should be carefully planned and implemented in a future PR for the omitted logs from maps, but fort he application as a whole.

PR #1329, opened 2024-08-11
Readded buttons and removed unnecessary buttons

williamw04:
# Description

There was a desire to use plotly buttons since there were certain desired features such as saving a png of the chart. However, enabling buttons displayed too many buttons and undesired features.

Partly Addresses #1160 & #1161

I enabled buttons on all chart component and removed buttons that doesn't correctly interact with OED, redundant or isn't useful. 

## Type of change

## Checklist

## Limitations
There is no option/button to display more buttons (advance buttons). The normal user will only use the download png button (downloads a png of a graph).

To Do:
Add a button to display advance buttons. In this case users should see a download png button, the plotly icon and the advance feature button when loading up a graph and clicking the advance feature button should display the advance buttons.


huss (commented):
Thanks to @williamw04 for addressing this issue. Code review found no issues and it works as expected. I think this can sit for a little while to see if someone can figure out how to hide/show the advanced options until a button is pushed (or something similar). If not, then I think this can be merged and leave that as an open issue.

huss:
For the record, I merged development so the DB is the latest.

williamw04:
Implemented advanced button feature to all graphics.

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -7,7 +7,9 @@
 import { connect } from 'react-redux';
 import { getComparePeriodLabels, getCompareChangeSummary, calculateCompareShift } from '../utils/calculateCompare';
 import translate from '../utils/translate';
+import * as React from 'react';
```
huss: I think this was added for useState. If so, could it just import that as done in src/client/app/components/HeaderButtonsComponent.tsx? I realize we don't consistently do this but we are trying to move in that direction.

src/client/app/components/LineChartComponent.tsx
```diff
@@ -86,8 +95,17 @@ export default function LineChartComponent() {
 				}}
 				config={{
 					responsive: true,
-					displayModeBar: false,
-					// Current Locale
+					displayModeBar: true,
+					modeBarButtonsToRemove: listOfButtons,
+					modeBarButtonsToAdd: [{
+						name: 'more-options',
+						title: 'More Options',
```
huss: This comment applies across all the components changed. Two things:
1. Could this be internationalized with translate() and data.ts.
2. I'm thinking making the text 'Toggle Options' might be better since it is the same text whether you are going to all or few icons. What do you think?

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -84,6 +86,14 @@ function mapStateToProps(state: RootState, ownProps: CompareChartContainerProps)
 		}
 	}
 
+	// Display Plotly Buttons Feature
+	// The number of items in defaultButtons and advancedButtons must differ as discussed below
+	const defaultButtons: Plotly.ModeBarDefaultButtons[] = ['zoom2d', 'pan2d', 'select2d', 'lasso2d', 'zoomIn2d', 'zoomOut2d', 'autoScale2d',
+		'resetScale2d'];
+	const advancedButtons: Plotly.ModeBarDefaultButtons[] = ['select2d', 'lasso2d', 'autoScale2d', 'resetScale2d'];
+	// Manage button states with useState
+	const	[listOfButtons, setListOfButtons] = React.useState(defaultButtons);
```
huss: When I click this icon it crashes OED. The error relates to using a Hook (useState, I think) outside a component. This does not use a component as the other graphics do as it is a container. It either needs a modified solution or the component needs to be converted. I doubt you want to do the later so either find a fix or remove it for now from this graphic and leave a message that it needs to be done where the desired buttons for each case is provided..

huss (changes requested):
Thanks to @williamw04 for figuring out how to make the icons dynamic on click. Overall, it works well. I have made a few comments in files. Also, I noticed that maps is not done. I don't recall if that was by design or because maps were not converted to the latest Redux. Right now I'm thinking it would be nice. If you agree, maybe a comment on PR #1314 that is updating React usage for maps with the desired buttons for each case would be good. Then it could be added to that.

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -84,6 +87,18 @@ function mapStateToProps(state: RootState, ownProps: CompareChartContainerProps)
 		}
 	}
 
+	// To Do: When I click this icon it crashes OED. The error relates to using a Hook (useState, I think)
```
huss: This should be TODO (instead of To Do). That way it shows up at a TODO item in many IDEs. Same to To Do below.

Really small: There is an extra space at the end of the line on this and the next line. The linting does not like this. This is why the tests are failing on GitHub.

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -84,6 +87,18 @@ function mapStateToProps(state: RootState, ownProps: CompareChartContainerProps)
...
+	// outside a component. This does not use a component as the other graphics do as it is 
+	// a container. It either needs a modified solution or the component needs to be converted.
+	// Only after the component has been converted uncomment the code below and in plotly config
+	// Display Plotly Buttons Feature:
+	// The number of items in defaultButtons and advancedButtons must differ as discussed below
+	const defaultButtons: Plotly.ModeBarDefaultButtons[] = ['zoom2d', 'pan2d', 'select2d', 'lasso2d', 'zoomIn2d', 'zoomOut2d', 'autoScale2d',
+		'resetScale2d'];
+	// const advancedButtons: Plotly.ModeBarDefaultButtons[] = ['select2d', 'lasso2d', 'autoScale2d', 'resetScale2d'];
+	// Manage button states with useState
+	// const	[listOfButtons, setListOfButtons] = useState(defaultButtons);
```
huss: It is within a comment but probably want a space and not a tab after const.

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -230,7 +245,19 @@ function mapStateToProps(state: RootState, ownProps: CompareChartContainerProps)
 		data: datasets,
 		layout,
 		config: {
-			displayModeBar: false,
+			displayModeBar: true,
+			modeBarButtonsToRemove: defaultButtons,
+			// TO DO: Removes line above and uncomment betlow. Read above for more info
```
huss: betlow -> below

src/client/app/components/LineChartComponent.tsx
```diff
@@ -86,8 +95,17 @@ export default function LineChartComponent() {
 				}}
 				config={{
 					responsive: true,
-					displayModeBar: false,
-					// Current Locale
+					displayModeBar: true,
+					modeBarButtonsToRemove: listOfButtons,
+					modeBarButtonsToAdd: [{
+						name: 'toggle-options',
+						title: 'Toggle Options',
```
huss: This is not internationalized. Look at useTranslate (src/client/app/components/unit/CreateUnitModalComponent.tsx has an example) for how to do the code. You also need to add the used key and text to data.ts in all three languages. When it works the text will change when the language is changed. This applies across all the pages with this icon. Let me know if you need help on this.

huss (changes requested):
Thanks to @williamw04 for the updated code. I've made a few comments to address. Also, I am getting a warning on compile of:

```web-1       | Failed to parse source map from '/usr/src/app/node_modules/@plotly/mapbox-gl/dist/mapbox-gl-unminified.js.map' file: Error: ENOENT: no such file or directory, open '/usr/src/app/node_modules/@plotly/mapbox-gl/dist/mapbox-gl-unminified.js.map'```

I don't think it was there before. I wonder if it relates to the change in the includes (but I have not looked at it).

File: src/client/app/translations/data.ts [hunk 8]
```diff
@@ -442,6 +442,7 @@ const LocaleTranslationData = {
 		"TimeSortTypes.meter": "meter value or default",
 		"today": "Today",
 		"toggle.link": "Toggle chart link",
+		"Toggle Option" : "Toggle Option",
```
huss: The key should be lowercase and words separated by a period so it is "toggle.option". Also, the text should be "Toggle Options" (pluralized).

src/client/app/components/LineChartComponent.tsx
```diff
@@ -86,8 +95,17 @@ export default function LineChartComponent() {
 				}}
 				config={{
 					responsive: true,
-					displayModeBar: false,
-					// Current Locale
+					displayModeBar: true,
+					modeBarButtonsToRemove: listOfButtons,
+					modeBarButtonsToAdd: [{
+						name: 'toggle-options',
+						title: translate('toggle option'),
```
huss: In this and all similar files this should be toggle.option for the translate key.

Since PR #1354 was merged, component functions should replace translate with useTranslate. This also applies to all related files.

src/client/app/containers/CompareChartContainer.ts
```diff
@@ -230,7 +245,19 @@ function mapStateToProps(state: RootState, ownProps: CompareChartContainerProps)
 		data: datasets,
 		layout,
 		config: {
-			displayModeBar: false,
+			displayModeBar: true,
+			modeBarButtonsToRemove: defaultButtons,
+			// TO DO: Removes line above and uncomment below. Read above for more info
```
huss: TO DO -> TODO

huss:
> Thanks to @williamw04 for the updated code. I've made a few comments to address. Also, I am getting a warning on compile of:
> 
> `web-1 | Failed to parse source map from '/usr/src/app/node_modules/@plotly/mapbox-gl/dist/mapbox-gl-unminified.js.map' file: Error: ENOENT: no such file or directory, open '/usr/src/app/node_modules/@plotly/mapbox-gl/dist/mapbox-gl-unminified.js.map'`
> 
> I don't think it was there before. I wonder if it relates to the change in the includes (but I have not looked at it).

I tried to resolve this. I found [this plotly GitHub issue](https://github.com/plotly/plotly.js/issues/7161) that might have been related (for different package). It was resolved very recently. I don't know what it means for webpack given the shift in plotly compiles.

I tried upgrading plotly packages to the latest with ``npm i plotly.js@2.35.2 @types/plotly.js@2.33.5 @types/react-plotly.js@2.6.3`` but this did not resolve it (actually there are now two warnings). It seems that the packages are old enough not to include recent changes.

There are ways to stop creating the source map but I did not try and had mixed feelings.

Does anyone have other thoughts/ideas?

huss (changes requested):
Thanks to @williamw04 for the changes. I've made a few comments to address. I also added a comment on the compile warning. It is not resolved but not sure what to do.

src/client/app/containers/CompareChartContainer.ts
huss: I fixed this.

huss:
The warning is not going away so going to live with it for now. I'm leaving this comment so people know about it.

huss (approved):
Thanks to @williamw04 for the changes and sorry I missed this was ready for review. Everything now looks good. Congratulations on another task finished.

Issue #1205, opened 2024-04-04
maps handling negative values better


huss:
### Describe the bug

Originally, OED did not have negative readings but it is now allowed. If you place a negative value on a map, you see no circle. If you place the cursor over the expected location you see the correct hover. This screenshot shows this:

![image](https://github.com/OpenEnergyDashboard/OED/assets/987062/fc11dac1-1295-47b8-a7c8-48ab2fc4327d)

The issue is likely that src/client/app/containers/MapChartContainer.ts finds the max size where it should probably use the max - min size or something like this.

Another option might be to use another visual for negative (lines through color circle or something else) so it is clear to the user that some values are negative. This would be combined with fixing the actual size.

If others have thoughts about the look then please state them. The ideas here would have larger negative values have smaller circles.

### To Reproduce

Steps to reproduce the behavior (assuming you have the usual developer testData loaded):
1. Create a new meter where the important values are:

![image](https://github.com/OpenEnergyDashboard/OED/assets/987062/2ff6bc1f-16a1-4afa-80dd-e188ad8dab76)

2. Place these values in a CSV file for the readings:

reading,start time,end time
-10,2023-01-01 00:00:00,2024-01-01 00:00:00

and do a CSV Readings upload where you specify the meter (from step 1), header row and refresh the daily/hourly readings, along with the CSV created in this step.
3. Go to the map and graph this meter.
4. Place cursor over Library to see

Note the provided CSV reading value is a constant -10 across all of 2023. It could be modified to get other values. If you want to place a circle in another location, looking at the location/GPS value for the amp meters will help.

### Expected behavior

The circle size should be visible and scale with usage.

### Additional context

None

Rakesh-Ranga-Buram:
The above mentioned file "src/client/app/containers/MapChartContainer.ts" is currently not in use as OED  migrated to use Redux Toolkit. This file is removed as part of #1314 . The correct file location is "src/client/app/components/MapChartComponent.tsx"

mmehta2669:
Please assign this to me

PR #1349, opened 2024-09-30
Adding Negative Functionallity to MapChartComponent

mmehta2669:
# Description

This changing the negative Functionality in the MapChartComponent. It does this bey changing the scaling to factor in the max-min, and adds and minShift that shifts the negative values into the positive with maintaining that the  radius decreases as the absolute value of the negative meter increases

Fixes [#1205]

## Type of change

## Checklist

## Limitations

I think that the color change for the negative meters could be more dynamic, however that may require a more refined color selection function


mmehta2669:
Fixes #1205

lawsj:
I see in the code that you added the various min functions to help set the size and scale of the circles that appear on the maps. This addition allows the user to have a more clear indication when a negative value is reported. Good job!

PR #1351, opened 2024-10-03
Compare line page

### LINKED ISSUE #953: new line comparison graphic

OED would like to add a line comparison graphic.

This is described in the [line comparison design document.](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/lineCompare/lineCompare.md)

We don't support this.

### Additional context

None

nmqng:
# Description

The purpose of this pull request is to add a comparison line graphic page based on the [design doc](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/lineCompare/lineCompare.md).

Fixes #953 

## Type of change

## Checklist

## Limitations

Currently, when a shifting to/from a leap year from/to a non-leap year, a warning show up to tell the user that the original graph and the shifted graph will not align appropriately. However, in the future, we would want to find a way that can make the two graphs align.


src/client/app/translations/data.ts
```diff
@@ -1010,7 +1022,20 @@ const LocaleTranslationData = {
 		"week": "Semaine",
 		"yes": " yes\u{26A1}",
 		"yesterday": "Hier",
-		"you.cannot.create.a.cyclic.group": "Vous ne pouvez pas créer un groupe cyclique"
+		"you.cannot.create.a.cyclic.group": "Vous ne pouvez pas créer un groupe cyclique",
+		"compare.line": "Compare line\u{26A1}",
+		"shift.date.interval": "Shift Date Interval\u{26A1}",
+		"a few seconds": "a few seconds\u{26A1}",
```
huss: This is not in en, it is not a normal OED key with periods (not spaces) between words and I cannot fiture out where it is used.

src/client/app/translations/data.ts
```diff
@@ -508,7 +508,19 @@ const LocaleTranslationData = {
 		"week": "Week",
 		"yes": "yes",
 		"yesterday": "Yesterday",
-		"you.cannot.create.a.cyclic.group": "You cannot create a cyclic group"
+		"you.cannot.create.a.cyclic.group": "You cannot create a cyclic group",
+		"compare.line": "Compare line",
```
huss: OED keeps this file alpabetical by key. If you want items to stay together then you have to make the key start similarly. I'm not saying that is needed but just making the point.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -88,6 +90,14 @@ export const graphSlice = createSlice({
 				state.current.queryTimeInterval = action.payload;
 			}
 		},
+		updateShiftTimeInterval: (state, action: PayloadAction<TimeInterval>) => {
+			if (action.payload.getIsBounded() || state.current.shiftTimeInterval.getIsBounded()) {
```
huss, line 97: Could you please add a comment on why the logic is this way.

src/client/app/components/UIOptionsComponent.tsx
```diff
@@ -80,6 +81,11 @@ export default function UIOptionsComponent() {
 			{/* UI options for radar graphic */}
 			{chartToRender == ChartTypes.radar}
 
+			{	/* Controls specific to the compare line chart */}
+			{chartToRender === ChartTypes.compareLine && <DateRangeComponent />}
+			{chartToRender === ChartTypes.compareLine && <CompareLineControlsComponent />}
+
+
```
huss: Extra blank line

src/client/app/components/MoreOptionsComponent.tsx
```diff
@@ -74,6 +74,13 @@ export default function MoreOptionsComponent() {
 							{chartToRender == ChartTypes.radar && <DateRangeComponent />}
 							{chartToRender == ChartTypes.radar && <AreaUnitSelectComponent />}
 							{chartToRender == ChartTypes.radar && <ChartLinkComponent />}
+
+							{/*More UI options for compare line */}
+							{chartToRender === ChartTypes.compareLine && <GraphicRateMenuComponent />}
+							{chartToRender === ChartTypes.compareLine && <AreaUnitSelectComponent />}
+							{chartToRender === ChartTypes.compareLine && <ErrorBarComponent />}
+							{chartToRender === ChartTypes.compareLine && <ExportComponent />}
+							{chartToRender === ChartTypes.compareLine && <ChartLinkComponent />}
```
huss, line 83: The chart link does not seem to reproduce the graphic. For one example it did not set the selected meter nor the shift date amount. This may not be a complete list but I think you need to modify src/client/app/redux/selectors/uiSelectors.ts & src/client/app/redux/slices/graphSlice.ts.

src/client/app/components/MoreOptionsComponent.tsx
```diff
@@ -74,6 +74,13 @@ export default function MoreOptionsComponent() {
 							{chartToRender == ChartTypes.radar && <DateRangeComponent />}
 							{chartToRender == ChartTypes.radar && <AreaUnitSelectComponent />}
 							{chartToRender == ChartTypes.radar && <ChartLinkComponent />}
+
+							{/*More UI options for compare line */}
+							{chartToRender === ChartTypes.compareLine && <GraphicRateMenuComponent />}
+							{chartToRender === ChartTypes.compareLine && <AreaUnitSelectComponent />}
+							{chartToRender === ChartTypes.compareLine && <ErrorBarComponent />}
+							{chartToRender === ChartTypes.compareLine && <ExportComponent />}
```
huss: Export does not seem to do anything. The design doc called for it to be absent so can just remove the option.

src/client/app/components/MoreOptionsComponent.tsx
```diff
@@ -74,6 +74,13 @@ export default function MoreOptionsComponent() {
 							{chartToRender == ChartTypes.radar && <DateRangeComponent />}
 							{chartToRender == ChartTypes.radar && <AreaUnitSelectComponent />}
 							{chartToRender == ChartTypes.radar && <ChartLinkComponent />}
+
+							{/*More UI options for compare line */}
+							{chartToRender === ChartTypes.compareLine && <GraphicRateMenuComponent />}
```
huss, line 80: I'm not seeing this menu but it probably should show. If I go to line with an appropriate flow meter then I see it and changes made impact the values shown on compare line.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as React from 'react';
+import { Input } from 'reactstrap';
+import { useAppDispatch, useAppSelector } from '../redux/reduxHooks';
+// eslint-disable-next-line max-len
```
huss: Maybe break the next line and get the automatic indenting on the second line rather than this eslint exception.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+import DateRangePicker from '@wojtekmaj/react-daterange-picker';
+import { dateRangeToTimeInterval, timeIntervalToDateRange } from '../utils/dateRangeCompatibility';
+import { selectSelectedLanguage } from '../redux/slices/appStateSlice';
+import { Value } from '@wojtekmaj/react-daterange-picker/dist/cjs/shared/types';
+import * as moment from 'moment';
+import { TimeInterval } from '../../../common/TimeInterval';
+import { showWarnNotification } from '../utils/notifications';
+
+/**
+ * @returns compare line control page
```
huss: Is this a page or a component that goes on a page?

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+}
+
+const labelStyle: React.CSSProperties = { fontWeight: 'bold', margin: 0 };
+
+/**
+ * shifting date function to find the shifted start date and shifted end date
+ * @param originalStart  start date of current graph data
+ * @param originalEnd end date of current graph data
+ * @param shiftType shifting amount in week, month, or year
```
huss, line 120: What happens if it is custom? The code deals with none but not in this comment.

The comments on these may be obsolete given the changes in the design document.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+ * @returns shifted start and shifted end dates for the new data
+ */
+export function shiftDateFunc(originalStart: moment.Moment, originalEnd: moment.Moment, shiftType: ShiftAmount) {
+	let shiftedStart: moment.Moment;
+	let shiftedEnd: moment.Moment;
+
+	const originalRangeDays = originalEnd.diff(originalStart, 'days');
+
+	switch (shiftType) {
```
huss: I think some comments on the logic would help.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+			}
+			if (originalEnd.isLeapYear() && originalEnd.month() === 1 && originalEnd.date() === 29) {
+				shiftedEnd = shiftedEnd.month(1).date(28);
+			}
+			if (shiftedEnd.isSameOrAfter(originalStart)) {
+				shiftedEnd = originalStart.clone().subtract(1, 'day');
+			}
+			break;
+
+		default:
```
huss: How does default happen? If this is normally an error then need to decide if this is a good return value vs. error.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+			shiftedEnd = originalEnd.clone().subtract(7, 'days');
+			break;
+
+		case 'month':
+			shiftedStart = originalStart.clone().subtract(1, 'months');
+			shiftedEnd = shiftedStart.clone().add(originalRangeDays, 'days');
+
+			if (shiftedEnd.isSameOrAfter(originalStart)) {
+				shiftedEnd = originalStart.clone().subtract(1, 'day');
+			} else if (originalStart.date() === 1 && originalEnd.date() === originalEnd.daysInMonth()) {
```
huss: This is a note that I need to recheck this once there are comments. I think it is trying to align the ends of months.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+				if (!(shiftedStart.date() === 1 && shiftedEnd.date() === shiftedEnd.daysInMonth())) {
+					shiftedEnd = shiftedStart.clone().endOf('month');
+				}
+			}
+			break;
+
+		case 'year':
```
huss: Another one to check after comments.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+		default:
+			shiftedStart = originalStart.clone();
+			shiftedEnd = originalEnd.clone();
+	}
+
+	return { shiftedStart, shiftedEnd };
+}
+
+/**
+ * This function check whether the original date range is leap year or the shifted date range is leap year.
```
huss: Is it actually if one is in a leap year but the other is not? See below about multiyear.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+	// Check for leap year shifting when new interval or meter/group is chosen
+	React.useEffect(() => {
+		const startDate = timeInterval.getStartTimestamp();
+		const endDate = timeInterval.getEndTimestamp();
+		if (startDate && endDate) {
+			// Check whether shifting to (or from) leap year to non leap year or not
+			checkLeapYearFunc(startDate, endDate, shiftOption);
+		}
+	}, [graphState.threeD.meterOrGroupID, timeInterval]);
```
huss: Why does this depend on the selected meter/group to graph? I'm unclear on how that changes the dates? Are there other dependencies (may need to recheck after changes made)?

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+		setCustomDateRange(timeIntervalToDateRange(shiftInterval));
+	}, [shiftInterval]);
+
+	// Check for leap year shifting when new interval or meter/group is chosen
+	React.useEffect(() => {
+		const startDate = timeInterval.getStartTimestamp();
+		const endDate = timeInterval.getEndTimestamp();
+		if (startDate && endDate) {
+			// Check whether shifting to (or from) leap year to non leap year or not
+			checkLeapYearFunc(startDate, endDate, shiftOption);
```
huss: It is interesting that every use seems to have the timeInterval that is separated into start/end and then passed to checkLeapYearFunc. Would it be cleaner to send the timeInterval as a single parameter? There is a comment in CompareLineChartCompare that relates to this idea.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+const labelStyle: React.CSSProperties = { fontWeight: 'bold', margin: 0 };
+
+/**
+ * shifting date function to find the shifted start date and shifted end date
+ * @param originalStart  start date of current graph data
+ * @param originalEnd end date of current graph data
+ * @param shiftType shifting amount in week, month, or year
+ * @returns shifted start and shifted end dates for the new data
+ */
+export function shiftDateFunc(originalStart: moment.Moment, originalEnd: moment.Moment, shiftType: ShiftAmount) {
```
huss: For this and the next function: I don't think OED generally puts Func at the end of the name, i.e., shiftDateFunc.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * This function check whether the original date range is leap year or the shifted date range is leap year.
+ * If it is true, it warns user about shifting to (or from) a leap year to non leap year.
+ * @param startDate original data start date
+ * @param endDate original data end date
+ * @param shiftOption shifting option
+ */
+function checkLeapYearFunc(startDate: moment.Moment, endDate: moment.Moment, shiftOption: ShiftAmount) {
+	const { shiftedStart, shiftedEnd } = shiftDateFunc(startDate, endDate, shiftOption);
```
huss: I'm curious why the shifted dates must be calculated here when they are needed for the graphing and maybe set for that too. I'm not saying it is an issue but was thinking about this.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+	const originalIsLeapYear = startDate.isLeapYear() || endDate.isLeapYear();
```
huss: What if the data covers multiple years and ones in the middle cross leap year?

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+	const shiftedIsLeapYear = shiftedStart.isLeapYear() || shiftedEnd.isLeapYear();
+
+	// Check if the original date range crosses Feb 29, which causes unaligned graph
+	const originalCrossFeb29 = (
```
huss: Similar to above, I'm wondering about other cases. What if the startData >= March 1 but a following year included in the graph is a leap year that crosses Feb. 29? Same for shifted function below.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+	// Handle changes in shift option (week, month, year, or custom)
+	const handleShiftOptionChange = (value: string) => {
+		if (value === 'custom') {
+			setShiftOption(ShiftAmount.custom);
+			dispatch(updateShiftAmount(ShiftAmount.custom));
+			setShowDatePicker(true);
+		} else {
+			setShowDatePicker(false);
+			const newShiftOption = value as ShiftAmount;
+			setShiftOption(newShiftOption);
```
huss: Why is there a local and global Redux copy of this is they are kept in sync? I'm trying to understand.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,235 @@
...
+			dispatch(updateShiftAmount(newShiftOption));
+
+			// notify user when original data or shift data cross leap year
+			const startDate = timeInterval.getStartTimestamp();
+			const endDate = timeInterval.getEndTimestamp();
+
+			if (startDate && endDate) {
```
huss: Would a more standard check be if the timeInterval is bounded? Not sure that is what you want.

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+		const endDate = timeInterval.getEndTimestamp();
+
+		if (startDate && endDate) {
+			setTimeIntervalStr([startDate.toISOString(), endDate.toISOString()]);
+
+			if (shiftAmount !== ShiftAmount.none && shiftAmount !== ShiftAmount.custom) {
+				updateShiftedInterval(startDate, endDate, shiftAmount);
+			}
+		}
+	}, [timeInterval, shiftAmount, updateShiftedInterval]);
```
huss: I'm curious why this depends on the updateShiftedInterval function. Is this standard for a callback function?

src/client/app/translations/data.ts
```diff
@@ -47,7 +47,7 @@ const LocaleTranslationData = {
 		"clipboard.copied": "Copied To Clipboard",
 		"clipboard.not.copied": "Failed to Copy To Clipboard",
 		"close": "Close",
-		"compare": "Compare",
+		"compare": "Compare bar",
```
huss: I have to put this comment here. The French and Spanish versions need to be updated to reflect the name change. It can be the English with the lightening bolt. Also, it would be good to change the key to compare.bar to be consistent with compare.line.

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+	const locale = useAppSelector(selectSelectedLanguage);
+	const shiftInterval = useAppSelector(selectShiftTimeInterval);
+	const shiftAmount = useAppSelector(selectShiftAmount);
+	const { args, shouldSkipQuery, argsDeps } = useAppSelector(selectCompareLineQueryArgs);
+
+	// getting the time interval of current data
+	const timeInterval = graphState.queryTimeInterval;
+
+	// Storing the time interval strings for the original data and the shifted data to use for range in plot
+	const [timeIntervalStr, setTimeIntervalStr] = React.useState(['', '']);
```
huss: I think 3D stores as an interval object and it has functions to convert to/from the date range picker. Would that make sense here? Would this make the initial value unbounded?

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+				skip: shouldSkipQuery,
+				selectFromResult: ({ data, ...rest }) => ({
+					...rest,
+					data: selectPlotlyGroupData(data ?? stableEmptyLineReadings,
+						{ ...argsDeps, compatibleEntities: [meterOrGroupID!] })
+				})
+			});
+
+	// Callback function to update the shifted interval based on current interval and shift amount
+	const updateShiftedInterval = React.useCallback((start: moment.Moment, end: moment.Moment, shift: ShiftAmount) => {
```
huss: I'm wondering about a callback vs a useEffect. I don't know which is better but this is the first time it has been used in OED (as far as I can tell). The updated design doc has some thoughts on the update.

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+						{ ...argsDeps, compatibleEntities: [meterOrGroupID!] })
+				})
+			});
+
+	if (isFetching || isFetchingNew) {
+		return <SpinnerComponent loading height={50} width={50} />;
+	}
+
+	// Check if there is at least one valid graph for current data and shifted data
+	const enoughData = data.find(data => data.x!.length > 1);
```
huss: Since these are only used together would it make sense to have a single boolean for either case?

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+	const enoughNewData = dataNew.find(dataNew => dataNew.x!.length > 1);
+
+	// Customize the layout of the plot
+	// See https://community.plotly.com/t/replacing-an-empty-graph-with-a-message/31497 for showing text `not plot.
+	if (!graphState.threeD.meterOrGroup && (data.length === 0 || dataNew.length === 0)) {
```
huss: How does this test on data length relate to enough data above (in terms of what it tests and how it is done)?

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+		return <><ThreeDPillComponent /><h1>{`${translate('select.meter.group')}`}</h1></>;
+	} else if (!enoughData || !enoughNewData) {
+		return <><ThreeDPillComponent /><h1>{`${translate('no.data.in.range')}`}</h1></>;
+	} else if (!timeInterval.getIsBounded()) {
```
huss: I don't know if it matters but am wondering if this check should come before the enough data part.

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	// Check if there is at least one valid graph for current data and shifted data
+	const enoughData = data.find(data => data.x!.length > 1);
+	const enoughNewData = dataNew.find(dataNew => dataNew.x!.length > 1);
+
+	// Customize the layout of the plot
+	// See https://community.plotly.com/t/replacing-an-empty-graph-with-a-message/31497 for showing text `not plot.
+	if (!graphState.threeD.meterOrGroup && (data.length === 0 || dataNew.length === 0)) {
+		return <><ThreeDPillComponent /><h1>{`${translate('select.meter.group')}`}</h1></>;
```
huss: ThreeDComponent uses setHelpLayout to do this. I'm thinking it would be a good idea to use that here. Note there is a TODO in src/client/app/components/RadarChartComponent.tsx about doing this there and someone should be working on that. I bring it up because the location of the used function is likely to move from 3D to utils/.

src/client/app/types/redux/graph.ts
```diff
@@ -55,6 +56,14 @@ export interface ThreeDState {
 	readingInterval: ReadingInterval;
 }
 
+export enum ShiftAmount {
+	week = 'week',
```
huss, line 62: The design document has day as an option. Is there a reason not to add that?

huss (changes requested):
I want to thank @nmqng for taking on this larger issue. I also ask your understanding for how long my review took as my efforts where focused elsewhere.

Overall, this does what the design document requested. I have made a number of comments to consider. In the big picture, I spent time considering how compare line functioned. In the end, based on your efforts, I concluded that changes in the logic were desired. This is due to my having a better understanding of the possibilities and not direct issues with your code. As a result, I have updated the design document with a new updated section to reflect my current thinking. I think the changes may simplify the code. I know you put in a lot of effort to get several of the current ideas implemented and I'm sorry for the changes. I also welcome your thoughts on what I wrote. If you want me to be actively involved in updating the code then please let me know.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,134 @@
...
+		setCustomDateRange(dateRangeToTimeInterval(value));
+		dispatch(updateShiftTimeInterval(dateRangeToTimeInterval(value)));
+	};
+
+	return (
+		<>
+			<div key='side-options'>
+				<p style={{ fontWeight: 'bold', margin: 0 }}>
+					<FormattedMessage id='shift.date.interval' />
+					{/* <TooltipMarkerComponent helpTextId='help.shift.date.interval' /> // TODO: Add later */}
```
huss: Uncomment

src/client/app/translations/data.ts
huss: Thanks for finding a way to alphabetize all the keys.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,134 @@
...
+				<Input
+					id='shiftDateInput'
+					name='shiftDateInput'
+					type='select'
+					value={shiftOption}
+					invalid={shiftAmount === ShiftAmount.none}
+					onChange={e => handleShiftOptionChange(e.target.value)}
+				>
+					<option value="none" hidden disabled>{translate('select.shift.amount')}</option>
+					<option value="day">{translate('1.day')}</option>
```
huss: Maybe look at src/client/app/components/IntervalControlsComponent.tsx for how it loops over object to do this. I think in other places it might loop over an enum. Would that make sense here or is it too specific to this case?

src/client/app/components/CompareLineControlsComponent.tsx
huss: I think it was converted to an if statement and the default is gone.

src/client/app/components/CompareLineControlsComponent.tsx
```diff
@@ -0,0 +1,134 @@
...
+			setShiftOption(ShiftAmount.custom);
+			dispatch(updateShiftAmount(ShiftAmount.custom));
+		} else {
+			const newShiftOption = value as ShiftAmount;
+			setShiftOption(newShiftOption);
+			dispatch(updateShiftAmount(newShiftOption));
+		}
+	};
+
+	// Update date when the data range picker is used in custome shifting option
```
huss, line 65: custome misspelled.

src/client/app/components/CompareLineControlsComponent.tsx
nmqng: As new implementation from the design document, comments related to this function and checkLeapYear() function are resolved.

huss (approved):
Review and testing indicate this is complete. There is an issue that the page redraws too many times so the messages are repeated. This appears to be a more general issue that impacts other pages (but not visible to the user). Attempts to fix have not yet worked. Thus, this is being merged and a new issue created.

Issue #1204, opened 2024-04-04
standardize translation method


huss:
### Is your feature request related to a problem? Please describe.

In many places OED uses translate() and in some it uses FormatMessage(). They both appear to return a string.

### Describe the solution you'd like

OED should pick the better of these two and use it consistently in the code. Maybe research the two and propose one to the project. Then make the needed edits once everyone agrees.

### Describe alternatives you've considered

Leave as is

### Additional context

None

huss:
This seems to relate to issue #1189 and can probably be done together.

lawsj:
Please assign this issue to me.

lawsj:
I found 15 instances of formatMessage() across the files.
I found 66 instances of translate() across the files.

General uses:

Both in react-intl library

translate():
Convert strings from one language to another, using a file/directory as a key.
ex) translate(“greeting”) returns “hello” for english or “hola” for Spanish

formatMessage():
Formatting strings from variables
Good for displaying messages that use variables, such as:error messages, user notifications, etc.
https://formatjs.io/docs/react-intl/api/	
ex) FormatMessage({id: ‘welcome’, defaultMessage: ‘Welcome, {name}!’}, {name: ‘Alice’}) returns “Welcome, Alice!”

https://react.i18next.com/latest/usetranslation-hook

Experimenting:
The changes from translate() to useTranslate() did not show errors, but all of the changes from formatMessage() to useTranslate() showed errors. These were the errors encountered when trying to implement this change:
1. Only allow translate() instead of useTranslate()
2. Intl would have to be removed from each instance, since the library is not used to call the useTranslate() function.

I think that it may be better to use the formatMessage() across the program if we want to continue to use the react-intl library. Translate() seems simpler, but it does not seem to be able to handle as many variables like the formatMessage() can. I would even propose to use translate() in some areas and use FormatMessage() strictly for error handling, which is how the instances seem to be used currently.

nneni:
I like how thorough your analysis of the two functions was, it was very enlightening and helped me form a better, more factual opinion of the efficiency of the use of either function. I agree with your statement at the end as well, the functions seem to each work better in different situations.

huss:
@lawsj Thanks for looking into this and your analysis. Would it be possible to meet to discuss your two options? (You can DM/email me to find a time.) I'd like to understand the tradeoff a little better so a final decision can be made. If not, I can ask questions here.

http-impreza:
I would like to take on this issue.

[...]

PR #1354, opened 2024-10-04
Standardize Translation Method

lawsj:
# Description

Made changes to use the new  "useTranslate()" instead of the old "translate()".

Partly Addresses #1204 

## Type of change

## Checklist

##Limitations

N/a


Rakesh-Ranga-Buram:
Everything looks good to me. Please convert the draft to PR.

Rakesh-Ranga-Buram:
@huss , this PR only include changes related to translate() method and not formatMessage() method. Since formatMessage() changes are already been done on #1314

src/client/app/components/BarControlsComponent.tsx
```diff
@@ -89,6 +89,7 @@ export default function BarControlsComponent() {
 			dispatch(graphSlice.actions.updateBarDuration(moment.duration(value, 'days')));
 		}
 	};
+	const translate = useTranslate();
```
huss: While this comment is here it applies to many files changed. The location of this line varies by file and is not, from my perspective, in the ideal location in the code. It is similar to ``const dispatch = useAppDispatch();`` that is at/near the top of the function. I think that is a good location for this type of code. I think all of these lines should be there (or in the block defining functions as in some files) or near the top of the function if dispatch is not present. If you feel otherwise then please let me know.

huss (changes requested):
Thanks to @lawsj for another contribution. Overall, this works as desired. I made one comment in a file to address. In addition, here are comments that could not be placed in a file:

- src/client/app/utils/translate.ts has the previously used translation function. I could not find it used (which is good). If so, I think this file should be removed to avoid unused code and to stop someone from using the old translate method.
- OED is crashing when I go to a number of admin pages. For example, going to the conversion page and then try to create a conversion does this. What is strange is I cannot see why any change made in this PR would cause this to happen. Thus, I removed the new translate change in this file but it still happened. Next I tried changing to the development branch and it no longer crashed. This seems to imply it was a change in the pull request but why is unclear. Modest debugging did not find the reason. Given this, I wanted to know if you (or anyone else) sees this same type of failure? I want to know if it is a general problem or something in my setup.

src/client/app/components/BarControlsComponent.tsx
Rakesh-Ranga-Buram: @huss , I too see the same type of failure.  After debugging, I see that placing `const translate = useTranslate();` at the top of the function was crashing in some components. However placing it somewhere below was rectifying the error. I was unable to technically debug why this was happening.
On the other hand, the translate changes in some components was resulting in error and that is why some admin pages were crashing. However in this case, altering the location of `const translate = useTranslate();` did not work out.

huss:
This is just for the record: We are continuing to try to determine the cause of the issue. It may well be something outside the direct changes made in this PR.

Rakesh-Ranga-Buram:
@huss , the code has been updated with the discussed changes. Can you please review this PR

huss:
I pushed commits to do the following:

1. Merge in development and address conflicts.
2. Added missing CSV menu translations. Not directly related to this PR but similar so did it.
3. Removed a couple of uses of useTranslate from maps because outside function component and causing issues.

huss (approved):
Thanks to @Rakesh-Ranga-Buram for the updated code. Review and testing found it works as desired at this time. Congratulations to @lawsj on their first accepted contribution to OED.

PR #1357, opened 2024-10-05
Adding negative function to maps

### LINKED ISSUE #1205: maps handling negative values better

mmehta2669:
# Description

Added changes to maps to handle negative values

Fixes #1205 

## Type of change

## Checklist

## Limitations
N/A

### DIFF
--- a/src/client/app/components/MapChartComponent.tsx
+++ b/src/client/app/components/MapChartComponent.tsx
@@ -8,9 +8,9 @@ import * as React from 'react';
 import Plot from 'react-plotly.js';
 import { useSelector } from 'react-redux';
 import {
-    selectAreaUnit, selectBarWidthDays,
-    selectGraphAreaNormalization, selectSelectedGroups,
-    selectSelectedMeters, selectSelectedUnit
+	selectAreaUnit, selectBarWidthDays,
+	selectGraphAreaNormalization, selectSelectedGroups,
+	selectSelectedMeters, selectSelectedUnit
 } from '../redux/slices/graphSlice';
 import { selectGroupDataById } from '../redux/api/groupsApi';
 import { selectMeterDataById } from '../redux/api/metersApi';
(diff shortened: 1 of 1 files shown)


mmehta2669:
Thanks @huss for the suggestions on the code. All the suggestions were really helpful to understand how Plotly works.

Rakesh-Ranga-Buram:
Everything looks fine to me. Please convert this draft to PR

src/client/app/components/MapChartComponent.tsx
```diff
@@ -289,22 +290,105 @@ export default function MapChartComponent() {
 					}
 				}
 			}
+
+			// TODO DEBUG Using amp 1, 2 & 3 from test data that within map. This arbitrarily changes the value for testing.
```
huss: Debugging statements that are commented out should be removed before the code is added to OED.

huss (changes requested):
Thanks to @mmehta2669 for another contribution. Review and testing found it works as desired. I have made one small comment to address.

huss (approved):
Thanks to @mmehta2669 for addressing my comment. Everything looks good. Congratulations on another accepted contribution.

PR #1360, opened 2024-10-16
Revised margins for graphics components

### LINKED ISSUE #1172: Consistent margin on graphics

Plotly was giving 3D a margin top for some reason. Setting all margins and padding to 0 seems to have fixed it. OED should be consistent about this.

Do for all others to ensure it doesn’t catch us off guard randomly in the future somehow. This assumes this is an okay change and good idea.

### Additional context

None.

hinyan17:
# Description

Modified margins in the layout objects of the specified components in src/client/app/components. This was done with the intent of making the graphs bigger by eliminating unnecessary space.

Implemented changes in Bar and Line: set top, bottom, and right margins to 0. Left margin was not removed because the y axis label would overlap with the units.
Implemented changes in Map: set bottom, left, and right margins to 0. Top margin was not removed because it would hide the name of the map. **Also made the same change to app/containers/MapChartContainer.ts to be consistent, which changed the size of the circles representing groups/meters over the buildings.

3D is unmodified because it was previously addressed. MultiCompare and Radar are unmodified because they have special, preset margin settings.

Fixes #1172

## Type of change

## Checklist

## Limitations

No remaining issues, unless it's decided that some margins should be further modified for whatever reason.

### DIFF
--- a/src/client/app/components/LineChartComponent.tsx
+++ b/src/client/app/components/LineChartComponent.tsx
@@ -78,6 +78,7 @@ export default function LineChartComponent() {
 				data={data}
 				style={{ width: '100%', height: '100%', minHeight: '700px' }}
 				layout={{
+					margin: { t: 0, b: 0, r: 0 }, // Eliminate top, bottom, and right margins
 					autosize: true, showlegend: true,
 					legend: { x: 0, y: 1.1, orientation: 'h' },
 					// 'fixedrange' on the yAxis means that dragging is only allowed on the xAxis which we utilize for selecting dateRanges
--- a/src/client/app/components/LineChartComponent.tsx
+++ b/src/client/app/components/LineChartComponent.tsx
@@ -78,7 +78,7 @@ export default function LineChartComponent() {
 				data={data}
 				style={{ width: '100%', height: '100%', minHeight: '700px' }}
 				layout={{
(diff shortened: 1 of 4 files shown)


src/client/app/components/LineChartComponent.tsx
```diff
@@ -78,6 +78,7 @@ export default function LineChartComponent() {
 				data={data}
 				style={{ width: '100%', height: '100%', minHeight: '700px' }}
 				layout={{
+					margin: { t: 0, b: 0, r: 0 }, // Eliminate top, bottom, and right margins
```
huss: I noticed a subtle issue where the grab slider rectangle on the bottom, far right was being obscured. I think setting the the right as r: 3 will fix this. The one of the left is losing its very left edge but changing the margin does not seem to fix so it is fine. Note the same change applies to bar as it has a slider too.

huss (changes requested):
Thanks to @hinyan17 for another contribution. Overall, it all looks good and addresses the issue. I found one small point that I think is easy to fix if you agree.

src/client/app/components/LineChartComponent.tsx
hinyan17: Fixed it, thanks for the suggestion.

huss (approved):
Thanks to @hinyan17 for the quick changes. Everything looks good and this is ready to merge. Thanks for another accepted contribution.

PR #1368, opened 2024-10-23
Added Admin Alert When Meter Is Set as Non-displayable (Resolved Issue#743)

### LINKED ISSUE #743: Giving warnings when setting displayable to false (groups)

There are cases where if you set a meter to be non-displayable, but a group containing this meter to be displayable, the group would be allowed to show its readings when graphing onto a line/bar/etc. This is accounted for and hidden away in the groups page, but the readings can be retrieved. 

We should give warnings to administrators such that we would recursively check everytime we edit a meter to be non-displayable (in addition to groups) where it would search all groups affected and to notify the user that this would expose hidden meters/groups. 

We could also set displayable so that if group's displayable is set to true, all meter/groups inside that group are set to displayable and cannot be changed until you set the group's displayable to false.

### Additional context

This is something that should be included in the Resource Generalization board where work is being done to recursively check other aspects of the application. Using the same algorithm and applying this change would be ideal.

danielshid:
# Description

I added an admin alert that pops up when a meter is set as non-displayable, which lists the groups the meter is used in and prompts the user to either confirm or cancel.

Fixes #743

## Type of change

## Checklist

## Limitations

N/A


src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -180,6 +180,34 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+		let save = true;
+		if (!JSON.parse(e.target.value)) {
+			// This will hold the overall message for the admin alert.
+			let msg = '';
+			// This will hold the names of groups that are affected.
+			let groups = '';
+			// Tells if the change should be cancelled.
+			// Checks for groups that include the meter being edited.
+			for (const groupId of Object.values(groupDataByID)) {
+				if (groupId.deepMeters.includes(meterState.id)) {
```
huss: This is my bad for not thinking of this earlier. If the group has its displayable as no/false then it should not be listed and should not impact saving the change. The logic is that if you cannot see the group then hiding a meter with that group does not change the situation.

huss (changes requested):
Thanks to @danielshid for another contribution. Review and testing found it works as specified. I did made one comment about how it functions that was not specified in the issue. I would appreciate you addressing it.

huss (approved):
Thanks to @danielshid for doing the requested update for a changed requirement. Review and testing found it works as desired. For the record, I modified the notice text to something I thought worked a little better (I just decided this). Congratulations on another accepted contribution.

PR #1379, opened 2024-11-03
updated ThreeD, Radar, Map area calculation method

### LINKED ISSUE #1200: consistent area calculation for graphics

src/client/app/components/LineChartComponent.tsx uses selectPlotlyMeterData (which then calls selectScalingFromEntity) to normalize by area. src/client/app/components/ThreeDComponent.tsx does it in that component.

OED should try to be consistent in how it does similar work. The area should be done as in the line chart. Other graphics should be checked to see if they do area in this way.

Leave as is

### Additional context

If any other important differences are found then they should either be fixed or have issues created.

hinyan17:
# Description
Updated the method of area calculation / normalization in ThreeD, Radar, and Map components. They used to calculate area individually, but now call functions from src/client/app/redux/selectors/entitySelectors.ts to do so. This is now consistent across all components. Line and Bar are unchanged because they already follow this method. MultiCompare doesn't need to.

Fixes #1200

## Type of change

## Checklist

## Limitations
none

### DIFF
--- a/src/client/app/components/MapChartComponent.tsx
+++ b/src/client/app/components/MapChartComponent.tsx
@@ -145,22 +145,20 @@ export default function MapChartComponent() {
 
 			for (const meterID of selectedMeters) {
 				// Get meter id number.
-				// Get meter GPS value.
-				const gps = meterDataById[meterID].gps;
+				const entity = meterDataById[meterID];
 				// filter meters with actual gps coordinates.
-				if (gps !== undefined && gps !== null && meterReadings !== undefined) {
-					let meterArea = meterDataById[meterID].area;
+				if (entity.gps !== undefined && entity.gps !== null && meterReadings !== undefined) {
 					// we either don't care about area, or we do in which case there needs to be a nonzero area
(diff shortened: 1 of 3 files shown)


src/client/app/components/MapChartComponent.tsx
```diff
@@ -151,10 +152,7 @@ export default function MapChartComponent() {
 					let meterArea = meterDataById[meterID].area;
```
huss: In RadarChartComponent, you used an entity and then consistently used that to get the needed attributes. I think that is better and could allow avoiding getting the entity again for areaUnit. If you agree then please make the change.

huss (changes requested):
Thanks to @hinyan17 for another contribution to OED. Review and testing found it works as expected. I have made one small comment to consider.

src/client/app/components/MapChartComponent.tsx
hinyan17: Sounds good, I've made the necessary changes for Map

huss (approved):
Thanks to @hinyan17 for the quick update. Everything looks good. Thanks for another accepted contribution.

PR #1388, opened 2024-11-17
Custom rate and named standard rate for units (Issue #1216)

### LINKED ISSUE #1216: standard choices for sec. in rate

On unit create/edit, there is a box to enter the Sec. in Rate as a number. It would be nice if the admin had a list of standard values and also could enter a custom value if desired.

The list of standard values would be: Per Second (1), Per Minute (60) and Per Hour (3600).

The UI to allow this list and a custom value is unclear. OED has places where there are fixed value menus using an enum but not then allowing a custom value.

Leave as is if this is difficult to do.

### Additional context

None.

gyordong:
# Description

Fixes #1216

Done in collaboration with @huss, @aduques, @AchsahJojo, @emilysarne.

## Type of change

## Checklist

- [X] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [X] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.


src/client/app/translations/data.ts
```diff
@@ -473,6 +476,7 @@ const LocaleTranslationData = {
 		"unit.preferred.display": "Preferred Display:",
 		"unit.represent": "Unit Represent:",
 		"unit.sec.in.rate": "Sec in Rate:",
+		"unit.sec.in.rate.enter": "Input seconds in rate and then hit enter:\u{26A1}",
```
huss: This is the en (English) section so the text is already in the correct language. Thus, no ``\u{26A1}`` is needed and should be removed.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -40,68 +42,117 @@ export default function CreateUnitModalComponent() {
 		// The units API expects these values to be undefined on call so that the database can assign their values.
 		id: -99
 	};
-
 	/* State */
```
huss: I'm just going to note that you removed blank lines added by others. I think the intend was to leave space between sections/items of code. It was probably best to leave most of these, esp. given you left some blank lines. Up to you if you want to put them back as it will be accepted either way. FYI.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -40,68 +42,117 @@ export default function CreateUnitModalComponent() {
...
+	const handleStandardNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		const { value } = e.target;
+		// Check if the custom value option is selected
+		if (value === CUSTOM_INPUT) {
+			setCustomRate(Number(rate));
+			setRate(CUSTOM_INPUT);
+			setShowCustomInput(true);
+		} else {
+			setRate(value);
+			setState({ ...state, [e.target.name]: Number(value) });
```
huss: This seems specific to secInRate (only used there) so should it be general or specific to that for the state update? That is what seems to happen in the handleCustomRateChange.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -40,68 +42,117 @@ export default function CreateUnitModalComponent() {
...
+		setRate(String(defaultValues.secInRate));
+		setShowCustomInput(false);
+	};
 	// Unlike edit, we decided to discard inputs when you choose to leave the page. The reasoning is
 	// that create starts from an empty template.
 
-	// Submit
-	const handleSubmit = () => {
+	// Save
+	const handleSave = () => {
```
huss: While this new name is okay it differs from the one in edit unit of ``handleSaveChanges``. It would be nice if they were consistent.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -40,68 +42,117 @@ export default function CreateUnitModalComponent() {
...
-		state.identifier = (!state.identifier || state.identifier.length === 0) ? state.name : state.identifier;
-		// set displayable to none if unit is meter
-		if (state.typeOfUnit == UnitType.meter && state.displayable != DisplayableType.none) {
-			state.displayable = DisplayableType.none;
-		}
-		// set unit to suffix if suffix is not empty
-		if (state.typeOfUnit != UnitType.suffix && state.suffix != '') {
-			state.typeOfUnit = UnitType.suffix;
-		}
+		const submitState = {
```
huss, line 161: While the change in how state is set here is nice, edit unit differs. This change should probably go there too. Note identifier is set here and done after save in edit so make consistent. I think I did this so it probably isn't your doing. If you prefer, I can make the change.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -111,180 +162,274 @@ export default function CreateUnitModalComponent() {
...
-						{/* Suffix input */}
-						<Col><FormGroup>
-							<Label for='suffix'>{translate('unit.suffix')}</Label>
+				<ModalBody>
+					<Container>
+						<Row xs="1" lg="2">
+							{/* Identifier input */}
+							<Col>
+								<FormGroup>
+									<Label for="identifier">{translate('identifier')}</Label>
```
huss: While OED is far from consistent in using single quotes over double quotes, the desire was for single quote when possible. I'm noting for the record and you don't have to change them back. At some point OED should systematically do this.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -59,8 +65,63 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
 	};
 
-	const handleNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-		setState({ ...state, [e.target.name]: Number(e.target.value) });
+	/**
+	 * Determines if the rate is custom.
+	 * @param rate The rate to check
+	 * @returns true if the rate is custom and false if it is a standard value.
+	 */
+	const isCustomRate = (rate: number) => {
```
huss: Crreate unit does not check for custom this way and it uses different names for some of the functions that follow. I'm wondering why create is not the same as edit. Since I made many of these changes I probably think the names are better (it may be my bias) but I do think they are more appropriate given the specific function. Do you have thoughts on this and which of the two methods might be best shared?

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -40,68 +42,117 @@ export default function CreateUnitModalComponent() {
 		// The units API expects these values to be undefined on call so that the database can assign their values.
 		id: -99
 	};
-
 	/* State */
 	// Unlike EditUnitModalComponent, there are no props so we don't pass show and close via props.
 	// Modal show
+	const CUSTOM_INPUT = '-99';
```
huss: Edit unit has near top of function. I think that is a better place for this type of const.

Also, I'm noticing now that -99 is used for the id and default graphic unit when no value. Given this, I think edit and create should use some other value for custom input. Maybe -77 (-88 looks too much like -99)? I'm hoping this is easy to do as it is a const.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -132,17 +193,33 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
-		setValidUnit(state.name !== '' && state.secInRate > 0 &&
-			(state.typeOfUnit !== UnitType.suffix || state.suffix !== ''));
-	}, [state.name, state.secInRate, state.typeOfUnit, state.suffix]);
+		// This checks:
+		// - Name cannot be blank
+		// - If type of unit is suffix their must be a suffix
+		// - The rate is set so not the custom input value. This happens if select custom value but don't input with enter.
+		// - The custom rate is a positive integer
+		const validUnit = state.name !== '' &&
+			(state.typeOfUnit !== UnitType.suffix || state.suffix !== '') && state.secInRate !== Number(CUSTOM_INPUT)
```
huss, line 200: This checks for not hitting enter on custom but create does not. This is kind to the user and nice to have.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -40,68 +42,117 @@ export default function CreateUnitModalComponent() {
...
 	/* End State */
 
 	// Reset the state to default values
+	// To be used for the discard changes and save button
 	const resetState = () => {
 		setState(defaultValues);
+		resetCustomRate();
 	};
 
+	const handleShow = () => setShowModal(true);
```
huss: Not your doing but could you add braces around the one function line of code? OED prefers them in all cases (see edit).

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -111,180 +162,274 @@ export default function CreateUnitModalComponent() {
...
+							</Col>
+							{/* Suffix input */}
+							<Col>
+								<FormGroup>
+									<Label for="suffix">{translate('unit.suffix')}</Label>
+									<Input
+										id="suffix"
+										name="suffix"
+										type="text"
+										autoComplete="off"
```
huss: This isn't in edit. Should it be?

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -279,157 +350,201 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+										invalid={state.typeOfUnit === UnitType.suffix && state.suffix === ''} />
+									<FormFeedback>
+										<FormattedMessage id="error.required" />
+									</FormFeedback>
+								</FormGroup>
+							</Col>
+						</Row>
+						{/* Note input */}
+						<FormGroup>
+							<Label for='note'>{translate('unit')}</Label>
```
huss: The translate should be to note (as in create).

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -279,157 +350,201 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
-							value={state.note}
-							placeholder='Note'
-							onChange={e => handleStringChange(e)} />
-					</FormGroup>
-				</Container></ModalBody>
+								id='note'
+								name='note'
+								type='textarea'
+								value={state.note}
+								placeholder='Note'
```
huss: There are several placeholder values in edit but not create. I think I prefer them to be gone so could you do that?

huss (changes requested):
I want to thank @gyordong, @aduques, @AchsahJojo & Emily Sarne for this contribution. Is Emily Sarne the GitHub user @emilysarne? I want to verify they have signed the CLA if they contributed to this work.

Testing found this works as desired. Most of my comments relate to differences between the code for edit and create for unit. I welcome your thoughts on my comments and you can address if you think a comment is fine. Please let me know if I can help in any way.

src/client/app/translations/data.ts
gyordong: done as requested

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: readded a line as requested

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: set back to single quotes as requested

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: added braces as requested

src/client/app/components/unit/EditUnitModalComponent.tsx
gyordong: removed as requested

src/client/app/components/unit/EditUnitModalComponent.tsx
gyordong: changed as requested

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: changed as requested to -77.

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: refactored to handleSaveChanges() as requested.

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: Not sure, since suffix is a unit type I think autocomplete might serve as quality of life for users who don't want to have to retype for meters that share the same unit. Then again, I'm not familiar with the purpose of suffix. I would love some additional info on how it works so I can give more informed feedback.

src/client/app/translations/data.ts
gyordong: done as requested.

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: set to only affect secInRate instead of a general target as requested

src/client/app/components/unit/EditUnitModalComponent.tsx
gyordong: I know you asked me to remove placeholders, but removing the placeholder='Identifier' in editunitmodalcomponent.tsx caused a testcase to break.
Should I just leave it in or keep the placeholder until someone fixes the testcase as TODO?
the test in question was: 
1) Test CSV Pipeline
         Testing files starting 'pipe3' doing 'Cumulative time readings' with 1 requests:

huss: @gyordong I thought this was strange and I looked at the code. The test code should be making a request to the server and not be impacted by the client code. Also, removing the placeholder text, I would think, would not impact the actual values on the client. I looked at this PR code on my machine and I see the commit where you removed the placeholder in EditUnitModalComponent. When I run the tests in src/server/test/web/csvPipelineTest.js they all pass. I even ran all the tests and they all pass. Thus, I'm not sure I am seeing what you see. Do you know what is up since I'm now confused.

gyordong: It passes now as well, even if with placeholder removed. May have just been a glitch on my local machine.

src/client/app/components/unit/EditUnitModalComponent.tsx
emilysarne: Should we consider aligning the implementation of isCustom in the Create file with isCustomRate in Edit for consistency? This approach is also more efficient, as isCustomRate uses predefined rates with Object.entries. If not, could you clarify the distinction in their expected behavior to ensure we're addressing both cases as intended?

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: Suffix is a string that is the end of a special type of unit's name. It seems unlikely that it would be reused or match a standard used string. I think it is probably best not to allow autocomplete in this case.

src/client/app/components/unit/EditUnitModalComponent.tsx
huss: I agree. I was pointing out the difference and, not very clearly, indicating my preference toward doing it the way edit unit did it. I think they should use similar techniques. I was just indicating I was open to which ones if you felt there was a reason to do it another way from edit.

src/client/app/components/unit/CreateUnitModalComponent.tsx
gyordong: removed as requested

src/client/app/components/unit/EditUnitModalComponent.tsx
aduques: After looking at comments on the CreateUnitModalComponent, I noticed that the design of the current implementation was to discard all values when leaving the page without saving. Given this, the custom rate would be discarded alongside the other values despite changing the similar line of code in CreateModalComponent. Should CreateUnitModalComponent be changed to still keep unsaved changes similarly to how it is implemented in EditUnitModalComponent?

src/client/app/components/unit/CreateUnitModalComponent.tsx
aduques: changed setting state in EditUnitModalComponent to be consistent with CreateUnitModalComponent

src/client/app/components/unit/EditUnitModalComponent.tsx
huss: I thought this test was to see if the user had started a custom input but not hit enter. This likely indicates they started to enter a value but never finalized it with an enter. This seems a potentially common mistake that OED is trying to catch to avoid. I didn't think it has to do with leaving the page and resetting state but maybe I'm missing something. What do you think?

gyordong: Hi Steve, I changed the code to check for enter. The only caveat is that I made it purposely allow users to save without clicking enter if they didn't change the custom value from the standard values automatically inputted when clicked (ex. day to custom value, box shows 86400, create unit can be clicked). Let me know if this doesn't seem ideal and I'll change it to require enter to create a unit under all circumstances.

huss: Sorry I missed this comment until now. I'm not sure which is best. It is unclear why the admin selected custom if they did not intend to change the value. However, I'm okay with this as long as both create and edit work the same.

gyordong: I think the difference won't be noticeable for the user. Should I change it to require a change of value or is it fine as is?

huss: I'm not sure if I understand but it is okay to use the value without enter if it is the standard value as long as both create and edit do the same thing. Is this what you were asking?

gyordong: Yes. Do you have a preference between the two implementations?

huss: I don't care too much but I might pick that if they do custom then they have to hit enter. One case I am thinking about is if they change the value away from a standard one and then put it back without hitting enter then would the code accept this second value because it is standard? This type of issue is why I'm leaning toward requiring enter for all custom values but maybe it isn't an issue.

gyordong: Okay, I've edited it so it requires both that the custom value isn't a standard rate and that enter is pressed. I believe this makes it line up with edit.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -45,63 +48,129 @@ export default function CreateUnitModalComponent() {
...
-
 	const handleBooleanChange = (e: React.ChangeEvent<HTMLInputElement>) => {
 		setState({ ...state, [e.target.name]: JSON.parse(e.target.value) });
 	};
-
-	const handleNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-		setState({ ...state, [e.target.name]: Number(e.target.value) });
+	const [canSave, setCanSave] = useState(false);
+	// Sets the starting rate for secInRate box, value of 3600 is chosen as default to result in Hour as default in dropdown box.
+	const isCustomRate = (rate: number) => {
```
huss: The comment on the line above belongs with the line after this function. Also, please copy the comment in edit about this function to be the comment for it.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -45,63 +48,129 @@ export default function CreateUnitModalComponent() {
...
+	UI events:
+		- When the user selects a new rate from the dropdown,`rate` is updated.
+		- If the user selects the custom value option, `showCustomInput` is set to true.
+		- When the user enters a custom value, `customRate` is updated.
+		- The initial value of `customRate` is set to the previously chosen value of `rate`
+		- Make sure that when submit button is clicked, that the state.secInRate is set to the correct value.
+  */
+	const handleStandardNumberChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		const { value } = e.target;
+		const isCustom = isCustomRate(Number(value));
```
huss: isCustom is used in both parts of the if statement below in the same way. Given this, I would move this line right after this line and before the if (and remove from the two parts of the if statement).

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -111,180 +180,273 @@ export default function CreateUnitModalComponent() {
...
+						<Row xs="1" lg="2">
+							{/* Seconds in rate input */}
+							<Col>
+								<FormGroup>
+									<Label for="secInRate">{translate('unit.sec.in.rate')}</Label>
+									<Input
+										id="secInRate"
+										name="secInRate"
+										type="select"
+										onChange={e => handleStandardNumberChange(e)}
```
huss: To be consistent with edit, I think this should be handleRateChange.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: I've done this change.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: This is gone given my other changes.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: Done with my edits.

huss (approved):
@gyordong The changes you made looked good and worked in almost every case. Since it was complex to note all my ideas and to move this forward, I pushed a commit that aligned unit edit and create. With this, I think this PR is now ready to merge. I'll let it sit for a little while to give everyone involved a chance to look at what I did and comment if they have nay concerns. If nothing needs changing then it will be merged as is. Congratulations on completing this work and I appreciate your efforts to do address all the comments given changes over time.

PR #1327, opened 2024-08-07
Initial LogMsg Database Model

ianLewis8:
# Description

The current commits I have done include the backbone for a database model that store the Log Messages generated by OED. These can then be queried and displayed to an admin user.

Partly Addresses #712 

## Type of change

- [x] Note merging this changes the database configuration.
- [x] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.github.io/developer/pr.html) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.github.io/developer/cla.html) and each author is listed in the Description section.

## Limitations

I am still working on the route for this database but there is also much more to be done. It needs to be only accessible by admins, the actual table could use indexing for faster lookup speed, etc.

PR #1390, opened 2024-11-20
Log message web page

### LINKED ISSUE #712: dealing with log files

OED wants non-IT people to be able to manage a site. While we can send daily email to the admin with all issues in the log file, it would be nice to allow admin to look at the log file on a web page.

We need to decide how to specify date ranges for the admin to look at. This may mean thinking about how we do logging and if we should adopt a standard package. We also should think through the fact that log files can get very large over time.

None.

### Additional context

None.

nmqng:
# Description

Continue [ianLewis8](https://github.com/ianLewis8)'s work on Draft Pull Request #1327.
The purpose of this pull request is to create the web page that allows admin to look at the log file on a web page.

Fixes #712

## Type of change

## Checklist


src/client/app/components/RouteComponent.tsx
```diff
@@ -58,7 +59,10 @@ const router = createBrowserRouter([
 					{ path: 'csvMeters', element: <MetersCSVUploadComponent /> },
 					{ path: 'maps', element: <MapsDetailContainer /> },
 					{ path: 'units', element: <UnitsDetailComponent /> },
-					{ path: 'users', element: <UsersDetailComponent /> }
+					{ path: 'conversions', element: <ConversionsDetailComponent /> },
```
huss: I think the conversions route was accidentally added twice.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+import { dateRangeToTimeInterval, timeIntervalToDateRange } from '../../utils/dateRangeCompatibility';
+import { Value } from '@wojtekmaj/react-daterange-picker/dist/cjs/shared/types';
+
+
+// number of log messages to display per page
+const PER_PAGE = 20;
+// initialize log message array to hold log messages
+const initialLogs: any[] = [];
+// log types for filtering
+const logTypes = ['ERROR', 'INFO', 'WARN', 'SILENT'];
```
huss: Normally OED uses an enum in a common place for items like this. I think it may be a good idea. Having said that, I notice src/server/log.js does not. Also, both server files (including src/server/sql/logmsg/create_log_types_enum.sql) have DEBUG as a type but this does not.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+	// Open modal with the full log message
+	const handleLogMessageClick = (logMessage: string) => {
+		setModalLogMessage(logMessage);
+		setModalOpen(true);
+	};
+
+	// Handle checkbox change for log type in the table
+	const handleTableCheckboxChange = (logType: string) => {
+		if (selectedTableLogTypes.includes(logType)) { // Remove log type if already selected
```
huss: OED convention is that comments be on a separate line. Normally it is above the line of code but with if-type statements it often goes inside the braces below that initial line.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	const paginatedLogs = showAllLogs ? filteredLogs : filteredLogs.slice((currentPage - 1) * PER_PAGE, currentPage * PER_PAGE);
+	const totalPages = Math.ceil(filteredLogs.length / PER_PAGE);
+
+	/**
+	 * Handle showing the log table by fetching from the server
+	 */
+	async function handleShowLogTable() {
+		// Number of logs being fetched must be between 1 and 1000
+		if (!logLimit || logLimit < 1 || logLimit > 1000) {
+			showWarnNotification(translate('log.limit.required'));
```
huss: This works fine but delays the msg until update is hit. In most forms with input OED uses invalid/FormFeedback to give immediate feedback. I'm not certain it will align correctly with the horizontal layout but it would be nice to use if possible.

src/client/app/translations/data.ts
```diff
@@ -516,7 +516,18 @@ const LocaleTranslationData = {
 		"week": "Week",
 		"yes": "yes",
 		"yesterday": "Yesterday",
-		"you.cannot.create.a.cyclic.group": "You cannot create a cyclic group"
+		"you.cannot.create.a.cyclic.group": "You cannot create a cyclic group",
+		"log.messages": "Log Messages",
```
huss: These should be alphabetically placed within each language. To keep together, the log specific ones might be best to start with "log.". On the other hand, "select.all", for example, is probably best under select alphabetically.

src/client/app/translations/data.ts
```diff
@@ -516,7 +516,18 @@ const LocaleTranslationData = {
...
+		"log.message": "Log Message",
+		"log.type": "Log Type",
+		"log.time": "Log Time",
+		"num.logs.display": "Number of logs to display",
+		"log.msg.update": "Update",
```
huss: The string "update" is already above. There is a difference in whether capitalized but it might be good to use one or the other.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	// Full log message to display in the modal
+	const [modalLogMessage, setModalLogMessage] = React.useState('');
+	// Showing all logs instead of paginated
+	const [showAllLogs, setShowAllLogs] = React.useState(false);
+	// Update button state
+	const [buttonAvailable, setButtonAvailable] = React.useState(false);
+
+	// Update the availability of the update button each time the selected log types, log limit, or date range changes
+	React.useEffect(() => {
+		setButtonAvailable(false);
```
huss, line 80: I'm uncertain in when the button is disabled as I could not easily get it not to show. It can be clicked even if invalid settings are detected. It would be good if that was not the case. Thus, I'm unsure about its dependency on change. Note may relate to when msgs are shown.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+// initialize log message array to hold log messages
+const initialLogs: any[] = [];
+// log types for filtering
+const logTypes = ['ERROR', 'INFO', 'WARN', 'SILENT'];
+
+/**
+ * React component that defines the log message page
+ * @returns LogMsgComponent element
+ */
+export default function LogMsgComponent() {
```
huss, line 43: I did not debug why this is happening but if you unselect all log types in the initial request then you get no logs (as expected). If you do it after having received logs to display then they still remain. I think it should be consistent. Seemingly okay options:
  1. Force admin to select at least one log type. This may be best as it stops silly input.
  2. Update log msgs if none selected.

src/client/app/translations/data.ts
```diff
@@ -516,7 +516,18 @@ const LocaleTranslationData = {
...
+		"from.1.to.1000": "from 1 to 1000",
+		"show.logs": "Show Log Table",
```
huss: I could not find where this key is used in the code (all 3 languages).

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+import * as React from 'react';
```
huss, line 7: OED normally has a blank line after the license comment.

src/server/sql/logmsg/create_logmsg_table.sql
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+--create logmsg table
+CREATE TABLE IF NOT EXISTS logmsg (
```
huss, line 6: This needs to be added to the 1.0.0-2.0.0 migration.

src/server/sql/logmsg/create_log_types_enum.sql
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+DO $$ BEGIN
```
huss, line 5: This needs to be added to the 1.0.0-2.0.0 migration.

src/server/sql/logmsg/get_all_logs.sql
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: I think we probably should rename this file because the same name is used in Logemail. Maybe call get_all_logmsg.sql. To be consistent (and to fix another name overlap), I think all the log should become logmsg in all sql files. The plural can rename, e.g., get_logmsgs_from_type.sql. The number done might change if some are removed.

This sql file along with get_logs_from_dates & get_logs_from_type are not used along with the associated model functions. They are covered by the more general get_logs_from_dates_and_type.sql. I propose we remove all three for the following reasons:

- We don't want to allow getting all msgs without a size limit.
- This is not a common admin request and the small difference in time that might be gained by the more restrictive calls does not seem valuable. If they did stay then they would need to be renamed and the limit added.
- Having calls we will likely never use only adds to maintenance and potential issues.

I know the recent changes created this situation where older functions should now be removed but I think it is best unless you feel otherwise.

src/server/routes/logs.js
```diff
@@ -8,6 +8,9 @@ const express = require('express');
 const { log } = require('../log');
 const validate = require('jsonschema').validate;
 const adminAuthenticator = require('./authenticator').adminAuthMiddleware;
+const LogMsg = require('../models/LogMsg');
```
huss, line 11: This is not a new situation but all these routes for logs should have test cases. I propose either doing it or adding an issue to get it done.

src/server/log.js
```diff
@@ -71,11 +73,49 @@ class Logger {
...
-			fs.appendFile(logFile, messageToLog, err => {
+			fs.appendFile(logFile, messageToLog, async err => {
 				if (err) {
 					console.error(`Failed to write to log file: ${err} (${err.stack})`); // tslint:disable-line no-console
 				}
 			});
+
+			// NOTE: for running the first time, uncomment the following code block to write all current logs to the database
+			// after that you can comment out it to write only new logs to the database
+			// TODO: This should be fix later to check whehter it should write new logs or all logs to the database when running
```
huss: whethter -> whether

src/server/log.js
```diff
@@ -71,11 +73,49 @@ class Logger {
...
 		// Always log to the logfile.
 		if (this.logToFile) {
-			fs.appendFile(logFile, messageToLog, err => {
+			fs.appendFile(logFile, messageToLog, async err => {
 				if (err) {
 					console.error(`Failed to write to log file: ${err} (${err.stack})`); // tslint:disable-line no-console
 				}
 			});
+
+			// NOTE: for running the first time, uncomment the following code block to write all current logs to the database
```
huss: What about making this a separate function (could be in a separate place?) that then has an npm command in package.json to migrate the log.txt file to the DB? Then sites (or developers) could be told to run it once to do this.

src/server/log.js
```diff
@@ -71,11 +73,49 @@ class Logger {
...
+			// 			}
+			// 		}
+			// 	}
+			// })
+
+
+			// Comment out the following code block when running the first time to write all logs to the database
+			// then uncomment it to write only new logs to the database later
+			// Write the new log to the database
+			const logMsg = new LogMsg(level.name, message, new Date(logTime));
```
huss: I think this should use logTime for two reasons. First, it means the text file and the DB have the same date/time stamp. Second, it uses moment that OED uses everywhere rather than Date. The type of the parameter  needs to be resolved or the type of the variable changed.

src/server/routes/logs.js
```diff
@@ -55,4 +79,24 @@ router.post('/error', async (req, res) => {
 	}
 });
 
+
+router.get('/logsmsg/getLogsByDateRangeAndType', async (req, res) => {
+	const validationResult = validate(req.query, validLogMsg);
+	if (!validationResult.valid) {
+		res.sendStatus(400);
```
huss, line 87: It would be good to log the error as done in other ones in the file for validation failure.

src/server/routes/logs.js
```diff
@@ -55,4 +79,24 @@ router.post('/error', async (req, res) => {
...
+	} else {
+		const conn = getConnection();
+		try {
+			const logLimit = parseInt(req.query.logLimit);
+			const timeInterval = TimeInterval.fromString(req.query.timeInterval);
+			const rows = await LogMsg.getLogsByDateRangeAndType(
+				timeInterval.startTimestamp, timeInterval.endTimestamp, req.query.logTypes.split('-'), logLimit, conn);
```
huss: Since the other arguments are prepared before the call should logTypes be too? Basically, create a variable to hold these values.

src/server/routes/logs.js
```diff
@@ -22,6 +25,27 @@ const validLog = {
 		}
 	}
 }
+
+const validLogMsg = {
+	type: 'object',
+	required: ['timeInterval', 'logTypes', 'logLimit'],
```
huss, line 31: Maybe add ``maxProperties: 3`` so know only the required ones are given.

src/server/models/LogMsg.js
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	/**
+	 * Returns a promise to get all of the logs from the database
+	 * @param conn the connection to be used.
+	 * @returns {Promise.<array.<logMsg>>}
+	 */
+	static async getAll(conn) {
+		const rows = await conn.any(sqlFile('logmsg/get_all_logs.sql'));
+		if (rows.length > 0) {
```
huss: This checks the returned length but the others do not. Is it necessary as I don't see it used in other files? It does not seem best since the return value is not set.

src/server/models/LogMsg.js
```diff
@@ -0,0 +1,120 @@
...
+	/**
+	 * Returns a promise to get all of the logs in between two dates.
+	 * @param {Date} startDate start date of the range to get logs
+	 * @param {Date} endDate end date of the range to get logs
+	 * @param {Array<string>} logTypes array of log types to get logs
+	 * @param {Number} logLimit the maximum number of logs to return
+	 * @param conn is the connection to use.
+	 * @returns {Promise.<array.<LogMsg>>}
+	 */
+	static async getLogsByDateRangeAndType(startDate = null, endDate = null, logTypes, logLimit = 100, conn) {
```
huss, line 71: null may be file but I'm wondering if the more common undefined would work?

src/client/app/utils/api/LogsApi.ts
```diff
@@ -25,4 +26,11 @@ export default class LogsApi {
 	public async error(log: LogData): Promise<void> {
 		return await this.backend.doPostRequest('/api/logs/error', log);
 	}
+
+	public async getLogsByDateRangeAndType(timeInterval: TimeInterval, logTypes: string[], logLimit: string): Promise<LogData[]> {
+		const request = await this.backend.doGetRequest('/api/logs/logsmsg/getLogsByDateRangeAndType',
+			{ timeInterval: timeInterval.toString(), logTypes: logTypes.join('-'), logLimit: logLimit });
```
huss: src/server/routes/groups.js seems to allow the passing of an array of items directly for childGroups & childMeters. Could something like that be used here to avoid the join (and then split in route)?

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+					<tbody style={bodyStyle}>
+						{paginatedLogs.map((log, index) => (
+							<tr key={index + 1}>
+								<td>{log.logType}</td>
+								<td
+									style={{ cursor: 'pointer' }}
+									onClick={() => handleLogMessageClick(log.logMessage)}
+								>{log.logMessage.length > 80 ? `${log.logMessage.slice(0, 80)} ...` : log.logMessage}</td>
+								<td>
+									{new Date(log.logTime).toLocaleString('en-US', {
```
huss: OED commonly uses moment for date and internationalization. Could that be used here?

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+
+	// Log messages state
+	const [logs, setLogs] = React.useState(initialLogs);
+	// Selected log types for filtering in the update log
+	const [selectedUpdateLogTypes, setSelectedUpdateLogTypes] = React.useState<string[]>(logTypes);
+	// Dropdown open state for log type in the header for filtering
+	const [typeTableDropdown, setTypeTableDropdown] = React.useState(false);
+	// Selected log types for filtering in the table
+	const [selectedTableLogTypes, setSelectedTableLogTypes] = React.useState<string[]>(logTypes);
+	// Dropdown open state for log type in the table for filtering
```
huss: This comment says table but the state name does not have table. Also, group related near each other if not already that way.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	const [updateLogDropdown, setUpdateLogDropdown] = React.useState(false);
+	// Sort order for date column in the table
+	const [dateSortOrder, setDateSortOrder] = React.useState<'asc' | 'desc'>('asc');
+	// Log messages date range state
+	const [logDateRange, setLogDateRange] = React.useState<TimeInterval>(TimeInterval.unbounded());
+	// Number of log messages to display
+	const [logLimit, setLogLimit] = React.useState(0);
```
huss: Do you think it is best to force the user to set the number (as here since 0) or put in some modest value (say 50?) and then the user can change? Not sure but wondering which is better.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	const [buttonAvailable, setButtonAvailable] = React.useState(false);
+
+	// Update the availability of the update button each time the selected log types, log limit, or date range changes
+	React.useEffect(() => {
+		setButtonAvailable(false);
+	}, [selectedUpdateLogTypes, logLimit, logDateRange]);
+
+	// Open modal with the full log message
+	const handleLogMessageClick = (logMessage: string) => {
+		setModalLogMessage(logMessage);
```
huss, line 86: Would it be easy to add the type and date/time to this so it is completely clear which msg popped up? It would be really nice if the popped up msg had a colored background but not necessary.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	const handleUpdateCheckboxChange = (logType: string) => {
+		if (selectedUpdateLogTypes.includes(logType)) {  // Remove log type if already selected
+			setSelectedUpdateLogTypes(selectedUpdateLogTypes.filter(type => type !== logType));
+		} else { // Add log type if not selected
+			setSelectedUpdateLogTypes([...selectedUpdateLogTypes, logType]);
+		}
+	};
+
+	// React effect to keep track of the "Select All" checkbox state
+	React.useEffect(() => {
```
huss, line 113: Since the update and table ones are done separately above should this be two separate useEffects? Not a big deal but I don't think both updates can happen.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	};
+	// Handle "Select All" checkbox change in the update log
+	const handleUpdateSelectAll = () => {
+		selectAllUpdate ? setSelectedUpdateLogTypes([]) : setSelectedUpdateLogTypes(logTypes);
+		setSelectAllUpdate(!selectAllUpdate);
+	};
+	// Handle sorting of logs by date
+	const handleDateSort = () => {
+		const newDateSortOrder = dateSortOrder === 'asc' ? 'desc' : 'asc';
+		const sortedLogs = [...logs].sort((a, b) => {
```
huss: Two thoughts. First, the use of Date vs moment. Second, while sort is used in some places so is orderBy (from Lodash). I think it may be able to deal with the order via parameter and also sort moment objects correctly.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,375 @@
...
+	const toggleUpdateLog = () => {
+		setUpdateLogDropdown(!updateLogDropdown);
+	};
+	// Handle showing all logs instead of paginated
+	const handleShowAllLogs = () => {
+		setShowAllLogs(!showAllLogs);
+	};
+
+	// Filter logs based on selected log types and date range
+	const filteredLogs = logs.filter(log => {
```
huss: I'm finding the dual use of the date picker for the update and display to be a little confusing. For type and number to display the update area does not impact the logs shown until update is hit. For the date picker it does. I think that difference may confuse users. If you agree then I'm okay with removing this filtering (since can update the logs quickly) or adding it to the log table area (maybe in the Log Time column header?) as a separate date picker.

huss (changes requested):
Thanks to @nmqng for addressing another issue. Overall, this works well and the UI is very nice. I've made some comment to consider. Also, there is now a merge conflict so let me know if you want me to deal with it (or you can). Let me know if anything is not clear.

nmqng:
> Thanks to @nmqng for addressing another issue. Overall, this works well and the UI is very nice. I've made some comment to consider. Also, there is now a merge conflict so let me know if you want me to deal with it (or you can). Let me know if anything is not clear.

Hi Steve, I think I can try to fix the merge conflict myself and I'll let you know if I need any help. I will have a look at your comments for this issue after I finish the compareLine one, which is about half completed.

src/server/models/LogMsg.js
nmqng: I use null here for consistency with the /src/server/models/Reading.js. However, I can change it to undefined if you want since it works fine

huss: Lets leave it as null. Thanks for looking at this.

src/client/app/components/admin/LogMsgComponent.tsx
nmqng: This filteredLogs was previously used for log time after the data is fetched. I forgot to remove it after our discussion about the UI of the page.

I think I will remove it since currently the behavior of Log Time column header is sorting the date after every click. I tried to add the filter in the column header but then every time I clicked on the date range picker or picked a date range it would active the sorting again. Please let me know if later on you'd like to have the filter back.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,361 @@
...
+	const [selectAllTable, setSelectAllTable] = React.useState(true);
+
+	// Update the availability of the update button each time the selected log types, log limit, or date range changes
+	React.useEffect(() => {
+		setButtonAvailable(false);
+	}, [selectedUpdateLogTypes, logLimit, logDateRange]);
+
+	// Open modal with the full log message
+	const handleLogMessageModal = (logType: string, logTime: string, logMessage: string) => {
+		setModelHeader(`[${logType}] ${moment(logTime).toLocaleString()}`);
```
huss: This did not internationalize the date for me. Also, it was shifting to local time rather than showing the server time that logged the error. I changed to ``${moment.parseZone(logTime).format('LL LTS [(and ]SSS[ms)]')`` and it worked as desired. Note this adds the ms here for detail if clicked on. I added at end since internationalization of the data/time moved the time around so I could not find a better solution. The moment import needs to be changed to ``import * as moment from 'moment-timezone';``

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,361 @@
...
+						{paginatedLogs.map((log, index) => (
+							<tr key={index + 1}>
+								<td>{log.logType}</td>
+								<td
+									style={{ cursor: 'pointer' }}
+									onClick={() => handleLogMessageModal(log.logType, log.logTime, log.logMessage)}
+								>
+									{log.logMessage.length > 80 ? `${log.logMessage.slice(0, 80)} ...` : log.logMessage}
+								</td>
+								<td>{moment(log.logTime).format('MM/DD/YYYY, hh:mm:ss.SS A')}</td>
```
huss: I think this date/time should be internationalized/time zone correct per other comment but with ``${moment.parseZone(logTime).format('LL LTS')}`` since no ms here.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,361 @@
...
+					</PaginationItem><PaginationItem disabled={currentPage === totalPages}>
+						<PaginationLink last onClick={() => handlePageChange(totalPages)} />
+					</PaginationItem>
+				</>
+			</Pagination >}
+
+			{/* Show all logs or in pages button */}
+			{logs.length > 0 &&
+				<Button color='primary' style={{ margin: '0% 40% 1%' }} onClick={() => setShowAllLogs(!showAllLogs)}>
+					{!showAllLogs ? `Show All Logs (${logs.length})` : 'Show in pages'}
```
huss: This should be translated.

src/server/routes/logs.js
```diff
@@ -55,4 +83,24 @@ router.post('/error', async (req, res) => {
...
+		const conn = getConnection();
+		try {
+			const logLimit = parseInt(req.query.logLimit);
+			const timeInterval = TimeInterval.fromString(req.query.timeInterval);
+			const logTypes = req.query.logTypes;
+			const rows = await LogMsg.getLogsByDateRangeAndType(
+				timeInterval.startTimestamp, timeInterval.endTimestamp, logTypes, logLimit, conn);
+			res.json(rows);
+		} catch (err) {
+			console.error(`Failed to fetch logs filter by date range and type: ${err}`);
```
huss: This should be log.error.

src/server/routes/logs.js
huss: Above this line lets put ``log.error('invalid request to getLogsByDateRangeAndType');``.

src/server/services/addLogMsg.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+const fs = require('fs');
+const logFile = require('../config').logFile;
+const LogMsg = require('../models/LogMsg');
+const { getConnection } = require('../db');
+const moment = require('moment');
+
+const addLogMsgToDB = async () => {
+    try {
```
huss, line 14: This did not work for me so I made these changes and it seems to work:

Screenshot 2024-12-07 081058

src/server/services/addLogMsg.js
```diff
@@ -0,0 +1,37 @@
...
+        const data = await fs.promises.readFile(logFile, 'utf8');
+        const logEntries = data.split('\n').filter(entry => entry.trim() !== '');
+        const conn = getConnection();
+
+        for (const entry of logEntries) {
+            const logParts = entry.match(/\[(.*?)@(.*?)\] (.*)/);
```
huss: I processed a file with almost 140k log messages. This did work which is great. I noticed that some message were truncated such as this one:

```
[ERROR@2024-04-06T16:03:30.167+00:00] Error while inserting new meter with detail "Key (name)=(a1) already exists."
Stacktrace: 
error: duplicate key value violates unique constraint "meters_name_key"
    at Parser.parseErrorMessage (/usr/src/app/node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (/usr/src/app/node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (/usr/src/app/node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (/usr/src/app/node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:514:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

I think the issue is that it is multiline as only the first line is shown and that is what is in the DB. Is it possible to easily fix this? I can provide more samples if needed.

src/client/app/components/admin/LogMsgComponent.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+// initialize log message array to hold log messages
+const initialLogs: any[] = [];
+
+/**
+ * React component that defines the log message page
+ * @returns LogMsgComponent element
+ */
+export default function LogMsgComponent() {
+	const locale = useAppSelector(selectSelectedLanguage);
```
huss, line 44: Unlike other pages, when you switch languages it changes the page to the default values without any logs. I'm not sure why and maybe this indicates some issue in how values are set. It is not a big deal but maybe look at a little to try to figure out.

src/server/models/LogMsg.js
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Returns a promise to insert this log into the database
+	 * @param conn the database connection to use
+	 * @returns {Promise.<>}
+	 */
+	async insert(conn) {
+		const logMsg = this;
+		await conn.none(sqlFile('logmsg/insert_new_logmsg.sql'), logMsg);
+	}
```
huss, line 60: The request to switch to moment lost the fractional sections on the value put in the DB. To fix I modified to: ``sqlFile('logmsg/insert_new_logmsg.sql'), { logType: logMsg.logType, logMessage: logMsg.logMessage, logTime: logMsg.logTime.format('YYYY-MM-DDTHH:mm:ss.SSS') }``

huss (changes requested):
Thanks to @nmqng for all the updates. This is looking good. I made a few comments to consider.

src/client/app/components/admin/LogMsgComponent.tsx
nmqng: Changing the language may reset the local change, which reset the log table. As we discussed, this comment will be marked as resolved.

huss (approved):
Thanks to @nmqng for addressing all the comments. Review and testing found it works as desired. Congratulations on your first accepted contribution to OED.

PR #1391, opened 2024-11-25
Obivus Modes Testing, logfile uploads

### LINKED ISSUE #673: further Obvius testing

We should have tests to verify that Obvius uploads store the correct information in the database.

src/server/test/db/configfileTests.js makes sure the parsing of an Obvius configuration file is correct. However, it does not verify that an actual upload puts the correct data into the database. It should create a meter with the correct name, type, note, etc.

src/server/test/db/loadLogfileToReadingsTests.js similarly checks the parsing and some DB info but not that the loaded readings are correct.

Manually testing was recently done but this is not the correct long-term solution. We missed a bug in OED from a lack of automated tests.

### Additional context

@huss can help a developer understand what should happen. Looking at the database after an upload will also show what is expected.

BrendenHaskins:
# Description

Adds regression testing to all currently implemented modes of the obvius model. Also addresses an issue with logfile uploads surrounding promise handling.
Contributions by: @lukeckk, @RaiuXL, @huss

Fixes #673

## Type of change

## Checklist

## Limitations

These tests do not cover unimplemented modes, such as file downloads from the model, or readings retrieval.


src/server/test/web/obviusTest.js
```diff
@@ -58,5 +62,210 @@ mocha.describe('Obvius API', () => {
...
+					.field('modbusdevice', '1234')
+					.attach('files', configFilePath);
+
+				const res = await chai.request(app)
+					.post('/api/obvius')
+					.field('username', obviusUser.username)
+					.field('password', obviusUser.password)
+					.field('mode', manifestRequestMode);
+
+				//logfile upload should respond with 200, success
```
huss: I might put this between the two requests so it is right after the config file upload since if that is not successful then the second request may cause issues.

src/server/test/web/obviusTest.js
```diff
@@ -58,5 +62,210 @@ mocha.describe('Obvius API', () => {
...
+					if (expMeterNames[i] != meterNames[i]) {
+						expectedNamesAreEqual = false;
+					}
+
+					if (expMeterIDs[i] != meterIDs[i]) {
+						expectedIDsAreEqual = false;
+					}
+				}
+
+				//assertion for type, displayable, and enabled
```
huss, line 273: While these tests work, they only tell if any meter failed and do not indicate which meter had the issue. How about doing a console.log in the loop above to print the name of the meter and the issue so it is recorded? Same for the other loop too.

src/server/test/web/obviusTest.js
```diff
@@ -58,5 +62,210 @@ mocha.describe('Obvius API', () => {
...
+
+				//the third line of the response should be the config file
+				expect(res.text.split("\n")[2]).equals(response);
+
+				//config file uploads should create accurate meter objects
+				const allMeters = await Meter.getAll(conn);
+
+				//mb-001.ini should make 28 meters
+				expect(allMeters.length).to.equal(28);
+
```
huss, line 222: I think changing all the uses of 28 to be the length of the exp array would be better. Here it means moving the exp arrays above this line (maybe higher to a logical place).

src/server/test/web/obviusTest.js
```diff
@@ -58,5 +62,210 @@ mocha.describe('Obvius API', () => {
...
+				//get "all" config files to compare to response
+				const allConfigfiles = await Configfile.getAll(conn);
+				let response = '';
+				for (f of allConfigfiles) {
+					response += `CONFIGFILE,${f.makeFilename()},${f.hash},${f.created.format('YYYY-MM-DD hh:mm:ss')}`;
+				}
+
+				//the third line of the response should be the config file
+				expect(res.text.split("\n")[2]).equals(response);
```
huss, line 215: I think it would be nice to make this test slightly stronger. It uses the returned info to get most of the values. The f.makeFilename is mb-001-mb-1234.ini which, I think, is a combination of the serialnumber & modbusdevice and some fixed text. What about creating the expected value directly instead of using the returned info?

I think the code could also create the hash but lets not do that now.

src/server/test/web/obviusTest.js
```diff
@@ -58,5 +62,210 @@ mocha.describe('Obvius API', () => {
...
+					'mb-001.16', 'mb-001.17', 'mb-001.18', 'mb-001.19', 'mb-001.20', 'mb-001.21', 'mb-001.22', 'mb-001.23',
+					'mb-001.24', 'mb-001.25', 'mb-001.26', 'mb-001.27'
+				];
+
+				const expMeterIDs = [
+					1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28
+				];
+				for (const meter of allMeters) {
+					//populate arrays with varying values in ascending order
+					let currentName = meter.name + "";
```
huss: Is the ``+ ""`` to force a string? Maybe use string directly to do this. If it stays then use single quotes since OED prefers that. This assumes it was not already a string (which I would have hoped it is but maybe not in JS).

huss (changes requested):
Thanks to @BrendenHaskins, @lukeckk & @RaiuXL for addressing this issue. Testing found this works well. I have made a few comments to consider. Also, it states it only partially fixes the issue. Could you let me know what aspects you think are not done?

src/server/test/web/obviusTest.js
BrendenHaskins: Going to give some pushback to this - and maybe I'll end up confidently incorrect. The test currently creates a string based on each config file in the database. That string is then compared to the response from the route. Given they are equal, It appears that that both moving pieces resulting from the request made are in agreement, which should in turn validate the test. This also removes the need to dynamically create a string representation of the meter without creating an object - let me know what you think.

src/server/test/web/obviusTest.js
BrendenHaskins: That was the idea, but as you said it isn't actually required and has been removed.

src/server/test/web/obviusTest.js
BrendenHaskins: That makes sense, it's now checking the length of the array, which has been added to a top level scope so multiple tests can access it.

src/server/test/web/obviusTest.js
BrendenHaskins: I've addressed this by creating more descriptive error messages involving the expected and actual names or IDs, and printing that message in case of the test failing.

src/server/test/web/obviusTest.js
BrendenHaskins: That's a good point - I've made the change.

BrendenHaskins:
I've addressed each of your concerns - as for your question: "Also, it states it only partially fixes the issue. Could you let me know what aspects you think are not done?"

I wanted to clearly communicate that some of the obvius modes are not currently implemented (downloads for example), so if they do end up being implemented in the future, they will not have coverage as a result.

src/server/test/web/obviusTest.js
huss: I appreciate you letting me know. I don't feel strongly about this but I went ahead and coded it up to show the difference:

Screenshot 2024-12-06 152923

Do you think this is a good idea or not?

huss:
> I've addressed each of your concerns - as for your question: "Also, it states it only partially fixes the issue. Could you let me know what aspects you think are not done?"
> 
> I wanted to clearly communicate that some of the obvius modes are not currently implemented (downloads for example), so if they do end up being implemented in the future, they will not have coverage as a result.

OED has no current plans to do the other modes. I think it is okay to consider this issue complete. It would be a new effort and testing issue if it might be done in the future. Thus, I think it can say "Fixes" for the issue. Thanks for being careful about linking issues.

src/server/test/web/obviusTest.js
BrendenHaskins: I completely misunderstood your suggestion - making these parameters will definitely help the code's readability and long term ease of maintenance. I've implemented the changes.

src/server/test/web/obviusTest.js
```diff
@@ -58,5 +75,210 @@ mocha.describe('Obvius API', () => {
...
+				let expectedNamesAreEqual = true;
+				let expectedIDsAreEqual = true;
+
+				//error message for more descriptive failures
+				let allErrorMessagesNames = "";
+				let allErrorMessagesIDs = "";
+
+
+				//both arrays should be contain the same sequence of values
+				for (let i = 0; i < 28; i++) {
```
huss: I would be good if this also used length rather than a value.

src/server/test/web/obviusTest.js
```diff
@@ -58,5 +75,210 @@ mocha.describe('Obvius API', () => {
...
+					if (expMeterNames[i] != meterNames[i]) {
+						expectedNamesAreEqual = false;
+						allErrorMessagesNames += "Meter failed name comparison, Expeceted: " + expMeterNames[i] + " Actual: " + meterNames[i] + "\n";
```
huss: Expeceted misspelled.

huss (changes requested):
Thanks to @BrendenHaskins, @lukeckk & @RaiuXL for addressing the comments. Overall, it works as expected. I've made two small comment that should be easy to address.

huss (approved):
Thanks for addressing the comments. Testing and review found everything works as expected. Congratulations to @BrendenHaskins, @lukeckk & @RaiuXL for another accepted contribution.

PR #1393, opened 2024-11-26
Added warnings to admin preferences component

hinyan17:
# Description

Added input validation and warnings for fields on the admin preferences form. Everything that wasn't multiple choice or drop down menu was checked. These fields are: meter reading frequency, min and max reading value, min and max date, reading gap, number of errors, warning file size, and file size limit.

Fixes the TODO associated with PreferencesComponent. (36 on the TODO list)

## Type of change

## Checklist

## Limitations
none


src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -43,6 +43,40 @@ export default function PreferencesComponent() {
...
+		return Number(localAdminPref.defaultMeterMinimumValue) >= Number(localAdminPref.defaultMeterMaximumValue);
+	};
+
+	const invalidDateLimits = () => {
+		const minMoment = moment(localAdminPref.defaultMeterMinimumDate);
+		const maxMoment = moment(localAdminPref.defaultMeterMaximumDate);
+		return !minMoment.isValid() || !maxMoment.isValid() || !minMoment.isBefore(maxMoment);
+	};
+
+	const invalidReadingGap = () => {
```
huss: These work fine but you might consider the min/max keys as in src/client/app/components/meters/CreateMeterModalComponent.tsx. They can even compare to another field and control the input value appropriately.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -176,6 +213,7 @@ export default function PreferencesComponent() {
 					value={localAdminPref.defaultMeterMinimumDate}
 					onChange={e => makeLocalChanges('defaultMeterMinimumDate', e.target.value)}
 					placeholder='YYYY-MM-DD HH:MM:SS'
+					invalid={invalidDateLimits()}
```
huss: It might be nice to use a FormFeedback tag to give the user feedback on the invalid state. src/client/app/components/meters/CreateMeterModalComponent.tsx does this in several places.

huss (changes requested):
Thanks to @hinyan17 for doing this improvement. Review and testing found it works fine. I've made a couple of comment to consider.

src/client/app/components/admin/PreferencesComponent.tsx
hinyan17: Made the necessary changes

src/client/app/components/admin/PreferencesComponent.tsx
hinyan17: Thanks for the suggestion, I just added the FormFeedbacks, and their behavior should be the same as in CreateMeterModalComponent. Let me know if any other changes should be made.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -143,6 +195,7 @@ export default function PreferencesComponent() {
 					type='text'
 					value={localAdminPref.defaultMeterReadingFrequency}
 					onChange={e => makeLocalChanges('defaultMeterReadingFrequency', e.target.value)}
+					invalid={invalidReadingFreq()}
```
huss: This check does not have a form feedback.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -187,7 +252,11 @@ export default function PreferencesComponent() {
 					value={localAdminPref.defaultMeterMaximumDate}
 					onChange={e => makeLocalChanges('defaultMeterMaximumDate', e.target.value)}
 					placeholder='YYYY-MM-DD HH:MM:SS'
+					invalid={invalidMaxDate()}
 				/>
+				<FormFeedback>
+					<FormattedMessage id="error.bounds" values={{ min: MAX_DATE, max: moment(localAdminPref.defaultMeterMinimumDate).utc().format() }} />
```
huss: I think the min & max are backward.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -154,7 +207,11 @@ export default function PreferencesComponent() {
 					value={localAdminPref.defaultMeterMinimumValue}
 					onChange={e => makeLocalChanges('defaultMeterMinimumValue', e.target.value)}
 					maxLength={50}
```
huss, line 210: Maybe take a look at src/client/app/components/meters/CreateMeterModalComponent.tsx where the min & max can constrain the values input by the arrows. You still need the checks but it helps some and would be parallel for similar items. This applies to all the min/min checks on this page.

huss (changes requested):
Thanks to @hinyan17 for addressing the comment. I'm sorry this sat so long - somehow I missed the update. I've made a couple of comments to consider.

src/client/app/components/admin/PreferencesComponent.tsx
hinyan17: I looked at data.ts and couldn't find a good string to let the user know that their input isn't in the right format for a duration or moment object. So for now I used error.unknown. Is this an acceptable change or would you prefer something else?

hinyan17:
While I was going through CreateMeterModalComponent I noticed that all of the invalid functions are called anonymously and inline. Also, the final check if the whole form is valid is done with a redux selector, isValidCreateMeter. Should I implement something similar for PreferencesComponent? It looks a lot cleaner than my current solution and is probably better for consistency.

src/client/app/components/admin/PreferencesComponent.tsx
huss: Could you add a new key of ``invalid.input`` in all three languages that translates to "invalid input". It seems better and can be used in other places in the future.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -308,8 +401,13 @@ export default function PreferencesComponent() {
 					type='number'
 					value={localAdminPref.defaultFileSizeLimit}
 					onChange={e => makeLocalChanges('defaultFileSizeLimit', e.target.value)}
+					min='0'
```
huss: How about changing this to ``min={Number(localAdminPref.defaultWarningFileSize)}`` so it more consistent with other min/max pairs and enforces the min limit?

huss:
> While I was going through CreateMeterModalComponent I noticed that all of the invalid functions are called anonymously and inline. Also, the final check if the whole form is valid is done with a redux selector, isValidCreateMeter. Should I implement something similar for PreferencesComponent? It looks a lot cleaner than my current solution and is probably better for consistency.

That seems fine if you want to do it.

huss (changes requested):
Thanks to @hinyan17 for the quick update. I responded to your two questions and made one small comment.

src/client/app/components/admin/PreferencesComponent.tsx
huss: I also modified the function to be consistent. I forgot to mention that.

huss (approved):
Everything now looks good. Thanks to @hinyan17 for another accepted contribution to OED.

Issue #1307, opened 2024-07-15
Enable reading rejection if outside bounds & work with units


huss:
This is covered by the [readingRangeRejection design document](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/readingRangeRejection/readingRangeRejection.md).

huss:
This was done by PR #1396 so I am closing the issue.

PR #1396, opened 2024-11-27
Max min values

Rakesh-Ranga-Buram:
# Description

This code is related to the create and edit unit changes to incorporate the max and min value checks, and also the changes to make the max and min value associated with the unit to reflect when unit is selected on meter(both create and edit).

Partly Addresses #1307 

## Type of change

## Checklist

## Limitations

The work related to the trinary changes related to disable checks is still pending


src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -267,6 +270,34 @@ export default function CreateUnitModalComponent() {
...
+					<Row xs='1' lg='2'>
+						{/* minVal input */}
+						<Col><FormGroup>
+							<Label for='minVal'>{translate('minVal')}</Label>
+							<Input id='minVal' name='minVal' type='number'
+								onChange={e => handleNumberChange(e)}
+								min={MIN_VAL}
+								max={state.maxVal}
+								defaultValue={state.minVal}
+								invalid={state?.minVal < MIN_VAL || state?.minVal > state?.maxVal} />
```
huss: This invalid check conditions should be added to validUnit setting so the save is disabled as done for other checks.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -267,6 +270,34 @@ export default function CreateUnitModalComponent() {
 							</FormFeedback>
 						</FormGroup></Col>
 					</Row>
+					<Row xs='1' lg='2'>
```
huss: The min/max exist on create/edit unit pages but not in UnitsDetailComponent.tsx. I think that is fine but the "Edit Unit" button on that page should become "Details/Edit Unit" to mirror meter where it both shows more info and allows editing. I could not put a comment in that file so doing here.

src/server/models/Unit.js
```diff
@@ -19,7 +19,7 @@ class Unit {
 	 * @param {*} preferredDisplay True if this unit is always displayed. If not, the user needs to ask to see (for future enhancement).
 	 * @param {*} note Note about this unit.
 	 */
-	constructor(id, name, identifier = name, unitRepresent, secInRate = 3600, typeOfUnit, suffix = '', displayable, preferredDisplay, note) {
+	constructor(id, name, identifier = name, unitRepresent, secInRate = 3600, typeOfUnit, suffix = '', displayable, preferredDisplay, note, minVal, maxVal) {
```
huss: This changed the signature of the unit constructor. There are a number of places where ``new Unit(...)`` is done but was not updated. The value of min/maxVal is undefined in these cases and that is causing issues. For example, I believe that is why a number of tests are failing. The two options are:
1. Give the two new parameters default values of the min/max on the create page. I think this would be okay in this case.
2. Change all the calls to have the desired min/max value. This is a little safer but since I think (almost) all cases that would be changed the min/max would not matter the default would be used. Given this, either solution can be used.

I tested doing this and it fixed the failing tests after other fixes applied. On my machine without the change all the tests seem to run so I'm unsure why only 14 ran on GitHub for the PR. Hopefully that will go away but needs to be verified (or addressed if still an issue).

src/server/util/insertData.js
```diff
@@ -427,8 +427,8 @@ async function insertMeters(metersToInsert, conn) {
 					await conn.none(query);
 				}
 				const conditionSet = {
-					minVal: meter.minVal,
-					maxVal: meter.maxVal,
+					minVal: minVal,
```
huss: I'm not sure what this is supposed to be but minVal & maxVal variables are not defined.

src/server/services/csvPipeline/uploadReadings.js
```diff
@@ -189,8 +189,8 @@ async function uploadReadings(req, res, filepath, conn) {
 	const mapRowToModel = row => { return row; }; // STUB function to satisfy the parameter of loadCsvInput.
 
 	const conditionSet = {
-		minVal: meter.minVal,
-		maxVal: meter.maxVal,
+		minVal: minVal,
```
huss: See other comment. minVal & maxVal are undefined.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -544,25 +559,25 @@ export default function CreateMeterModalComponent(props: CreateMeterModalProps):
...
 						{/* maxVal input */}
 						<Col><FormGroup>
-							<Label for='maxVal'>{translate('meter.maxVal')}</Label>
+							<Label for='maxVal'>{translate('maxVal')}</Label>
 							<Input id='maxVal' name='maxVal' type='number'
 								onChange={e => handleNumberChange(e)}
 								min={meterDetails.minVal}
 								max={MAX_VAL}
-								defaultValue={meterDetails.maxVal}
+								value={meterDetails.maxVal}
```
huss, line 580: The change from defaultValue to value made it impossible to type in a negative value. The arrows still allow. This issue is on min, edit meter as well as edit unit.

src/server/services/obvius/loadLogfileToReadings.js
```diff
@@ -113,8 +113,8 @@ async function loadLogfileToReadings(serialNumber, ipAddress, logfile, conn) {
 				// Unsure if previous values should not change but going to assume want the latest one sent.
 				shouldUpdate = true,
 				conditionSet = {
-					minVal: meter.minVal,
-					maxVal: meter.maxVal,
+					minVal: minVal,
+					maxVal: maxVal,
 					minDate: meter.minDate,
```
huss: Another place where min/maxVal is not set.

src/server/services/readMamacData.js
```diff
@@ -106,8 +106,8 @@ async function readMamacData(meter, conn) {
 		// Previous Mamac values should not change.
 		shouldUpdate = false,
 		conditionSet = {
-			minVal: meter.minVal,
-			maxVal: meter.maxVal,
+			minVal: minVal,
```
huss: Another place min/maxVal not set.

src/server/services/eGauge/readEgaugeData.js
```diff
@@ -43,8 +43,8 @@ async function readEgaugeData(meter, conn) {
 		isEndOnly = false, //true
 		shouldUpdate = false,
 		conditionSet = {
-			minVal: meter.minVal,
-			maxVal: meter.maxVal,
+			minVal: minVal,
```
huss: Another place min/maxVal not set.

src/server/models/Unit.js
```diff
@@ -19,7 +19,7 @@ class Unit {
 	 * @param {*} preferredDisplay True if this unit is always displayed. If not, the user needs to ask to see (for future enhancement).
 	 * @param {*} note Note about this unit.
```
huss, line 20: The JSDoc needs to be updated.

src/client/app/utils/input.ts
```diff
@@ -82,7 +82,9 @@ export const NoUnit: UnitData = {
 	suffix: '',
 	displayable: DisplayableType.none,
 	preferredDisplay: false,
-	note: ''
+	note: '',
+	minVal: -Infinity,
```
huss: Other places use ``adminSelectors`` to get the max/min allowed. I think it makes sense to do that here if possible. My only concern is this is outside a React component but it may work fine since not altering a value.

src/client/app/types/redux/units.ts
```diff
@@ -31,20 +31,22 @@ export interface UnitData {
...
-	identifier: string;
-	unitRepresent: string;
-	secInRate: number;
-	typeOfUnit: UnitType;
-	suffix: string;
-	displayable: DisplayableType;
-	preferredDisplay: boolean;
-	note: string;
-}
+// export interface UnitEditData {
```
huss: Since this interface seems not to be used it should be removed rather than commented out.

src/client/app/translations/data.ts
```diff
@@ -339,8 +339,8 @@ const LocaleTranslationData = {
 		"meter.enabled": "Updates:",
 		"meter.endOnlyTime": "Only End Times:",
 		"meter.endTimeStamp": "End Time Stamp:",
-		"meter.minVal": "Minimum Reading Value Check",
-		"meter.maxVal": "Maximum Reading Value Check",
+		"minVal": "Minimum Reading Value Check",
```
huss: This should be placed alphabetically in all languages. Also, it does not follow the key naming conventions. I know the old value was something similar to this but lets use ``min.value``. Similar for max.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -67,6 +67,17 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 
 	const [validMeter, setValidMeter] = useState(isValidMeter(localMeterEdits));
 
+	useEffect(() => {
```
huss: I'm unclear on the relationship of this useEffect and handleUnitChange. They seem to do similar work. I tried commenting this out and it did not seem to matter. It is not in create unit. I may be missing something so let me know.

Also, a comment on what this is doing would be nice. For example, it is getting the min/max reading value from the unit to make that the starting value for the meter when a unit is selected.

huss (changes requested):
Thanks to @lawsj, @mmehta2669 & @Rakesh-Ranga-Buram for working on this issue. Review and testing found it generally works but I have made several comments where issues where found. One comment that could not be easily placed into the code was:

src/server/util/compareUnits.js needs to have expectUnitToBeEquivalent updated to test min/max. src/server/test/db/unitTests.js & src/server/test/web/unitsTest.js need to be updated so it properly tests these values (and not just default values for all of them).

Please let me know if anything is not clear or you have thoughts.

src/client/app/components/meters/CreateMeterModalComponent.tsx
Rakesh-Ranga-Buram: Converting it to value is also not allowing to type a negative value. Removing a value in that field is setting the value to zero and then typing a negative value is impossible. Also please let me know whether that should be value or defaultValue in all the cases.

Rakesh-Ranga-Buram:
> @huss, running tests on my local shows some errors and I guess the errors are related to the default_disable_checks on preferences. Should I go ahead and modify that column from boolean to the new enum created?

I think your next comment means this is resolved but let me know if it is not. It is correct that, in general, items that are an enum (or equivalent object in JS) should be matched with an enum in the database so they are equivalent. Also, the newer tests in src/server/test/db/enumTests.js to verify the enum in the database and JS are equivalent should be updated for any new enums added.

Rakesh-Ranga-Buram:
@huss , I added the enum changes on preferences. This seems to be solving the issues I mentioned earlier.

src/client/app/components/meters/CreateMeterModalComponent.tsx
huss: I have not tried to implement this but my thinking is to look at src/client/app/components/admin/PreferencesComponent.tsx (newest development version) on setting the value with checks. Also, src/client/app/redux/selectors/adminSelectors.ts has the default values for meters. If a value was set for min/max rather than just the type I think the initial state would have the default value so it would work. Let me know if this makes sense/works or if I need to look at this more.

Rakesh-Ranga-Buram:
> > @huss, running tests on my local shows some errors and I guess the errors are related to the default_disable_checks on preferences. Should I go ahead and modify that column from boolean to the new enum created?
> 
> I think your next comment means this is resolved but let me know if it is not. It is correct that, in general, items that are an enum (or equivalent object in JS) should be matched with an enum in the database so they are equivalent. Also, the newer tests in src/server/test/db/enumTests.js to verify the enum in the database and JS are equivalent should be updated for any new enums added.

@huss , yes that is resolved. I will update src/server/test/db/enumTests.js and update this PR shortly

Rakesh-Ranga-Buram:
@huss , I just resolved conflicts but the tests are failing now, earlier on my branch the tests were all passing

Rakesh-Ranga-Buram:
@huss ,now all the changes have been made

Rakesh-Ranga-Buram:
@huss , the pipeline changes have been made, Can you please review this PR

huss:
> @huss , the pipeline changes have been made, Can you please review this PR

@Rakesh-Ranga-Buram I went to re-review this PR and noticed that two files have a merge conflict. Will you be resolving that? That will allow the CI to run for the tests.

Rakesh-Ranga-Buram:
> > @huss , the pipeline changes have been made, Can you please review this PR
> 
> @Rakesh-Ranga-Buram I went to re-review this PR and noticed that two files have a merge conflict. Will you be resolving that? That will allow the CI to run for the tests.

@huss ,I have resolved the conflicts and all the tests are passing now

src/server/test/db/enumTests.js
```diff
@@ -19,65 +19,65 @@ const { mocha, expect } = require('../common');
 
 mocha.describe('Enums JS to DB', () => {
```
huss, line 20: Tab indenting should be used. It was a mix before and I just noticed. Easy to do in VSC so let me know if you prefer that I do it.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -423,19 +433,59 @@ export default function CreateUnitModalComponent() {
 								</FormGroup>
 							</Col>
 						</Row>
+						<Row xs='1' lg='2'>
+							{/* minVal input */}
+							<Col><FormGroup>
+								<Label for='minVal'>{translate('min.val')}</Label>
```
huss: Should be ``min.value``.

src/server/migrations/1.0.0-2.0.0/sql/units/alter_units_table_add_columns.sql
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ALTER TABLE units
+    ADD COLUMN IF NOT EXISTS min_val FLOAT NOT NULL DEFAULT -9007199254740991 CHECK (min_val::FLOAT >= -9007199254740991),
```
huss: I'm unclear on the value of these checks as it seems to verify it is not out of bounds of allowed min/max float. Would it make sense to remove these checks and check that max_val >= min_val? This also applies to meters (and preferences but those seem likely to go away).

src/server/migrations/1.0.0-2.0.0/index.js
```diff
@@ -20,5 +20,9 @@ module.exports = {
         // It should not matter but first rename cik and then do units.
         await db.none(sqlFile('../migrations/1.0.0-2.0.0/sql/cik/alter_cik_table.sql'));
         await db.none(sqlFile('../migrations/1.0.0-2.0.0/sql/units/alter_units_table.sql'));
+        await db.none(sqlFile('../migrations/1.0.0-2.0.0/sql/units/alter_units_table_add_columns.sql'));
+        await db.none(sqlFile('../migrations/1.0.0-2.0.0/sql/units/add_disable_checks_types.sql'));
```
huss, line 23: It would be best to add the DB types for checks before the column using it is added in the line above.

src/server/migrations/1.0.0-2.0.0/sql/units/alter_units_table_add_columns.sql
```diff
@@ -0,0 +1,9 @@
...
+    ADD COLUMN IF NOT EXISTS max_val FLOAT NOT NULL DEFAULT 9007199254740991 CHECK (max_val::FLOAT <= 9007199254740991),
+    ADD COLUMN IF NOT EXISTS disable_checks disable_checks_type NOT NULL DEFAULT 'reject_none'
```
huss: I think the current default is reject_all and that is what is in the design document for meters so units should follow that. Applies to create_units_table.sql too. meters & preferences should do the same.

src/server/migrations/1.0.0-2.0.0/sql/preferences/alter_preferences_default_meter_disable_checks.sql
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+-- Add a temporary column with the new type
+ALTER TABLE preferences 
+    ADD COLUMN default_meter_disable_checks_temp disable_checks_type NOT NULL DEFAULT 'reject_none';
```
huss: This value probably should not matter but lets make it 'reject_all' to be safe/consistent. (obsolete since another file impacted by removal)

src/server/sql/preferences/create_preferences_table.sql
```diff
@@ -21,6 +21,6 @@ CREATE TABLE IF NOT EXISTS preferences (
     default_meter_maximum_date TIMESTAMP NOT NULL,
 	default_meter_reading_gap REAL NOT NULL,
     default_meter_maximum_errors INTEGER NOT NULL,
-	default_meter_disable_checks BOOLEAN NOT NULL,
+	default_meter_disable_checks disable_checks_type NOT NULL DEFAULT 'reject_none',
```
huss: I could not easily add this in the right place in some files so I'm putting this comment here. The design doc has

> The value min/max and default enabled will be removed from the site settings page and from the related routes, DB, model, etc.

The idea is that the value comes from units by default so this is not longer needed. Doing this will change a number of files. Let me know if you think this is a bad idea.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -125,13 +133,15 @@ export default function CreateUnitModalComponent() {
 
 	// Keeps canSave state up to date. Checks if valid and if edit made.
 	useEffect(() => {
+
```
huss: Please remove blank line.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -125,13 +133,15 @@ export default function CreateUnitModalComponent() {
...
 		// This checks:
 		// - Name cannot be blank
 		// - If type of unit is suffix there must be a suffix
 		// - The rate is set so not the custom input value. This happens if select custom value but don't input with enter.
 		// - The custom rate is a positive integer
 		const validUnit = state.name !== '' &&
 			(state.typeOfUnit !== UnitType.suffix || state.suffix !== '') && state.secInRate !== Number(CUSTOM_INPUT)
+			&& (state?.minVal < MIN_VAL || state?.minVal > state?.maxVal)
```
huss: First, edit unit does not have this logic on min/max but should. Second, I think the logic is off. I made it ``&& state?.minVal >= MIN_VAL && state?.maxVal <= MAX_VAL && state?.minVal <= state?.maxVal`` to mimic the form feedback and to get it the be valid (the current logic is more for invalid which is not the check).

src/server/routes/meters.js
```diff
@@ -262,6 +266,7 @@ router.post('/edit', requiredAdmin('edit meters'), async (req, res) => {
 		const conn = getConnection();
 		try {
 			const meter = await Meter.getByID(req.body.id, conn);
+			console.log(req.body);
```
huss: This debug statement should be removed.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -660,17 +660,29 @@ async function processData(rows, meterID, timeSort = MeterTimeSortTypesJS.increa
...
 			({ msgTotal, msgTotalWarning } = appendMsgTotal(msgTotal, errMsg, msgTotalWarning));
-			// This empties the result array. Should be fast and okay with const.
-			result.splice(0, result.length);
-			isAllReadingsOk = false;
-			return { result, isAllReadingsOk, msgTotal };
+
+			// Handle 'reject_bad' specifically
+			if (conditionSet['disableChecks'] === 'reject_bad') {
+				// This removes invalid readings but keeps the valid ones.
+				result = result.filter(reading => validateSingleReading(reading, conditionSet));
```
huss, line 673: ``validateSingleReading`` function does not seem to exist. This causes uploads to fail if you are only rejecting bad readings.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -76,6 +89,20 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
...
+	const handleUnitChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+		const selectedUnitId = Number(e.target.value);
+		const selectedUnit = unitDataById[selectedUnitId];
+		if (selectedUnit) {
+			setLocalMeterEdits({
+				...localMeterEdits,
+				unitId: selectedUnitId,
+				minVal: selectedUnit.minVal,
+				maxVal: selectedUnit.maxVal,
+				disableChecks: selectedUnit.disableChecks
```
huss: I don't think min/maxVal nor disableChecks needs to be set here since the useEffect above fires when unitId is updated.

src/client/app/components/meters/EditMeterModalComponent.tsx
huss: There is a new comment on the update so this is resolved.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -41,7 +42,10 @@ export default function CreateUnitModalComponent() {
 		// The client code makes the id for the selected unit and default graphic unit be -99
 		// so it can tell it is not yet assigned and do the correct logic for that case.
 		// The units API expects these values to be undefined on call so that the database can assign their values.
-		id: -99
+		id: -99,
+		minVal: MIN_VAL,
+		maxVal: MAX_VAL,
+		disableChecks: DisableChecksType.reject_none
```
huss: I think per other comments this should be reject all.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -302,8 +303,8 @@ export default function PreferencesComponent() {
 					type='select'
 					value={localAdminPref.defaultMeterDisableChecks?.toString()}
 					onChange={e => makeLocalChanges('defaultMeterDisableChecks', e.target.value)}>
-					{Object.keys(TrueFalseType).map(key => {
-						return (<option value={key} key={key}>{translate(`TrueFalseType.${key}`)}</option>);
+					{Object.keys(DisableChecksType).map(key => {
```
huss: I believe the design doc called for removing min, max and this since it is now controlled by the unit for the default. Note this will change the DB and other files. I did a quick search and they probably are not used anywhere so hopefully removing does not cause issues.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -423,19 +433,59 @@ export default function CreateUnitModalComponent() {
...
-								id="note"
-								name="note"
-								type="textarea"
-								value={state.note}
+								id='note'
+								name='note'
+								type='textarea'
 								onChange={e => handleStringChange(e)}
-							/>
+								value={state.note} />
```
huss: value seems above onChange in other places so maybe put back above it.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -423,19 +433,59 @@ export default function CreateUnitModalComponent() {
...
 						</FormGroup>
-					</Container>
-				</ModalBody>
+					</Container></ModalBody>
```
huss: ModalBody is a tag own its on line when started so probably should put this on a separate line as it was before.

src/client/app/redux/selectors/adminSelectors.ts
```diff
@@ -312,7 +312,7 @@ export const selectDefaultCreateMeterValues = createAppSelector(
 			minDate: adminPreferences.defaultMeterMinimumDate,
 			maxDate: adminPreferences.defaultMeterMaximumDate,
 			maxError: adminPreferences.defaultMeterMaximumErrors,
-			disableChecks: adminPreferences.defaultMeterDisableChecks
+			disableChecks: DisableChecksType.reject_none
```
huss: reject all per other comments.

src/client/app/redux/slices/adminSlice.ts
```diff
@@ -34,7 +35,7 @@ export const defaultAdminState: AdminState = {
 	defaultMeterMaximumDate: moment(0).utc().add(5000, 'years').format('YYYY-MM-DD HH:mm:ssZ'),
 	defaultMeterReadingGap: 0,
 	defaultMeterMaximumErrors: 75,
-	defaultMeterDisableChecks: false,
+	defaultMeterDisableChecks: DisableChecksType.reject_none,
```
huss: reject all per other comments.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: Newer comment so resolved.

src/client/app/types/redux/units.ts
```diff
@@ -20,6 +20,11 @@ export enum UnitRepresentType {
 	raw = 'raw'
 }
 
+export enum DisableChecksType {
+	reject_bad = 'reject_bad',
+	reject_all = 'reject_all',
+	reject_none = 'reject_none'
+}
 export interface UnitData {
```
huss, line 29: Please add blank line here.

src/client/app/utils/input.ts
```diff
@@ -82,7 +83,10 @@ export const NoUnit: UnitData = {
 	suffix: '',
 	displayable: DisplayableType.none,
 	preferredDisplay: false,
-	note: ''
+	note: '',
+	minVal: MIN_VAL,
+	maxVal: MAX_VAL,
+	disableChecks: DisableChecksType.reject_none
```
huss: reject all per other comments.

src/server/migrations/1.0.0-2.0.0/sql/preferences/add_preferences_pipeline_checks.sql
```diff
@@ -12,7 +12,7 @@ ALTER TABLE preferences
     ADD COLUMN IF NOT EXISTS default_meter_maximum_date TIMESTAMP NOT NULL DEFAULT '6970-01-01 00:00:00+00:00',
 	ADD COLUMN IF NOT EXISTS default_meter_reading_gap REAL NOT NULL DEFAULT 0,
     ADD COLUMN IF NOT EXISTS default_meter_maximum_errors INTEGER NOT NULL DEFAULT 75,
-	ADD COLUMN IF NOT EXISTS default_meter_disable_checks BOOLEAN NOT NULL DEFAULT false
+	ADD COLUMN IF NOT EXISTS default_meter_disable_checks disable_checks_type NOT NULL DEFAULT 'reject_none'
```
huss: Just noting this is another file impacted by removing the min/max/disable preference.

src/server/models/Meter.js
```diff
@@ -60,7 +60,7 @@ class Meter {
 		reading = 0.0, startTimestamp = moment(0).utc().format('YYYY-MM-DD HH:mm:ssZ'), endTimestamp = moment(0).utc().format('YYYY-MM-DD HH:mm:ssZ'),
 		previousEnd = moment(0).utc(), unitId = -99, defaultGraphicUnit = -99, areaUnit = Unit.areaUnitType.NONE, readingFrequency = '00:15:00',
 		minVal = Number.MIN_SAFE_INTEGER, maxVal = Number.MAX_SAFE_INTEGER, minDate = moment(0).utc().format('YYYY-MM-DD HH:mm:ssZ'),
-		maxDate = moment(0).utc().add(5000, 'years').format('YYYY-MM-DD HH:mm:ssZ'), maxError = 75, disableChecks = false) {
+		maxDate = moment(0).utc().add(5000, 'years').format('YYYY-MM-DD HH:mm:ssZ'), maxError = 75, disableChecks = Meter.disableChecksType.REJECT_NONE) {
```
huss: I think this might be better as reject all (I'm hoping that still works with current code.)

src/server/models/Meter.js
```diff
@@ -313,6 +313,14 @@ class Meter {
 		return Reading.getAllByMeterID(this.id, conn);
 	}
 
+	/**
```
huss: I think enums are generally after the create as the one above so this should be moved there but that probably does not matter given comment below.

This duplicates the code in unit.js. Units must (and are) created first in src/server/models/database.js so I think this should be removed here and in the database model.

src/server/models/Unit.js
```diff
@@ -18,8 +18,11 @@ class Unit {
 	 * @param {*} displayable Can be none, all, or admin. Restrict the type of user that can see this unit.
 	 * @param {*} preferredDisplay True if this unit is always displayed. If not, the user needs to ask to see (for future enhancement).
 	 * @param {*} note Note about this unit.
+	 * @param minVal inclusive minimum acceptable reading value (won't be rejected)
+	 * @param maxVal inclusive maximum acceptable reading value (won't be rejected)
+	 * @param disableChecks disable checks on meter for conditionSet 
 	 */
-	constructor(id, name, identifier = name, unitRepresent, secInRate = 3600, typeOfUnit, suffix = '', displayable, preferredDisplay, note) {
+	constructor(id, name, identifier = name, unitRepresent, secInRate = 3600, typeOfUnit, suffix = '', displayable, preferredDisplay, note, minVal = Number.MIN_SAFE_INTEGER, maxVal = Number.MAX_SAFE_INTEGER, disableChecks = Unit.disableChecksType.REJECT_NONE) {
```
huss: reject all

src/server/models/Unit.js
```diff
@@ -18,8 +18,11 @@ class Unit {
 	 * @param {*} displayable Can be none, all, or admin. Restrict the type of user that can see this unit.
 	 * @param {*} preferredDisplay True if this unit is always displayed. If not, the user needs to ask to see (for future enhancement).
 	 * @param {*} note Note about this unit.
+	 * @param minVal inclusive minimum acceptable reading value (won't be rejected)
+	 * @param maxVal inclusive maximum acceptable reading value (won't be rejected)
+	 * @param disableChecks disable checks on meter for conditionSet
```
huss: I think this parameter description might be better if indicates it has three possible values.

src/server/routes/meters.js
```diff
@@ -63,7 +63,7 @@ function formatMeterForResponse(meter, hasFullAccess) {
 		minDate: null,
 		maxDate: null,
 		maxError: null,
-		disableChecks: null,
+		disableChecks: meter.disableChecks,
```
huss: This is admin/csv user only information so it should not be returned here. The original value was null and that is what OED usually uses. I'm not sure if that was causing issues but it should be put back and then made sure it is not accessed by non-admin/csv user due to the null (may not be since I'm unsure why a non-admin/csv user would need that).

src/server/routes/preferences.js
```diff
@@ -86,7 +87,9 @@ router.post('/', async (req, res) => {
 					type: 'number'
 				},
 				defaultMeterDisableChecks: {
-					type: 'boolean'
+					type: 'string',
```
huss: This and min/max is impacted if removed from site preferences.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -660,17 +660,29 @@ async function processData(rows, meterID, timeSort = MeterTimeSortTypesJS.increa
 		prevEndTimestampTz = endTimestampTz;
 	}
 	// Validate data if conditions given
```
huss: Seems to overlap new comment so can it be removed?

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -660,17 +660,29 @@ async function processData(rows, meterID, timeSort = MeterTimeSortTypesJS.increa
...
-			isAllReadingsOk = false;
-			return { result, isAllReadingsOk, msgTotal };
+
+			// Handle 'reject_bad' specifically
+			if (conditionSet['disableChecks'] === 'reject_bad') {
+				// This removes invalid readings but keeps the valid ones.
+				result = result.filter(reading => validateSingleReading(reading, conditionSet));
+			} else {
+				// Default behavior: reject all readings
+				result.splice(0, result.length); // Empties the result array
```
huss: OED prefers comments above and not at the end of lines.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -660,17 +660,29 @@ async function processData(rows, meterID, timeSort = MeterTimeSortTypesJS.increa
...
+			}
+
+			if (result.length === 0) {
```
huss, line 686: I suggest putting the setting of ``isAllReadingsOk = false`` inside the else above and removing this ``if``. Now that I think about it, both cases should return false since some values were rejected. That is what OED does whenever any data is dropped.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -15,12 +15,36 @@ const { log } = require('../../log');
...
 	/* tslint:disable:no-string-literal */
-	const { validDates, errMsg: errMsgDate } = checkDate(arrayToValidate, conditionSet['minDate'], conditionSet['maxDate'], conditionSet['maxError'] / 2, meterIdentifier);
-	const { validValues, errMsg: errMsgValue } = checkValue(arrayToValidate, conditionSet['minVal'], conditionSet['maxVal'], conditionSet['maxError'] / 2, meterIdentifier);
+	const { validDates, rejectedDates, errMsg: errMsgDate } = checkDate(arrayToValidate, conditionSet['minDate'], conditionSet['maxDate'], conditionSet['maxError'] / 2, meterIdentifier);
+	const { validValues, rejectedValues, errMsg: errMsgValue } = checkValue(arrayToValidate, conditionSet['minVal'], conditionSet['maxVal'], conditionSet['maxError'] / 2, meterIdentifier);
 	/* tslint:enable:no-string-literal */
 	const errMsg = errMsgDate + errMsgValue;
+
+	 // Handle the disableChecks logic
+	 if (conditionSet['disableChecks'] === 'reject_none') {
```
huss: I think this comes after the checks since issues are still logged so a comment about that would be good. I'm not sure why someone wants to check but not reject but let's leave it.

As a rule, I think avoiding returns in the middle of a function is a good idea when it isn't too hard. Would it be possible to easily change the logic to set the needed values and only return at the end. For example, make the rest part of an else until the return?

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -15,12 +15,36 @@ const { log } = require('../../log');
...
+        return {
+            validReadings: true,
+            rejectedReadings: [],
+            errMsg: '', // No validation errors since checks are disabled
+        };
+    }
+
+    let validReadings = validDates && validValues;
```
huss: As currently written this should be a const but that is probably going away.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -15,12 +15,36 @@ const { log } = require('../../log');
...
+    let rejectedReadings = [];
+
+    if (conditionSet['disableChecks'] === 'reject_bad') {
+        // Remove only invalid readings and return the rejected ones
+        rejectedReadings = [...rejectedDates, ...rejectedValues];
+        arrayToValidate = arrayToValidate.filter(reading => !rejectedReadings.includes(reading));
+    } else if (conditionSet['disableChecks'] === 'reject_all' && !validReadings) {
+        // Reject all readings if validation fails
+        rejectedReadings = [...arrayToValidate];
+        arrayToValidate.length = 0; // Clear the array
```
huss: I think OED uses slice in other places to do this and it is nice to be consistent. Also, comment above line of code.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -36,35 +60,27 @@ function validateReadings(arrayToValidate, conditionSet, meterIdentifier = undef
 function checkDate(arrayToValidate, minDate, maxDate, maxError, meterIdentifier) {
```
huss, line 65: This never had return values in the JSDoc but it should.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -36,35 +60,27 @@ function validateReadings(arrayToValidate, conditionSet, meterIdentifier = undef
...
 	let validDates = true;
 	let errMsg = '';
+	const rejectedDates = [];
 	if (minDate === null && maxDate === null) {
-		return { validDates, errMsg };
+		return { validDates, rejectedDates, errMsg };
```
huss: See above for return in middle of function.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -36,35 +60,27 @@ function validateReadings(arrayToValidate, conditionSet, meterIdentifier = undef
...
 	}
 	let readingNumber = 0;
 	for (const reading of arrayToValidate) {
 		readingNumber++;
 		if (maxError <= 0) {
 			break;
 		}
-		if (reading.startTimestamp < minDate) {
+		if (reading.startTimestamp < minDate || reading.endTimestamp > maxDate) {
```
huss: While I see why you probably wanted to combine these two messages, this is an existing API response so changing it could break scripts, etc. for sites. Unless you want to advocate that this is still a good idea, the messages should be put back the way they were.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -78,31 +94,24 @@ function checkDate(arrayToValidate, minDate, maxDate, maxError, meterIdentifier)
...
-			log.error(newErrMsg);
-			errMsg += '<br>' + newErrMsg + '<br>';
-			--maxError;
-			validValues = false;
-		} else if (reading.reading > maxVal) {
+		if (reading.reading < minVal || reading.reading > maxVal) {
 			const newErrMsg = `error when checking reading value for #${readingNumber} on meter ${meterIdentifier}: ` +
-			`value ${reading.reading} is larger than upper bound ${maxVal} ` +
-			`with startTimestamp ${reading.startTimestamp} and endTimestamp ${reading.endTimestamp}`;
+                `value ${reading.reading} is out of bounds (min: ${minVal}, max: ${maxVal}).`;
```
huss: The new code seems to use spaces instead of tab indenting.

src/server/services/pipeline-in-progress/validateReadings.js
```diff
@@ -78,31 +94,24 @@ function checkDate(arrayToValidate, minDate, maxDate, maxError, meterIdentifier)
...
-		if (reading.reading < minVal) {
-			const newErrMsg = `error when checking reading value for #${readingNumber} on meter ${meterIdentifier}: ` +
-			`value ${reading.reading} is smaller than lower bound ${minVal} ` +
-			`with startTimestamp ${reading.startTimestamp} and endTimestamp ${reading.endTimestamp}`;
-			log.error(newErrMsg);
-			errMsg += '<br>' + newErrMsg + '<br>';
-			--maxError;
-			validValues = false;
-		} else if (reading.reading > maxVal) {
+		if (reading.reading < minVal || reading.reading > maxVal) {
```
huss: See comment on date check about changing messages.

src/server/sql/meter/create_meters_table.sql
```diff
@@ -70,5 +70,5 @@ CREATE TABLE IF NOT EXISTS meters (
     min_date TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:00+00:00',
     max_date TIMESTAMP NOT NULL DEFAULT '6970-01-01 00:00:00+00:00',
     max_error INTEGER NOT NULL DEFAULT 75,
-    disable_checks BOOLEAN DEFAULT false
+    disable_checks disable_checks_type NOT NULL DEFAULT 'reject_none'
```
huss: reject all

src/server/sql/preferences/insert_default_row.sql
```diff
@@ -21,7 +21,7 @@ IF NOT EXISTS(SELECT *
 	default_meter_disable_checks, default_help_url) 
 	VALUES ('', 'line', FALSE, 'en', NULL, 5, 25, FALSE, 'meters', '00:15:00', 
 	-9007199254740991, 9007199254740991, '1970-01-01 00:00:00+00:00', '6970-01-01 00:00:00+00:00',
-	0, 75, FALSE, 'https://openenergydashboard.org/');
+	0, 75, 'reject_all', 'https://openenergydashboard.org/');
```
huss: This may need to go for min/max/reject.

src/server/sql/unit/create_units_table.sql
```diff
@@ -12,5 +12,8 @@ CREATE TABLE IF NOT EXISTS units (
 	suffix VARCHAR(50) DEFAULT '',
 	displayable displayable_type NOT NULL,
 	preferred_display BOOLEAN NOT NULL,
-	note TEXT
+	note TEXT,
+	min_val FLOAT NOT NULL DEFAULT -9007199254740991 CHECK (min_val::FLOAT >= -9007199254740991),
```
huss: See the alter unit file for comments about this. Need to be consistent with any changes.

src/server/sql/unit/insert_new_unit.sql
```diff
@@ -2,6 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, suffix, displayable, preferred_display, note)
-VALUES (${name}, ${identifier}, ${unitRepresent}, ${secInRate}, ${typeOfUnit}, ${suffix}, ${displayable}, ${preferredDisplay}, ${note})
+INSERT INTO units(name, identifier, unit_represent, sec_in_rate, type_of_unit, suffix, displayable, preferred_display, note, max_val, min_val, disable_checks)
```
huss: min comes before max in the table creation which makes sense to me. Here it is the other way. It would be nice if min came first.

src/server/sql/unit/update_unit.sql
```diff
@@ -11,5 +11,8 @@ UPDATE units
 		suffix = ${suffix},
 		displayable = ${displayable},
 		preferred_display = ${preferredDisplay},
-		note = ${note}
+		note = ${note},
+		max_val = ${maxVal},
```
huss: See insert unit file for min before max.

huss (changes requested):
Thanks to @Rakesh-Ranga-Buram for addressing previous comments. I've made more comments to consider. To keep the process cleaner, I resolved all the old comments and made new ones about the same area when needed. Also, the following comment from before does not seem to be addressed:

> src/server/util/compareUnits.js needs to have expectUnitToBeEquivalent updated to test min/max. src/server/test/db/unitTests.js & src/server/test/web/unitsTest.js need to be updated so it properly tests these values (and not just default values for all of them).

Finally, I tested uploads to a modest extent. The reject bad seems to have an issue (see comment) so full testing is awaiting resolution.

Please let me know if anything is not clear, you have thoughts or need help.

Rakesh-Ranga-Buram:
Hi @huss , I addressed all the comments except for the tab-indenting and removing max-min-reject from meters and preferences. On the tab-indenting, I normally use prettier on save but do you feel I should use any other app or please let me know the preferred one. On removing the min-max-reject, I thought to handle it separately as it includes touching a lot of files, I will merge those changes shortly.

huss:
> On the tab-indenting, I normally use prettier on save but do you feel I should use any other app or please let me know the preferred one.

OED uses tabs because it allows each developer to choose the space associated with each indent. If you use standard VSC formatting (no extension needed) then it will do this if set up properly. See tabs vs spaces in the OED developer first steps web page.

Rakesh-Ranga-Buram:
@huss, Can you please review the code now

src/server/services/pipeline-in-progress/processData.js
huss: I was wrong about this. Code has been changed so it works correctly with a comment.

huss (approved):
I'm sorry this took a while to review. I have now carefully tested and reviewed the code. To move this along, I push a number of changes. It all works as desired from what I can see. I'm going to let this sit for a couple of days to allow @Rakesh-Ranga-Buram to make any comments on my changes.

Rakesh-Ranga-Buram:
> I'm sorry this took a while to review. I have now carefully tested and reviewed the code. To move this along, I push a number of changes. It all works as desired from what I can see. I'm going to let this sit for a couple of days to allow @Rakesh-Ranga-Buram to make any comments on my changes.

Hi @huss , thanks for all the changes. I tested it multiple times and it works as expected. Also from your changes I got to know about the things that I missed out. I think now the code is good to be merged to `development`

Issue #557, opened 2021-02-14
verify site is responding to requests


huss:
OED has a script to check if a site is responding. In the past, OED would stop responding from time to time so having a regular check was desired. This script has not been modified to make it easy for a site to use it and get notifications.

huss:
See closed PR #448 for JS file that can check the site status.

huss:
The PHP and .js is in PR #452 but that was closed (not merged). There is also the harf repo that checks status and allows restart but does not seem too specific to the OED (at least yet). Using one of these or something else would be appropriate.

lukeckk:
I'd like to work on this one

lukeckk:
@huss 

Could I get some clarification on which URL am I checking? Is it "https://oed.beloit.edu/" ? 

Also, is there a preferred language I should use to write the script?

Thank you

huss:
@lukeckk 

> Could I get some clarification on which URL am I checking? Is it "https://oed.beloit.edu/" ?

The URL depends on the site running the test. Thus, it should be supplied in a easy way when the test is started or by an easy edit.

> Also, is there a preferred language I should use to write the script?

We have used bash for other scripts and that would be nice. It is automatically included with Linux so easy for the site to use. If you have other ideas then let us know.

lukeckk:
I was thinking about adding a file containing js code, then make a HEAD request using fetch or axios to get to the result. User can run the file if they want to know about the status. Or I can create a script to adjust the running frequency depending on your needs. I'm attaching the code below. Let me know what you think, if you prefer bash scripts, I am fine with that too.

![image](https://github.com/user-attachments/assets/bc591d0f-1990-41f8-aa80-3e0d380fb25d)

huss:
@lukeckk I'm not sure I have a strong preference for either solution. Since fetch seems built-in it would not need a dependency so that is nice.

If you look at package.json, there are scripts. Some pass parameters via the npm command, e.g., ``generateSineSquaredTestingData``. This might be a way to easily allow the user to run the command via docker/npm or in a script. I think there is a way to pass parameters from the ``npm run`` so the server URL could be specified but I don't know for sure or the details. If you have other thoughts on how to do this then we are open to them. Whatever is decided, OED needs to make it easy on a site to do this such as the cron jobs currently used for the reading views. In a similar way it should run periodically. It also needs to work if OED is not running properly.

[...]

PR #1398, opened 2024-12-02
Script to check website status and send email to users if website is down

lukeckk:
# Description

Created a file named 'checkWebsiteStatus.js' that contains the script to check website status and sends email to users when the website is down. 

Partly Addresses #557 

## Type of change

## Checklist

## Limitations

Currently, I am unable to see if the email-sending part works. I verified that it prints 'The server is down' in the console when the website is down.  I replaced some variables in docker-compose.yml with my emails and credentials, restarted the server and ran the script with command 'npm run status' but no email was received. 

Could you run the script to see if it works on your end since your email is already in the database? Feel free give any feedback if you notice anything wrong. 

The cron schedule is currently excluded for ease of testing.

### DIFF
--- a/src/scripts/checkWebsiteStatusCron.bash
+++ b/src/scripts/checkWebsiteStatusCron.bash
@@ -1,9 +1,16 @@
-# This aggregates the readings data at hour level.
+#!/bin/bash
+# *
+# * This Source Code Form is subject to the terms of the Mozilla Public
+# * License, v. 2.0. If a copy of the MPL was not distributed with this
+# * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# *
+
+# This check the website status at hour level.
 # This should be copied to /etc/ or /etc/cron.hourly/ or /etc/cron.daily and the copy renamed so that its function will be clear to admins.
 
 # Check if a URL is provided as an argument
 if [ -z "$1" ]; then
-    echo "Error: No URL provided. Usage: $0"
+    echo "Error: No URL provided. Usage: $0 URL-to-check"
     exit 1
 fi
 
--- a/src/scripts/checkWebsiteStatusCron.bash
+++ b/src/scripts/checkWebsiteStatusCron.bash
@@ -1,0 +1,16 @@
+# This aggregates the readings data at hour level.
+# This should be copied to /etc/ or /etc/cron.hourly/ or /etc/cron.daily and the copy renamed so that its function will be clear to admins.
+
+# Check if a URL is provided as an argument
+if [ -z "$1" ]; then
+    echo "Error: No URL provided. Usage: $0"
+    exit 1
+fi
+
+URL=$1
+
+# The absolute path the project root directory (OED)
+cd '/example/path/to/project/OED'
+
+# The following line should NOT need to be edited except by devs or if you have an old system with only docker-compose.
+docker compose run --rm web npm run --silent checkWebsiteStatus -- $URL &>> /dev/null &
--- a/src/server/services/checkWebsiteStatus.js
+++ b/src/server/services/checkWebsiteStatus.js
@@ -2,11 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-	const cron = require('node-cron');
-	const { getConnection } = require('../db');
 	const { log } = require('../log');
-	const { logMailer } = require('../logMailer');
-	const LogEmail = require('../models/LogEmail');
 	const WEBSITE_URL = process.argv[2];
 	
 	async function checkWebsite() {
--- a/src/server/services/checkWebsiteStatus.js
+++ b/src/server/services/checkWebsiteStatus.js
@@ -2,23 +2,22 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-	const { log } = require('../log');
-	const WEBSITE_URL = process.argv[2];
-	
-	async function checkWebsite() {
-		try {
-			const response = await fetch(WEBSITE_URL, { method: 'HEAD' });
-	
-			if (!response.ok) {
-				const errorMessage = `The server at ${WEBSITE_URL} is down.`;
-				// Log the error using Logger class
-				log.error(errorMessage);	
-			}
-		} catch (error) {
-			const errorMessage = `Failed to reach ${WEBSITE_URL}. Error: ${error.message}`;
-	  log.error(errorMessage, error);
+const { log } = require('../log');
+const WEBSITE_URL = process.argv[2];
+
+async function checkWebsite() {
+	try {
+		const response = await fetch(WEBSITE_URL, { method: 'HEAD' });
+
+		if (!response.ok) {
+			const errorMessage = `The server at ${WEBSITE_URL} is down.`;
+			// Log the error using Logger class
+			log.error(errorMessage);
 		}
+	} catch (error) {
+		const errorMessage = `Failed to reach ${WEBSITE_URL}. Error: ${error.message}`;
+		log.error(errorMessage, error);
 	}
-	
-	checkWebsite();
-	
+}
+
+checkWebsite();
--- a/src/server/services/checkWebsiteStatus.js
+++ b/src/server/services/checkWebsiteStatus.js
@@ -1,52 +1,28 @@
-const cron = require('node-cron');
-const { getConnection } = require('../db');
-const { logMailer } = require('../logMailer');
-const LogEmail = require('../models/LogEmail');
-const colors = require('colors');
-  
-async function checkWebsite() {
-  const conn = getConnection();
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-  try {
-    const response = await fetch('https://httpstat.us/500/', { method: 'HEAD' });
-
-    if (response.ok) {
-      console.log('The server is up'.green.inverse);
-    } else {
-      console.error('The server is down'.red.inverse);
-
-      const errorMessage = 'The server at https://httpstat.us/500/ is down.';
-
-      // Log the error in the database
-      const log = new LogEmail(undefined, errorMessage);
-      await log.insert(conn);
-
-      // Send a logging email
-      await logMailer(conn);
-    }
-  } 
-  catch (error) {
-    console.error('Error:', error);
-
-    const errorMessage = `Failed to reach https://httpstat.us/500/. Error: ${error.message}`;
-
-    // Log the error in the database
-    const log = new LogEmail(undefined, errorMessage);
-    await log.insert(conn);
-
-    // Send a logging email
-    await logMailer(conn);
-  } finally {
-    if (conn.close) {
-      await conn.close();
-    }
-  }
-}
-
-// Schedule the task to run every hour
-// cron.schedule('0 * * * *', () => {
-//   console.log('Running website status check...');
-//   checkWebsite();
-// });
-
-checkWebsite();
+	const cron = require('node-cron');
+	const { getConnection } = require('../db');
+	const { log } = require('../log');
+	const { logMailer } = require('../logMailer');
+	const LogEmail = require('../models/LogEmail');
+	const WEBSITE_URL = process.argv[2];
+	
+	async function checkWebsite() {
+		try {
+			const response = await fetch(WEBSITE_URL, { method: 'HEAD' });
+	
+			if (!response.ok) {
+				const errorMessage = `The server at ${WEBSITE_URL} is down.`;
+				// Log the error using Logger class
+				log.error(errorMessage);	
+			}
+		} catch (error) {
+			const errorMessage = `Failed to reach ${WEBSITE_URL}. Error: ${error.message}`;
+	  log.error(errorMessage, error);
+		}
+	}
+	
+	checkWebsite();
+	
--- a/src/server/services/checkWebsiteStatus.js
+++ b/src/server/services/checkWebsiteStatus.js
@@ -1,0 +1,52 @@
+const cron = require('node-cron');
+const { getConnection } = require('../db');
+const { logMailer } = require('../logMailer');
+const LogEmail = require('../models/LogEmail');
+const colors = require('colors');
+  
+async function checkWebsite() {
+  const conn = getConnection();
+
+  try {
+    const response = await fetch('https://httpstat.us/500/', { method: 'HEAD' });
+
+    if (response.ok) {
+      console.log('The server is up'.green.inverse);
+    } else {
+      console.error('The server is down'.red.inverse);
+
+      const errorMessage = 'The server at https://httpstat.us/500/ is down.';
+
+      // Log the error in the database
+      const log = new LogEmail(undefined, errorMessage);
+      await log.insert(conn);
+
+      // Send a logging email
+      await logMailer(conn);
+    }
+  } 
+  catch (error) {
+    console.error('Error:', error);
+
+    const errorMessage = `Failed to reach https://httpstat.us/500/. Error: ${error.message}`;
+
+    // Log the error in the database
+    const log = new LogEmail(undefined, errorMessage);
+    await log.insert(conn);
+
+    // Send a logging email
+    await logMailer(conn);
+  } finally {
+    if (conn.close) {
+      await conn.close();
+    }
+  }
+}
+
+// Schedule the task to run every hour
+// cron.schedule('0 * * * *', () => {
+//   console.log('Running website status check...');
+//   checkWebsite();
+// });
+
+checkWebsite();
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
 				"bcryptjs": "~2.4.3",
 				"body-parser": "~1.20.2",
 				"bootstrap": "~5.3.3",
+				"colors": "~1.4.0",
 				"csv": "~5.3.2",
 				"csv-stringify": "~5.6.5",
 				"d3": "~7.8.5",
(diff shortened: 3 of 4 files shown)


src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
+const cron = require('node-cron');
```
huss: The standard OED license comment needs to be added. See src/server/services/createDB.js for an example (3 lines of comment and one blank line).

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
...
+const { getConnection } = require('../db');
+const { logMailer } = require('../logMailer');
+const LogEmail = require('../models/LogEmail');
+const colors = require('colors');
+  
+async function checkWebsite() {
+  const conn = getConnection();
+
+  try {
+    const response = await fetch('https://httpstat.us/500/', { method: 'HEAD' });
```
huss: This is great for development but the URL needs to be provided somehow. I'm open to ideas on how to do this. I don't think OED has yet to create an npm command that passes the value to a function but maybe that is possible so it would be ``npm run status http://myodedsite.myorg.org:3000``. Thoughts?

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
+const cron = require('node-cron');
+const { getConnection } = require('../db');
+const { logMailer } = require('../logMailer');
+const LogEmail = require('../models/LogEmail');
+const colors = require('colors');
+
```
huss: Two formatting items:

1. There should be no white spaces at the end of lines (or as the entire line).
2. OED uses tab and not spaces for indenting.

If you are using VSC then you should be able to format the document to fix the white spaces. If you click on the item normally in the bottom, right of VSC window (it might say "Spaces: 2" or something like that) then you can change to using tabs and also convert the spaces to tabs. Let me know if you need help with this.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
...
+const colors = require('colors');
+  
+async function checkWebsite() {
+  const conn = getConnection();
+
+  try {
+    const response = await fetch('https://httpstat.us/500/', { method: 'HEAD' });
+
+    if (response.ok) {
+      console.log('The server is up'.green.inverse);
```
huss: Might be nice to include the URL tested in the message as done when it is down for logging.

In the larger scope, OED normally does not print to console.log nor directly call LogEmail. I suggest you look at src/server/log.js for the standard logging commands. When I tested this I had to do ``const { log : logger } = require('../log');`` since log is used in other ways here but that may go away when you don't directly call LogEmail. This will also help as changes are coming to how logging is done and using the standard functions means it should seamlessly integrate. If you agree these should be used then the colors package can be removed. It is nice but given its limited use it is not worth adding and not the standard OED way.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
...
+      const errorMessage = 'The server at https://httpstat.us/500/ is down.';
+
+      // Log the error in the database
+      const log = new LogEmail(undefined, errorMessage);
+      await log.insert(conn);
+
+      // Send a logging email
+      await logMailer(conn);
+    }
+  }
```
huss: Maybe put the catch right after the close brace to be consistent with the finally formatting. Most of the OED code base does it this way.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
+const cron = require('node-cron');
+const { getConnection } = require('../db');
+const { logMailer } = require('../logMailer');
+const LogEmail = require('../models/LogEmail');
+const colors = require('colors');
+  
+async function checkWebsite() {
+  const conn = getConnection();
```
huss: While I'm normally an advocate for getting the DB connection up front, in this case it might cause an issue. If OED is down (including the server) then I'm uncertain if this might fail. Should be it delayed until after the check and console.log for that reason? When console.log/err goes away then this may not be needed. I believe the standard logging code deals with this automatically and does not need a conn.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
...
+    // Send a logging email
+    await logMailer(conn);
+  } finally {
+    if (conn.close) {
+      await conn.close();
+    }
+  }
+}
+
+// Schedule the task to run every hour
```
huss: I like this idea of starting cron inside of OED code. What I'm wondering is how a site would easily run this. It worked fine by doing ``npm run status`` inside an OED Docker web container. However, it stopped if I terminated OED so it does not tell you OED is down when that happens. The only way run outside the Docker container that I know is to require installing node on the server. That is extra work and OED does not want to modify the server software. Do you know how to do this in another way or make sure it does not stop if OED is down? Up to now OED tells sites to set up a cron job via Linux/Unix on their server to run tasks. If this new way was easy to do then OED would consider using it and switching other cron uses to this.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const errorMessage = `Failed to reach https://httpstat.us/500/. Error: ${error.message}`;
+
+    // Log the error in the database
+    const log = new LogEmail(undefined, errorMessage);
+    await log.insert(conn);
+
+    // Send a logging email
+    await logMailer(conn);
+  } finally {
```
huss: My recollection is the conn comes from a pool of connections that can be reused. I cannot see anywhere else that conn.close is called so I don't think this is necessary. Let me know if you think otherwise.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,52 @@
...
+      // Log the error in the database
+      const log = new LogEmail(undefined, errorMessage);
+      await log.insert(conn);
+
+      // Send a logging email
+      await logMailer(conn);
+    }
+  } 
+  catch (error) {
+    console.error('Error:', error);
```
huss: I'm trying to understand what error would cause this. If it was the fetch then maybe this makes sense. On the other hand, if the error occurred while logging then it may happen again when that is tried. Could you help me understand better?

huss (changes requested):
Thanks to @lukeckk for working on this. It is a more comprehensive solution than I had thought about and that is nice. I know this is a draft PR but I'm providing some thoughts. I know there are a number of comments but that does not indicate that the work is not good and valuable. I just wanted to think about it holistically given the comprehensive nature of the solution. Know that I'm happy to discuss and get any thoughts you have.

src/server/services/checkWebsiteStatus.js
lukeckk: I included this catch error to catch non-http errors such as when internet is unavailable, invalid url, and etc. Not sure if this is necessary but I thought I would include it and get some feedback. What do you think?

src/server/services/checkWebsiteStatus.js
lukeckk: You are right that the script will stop running once your terminate the docker container. I initially did it think way in order to make it easy for user to run without having to enter cron commands.

To fix this, there are 2 ways to do it that I can think of:
1. Have user enter the cron command below, it will then run the script regardless docker container is open or not.

crontab -e
0 * * * * /usr/bin/node /absolute/path/to/src/server/services/checkWebsiteStatus.js >> /path/to/logfile.log 2>&1


2. Create another docker container that contains the cron job and have it run all the time, which is more complex to 1st option.

what do you think?

src/server/services/checkWebsiteStatus.js
lukeckk: I agree with you to move conn after verify website status.

But I believe I still need a conn to pass into logMail method as parameter

src/server/services/checkWebsiteStatus.js
lukeckk: What would be the url for production? is it https://openenergydashboard.org/index.html ?

src/server/services/checkWebsiteStatus.js
lukeckk: You are correct that I 'finally' block isnt necessary, I remove conn.close but forgot to remove this. Thanks for noticing

src/server/services/checkWebsiteStatus.js
huss: I agree that option 1 is better. Maybe take a look at our [site installation directions](https://openenergydashboard.org/helpV1_0_0/adminInstallation/) starting at step 10 to see how we already set up for cron jobs. OED provides a cron script that the site edits and then uses for cron. Let me know if you think that is good for this situation.

src/server/services/checkWebsiteStatus.js
huss: The URL is unique for each site. Thus, the side needs to be able to provide. If we do this via a cron script (see other comment) then it could be a value set in that script file. An alternative is to all it to be passed via arguments to the script but I don't know how off the top of my head for npm scripts.

src/server/services/checkWebsiteStatus.js
huss: See other comment about not directly calling logMail. If you use the OED standard logging method then no conn is needed as it takes care of this. It might help with this but let me know if not so we can figure out what to do.

src/server/services/checkWebsiteStatus.js
huss: That makes sense. If the code changes to use the standard OED logging then a lot of this will be minimized. Another thought is to limit the try/catch to the fetch so the source of the exception is clearer. A second try/catch could be used for other code but I don't know if that is needed here. If the ``response`` variable is declared before the try/catch then maybe it could be analyzed after the try/catch which you can make only execute if there is no exception to the catch block. What do you think?

src/server/services/checkWebsiteStatus.js
lukeckk: is the OED standard logging method located in 'sendLogEmail.js' ?? could you direct me to the file?

huss: Sorry. Check out src/server/log.js. It will print to console, add to log email, add date and add to log.txt (soon to go into DB). The code normally calls the info, warn, error methods that call the main one.

src/server/services/checkWebsiteStatus.js
lukeckk: Oh I thought there was only one URL we are verifying. If that is the case. how about using command-line arguments like below? 
Then users can type in the unique url after the cron command, the system will recognize the url as the second argument and store it as WEBSITE_URL variable and verify its status

command example:
0 * * * * src/server/services/checkWebsiteStatus.js https://example.com >> /path/to/logfile.log 2>&1


image

src/server/services/checkWebsiteStatus.js
lukeckk: I just took a look and it seems like they are the cron command in a bash file to simplify the process of running the script. One thing I dont understand is, take sendLogEmailCron.bash for example, the comment says 'This should be copied to /etc/cron.daily/'

So is there such path as '/etc/cron.daily/' in the codespace? i couldnt seem to find it

src/server/services/checkWebsiteStatus.js
huss: That is fine. In the OED base scripts to run cron jobs we often have values set at the top to provide local information. That is fine but it is also okay to use the command line arg if it is present and default to the one in the script if not.

src/server/services/checkWebsiteStatus.js
huss: That is in the Linux/Unix file system and not in the OED files/container. It is a common place where all scripts in that directory are automatically run at the frequency specified by that directory. In the one you gave it is every day. There is also an hourly directory. Using this guarantees it runs whenever the server is up (even after a restart). Help?

lukeckk: Ah ok i see. Should i create a new bash file and reuse the command and only change the file name then?

image

For example, I will replace 'sendLogEmail' with 'checkWebsiteStatus'. Of couse 'checkWebsiteStatus' will be in package.json with the relative path of 'checkWebsiteStatus.js'

huss: I think that is correct (I can verify when I see it). When you say replace I assume you mean you have a copy of one into a new file that you edit and leave the other one untouched.

lukeckk: Correct, I am creating a new file and leaving the other one untouched :)

src/server/services/checkWebsiteStatus.js
lukeckk: I see. Thank you for bringing this to my attention. 
I can picture how the script will look like using the existing OED functions. I will list out all the steps below and correct me if im wrong or if there is anything to change. I hope to get this merged by Monday 12/9 so I can showcase our work for my college :)

Step 1:
create a new file containing website checking function named 'checkWebsiteStatus.js' 

Step 2:
use js fetch function to verify website status by sending a head request. 

Step 3:
if website is down, call the function in log.js to start database connection and log it in the database

Step 4:
call function in sendLogEmail.js to send email to all users 

Step 5:
insert an object in package.json under 'scripts', with key being 'checkWebsiteStatus' and value being the relative path to 'checkWebsiteStatus.js' 

Step 6:
create a new bash file named 'checkWebsiteStatusCron.bash' containing the similiar format as below and replace 'sendLogEmail' with 'checkWebsiteStatus'

image

Step 7:
user can run the following cron command to set up hourly check

 chmod +x /etc/cron.hourly/checkWebsiteStatusCron.bash

huss: First, good luck with your showcase. You are welcome to show whatever you have done even if it is not yet merged. Having said that, we can continue to push to get this completed soon.

My only comment right now is steps 3 & 4. If the code calls the error function in src/server/log.js then it will log to appropriate places (including the upcoming change to support an admin page to look at log message) and prepare the message to send via email if appropriate. I don't think you need to do step 4 because properly doing step 3 will cover that. Does this make sense?

lukeckk: You are correct. I got myself confused since I didnt see 'sendLogEmail' function being implemented, but it is running as a cron job to check the database for appropriate message and send.

That makes sense. I will skip step 4 then.

lukeckk: Just submitted another pull request with the changes we discussed. The commands for user to execute the cron job and run the scripts are as below:

chmod +x /etc/cron.hourly/checkWebsiteStatusCron.bash

/etc/cron.hourly/checkWebsiteStatusCron.bash www.example.com

src/scripts/checkWebsiteStatusCron.bash
```diff
@@ -0,0 +1,16 @@
+# This aggregates the readings data at hour level.
```
huss: Could you please add the copyright notice similarly to src/scripts/checkMeters.sh.

Also, the comment should be updated to reflect what the script does.

src/scripts/checkWebsiteStatusCron.bash
```diff
@@ -0,0 +1,16 @@
...
+# This should be copied to /etc/ or /etc/cron.hourly/ or /etc/cron.daily and the copy renamed so that its function will be clear to admins.
+
+# Check if a URL is provided as an argument
+if [ -z "$1" ]; then
+    echo "Error: No URL provided. Usage: $0"
+    exit 1
```
huss, line 14: I think this message would be better as ``echo "Error: No URL provided. Usage: $0 URL-to-check"``.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+	const cron = require('node-cron');
+	const { getConnection } = require('../db');
```
huss: I don't think this is needed any more.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,28 @@
...
+	const { log } = require('../log');
+	const { logMailer } = require('../logMailer');
```
huss: I don't think this is needed any more.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,28 @@
...
+	const LogEmail = require('../models/LogEmail');
```
huss: I don't think this is needed any more.

src/server/services/checkWebsiteStatus.js
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+	const cron = require('node-cron');
```
huss: I think this was removed from usage. I also needs to be uninstalled so it is gone from the package files.

src/server/services/checkWebsiteStatus.js
huss: The indenting is still off and some extra spaces in places. Would you like me to fix this up or can you?

huss:
I'm leaving a comment that I realized the site person may not get notified if the site is down. If the server is in a bad state then the log email may not be sent. This is not different from other log emails but noting it here since it means the hope to notify the situation is lost. I'm okay with this situation and I think it may require creating code separate from OED to avoid this.

huss (changes requested):
Thanks to @lukeckk for addressing the comments. Testing found it works as expected. I have made a few comments to finalize the code cleanup. I'm sorry we did not get this merged for your presentation but it should be close. Please let me know if I can help.

lukeckk:
@huss Thank you very pointing out the comments and the unnecessary imports. I have made some updated according to your feedback. Regarding the indenting, could you fix it for me please? I gave it a few attempts but cant seem to get the spaces right. Thank you so much

huss (approved):
Thanks to @lukeckk for quickly addressing all the comments. Everything now looks good. I am also noting that you provided documentation to me and I plan to add it to the website. Congratulations on another accepted contribution.

Issue #905, opened 2023-05-20
protect admin when deleting a conversion


huss:
### Is your feature request related to a problem? Please describe.

Admins can delete a conversion. This can cause undesirable effects including:

- If this is the only conversion for a meter using that unit then you can no longer graph it.
- You can orphan a unit if this is the only conversion to it.
- This can change the compatible units. This might cause a default graphic unit for a meter to become invalid. For a group, it could make the default graphic unit or the members invalid.

There may be other cases but hopefully this covers them.

### Describe the solution you'd like

Checks need to be added to warning the admin. As was done for groups, the default graphic unit is set to no unit if needed but not allowed if there is no remaining default graphic unit.  If members of a group become invalid then the change is not allowed (as is done in group edit page).

### Describe alternatives you've considered

The admin is on their own and that is not the philosophy of OED.

### Additional context

None

huss:
After talking with @mrathgeber and thinking about this, I have decided that deleting a conversion involving any unit that is compatible with any used meter unit is problematic. This same type of issue (but more complex here) came up about deleting a unit (see issue #1020) and it was decided you cannot delete a unit that is used in places. The same ideas will be used here. 

The basic outline of a solution is:

1. Use Redux meter state to find all the meter ids that are used. Technically if two meters have the same meter unit then you only need to check one of them. However, this should be fast without this, easier and will give the admin better warning messages.
2. Use unitsCompatibleWithMeters() in src/client/app/utils/determineCompatibleUnits.ts where supply the meter ids just found  to the function. In principle you could do them all at once but to make the messages for the admin better the meter ids are done one at a time (each argument is a set of size 1).
3. See if the returned set of compatible units contains the source unit id for the conversion that is being deleted. If so, it may be an issue to delete the conversion so it will not be allowed. Add this meter to the message. The start of the message can be something along the lines of "Deleting this conversion may cause the compatible units to change on the following meter(s)  so deleting it is not allowed".
4. Once all meters have been processed, pop up the message if any meter is an issue. See validateGroupPostAddChild() in src/client/app/components/groups/EditGroupModalComponent.tsx for how it looks and works there as what is desired. In this case the delete is not executed. Otherwise the delete happens.

A few notes:

- Checking all meters means that any groups containing this meter are indirectly protected. However, the admin will not know about these groups in the message. That could be added if desired.
- It is possible that the delete would be okay because the source unit was also included due to a different conversion. The proposed solution is conservative to avoid any possible issue.
[...]

PR #1400, opened 2024-12-03
Added Admin Warnings For Deleting Conversions

danielshid:
# Description

I added warnings to inform admins of the impact their actions will have on meter units and possible graphing units when choosing to delete conversions.

Partially Addresses #905

## Type of change

## Checklist

## Limitations

I was not able to cover cases for default graphic units and groups.


src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 	};
 	/* End State */
 
+	const getConversionCount = (unit: UnitData, conversions: ConversionData[]) => {
```
huss, line 72: Each function should have a comment (similar to JSDoc) to tell what it does and the params. It would also be nice to have comments inside the functions to tell what is being done in the big picture sense.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+		let count = 0;
+		if (unit === unitDataById[state.sourceId]) {
```
huss: I think these types of compare of entire unit objects work but OED more commonly uses the .id for the check.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+		let msg = '';
+		let cancel = false;
+		if (source.typeOfUnit === UnitType.meter) {
+			const count = getConversionCount(source, conversionDetails);
+			if (count === 1) {
+				msg += `Deleting this meter conversion will orphan ${unitDataById[state.destinationId].name}.\n`;
+			} else {
+				msg += 'Deleting this meter conversion will make the following meter(s) ungraphable:\n';
+				for (const meterId of Object.values(meterDataById)) {
+					if (meterId.unitId == source.id) {
```
huss: Should this be ===?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+
+	const checkState = () => {
+		const source = unitDataById[state.sourceId];
+		const dest = unitDataById[state.destinationId];
+		let msg = '';
+		let cancel = false;
+		if (source.typeOfUnit === UnitType.meter) {
+			const count = getConversionCount(source, conversionDetails);
+			if (count === 1) {
+				msg += `Deleting this meter conversion will orphan ${unitDataById[state.destinationId].name}.\n`;
```
huss: All strings should be translated.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			} else {
+				msg += 'Deleting this meter conversion will make the following meter(s) ungraphable:\n';
```
huss: I'm not sure the logic is what was desired. If there is only this conversion and there are meter(s) using the source unit then tell the admin about each meter and cancel the delete. If there are multiple conversions then tell the admin about each meter using the source unit and warn that the number of graphable units for the meter is reduced.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			const count = getConversionCount(source, conversionDetails);
+			if (count === 1) {
+				msg += `Deleting this suffix conversion will disable use of ${source.name}.\n`;
+			}
+		} else if (source.typeOfUnit === UnitType.unit && dest.typeOfUnit === UnitType.unit) {
+			const destCount = getConversionCount(dest, conversionDetails);
+			if (destCount === 1) {
+				msg += `Deleting this unit conversion will orphan ${unitDataById[state.destinationId].name}.\n`;
+			}
+			if (state.bidirectional) {
```
huss: I think the description of the cases I gave you was confusing. What needs to be determined (including the check right above) is if there is only this conversion for the unit. To check this, the code needs to check if that unit is a source or if the conversion is bidirectional then it can be the destination (excluding the one being deleted). The idea is if it is the destination then it can could get to another unit if allowed which means it must be bidirectional. This check is looking at the bidirectional status of the one being deleted and not the other possibilities.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+				const sourceCount = getConversionCount(source, conversionDetails);
+				if (sourceCount === 1) {
+					msg += `Deleting this unit conversion will orphan ${unitDataById[state.destinationId].name}.\n`;
+				}
+			}
+			if (msg === '') {
+				msg += 'Deleting this unit conversion between two units of type unit will have consequences.\n';
```
huss: I think change "will" to "may" since it can happen. Overall, I think the message might be better as: "Deleting this conversion between two units of type unit may reduce the possible graphing units for some meter(s) but cannot be known until after the deletion happens." I also would like a TODO put here to indicate that OED should try to add a check after the deletion to see if a change happens to notify the admin of the consequence of this delete. In a perfect world the admin could ask to have the delete undone.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			}
+		}
+
+		if (msg === '') {
+			handleDeleteConfirmationModalOpen();
+		} else if (cancel) {
+			setDeleteConfirmationMessage(msg + 'This conversion cannot be deleted.\n');
```
huss: Could the new part of the msg be: "This conversion cannot be deleted until the significant consequences of doing this are addressed."

src/client/app/styles/modal.css
```diff
@@ -17,4 +17,8 @@ select option[value="false"] {
 
 select option[value="true"] {
 	color: green;
-}
\ No newline at end of file
+}
+
+.confirmation-message {
+	white-space: pre-wrap;
+  }
```
huss: Please remove extra spaces at start of line.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -59,10 +63,80 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			}
+		} else if (source.typeOfUnit === UnitType.suffix) {
+			const count = getConversionCount(source, conversionDetails);
+			if (count === 1) {
+				msg += `Deleting this suffix conversion will disable use of ${source.name}.\n`;
+			}
+		} else if (source.typeOfUnit === UnitType.unit && dest.typeOfUnit === UnitType.unit) {
+			const destCount = getConversionCount(dest, conversionDetails);
+			if (destCount === 1) {
+				msg += `Deleting this unit conversion will orphan ${unitDataById[state.destinationId].name}.\n`;
```
huss: Could we reword so it has "will orphan unit <unit name>" (so adding word unit). Also, generally, it would be nice if there were quotes around the name in all messages since unit names can have spaces and multiple words.

huss (changes requested):
Thanks to @danielshid for working on this issue. Overall, it is fine. I have made some comments about details and style for you to consider. Given the cases are complex, this is not surprising. I did some testing and it works as desired (modulo the comments). I'll do a final check after the comments are addressed. Let me know if I can help in any way.

huss:
I've modified the code to fix several issues that careful testing found with the design given in the issue on GitHub. This was my fault and not in the code provided. I'm going to let this sit a little while in case anyone else (including @danielshid who did this PR) has thoughts on the modifications.

huss (approved):
Thanks to @danielshid for continued work on this after the time expected. I'm sorry this sat for a couple of weeks while I was focused elsewhere.

All your changes looked good and worked as requested. Please see my comment on a commit I made to update the behavior that is outside what was requested.

I'm approving the work but will wait to merge per my other comment. Congratulations to @danielshid for another successful contribution to OED.

PR #1402, opened 2024-12-03
Is valid gps input new & setLayout

zhenga5:
# Description

Changing IsValidGPSInput to include returning a message about why there is something invalid particularly involving GPS inputs whether the input lacks a comma, has too many commas, or the GPS coordinate is out of range. Additionally changed setLayout to a separate utils file and imported it to existing files that use the setLayout function or variations of it.

## Type of change

## Checklist

## Limitations

Shouldn't be any issues regarding IsValidGPSInput. It only changed the return type to {boolean, string}


src/client/app/utils/calibration.ts
```diff
@@ -106,25 +107,33 @@ export function itemDisplayableOnMap(size: Dimensions, point: CartesianPoint): b
  * @param input The string to check for GPS values
  * @returns true if string is GPS and false otherwise.
  */
-export function isValidGPSInput(input: string): boolean {
+export function isValidGPSInput(input: string){
+	let message = '';
+	let validGps = true;
 	if (input.indexOf(',') === -1) { // if there is no comma
 		// TODO It would be nice to tell user that comma is missing but need to check all uses to be sure don't get ''.
```
danielshid: Remove TODO comment.

src/client/app/utils/calibration.ts
```diff
@@ -106,25 +107,33 @@ export function itemDisplayableOnMap(size: Dimensions, point: CartesianPoint): b
...
+	if(validGps){
+		// Works if value is not a number since parseFloat returns a NaN so treated as invalid later.
+		const array = input.split(',').map((value: string) => parseFloat(value));
+		const latitudeIndex = 0;
+		const longitudeIndex = 1;
+		const latitudeConstraint = array[latitudeIndex] >= -90 && array[latitudeIndex] <= 90;
+		const longitudeConstraint = array[longitudeIndex] >= -180 && array[longitudeIndex] <= 180;
+		const result = latitudeConstraint && longitudeConstraint;
+		if (!result) {
+			// TODO It would be nice to return the error and then notify as desired.
```
danielshid: Remove TODO comment.

danielshid:
I noticed that the invalid GPS warnings appear only after the user saves the meter. The meter still stores the invalid GPS input. Unless Steve thinks otherwise, I think the warnings should appear as the user is editing the meter.

danielshid:
The pull request also still has text in ( ).

huss:
> I noticed that the invalid GPS warnings appear only after the user saves the meter. The meter still stores the invalid GPS input. Unless Steve thinks otherwise, I think the warnings should appear as the user is editing the meter.

The potential issue with checking in real time is that you will get errors while typing the GPS value. I think that is why we waited until the save is done. I think it is fine to leave it the way it is.

src/client/app/components/maps/MapCalibrationInfoDisplayComponent.tsx
```diff
@@ -88,7 +88,8 @@ export default class MapCalibrationInfoDisplayComponent extends React.Component<
 		const longitudeIndex = 1;
 		if (this.props.currentCartesianDisplay === 'x: undefined, y: undefined') { return; }
 		const input = this.state.value;
-		if (isValidGPSInput(input)) {
+		const {validGps} = isValidGPSInput(input);
+		if (validGps) {
```
huss, line 92: This isn't your doing but would it work if you had an else clause that showed the error if the GPS is invalid? Up to now the code simply does nothing.

huss (changes requested):
Thanks to @zhenga5 for working on both of these TODO issues. Review and testing found it works fine and addressed the comments by @danielshid.

I cannot comment inside file without changes so here is a thought:

The isValidGPSInput in src/server/services/csvPipeline/uploadMeters.js is very similar to the modified one you changed in src/client/app/utils/calibration.ts. It was not your doing that there are two version but I think it would be good to remove the other one that is scoped to that file and use the modified one from your work.

I also made one other comment. Note both of these were not part of the original issue but related when I looked at the code. If you are willing to make these changes that would be very helpful. If not, then let me know and I can do them.

huss:
Note I pushed a commit to address formatting issues where most are outside the PR changes but in the same files.

huss:
> The isValidGPSInput in src/server/services/csvPipeline/uploadMeters.js is very similar to the modified one you changed in src/client/app/utils/calibration.ts. It was not your doing that there are two version but I think it would be good to remove the other one that is scoped to that file and use the modified one from your work.

I realized one was on the server so JS and not TS. Thus, I synchronized them and put in a TODO.

huss:
I addressed all the comments.

huss (approved):
All good now.

Issue #1217, opened 2024-04-09
CSV meter upload validates values


huss:
### Is your feature request related to a problem? Please describe.

The create meter page stops admins from entering bad values by constraining choices and/or checking before saving. It would be good if the CSV meter upload page did similar checks.

### Describe the solution you'd like

See above.

### Describe alternatives you've considered

No checking

### Additional context

None

mikedivine:
I will work on this issue.

ethanernst11:
I can work on this issue.

huss:
> I can work on this issue.

I think this is okay since @mikedivine is not working on this at this time.  If you need any help on this then let me know.

ethanernst11:
I have a quick question. Where can I find the code for the create meter page that stops admins from entering bad values?

huss:
> I have a quick question. Where can I find the code for the create meter page that stops admins from entering bad values?

See src/client/app/components/meters/CreateMeterModalComponent.tsx (there is an edit equivalent). There are a number of checks here that are done in different ways (different ways is not needed for this issue). As examples, it validates GPS, can't have cumulative reset if not cumulative, set identifier if name given, no blank name, .... Let me know if you need anything else.

ethanernst11:
Is the csv meter upload also creating a meter?

huss:
> Is the csv meter upload also creating a meter?

In general, the CSV meter upload creates meter(s). It is possible to edit a single meter but that is less common. If you have not seen it yet, you may find the documentation at https://openenergydashboard.org/helpV1_0_0/adminCsvImport/ useful. It is for v1.0.0 so is a little out of date with development but still useful.

ethanernst11:
I apologize but this ended up being more complicated than expected and I do not have enough time currently to work on it. So I will not be able to continue working on this issue for finish it. I am sorry for any inconvenience.

huss:
> I apologize but this ended up being more complicated than expected and I do not have enough time currently to work on it. So I will not be able to continue working on this issue for finish it. I am sorry for any inconvenience.

It's fine and thanks for letting us know. If you want to work on something in the future then please just let us know (and we can help find a task that suits your situation if that helps).

SageMar:
If it's alright, @Cmatthews444 and I will tackle this issue

huss:
[...]

PR #1403, opened 2024-12-05
Issue 1217 - csv upload validation

SageMar:
# Description
"The create meter page stops admins from entering bad values by constraining choices and/or checking before saving. It would be good if the CSV meter upload page did similar checks." After a talk with Steve, it was decided that this validation should be done in JavaScript on the back-end, so these validations are inside the uploadMeters.js file.

Done in collaboration with @cmatthews444 

Partly Addresses #1217

This adds validations for the following variables: 
- Enabled
- Disabled
- Type
- Timezone
- Area
- Cumulative
- Reset
- Timesort
- End Only
- Area Unit
- Min Val
- Max Val
- Disabled Checks

## Type of change

## Checklist

## Limitations
This does not cover every single validation done on a CSV file - it only covers those listed above.


src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +293,64 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+	for (const [index, name] of Object.entries(booleanFields)) {
+		let value = meter[index];
+
+		// allows upper/lower case.
+		if (typeof value === 'string') {
+			value = value.toLowerCase();
+		}
+
+		// Validates read values to either false or true
+		if (value !== 'true' && value !== 'false' && value !== true && value !== false) {
```
huss: I know this is a draft PR but a number of the tests were failing. I debugged one of them. The issue is that endOnlyTime on a meter can be undefined. In src/server/sql/meter/create_meters_table.sql it will set it to a default value if undefined. Thus, some tests need to allow for this.

I also noticed that a few inputs go through normalizeBoolean() that allows for yes/no in addition to true/false. FYI.

Please let me know if you need any more information or help.

SageMar: Thank you for those comments, they're very helpful. We took a short break over the holidays so we'll get to work resolving those ASAP. 

Happy new year!

huss:
@SageMar & @cmatthews444 I saw you made some changes since my comments and wanted to see where this PR is at. Are you planning to do more work on this, do you need any help or is that status something else? Thanks.

SageMar:
@huss We are definitely still here finishing it up. It's been difficult to find the time between other projects this quarter, but we're aiming to fix the last errors and at least get in the validations we listed in the opening post for this PR. 

When we last left off with the code, we had 3 errors still occurring which we were working to figure out what they mean. It would actually help a ton if you're able to give clarification on them as they're not quite as straightforward as the others we got. If it's easier to chat about this over Discord we can schedule a time on Friday (if you're available, I believe we can both make ourselves available any time on Friday other than 1pm PST), but if they're not too complicated to sum up here that would be great. 

1) Meter testing files starting 'pipe100' doing 'Second meter upload where incorrectly provides meter identifier so fails' with 2 

2) Meter testing files starting 'pipe101' doing 'Second meter with same name so fails but first meter exists' with 1 requests

3) Meter testing files starting 'pipe103' doing 'Uploading meters using Email. First succeeds then the second meter upload fails because it incorrectly provides meter identifier' with 2 requests

huss:
@SageMar 

> 1. Meter testing files starting 'pipe100' doing 'Second meter upload where incorrectly provides meter identifier so fails' with 2
> 3. Meter testing files starting 'pipe103' doing 'Uploading meters using Email. First succeeds then the second meter upload fails because it incorrectly provides meter identifier' with 2 requests

Both of these failed because it expected OED to report a success code (200) but it received an error code (400). I would debug this by adding a console.log to the code to print out the response text since that might indicate what went wrong.

> 2. Meter testing files starting 'pipe101' doing 'Second meter with same name so fails but first meter exists' with 1 requests

This one is curious because the returned error is ``Failed to upload meters due to internal OED Error: For meter pipe101 the area entry of 33 is invalid. Area must be a number greater than 0.`` but the value of 33 is greater than 0. I'm guessing you added a test that the area is valid. It may not be working as desired and/or it is returning new info so the test needs to be updated. I would first want to figure out why the test seems to have a message that is inconsistent on 33 > 0.

Do you want to work on these yourselves? I can also help (either working with you or by myself). Let me know what you prefer and if you want/need more information.

SageMar:
> Do you want to work on these yourselves? I can also help (either working with you or by myself). Let me know what you prefer and if you want/need more information.

@huss 

I think we'll at least give it a try ourselves with that information. If we're still stuck by Tuesday, I'll reach out again! Thank you!

SageMar:
@huss In case you were curious about what caused the issues, it seems the confusing error on pipe101 was what effected all of them. For come reason those files in particular had the number under area coming in as undefined, so using parseInt() on the value ended up fixing all three. 

It passed tests and appears to be working for everything on our end, but let us know if there's anything else that needs to be changed! Thanks for always being so helpful.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,21 +42,71 @@ async function uploadMeters(req, res, filepath, conn) {
...
 				}
 				// Need to reverse latitude & longitude because standard GPS gives in that order but a GPSPoint for the
 				// DB is longitude, latitude.
 				meter[6] = switchGPS(gpsInput);
 			}
 
+			// Verify area unit - have to parse float to ensure it can read in integer numbers like "33"
+			const areaInput = parseFloat(meter[9]);
+			if (areaInput) {
+				if (!isValidArea(areaInput)){
```
huss: This is a small point but here and a few other lines don't have spaces: ``){`` vs ``) }``. Formatting with VSC should clear this up. I can do it if you prefer.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -8,17 +8,18 @@ const Meter = require('../../models/Meter');
 const readCsv = require('../pipeline-in-progress/readCsv');
 const Unit = require('../../models/Unit');
 const { normalizeBoolean } = require('./validateCsvUploadParams');
+const { type } = require('os');
```
huss: I think this include may have been accidentally added as I don't think it is used.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -152,6 +203,78 @@ function switchGPS(gpsString) {
...
+	}
+}
+
+/**
+ * Checks if the area unit provided is an option
+ * @param areaUnit the provided area for the meter
+ * @returns true or false
+ */
+function isValidAreaUnit(areaUnit) {
+	const validTypes = ['feet', 'meters', 'none'];
```
huss: I think using the standard enum would be better. See src/server/models/Unit.js for ``areaUnitType``. This will allow it to adjust if the allowed values are changed. Also, the error msg above with these values should be constructed by looping over the enum. Also for potential future changes the comment(s) listing the values could be updated about the enum rather than specifics.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -152,6 +203,78 @@ function switchGPS(gpsString) {
...
+}
+
+/**
+ * Checks if the time sort value provided is accurate (should be increasing or decreasing)
+ * @param timeSortValue the provided time sort
+ * @returns true or false
+ */
+function isValidTimeSort(timeSortValue) {
+	// must be one of the three values
+	if (timeSortValue == 'increasing' || timeSortValue == 'decreasing'){
```
huss: Similar to above about enum, see src/server/services/csvPipeline/validateCsvUploadParams.js for ``MeterTimeSortTypesJS``.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -152,6 +203,78 @@ function switchGPS(gpsString) {
...
+	}
+}
+
+/**
+ * Checks if the meter type provided is one of the 5 options allowed when creating a meter.
+ * @param meterTypeString the string for the meter type
+ * @returns true or false
+ */
+function isValidMeterType(meterTypeString) {
+	const validTypes = ['egauge', 'mamac', 'metasys', 'obvius', 'other'];
```
huss: Same enum and see src/server/models/Meter.js for `Meter.type``.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +293,78 @@ async function getUnitId(unitName, expectedUnitType, conn) {
 	return unit.id;
 }
 
-module.exports = uploadMeters;
+
```
huss, line 301: Maybe remove second blank line.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -152,6 +203,78 @@ function switchGPS(gpsString) {
...
+/**
+ * Checks the provided time zone and if it is a real time zone.
+ * @param zone the provided time zone from the csv
+ * @returns true or false
+ */
+function isValidTimeZone(zone) {
+	// check against the built in timezones, must use a try catch since it does not return a boolean
+	try {
+		new Intl.DateTimeFormat(undefined, {timeZone : zone});
```
huss: The src/client/app/components/TimeZoneSelect.tsx front-end component uses ``moment.tz.names()`` to get the name choices. It would be good to be consistent and use that here if we can. It may mean the try/catch is not needed.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +293,78 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
 
-module.exports = uploadMeters;
+
+/**
+ * Validates all boolean-like fields for a given meter row.
+ * @param {Array} meter - A single row from the CSV file.
+ * @param {number} rowIndex - The current row index for error reporting.
+ */
+function validateBooleanFields(meter, rowIndex) {
+// all inputs that involve a true or false all bieng validated together.
```
huss: bieng ->> being

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +293,78 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+	}
+}
+
+function validateMinMaxValues(meter, rowIndex) {
+	const minValue = Number(meter[27]);
+	const maxValue = Number(meter[28]);
+
+	if (isNaN(minValue)){
+		// do nothing, pass it through
+	} else if (isNaN(minValue) || minValue < -9007199254740991 || minValue >= maxValue){
```
huss: I think OED is allowing minValue = maxValue. This applies below.

This did raise the question in my mind that the maxValue is used before it is validated. Should they be checked for isNan, isNumber first, then min/max value and finally if min > max?

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,21 +42,71 @@ async function uploadMeters(req, res, filepath, conn) {
...
+				if (!isValidTimeSort(timeSortValue)){
+					let msg = `For meter ${meter[0]} the time sort ${timeSortValue} is invalid. Valid options are increasing or decreasing.`;
+					throw new CSVPipelineError(msg, undefined, 500);
+				}
+			}
+
+			const timezone = meter[5];
+			if (timezone){
+				if (!isValidTimeZone(timezone)){
+					let msg = `For meter ${meter[0]}, ${timeSortValue} is not a valid time zone.`;
```
huss: I think ``timeSortValue`` should be for time zone instead.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,21 +42,71 @@ async function uploadMeters(req, res, filepath, conn) {
...
+					let msg = `For meter ${meter[0]} the gps coordinates of ${gpsInput} are invalid.`;
 					throw new CSVPipelineError(msg, undefined, 500);
 				}
 				// Need to reverse latitude & longitude because standard GPS gives in that order but a GPSPoint for the
 				// DB is longitude, latitude.
 				meter[6] = switchGPS(gpsInput);
 			}
 
+			// Verify area unit - have to parse float to ensure it can read in integer numbers like "33"
+			const areaInput = parseFloat(meter[9]);
```
huss: Just noting you parseFloat before the check for being a number and if it is undefined.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,21 +42,71 @@ async function uploadMeters(req, res, filepath, conn) {
...
+			if (areaInput) {
```
huss, line 68: This check can fail in certain circumstances. I tested and 0 (zero) and -1 returns false so the area check does not happen. I also notice that 0.1 or any value with a decimal part does not work but it should. Finally, this allowed letters for numbers to pass the test. This may impact other tests you added. I did not try a fix so let me know if you have issues in resolving.

huss (changes requested):
Thanks to @SageMar & @cmatthews444 for continued work on this issue. I reviewed/tested and made some comments. Please let me know if you need any help with this.

cmatthews444:
@huss We are planning on working on this later this week, will reach and let you know if we need help from there.

SageMar:
@huss Do you have any time to meet with us on Friday in the Discord channel? Just looking to get a little clarification on some of the fixes. We're both available anytime after 11AM PST. Thanks!

huss:
> @huss Do you have any time to meet with us on Friday in the Discord channel? Just looking to get a little clarification on some of the fixes. We're both available anytime after 11AM PST. Thanks!

@SageMar Sure. Would 12:00 PT work for you?

SageMar:
> >
> 
> @SageMar Sure. Would 12:00 PT work for you?

Yup, that works great! See you then

huss:
@SageMar & @cmatthews444 I wanted to find out the status of this PR. Is it ready or is more work needed? Sorry that I lost track of this.

SageMar:
> @SageMar & @cmatthews444 I wanted to find out the status of this PR. Is it ready or is more work needed? Sorry that I lost track of this.

Yes, it is ready! I had forgotten to push up the last change to make timezone validation use moment.tz, but that's included now. I believe we fixed all the comments you had before. Thank you for your patience with us!

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +298,78 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+
+function validateMinMaxValues(meter, rowIndex) {
+	const minValue = Number(meter[27]);
+	const maxValue = Number(meter[28]);
+
+	if (isNaN(minValue) && isNaN(maxValue)) {
+		// do nothing, pass it through
+	} else if (isNaN(minValue) || minValue < -9007199254740991 || minValue > maxValue) {
+		throw new CSVPipelineError(
+			`Invalid min/max values in row ${rowIndex + 1}: min="${meter[27]}", max="${meter[28]}". ` +
```
huss, line 356: Here and below you directly use the meter value. Since you assigned it a variable above I think it would be best to use that to be consistent.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +298,78 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+			);
+		}
+	}
+}
+
+function validateMinMaxValues(meter, rowIndex) {
+	const minValue = Number(meter[27]);
+	const maxValue = Number(meter[28]);
+
+	if (isNaN(minValue) && isNaN(maxValue)) {
```
huss, line 352: I feel the logic is tricky. For example, assume only maxValue is a NaN. The first else if will compare minValue > maxValue in the first else if but maxValue is a NaN. Also, the second if below stops the else if so I don't see how the exception can occur in this case. Maybe this logic is better and simpler to understand (open to ideas):

```js
// Check minValue is okay. If maxValue is a NaN then skip last check but it will throw an exception below.
if (isNaN(minValue) || minValue < -9007199254740991 || (!isNaN(maxValue) && minValue > maxValue)) {
    exception
}
// Check maxValue is okay.
// A check for minValue > maxValue is not needed since it would have caused an exception in the
// if above unless minValue or maxValue is a NaN but those cause the exception anyway.
if (isNaN(maxValue) || maxValue > 9007199254740991) {
    exception
}
```

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -170,4 +298,78 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+ */
+function validateBooleanFields(meter, rowIndex) {
+	// all inputs that involve a true or false all being validated together.
+	const booleanFields = {
+		2: 'enabled',
+		3: 'displayable',
+		10: 'cumulative',
+		11: 'reset',
+		18: 'end only',
+		32: 'disableChecks'
```
huss, line 315: It appears your branch is out of date with development. The definition of disable checks was changed to be an enum. Your code needs to be updated to reflect this. Note you need to merge development and update your DB to do this so let me know if you need help.

huss (changes requested):
Thanks to @SageMar & @cmatthews444 for continued work on this issue. The changes are good where I made a couple of comments to address.

I did systematic testing of the code and found these checks are not being made:

- gap should be zero or greater.
- variation should be zero or greater.
- duplicates should be 1 to 9.
- if area unit is none then area cannot be specified
- max error must be 0 to 75
- the max date cannot be before the min date; a bonus would be to validate they are valid dates but the check may fail anyway if they are not and that is fine

I realize this is a non-trivial number of changes so let me know if doing them is an issue.

huss:
@SageMar & @cmatthews444 I saw your commit on 7 June and appreciate your long efforts on this work. I wanted to see if are or are planning to continue work on this. Please leave a comment with any information. If OED does not hear anything by 1 Aug. then we will try to see about someone else finishing up the needed items to address.

SageMar:
> @SageMar & @cmatthews444 I saw your commit on 7 June and appreciate your long efforts on this work. I wanted to see if are or are planning to continue work on this. Please leave a comment with any information. If OED does not hear anything by 1 Aug. then we will try to see about someone else finishing up the needed items to address.

Thank you for your patience with us through this whole process - I had hoped we would be able to do that last round of changes, but unfortunately we have both been extremely busy. It may be best to pass it on to someone else now in order to get it done in a reasonable time frame. 

Thank you again for working with us for so long!

huss:
Thanks for the quick response. I've unassigned the issue and will look for someone else to work on this.

src/server/services/csvPipeline/uploadMeters.js
huss: The updated PR has other thoughts on this so it is resolved.

PR #1408, opened 2024-12-10
Issue #590 Automatic Site Backup

### LINKED ISSUE #590: automatic site backup
We have a shell command to backup a database by using:
`sudo docker-compose exec database pg_dump -U oed > /home/eddy/Database_Dumps/<site name>/dump_$(date +%Y-%m-%d"_"%H_%M_%S.sql)`
It would be nice to have a script provided by OED to do this so it could be put into a cron job to happen regularly. Would also want it described somewhere on the admin and/or install pages.

vladoleksiyenko:
# Description

Fixed the script that was previously mentioned in the closed PR 759. This along with the cron job template that will be mentioned in the #590 issue will allow automatic backup of the website.

Fixes #590

## Collaborators
- Matt Miss @MattMiss
- Vlad Oleksiyenko @vladoleksiyenko
- Brandon Viorato @brandonviorato
- Eugene Faison @Gene7Him

## Type of change

## Checklist

- [x] I have followed the [OED pull request]
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement] and each author is listed in the Description section.

## Limitations

The cron template will be mentioned in the comments, that documentation will be need to be added to the webpage.

### DIFF
--- a/src/scripts/backupScript.sh
+++ b/src/scripts/backupScript.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# 
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+# Input the pathname for the desired backup directory
+# ie PATH="/home/<your_username>/path_to/database_dumps/dump.sql"
+# This path MUST exist
+
+# This could probably be programmatically populated. Currently needs to be set manually
+db_dump_path="/home/<your_username>/database_dumps" #INPUT REQUIRED
+
+# Generate a timestamp to append to the dump file. This line has errors at the moment.
+date=`date +%Y-%m-%d_%H_%M_%S`
+
+# Set the final path for the backup file
+final_path="${db_dump_path}/dump_${date}.sql"
+
+# Perform the backup using pg_dump
+docker compose exec database pg_dump -U oed > "$final_path"
--- a/src/scripts/backupScript.sh
+++ b/src/scripts/backupScript.sh
@@ -3,6 +3,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 # The OED Docker database container must be running for this script to work
 
 # Input the pathname for the desired backup directory
@@ -27,3 +28,5 @@ final_path="${db_dump_path}/dump_${date}.sql"
 
 # Perform the backup using pg_dump
 docker compose exec database pg_dump -U oed > "$final_path"
+
+echo "OED database backup placed in ${final_path}"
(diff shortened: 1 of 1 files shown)


src/scripts/backupScript.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# 
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+# Input the pathname for the desired backup directory
+# ie PATH="/home/<your_username>/path_to/database_dumps/dump.sql"
```
huss: This isn't really an explicit path but a file location including the path. Maybe update the comment. Maybe also note that it will be the file listed but with the date and give an example.

Also, if the directory does not exist then this fails. At a minimum there should be a comment if that is the case. An alternative is to check if there it exists and create if it does not. Helpful for the first time this runs.

src/scripts/backupScript.sh
```diff
@@ -0,0 +1,22 @@
...
+# This path MUST exist
+
+# This could probably be programmatically populated. Currently needs to be set manually
+db_dump_path="/home/<your_username>/database_dumps" #INPUT REQUIRED
+
+# Generate a timestamp to append to the dump file. This line has errors at the moment.
```
huss: This is working fine for me. Is there still an issue you see?

src/scripts/backupScript.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# 
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+# Input the pathname for the desired backup directory
```
huss, line 9: It might be good to put in a comment at the start that the OED database Docker container must be running for this to work. It also assumes it is the only database container running (seems likely).

huss (changes requested):
Thanks to @MattMiss, @vladoleksiyenko, @brandonviorato & @Gene7Him for addressing this issue. Testing found it mostly works as expected. I make a few comment to consider. ~~Also, OED is likely to create a cron job template to allow sites to automatically run this script at desired intervals. Let me know if you want to be involved in that and I can give more details.~~ (removed since I realized you had put a file on the issue)

src/scripts/backupScript.sh
vladoleksiyenko: Added an if statement that checks to see if the file path exists, and lets the user know if it exists. If not, it attempts to create the database_dumps directory. If this is unsuccessful, a message will then be displayed saying it failed. Thank you for bringing this up.

src/scripts/backupScript.sh
vladoleksiyenko: Apologies, this was a comment that we forgot to remove originally, there aren't any issues with this. Thank you for noticing that.

src/scripts/backupScript.sh
vladoleksiyenko: I've included a message at the top that mentions this now, thank you for that suggestion.

huss (approved):
Thanks for addressing all the comments. Review and testing found it works as desired. Congratulations to  @MattMiss, @vladoleksiyenko, @brandonviorato & @Gene7Him on have this contribution accepted by OED.

Issue #1187, opened 2024-04-03
horizontal shift when show error bars


huss:
### Describe the bug

When error bars are selected, the dates on the x-axis expand so the graphic is compress somewhat toward the center of the graphic. Normalizing by area does not do this.

### To Reproduce

Any meter/group on a line graphic that supports error bars should show this.

### Expected behavior

The shift is visually unappealing. It is unclear if the width of the horizontal error bar is causing Plotly to automatically do this. If so, the fix may be difficult. If so, we can reconsider what to do.

### Additional context

None

sainmas:
I would like to work on this issue.

huss:
> I would like to work on this issue.

@sainmas  Great. I've assigned you the issue. Please let me know if you need anything.

sainmas:
Hey Huss, I just have a question as to the bug at hand here. 
Just so I have a perfect understanding, is this the issue at hand? 
This image is before I click the "Show error bars" button. 
![chrome_rQhF5ROsq7](https://github.com/user-attachments/assets/3f181930-7662-4e93-a845-d31edf9d84d7)
Down underneath is an image of after I click on the button. The fact that there is empty space to the left of the graph? 
![chrome_loq4sQWQUH](https://github.com/user-attachments/assets/530e5949-b6b0-47b2-9e9a-d915e40c70c0)

huss:
The issue is hard to see due to including the Electric Utility group. It is hard to see but there are a few values graphed on the far, right edge of the graphic because they span only a few, very different dates than the Cos 23 meter. Here is what I see if I only graph that meter:

![Screenshot 2024-11-19 172806](https://github.com/user-attachments/assets/b23b3c83-5add-4a50-95c1-8f4b4101ad89)

and this what it looks like after you add error bars:

![Screenshot 2024-11-19 172824](https://github.com/user-attachments/assets/45ae82a2-8974-4678-8485-a3150ffe4718)

Hopefully this is easier to see the shift of the original graph to the left and right when error bars are shown. Let me know if this is not clear or you need anything else.

sainmas:
Hello @huss ,
I've tried multiple different attempts at trying to figure out where the file would be that interacts with the x-axis at the bottom with no real success. I think I have narrowed it down to these two lines are what I assume cause the issue. (these are in src/client/app/redux/selectors/lineChartSelectors.ts )
[...]

PR #1409, opened 2024-12-10
Issue 1187 - Horizontal shift when show error bars

sainmas:
# Description

When creating a line bar graph and "Show error bars" is checked, the graph horizontally shrinks. 
This request changes a check to allow the use of the x-axis values. Which can be used in a range modifier that affects the size of the x-axis. This range modifier allows error bars to not adjust the size of the graph.

Mason Sain - @sainmas
Pedro Valdovinos Reyes - @PValdovinos

Fixes #[1187](https://github.com/OpenEnergyDashboard/OED/issues/1187)

## Type of change

## Checklist

## Limitations

One potential issue that remains is the fact that the rangeslider still shows grey area on the right and left.

### DIFF
--- a/src/client/app/components/LineChartComponent.tsx
+++ b/src/client/app/components/LineChartComponent.tsx
@@ -73,14 +73,13 @@ export default function LineChartComponent() {
 		return <SpinnerComponent loading height={50} width={50} />;
 	}
 
-
-	// Check if there is at least one valid graph
-	const enoughData = data.find(data => data.x!.length > 1);
 	// Customize the layout of the plot
 	// See https://community.plotly.com/t/replacing-an-empty-graph-with-a-message/31497 for showing text not plot.
 	if (data.length === 0) {
 		return <h1>{`${translate('select.meter.group')}`}	</h1>;
-	} else if (!enoughData) {
+	}
+	// Checks if there is enough data for at least one valid graph
+	else if (!data[0].x) {
 		return <h1>{`${translate('no.data.in.range')}`}</h1>;
 	} else {
 		return (
@@ -93,7 +92,8 @@ export default function LineChartComponent() {
 					legend: { x: 0, y: 1.1, orientation: 'h' },
 					// 'fixedrange' on the yAxis means that dragging is only allowed on the xAxis which we utilize for selecting dateRanges
 					yaxis: { title: unitLabel, gridcolor: '#ddd', fixedrange: true },
-					xaxis: { rangeslider: { visible: true }, showgrid: true, gridcolor: '#ddd' }
+					//Range must be between these values to allow error bars to show properly
+					xaxis: { rangeslider: { visible: true }, showgrid: true, gridcolor: '#ddd', range: [data[0].x[0], data[0].x[data[0].x.length - 1]]}
 				}}
 				config={{
 					responsive: true,


src/client/app/components/LineChartComponent.tsx
```diff
@@ -93,7 +92,8 @@ export default function LineChartComponent() {
 					legend: { x: 0, y: 1.1, orientation: 'h' },
 					// 'fixedrange' on the yAxis means that dragging is only allowed on the xAxis which we utilize for selecting dateRanges
 					yaxis: { title: unitLabel, gridcolor: '#ddd', fixedrange: true },
-					xaxis: { rangeslider: { visible: true }, showgrid: true, gridcolor: '#ddd' }
+					//Range must be between these values to allow error bars to show properly
+					xaxis: { rangeslider: { visible: true }, showgrid: true, gridcolor: '#ddd', range: [data[0].x[0], data[0].x[data[0].x.length - 1]]}
```
huss, line 96: I had an issue where the slider is wrong if there were multiple lines that span different ranges. I think this probably needs to be the min of all first points and the max of all end points to be sure it spans the entire range of all lines.

src/client/app/components/LineChartComponent.tsx
```diff
@@ -73,14 +73,13 @@ export default function LineChartComponent() {
...
-	// Check if there is at least one valid graph
-	const enoughData = data.find(data => data.x!.length > 1);
 	// Customize the layout of the plot
 	// See https://community.plotly.com/t/replacing-an-empty-graph-with-a-message/31497 for showing text not plot.
 	if (data.length === 0) {
 		return <h1>{`${translate('select.meter.group')}`}	</h1>;
-	} else if (!enoughData) {
+	}
+	// Checks if there is enough data for at least one valid graph
+	else if (!data[0].x) {
```
huss, line 82: This change seems to have stopped the message about no data in range for the meter. I found this by changing the range outside all points.

huss (changes requested):
Thanks to  @sainmas & @PValdovinos for working on this issue. I have made a couple of comments to consider.

Having seem this work, I'm starting to suspect that the Plotly auto range on the x-axis is the issue. I noticed that this solution cuts off the horizontal part of the error bars that are beyond the original range of the data. Thus, Plotly may be shifting the x-axis values to let them show. I'm okay with losing them to stop the shift in points.

I wanted to ask if any other solution was found. OED has, to a certain extent, tried to avoid directly setting ranges and let Plotly do it automatically. However, per my comment above, this may cause the shift. If this is the only option then I'm open to the change.

As for the slider range noted in the limitations, I have a feeling (without testing) that the slider range calculation is not in sync with the new x-axis range value. If it spans more values then it might be gray in the area between. OED has had and still seems to have some issues with the slider range. It might be nice to track this down but you don't have to do this if you don't want to.

huss:
@sainmas & @PValdovinos It has been a while since I put in some comments on the PR. I wanted to touch base to see if you have plans to address/comment on them. Thanks.

huss:
@sainmas & @PValdovinos I wanted to make a final check if you intend to work on this any more. If OED does not hear by 6/9 then this PR will be closed. Thanks for your help to date.

PR #1418, opened 2024-12-29
fix radar graph fails if no pre-chosen graph

### LINKED ISSUE #1404: radar graphic fails if meter/group not already graphed with line graphic
### Describe the bug

If a meter or group is selected for graphing in a radar graphic when no previously graphed meter or group in a line graphic then the code crashes where the buttons to restart are shown. So far testing has found that what is important is that the line graphic has never had a meter graphed before a meter is graphed in radar. It does not seem to matter if that meter is currently being displayed or not. It also does not seem to matter which meter is selected - even if it is not the ones being graphed in radar. The same behavior is seen in graphing groups. Selecting a meter in advance does not avoid issues with groups and the other way around.

### To Reproduce

1. Refresh the OED page.
2. Go to the radar graphic page.
3. Select any meter/group.
4. The code should crash.

Selecting a meter/group on the line graphic page after step 1 but before step 2 should avoid the error.

Note there are React warning/errors in the console about the order and number of hooks done.

### Expected behavior

The code should not have an error and display the radar graphic as it does when a meter/group is first selected in a line graphic.

nmqng:
# Description

This PR changed the `defaultButtons` and `advancedButtons` to be global and make the `useState()` declaration upfront to make sure it is not re-render every time the page is rendered.

Fixes #1404 

## Type of change

## Checklist

### DIFF
--- a/src/client/app/components/RadarChartComponent.tsx
+++ b/src/client/app/components/RadarChartComponent.tsx
@@ -27,6 +27,15 @@ import { lineUnitLabel } from '../utils/graphics';
 import { useTranslate } from '../redux/componentHooks';
 import SpinnerComponent from './SpinnerComponent';
 
+
+// Display Plotly Buttons Feature
+// The number of items in defaultButtons and advancedButtons must differ as discussed below
+const defaultButtons: Plotly.ModeBarDefaultButtons[] = [
+	'zoom2d', 'pan2d', 'select2d', 'lasso2d', 'zoomIn2d',
+	'zoomOut2d', 'autoScale2d', 'resetScale2d'
+];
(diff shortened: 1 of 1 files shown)


huss (approved):
Thanks to @nmqng for addressing this issue. Review and testing found it works as desired. It is ready to merge.

For the record, I was curious if a similar change would impact other graphics. I tried bar but it did not seem to change the result nor number of renders.

nmqng:
> For the record, I was curious if a similar change would impact other graphics. I tried bar but it did not seem to change the result nor number of renders.

I don't think this change could apply to other graphics since the problem of radar graphic was the use of `useState()` after some renders were made. Thus, I moved it before any render is made. I checked other graphics and note that for now all declarations of `useState()` are in the correct position (before render is made), so those can't apply the change made in this issue.

PR #1458, opened 2025-04-05
Added warning if new created conversion is both 0. Issues #1444

### LINKED ISSUE #1444: warn admin if conversion is zero

There are unusual cases where you want both the slope and intercept of a conversion to be zero but it is not common. The initial values for creating a conversion have these values as zero so it happens if they are not changed. It would be better to warn the admin if the values are both zero if saved.

When the admin saves a new conversion (create) then warn if both values are zero and offer to cancel/confirm the save. For example, the conversion page has a confirmation popup for deleting a conversion (and for other cases) that can be used as the model.

Edit isn't likely to be an issue since you have to change at least one value to zero but putting the same type of warning would be a good idea.

None other than don't warn.

### Additional context

None

TheREALNguyenAnh:
# Description

Used the ConfirmActionModal to make a warning message similar to one when deleting a conversion. The files edited is CreateConversionModalComponent.tsx and data.ts

Fixes #1444

## Type of change

## Checklist

## Limitations

The translation for Spanish added to data.ts is machine translated, and I have no clue if it is appropriate. I did not do translation for French.


src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -72,14 +78,35 @@ export default function CreateConversionModalComponent() {
...
 	const resetState = () => {
 		setConversionState(defaultValues);
 	};
+	/* End Warning Modal */
 
 	// Submit
 	const handleSubmit = () => {
-		if (validConversion) {
+		// Show warning modal if slope and intercept are both 0
+		if (conversionState.slope === 0 && conversionState.intercept === 0){
```
huss: There should be a space between ``){``. There is at least one other place in the file with issues. In general, it is good to use the VSC format document to get all the formatting correct.

src/client/app/translations/data.ts
```diff
@@ -90,6 +90,7 @@ const LocaleTranslationData = {
 		"conversion.select.destination": "Select a destination unit",
 		"conversion.select.source": "Select a source unit",
 		"conversion.slope": "Slope:",
+		"conversion.slope.intercept.zero": "Both the slope and the intercept are zero. Are you sure?",
```
huss: To make this clearer and more consistent with other messages, I suggest it become:

"Both the slope and the intercept are zero. Are you sure you want to save this conversion?"

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -17,6 +17,7 @@ import { TrueFalseType } from '../../types/items';
 import { showErrorNotification } from '../../utils/notifications';
 import { useTranslate } from '../../redux/componentHooks';
 import TooltipMarkerComponent from '../TooltipMarkerComponent';
+import ConfirmActionModalComponent from '../ConfirmActionModalComponent';
```
huss, line 20: A general comment: I think this same type of check should apply when editing a conversion per the ideas in the issue.

huss (changes requested):
Thanks to @TheREALNguyenAnh for another contribution. Overall, it works as desired. I've made a few comments to consider.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -99,6 +126,14 @@ export default function CreateConversionModalComponent() {
 
 	return (
 		<>
+			<ConfirmActionModalComponent
+				show={showWarningModal}
+				actionConfirmMessage={warningMessage}
+				handleClose={handleWarningCancel}
+				actionFunction={handleWarningConfirm}
+				actionConfirmText={translate('confirm')}
```
huss: I just noticed that confirm is not yet a translated string in data.ts. There is confirm.action if that would work or add a new one. cancel seems fine.

huss (changes requested):
Thanks to @TheREALNguyenAnh for the updates. The changes look good. I made one comment on a translation that I only noticed this time.

huss (approved):
Thanks for the quick update. Everything looks good. Thanks to @TheREALNguyenAnh for another accepted contribution.

PR #1464, opened 2025-04-08
Prohibit bidirectional conversions when source is a meter

### LINKED ISSUE #1443: bidirectional meter conversion allowed
### Describe the bug

The design of OED only has meters as sources of conversions. It was assumed these would be one directional (not bidirectional). The current conversion pages allow for it to be bidirectional.

### To Reproduce

1. Go to the conversion page and click to create a conversion.
2. Select a meter unit as the source and any allowed unit as the destination.
3. Leave it as bidirectional.
4. Save to see if will not fail.

The same idea applies to edit of conversions.

### Expected behavior

Other OED pages warn that a bad choice will b changed by OED. For example, if a type of unit is set to meter for a unit then the displayable will be reset to none if that is not already chosen. The same warnings and behavior should apply for bidirectional on meter conversions.

### Additional context

None

esummerh:
# Description

This PR improves conversion creation and editing methods by preventing bidirectional conversions from being allowed when the source is a meter unit. This behavior aligns with OED's assumption that meter conversions should only be unidirectional. This implementation aligns significantly with unit meter selection, as was described in the issue specification.
The changes include:
- When a meter is selected as the source unit, the "bidirectional" field displays a visual warning that OED will set the conversion as unidirectional upon saving.
- On save, bidirectional is forced to false, independent of user input.
- These changes are present in both the create conversion and edit conversion files.

Fixes #1443 

## Type of change

## Checklist

## Limitations

N/A


huss (commented):
Thanks to @esummerh for another contribution. Testing found it works correctly. However, it uses different logic than the unit component. Was there a reason that I may not see for doing this? I'm asking  because I have these thoughts on the new way:

- It changed the handle event methods. These are normally the same across all files and handle to generic inputs. If they do need to be changed it might be good to rename them.
- I like that it centralizes the decision but it does use more state. What if there was a function to determine if there is an issue that is used in save and the form input so it is the same/centralized but not with state as is done in unit.
- I suspect that the direct form feedback as is in unit may require less logic.
- Consistent methodologies generally make the code easier to maintain.

I welcome your thoughts on this.

esummerh:
Hi @huss - Thank you for the feedback! You're right that the implementation ending up diverging a bit from the unit component. My intention was to centralize the decision, but I can see why using more states is a bit messy. I like the idea of creating a separate function for validating instead of relying on many states. I will go back and look more closely at how the unit component does this so the code is more concise. Let me know if you have any other suggestions!

esummerh:
Hi @huss - I pushed more commits that contain the suggestions you mentioned. Now, there are no additional unnecessary states and a function called isMeterSource() is used instead of changing the handle event methods. Let me know if anything needs refactoring. Thanks!

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -213,11 +220,21 @@ export default function CreateConversionModalComponent() {
...
+								value={String(conversionState.bidirectional)}
+								invalid={isMeterSource() && conversionState.bidirectional === true}>
 								{Object.keys(TrueFalseType).map(key => {
 									return (<option value={key} key={key}>{translate(`TrueFalseType.${key}`)}</option>);
 								})}
 							</Input>
+							{isMeterSource() && conversionState.bidirectional === true && (
+								<FormFeedback className='d-block'>
+									<FormattedMessage
+										id="conversion.bidirectional.disabled.meter"
```
huss: In OED when a translation is done, the key is placed in data.ts in the correct alphabetical place and followed by the translation. This needs to be done for each language where a special symbol is used if it is not yet translated (normal for new keys outside English). In this case the translation for English is your defaultMessage and then it can be removed from here. If it works then if you select a different language in the Options menu then it will show the translated string (or the English with the special symbol). The discard.changes below is a good example of how to do this.

huss (changes requested):
Thanks to @esummerh for making the changes. I made one comment on how OED does translations to address. It applies to both create and edit.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
esummerh: Hi @huss - I just pushed the changes you suggested above. On my end this seems to have worked. Let me know if anything else should be changed!

huss (approved):
Thanks for the update and everything looks good. OED appreciates @esummerh continued contributions.

Issue #1448, opened 2025-03-19
enhance suffix unit conversions


huss:
### Is your feature request related to a problem? Please describe.

There appears to be multiple issues/considerations conversions that involve suffix units:

- A suffix unit can be in a bidirectional conversion. OED is curious if there is any use case where this makes sense. If not, it should not be allowed.
- When you delete a conversion involving a suffix unit it works but all the autocreated units and conversions stay. The admin can delete the conversions one-by-one and then do the same with all the units. This seems unnecessarily complex and OED should probably do this automatically. OED should warn the admin of the units/conversions that will be eliminated and prompt for approval.

There may be other issues/items to consider. A careful analysis of what would be needed and how to do it needs to be done and implemented. @huss is willing to work with a developer to work this out.

### Describe the solution you'd like

It should work as described above.

### Describe alternatives you've considered

The manual method described does work and admins can avoid bidirectional.

### Additional context

None

esummerh:
Hi @huss - I've submitted a few PRs for smaller UI-related issues and found this task interesting because it involves a bit more design and analysis. I'd like to give this issue a try if it is still open to contributors.

huss:
@esummerh You are welcome to work on this issue. As you noted, it is less settled. Know that I'm available to discuss via message or a video call it that would help at any point.

huss:
@esummerh You are welcome to work on this issue. As you noted, it is less settled. Know that I'm available to discuss via message or a video call it that would help at any point.

esummerh:
@huss - Great! Thanks!

esummerh:
Hi @huss - I wanted to start working on this issue and I have a few questions. For the bidirectional conversion portion, does OED allow suffix units to act as both a source and a destination in conversions, or should it be restricted to only one role (i.e. source)? Could you also elaborate a bit on why bidirectional conversions involving suffix units could be problematic for OED? I just want to make sure I understand the design intent regarding this and any edge cases before I implement anything.

huss:
@esummerh 

> For the bidirectional conversion portion, does OED allow suffix units to act as both a source and a destination in conversions, or should it be restricted to only one role (i.e. source)?

I don't know of any use for bidirectional. Since no one has brought one up in this issue I think OED should be restricted to one direction. Note it is okay for the suffix unit to be either the source or the destination.

> Could you also elaborate a bit on why bidirectional conversions involving suffix units could be problematic for OED?

[...]

PR #1468, opened 2025-04-11
Suffix bidirectional behavior and deletions

esummerh:
# Description

This change prevents users from creating bidirectional conversions when either the source or destination is a suffix unit. A warning message is shown if 'yes' is selected in this instance, and if the user tries to create or edit a conversion where bidirectional is marked as 'yes' despite not being allowed, OED defaults bidirectional to 'no' in order to control conversion creation/alteration behavior. 

Partly Addresses #1448 

## Type of change

## Checklist

## Limitations

This PR has not yet implemented the second change listed in issue 1448. Before making any changes in this regard, I wanted to clarify which files other than EditConversionModalComponent.tsx should be altered to implement this. Any clarification in this regard would be very helpful. Thanks!


src/client/app/translations/data.ts
```diff
@@ -65,6 +65,7 @@ const LocaleTranslationData = {
 		"conversion": "Conversion",
 		"conversion.bidirectional": "Bidirectional:",
 		"conversion.bidirectional.disabled.meter": "Selecting meter units as sources will set bidirectional to false",
+		"conversion.bidirectional.disabled.suffix": "Selecting suffix units as sources or destinations will set bidirectional to false",
```
huss: While the code uses true/false, the UI shows yes/no. Thus, I think this message should have no instead of false in all languages.

huss (changes requested):
Thanks to @esummerh for another contribution. Review and testing found it works as desired. I made one small comment on a translation to address.

src/client/app/translations/data.ts
esummerh: @huss - I will make those changes now - I have been working on writing code to address the deletion portion of the issue, but it is not completely finished. Would you rather I make the change you suggested above and push a version without the deletion implementation I started on so the first part of the issue can be merged and then I can open another PR for the second part of the issue, or should I use this PR for both subtasks?

huss:
@esummerh Thanks. I think it is fine as a separate PR as this fix is also desired independent of the other part of the issue.

esummerh:
Hello @huss - I made the requested change and pushed to the PR. Let me know if anything else needs updating. Once I get further along on the second task I'll push a new PR.

huss (approved):
@esummerh Thanks for another quick update. Everything looks great. OED appreciates your continued contributions.

PR #1471, opened 2025-04-14
run spinners while view reading is refreshing. turn some use of translate() in utils to the redux hook useTranslate()

### LINKED ISSUE #1170: changing unit rate does not update graphing values until refresh
### Describe the bug

If you update the rate on a flow unit that is currently being graphed in a line unit does not update the values on the graph (or Redux state). A web page refresh does show the new values.

### To Reproduce

1. Line graph of the Water Gallon flow 1-5 per minute.
2. Go to units page, edit the Water_Gallon_Per_Minute unit and change rate from 60 to 600.
3. Go back to line graphic and the values will not have changed.
4. Refresh the web page and the values will be 1/10 what it was before and what was expected.

### Additional context

I looked at this a little:
- src/client/app/components/unit/EditUnitModalComponent.tsx does a shouldRefreshReadingViews if the rate changes.
- src/client/app/redux/api/conversionsApi.ts refresh invalidates the Readings tag.
- src/client/app/redux/api/readingsApi.ts has the Readings tag for line.

Given this, I'm unsure why the Redux state does not update but did not look in detail.

TheREALNguyenAnh:
# Description

Change the use of translate() in components to useTranslate() redux hooks issue #1204 . Add a state to AppStateSlice that would keep track of conversion readings refresh globally. This is then used for when editing units because if the unit is flow or raw, we need to refresh, and currently do not know how to keep track of it otherwise. Users now have to wait until the reading finishes refreshing before they can return to the graphs. 

Fixes #1170

## Type of change

## Checklist

## Limitations

Currently using an app state store in AppStateSlice to keep track of refreshes, felt like there should be a more elegant way of doing this.


huss (approved):
Review and testing found everything seems fine. Thanks to @TheREALNguyenAnh for another successful contribution.

PR #1472, opened 2025-04-15
Added languageManuallySet flag to prevent overwriting selectedLanguag…

shaikhibrahim2000:
This PR addresses three related issues in the admin panel:

1. No Success Notification
When the current language is not English, the success notification after saving changes is not shown.

2. Language Switches to Default on Save
If the selected language is not the default one, saving changes causes it to reset to the default language unexpectedly.

3. Notification Flashes or Disappears Quickly
In some cases (especially when changing the default language), the success notification appears briefly and disappears due to a page refresh.

**Expected Behavior (Now Working):**
The success notification always shows clearly after saving changes.

The user-selected language persists after saving (does not reset to default).

No unnecessary page refresh that would cause the notification to disappear.

**Additional Note:**
To ensure the selected language persists correctly, the user must first select a language via the Options → Language menu. This marks the language as manually chosen (languageManuallySet = true), preventing it from being overwritten by the default setting on future saves.

## Type of change

## Checklist

## Limitations


huss:
Note that it seems that changing the language causes the Toast messages to disappear. When the default language is changed and the language was not set manually the language changes to the new default language. As a result the Toast notification is lost. No fix is known so this will need to be the behavior.

huss (approved):
Thanks to @shaikhibrahim2000 for another contribution. Review and testing found it is fine.

PR #1474, opened 2025-04-15
Issues 853. slider range in graph link

### LINKED ISSUE #853: chart link slider range does not work
### Describe the bug

If you zoom a line graphic with the slider at the bottom of the graph, you can zoom in on a subpart of the data. If you click the chartlink (must open after done) then it includes the value in the link, e.g., sliderRange=2020-08-17T16:06:41Z_2021-07-15T20:23:26Z. If you use this link then you get the same graphic but the slider range is not preserved.  The code goes through:
- src/client/app/components/RouteComponent.tsx parses this value and set the sliderRange option.
- src/client/app/actions/graph.ts processes this option and dispatches changeRangeSliderIfNeeded
- This dispatches changeRangeSlider which goes to src/client/app/reducers/graph.ts and modifies rangeSliderInterval in the graph state.
The issue is that all of this that seems to work fine (looking at the state.graph.rangeSliderInterval after using the chart link seems to have the correct value) but it does not actually set the Plotly range slider. I suspect that we don't ever (yet) use this state to set the slider in Plotly and it only tracked where it was. If possible, we should find a way to set the slider to this value so it completely reproduces the graphic.

### Expected behavior

It set the slider!

### Additional context

None.

### LINKED ISSUE #1181: chosen slider range

The slider range (zoom) is not, in general, an attribute that remains active after new selections. This is different than when the range is set by reloading the data. For example, normalizing a line graphic by area or changing to another graphic resets the slider range. This may also relate to the fact that the zoom does not work with chart links (see issue #853).

OED needs to decide the desired behavior and implement it. One choice is that the slider range sticks but there is an easy way to reset. Other choices include the current behavior or something in-between. Input is desired. Right now I find some of the resets to be counterintuitive for the user.

See above.

### Additional context

None

TheREALNguyenAnh:
# Description

Removed the comments from the generation and processing of the slider range. And use it for the initial slider range.

Fixes #853 and Fixes #1181

## Type of change

## Checklist


src/client/app/components/LineChartComponent.tsx
```diff
@@ -19,6 +19,8 @@ import { selectSelectedLanguage } from '../redux/slices/appStateSlice';
 import Locales from '../types/locales';
 import { useTranslate } from '../redux/componentHooks';
 import SpinnerComponent from './SpinnerComponent';
+import { selectSliderRangeInterval } from '../redux/slices/graphSlice';
+
```
huss: remove blank line between imports

src/client/app/components/LineChartComponent.tsx
```diff
@@ -93,7 +97,13 @@ export default function LineChartComponent() {
 					legend: { x: 0, y: 1.1, orientation: 'h' },
 					// 'fixedrange' on the yAxis means that dragging is only allowed on the xAxis which we utilize for selecting dateRanges
 					yaxis: { title: unitLabel, gridcolor: '#ddd', fixedrange: true },
-					xaxis: { rangeslider: { visible: true }, showgrid: true, gridcolor: '#ddd' }
+					xaxis: {
+						rangeslider: { visible: true },
+						range: [sliderRangeInterval.getStartTimestamp()?.toISOString()
+							, sliderRangeInterval.getEndTimestamp()?.toISOString()],
```
huss: It would be better to have the comment on the previous line.

src/client/app/redux/selectors/uiSelectors.ts
```diff
@@ -485,7 +485,7 @@ export const selectChartLink = createAppSelector(
 				break;
 			case ChartTypes.line:
 				// TODO Omitted for the time being re-implement slider range later.
```
huss: Is this TODO still needed?

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -366,6 +365,7 @@ export const graphSlice = createSlice({
 							case 'shiftTimeInterval':
 								current.shiftTimeInterval = TimeInterval.fromString(value);
 								break;
+
```
huss: Please remove new blank line.

huss (changes requested):
Thanks to @TheREALNguyenAnh for another contribution. I have made a few minor comments about formatting. This works as described in the issue.  However, I noticed that the slider range is included in chart link but it is all and not the desired value. I think a similar change to what was done for line is needed. The issue did not mention the slider for bar but it should also work there.

I note that this also fixed the issue that the slider range would reset when certain options such as error bars was selected.

huss (approved):
Thanks for the update. Review and testing found it works as desired. Thanks to @TheREALNguyenAnh on another accepted OED contribution.

Issue #1191, opened 2024-04-03
reimplement unsaved warning


huss:
### Describe the bug

Before RTK, OED used UnsavedWarning (searching for this will show where) to let the admin know when they might lose a change because it was unsaved. This was commented out to add in later after the basic RTK conversion was done.

### Expected behavior

These warning should be re-implemented and enabled in appropriate files.

OED currently does not, in general, use this in the new modal pages with cards. Thus, it is possible this will go away at some point. However, OED should decide if it should be used in some cases.

### Additional context

None

gyordong:
Hi Steve, can @aduques and I work on this issue?

huss:
> Hi Steve, can @aduques and I work on this issue?

@gyordong & @aduques You are welcome to work on this. I'm not sure how easy/hard it will be with the transition to using the Redux Toolkit. If it can be made to work we can discuss how it might be used with the modals. For example, would the warning pop up if the modal is closed (such as clicking off the modal) and there are unsaved changes? Please let me know if you need anything.

aduques:
As it has been a couple of weeks since looking at this issue, I would like to verify the intended behavior of this feature. There is currently an unsaved warning that appears whenever the user tries to leave the Admin Settings page with unsaved changes. The pop-up can be closed out when clicking anywhere on the screen or by clicking “Cancel,” “Leave,” or “Save All.” This is currently implemented using React. This issue requires us to revert back and implement the unsaved warning feature using the corresponding Redux reducers, which have been commented out. Once the unsaved warning feature in the Admin Settings page is implemented with Redux, should similar pages, such as MapCalibrationComponent and MapsDetailComponent, have the same functionality and implementation?

At the moment, my partner and I will look to see how converting back to Redux can be done by looking through previous commits that have the Redux implemented and figure out how it can be carried over to the current version.

huss:
Overall, I think what you stated is correct. It seems the src/client/app/components/UnsavedWarningComponent.tsx has had some changes so I'm not sure what might be needed to make it RTK compatible. src/client/app/types/redux/unsavedWarning.ts seems old but I'm unsure where that is used. src/client/app/components/HeaderButtonsComponent.tsx is involved and uses the old style in src/client/app/types/redux/actions.ts. I don't know if this helps but it is what I readily found.

As for maps, PR #1314 upgrades this to modals and RTK usage. It would be the starting point for any other changes. I am behind in getting that reviewed.

[...]

PR #1504, opened 2025-07-04
Issue1191

aduques:
# Description

Assignees: @aduques

These changes address Issue #1191. These changes aim to present admin users with an unsaved changes warning popup whenever they attempt to leave a create/edit modal with unsaved changes. The implementation of the unsaved warning is modeled after the existing UnsavedWarningComponent.tsx, but has been simplified for the current implementation. In the current implementation, whenever any field is modified in the related create/edit modal page, a boolean value that is responsible for displaying the unsaved warning is updated. 

The unsaved warning component maintains it’s desired functionality: 
- when pressing ‘Cancel,’ the user closes out of the Unsaved Warning popup
- when pressing ‘Leave,’ the user closes out of the Unsaved Warning popup + closes out of the modal
- when pressing ‘Save all,’ the user closes out of the Unsaved Warning popup + closes out of the modal + saves the edits/ creates the object (given that all of the necessary fields have been completed)

Motivation: I wanted to work on this issue to gain further experience with working with open source projects and maintain familiarity with industry-relevant technologies. Working on an issue for a project that I am already familiar makes the problem-solving process required much more easier.

Partly Addresses  #1191

## Type of change

## Checklist

## Limitations

The Simple Unsaved Warning Component does not behave the way I expect it in the following page:
- CSV Meters
- CSV Readings
- Maps

These pages differed from the rest as the create/pages do not make use of modals. This is a problem as I was unable to figure out how to make the Unsaved Warning’s ‘Leave’ and ‘Save all’ buttons to function as they would on the other pages.

The problems are as follows:
- When pressing ‘Leave,’ the user closes out of the Unsaved Warning, the boolean values related to Unsaved Warning is reset, but the changes itself are not reset (ex. If a checkbox is checked, pressing ‘Leave’ doesn’t reset the box).
- The implementation attempted to make use of a blocker and useNavigator to prevent the user from leaving the page with unsaved changes. If the user handles the unsaved warning, then they would be able to proceed with leaving to their desired page. However, I was unable to get this to work, so the current implementation requires the user to try to leave the page again after handling the unsaved warning.
- In the Maps page, the unsaved warning does not work if the user simply creates a map without saving the map edits. Instead, the unsaved warning only works if the user edits a field such as “Map Name” or “Notes.”

Limitations:
- Test code for these specific additions have not been written. The additions have only been tested using ‘npm run test’
- For pages such as CSV Meters, CSV Readings, Maps, Admin Settings, the user is able to log out using the log out button from HeaderButtonsComponent.tsx. This is a problem as the user can bypass the unsaved warning in this manner. I was unable to figure out how to get HeaderButtonsComponent.tsx to be able to keep track of the unsaved changes state and make it so that the user is unable to interact with its buttons while unsaved changes are present.

These limitations are as to why the changes are considered to be “Partly addressing” the issue.


src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -20,12 +20,30 @@ import TooltipMarkerComponent from '../TooltipMarkerComponent';
 import ConfirmActionModalComponent from '../ConfirmActionModalComponent';
 import { UnitType } from '../../types/redux/units';
```
huss: Here and in other files: OED would prefer not to have blank lines between imports.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -177,7 +214,10 @@ export default function CreateConversionModalComponent() {
 										name='sourceId'
 										type='select'
 										value={conversionState.sourceId}
-										onChange={e => handleNumberChange(e)}
+										onChange={e => {
+											handleNumberChange(e);
+											setHasUnsavedChanges(true); // Mark as unsaved
```
huss: This is a comment that applies across many files. The code puts a function call whenever a value is changed. I have two thoughts on this:

1. It means a developer must add this in every instance which is work (which you did) and it can lead to bugs if they forget.
2. If you change a value and then put it back then you are still asked if you want to save, cancel or leave. In reality, there is nothing new to save.

I suggest you take a look at src/client/app/components/unit/EditUnitModalComponent.tsx for the use of canSave. What I prefer about it is:

- It uses a useEffect on the local page state so it happens automatically whenever a value is changed since OED stores changes in local state until saved.
- It compares the original state to the local state to see if there are changes. Thus, changing an resetting a value does not trigger the warning popup in your case.

On this page it is used to determine if the save button should be enabled and does other checks. I believe the editMade variable could be used to set the setHasUnsavedChanges since it is true/false for the same reasons.

OED really should do this save changes checks consistently across the pages and in this way. Unfortunately it does not, e.g., src/client/app/components/conversion/EditConversionModalComponent.tsx. If you did this for your unsaved check then this would be a good start toward this and also make your code simpler. Please let me know if you feel otherwise or have any questions/thoughts.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -19,6 +19,11 @@ import { selectVisibleMeterAndGroupData } from '../../redux/selectors/adminSelec
 import SpinnerComponent from '../SpinnerComponent';
 import { tooltipBaseStyle } from '../../styles/modalStyle';
 
+import { SimpleUnsavedWarningComponent } from '../SimpleUnsavedWarningComponent';
+import { useState } from 'react';
+import { useBlocker } from 'react-router-dom';
+//import { useNavigate } from 'react-router-dom';
```
huss: Should this commented out code go?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -102,126 +107,218 @@ export default function MetersCSVUploadComponent() {
 		display: 'flex'
 	};
 
+	// boolean that updates if any change is made to any meter modal
```
huss, line 117: A couple of general notes that apply to both CSV pages:

1. I'm unsure what the save button does. There is no save for CSV but a button to submit the request. If possible, it would be best to remove that option from these pages. I think doing cancel and then save is fine for this page. If not, doing the submit would be okay.
2. If you make a change, try to leave, cancel and try to leave again it lets you. Did not look into why.

I think it is less important about unsaved changes on these pages so the fact that it works reasonably well is okay.

src/client/app/components/maps/MapsDetailComponent.tsx
```diff
@@ -20,30 +24,124 @@ interface MapsDetailProps {
...
-
-		const buttonContainerStyle: React.CSSProperties = {
-			minWidth: '150px',
-			width: '10%',
-			marginLeft: '40%',
-			marginRight: '40%'
-		};
-
-		return (
+/**
```
huss: I have not carefully reviewed the changes to maps. I do wonder about some of the changes with commented out code. Note PR #1314 will updates maps to the new modal look and the Redux Toolkit. Do you think it makes sense to delay these changes until then since they would likely align with the other pages with cards/modals?

huss (changes requested):
Thanks to @aduques for this contribution. Overall it works as desired. I have made a few comments to consider. Given the changes might update a large number of the files, I'm waiting to do the final testing for later (but don't think it will find much). I know you have worked hard on this so let me know if you have any thoughts/questions or would like help in finishing this up.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: Thank you for pointing to specific examples, I am currently looking into this suggestion.

src/client/app/components/maps/MapsDetailComponent.tsx
aduques: I believe it makes sense to leave as the bugs that I encountered with my implementation would be solved if we were to wait for the updates to the map page. As I have mentioned before, I have only experienced this minor bug on the maps page as it is the only one that does not use the modal design. I can comment out the entirety of the unsaved warning implementation on MapsDetailComponent.tsx, as well as the changes with commented out code, so that it can be referred to in the future.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: The commented out code was included with my attempt to make use of useNavigator to prevent users from leaving the page with unsaved changes. I can definitely remove these commented out code from the final commit to avoid any confusion if desired. I can do the same for the comments left in MapsDetailComponent.tsx and ReadingsCSVUploadComponent.tsx as they all have the same notes if desired.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: Screenshot 2025-07-21 at 7 41 40 PM

To clarify, for both CSV pages, I should only keep the 'cancel' and 'leave' buttons?

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I have noted this and will make the changes in my next commit.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I agree that making use of the canSave provided a more clean user experience (The user is no longer prompted with the unsaved warning despite clearing the changes made). Currently, I have made the changes to all of the related pages besides the MapsDetailComponent.tsx. I have ran the updated code with the 'npm run test' test cases, but I plan to test the changes again myself before making the final commits.

I did not notice any functionality issues or bugs in the reimplementation, but please let me know if you find any during final testing so I can address them as soon as possible.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -136,13 +155,64 @@ export default function CreateConversionModalComponent() {
...
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks:
+		// - Source ID has to be a number
+		// - Destination ID has to be a number
+		// - Bidirectional has to be a boolean
+		// - Slope has to be a number
+		// - Intercept has to be a number
+		// - Note cannot be blank
+		const validConversion = !isNaN(conversionState.sourceId)
```
huss: Overall I like this check. A few thoughts:

- I believe both Id's start as -999 and must be an integer when a menu item is selected. Thus, I think these tests should go since the check for a change below will stop a save if this is not updated from the default value.
- bidirectional must be a boolean so, for similar reasons, I think this test can go.
- While slope and intercept are input boxes they are limited to numbers. I found no way not to have a valid number so these checks should go.
- It is valid not to have a note (empty) so this check should go.

Given all of these I think the could completely go if you agree.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -136,13 +155,64 @@ export default function CreateConversionModalComponent() {
...
+			&& typeof conversionState.bidirectional === 'boolean'
+			&& !isNaN(conversionState.slope)
+			&& !isNaN(conversionState.intercept)
+			&& conversionState.note !== '';
+
+		// Compare the local changes to the default values
+		const editMade =
+			conversionState.sourceId !== defaultValues.sourceId
+			|| conversionState.destinationId !== defaultValues.destinationId
+			|| conversionState.bidirectional !== defaultValues.bidirectional
```
huss, line 177: Here are a few rules about the need for a change that you probably did not know:

- bidirectional starts true and it is fine if it is left that way. With this check an admin cannot create a conversion with this as true.
- While it is unusual to have slope & intercept both as zero that is allowed. Note the save checks warns the admin about this. It is also common that intercept remains zero.
- As above, the note can be empty.

Given these rules, all four of these checks should go.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -309,13 +327,64 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks:
+		// - Source ID has to be a number
+		// - Destination ID has to be a number
+		// - Bidirectional has to be a boolean
+		// - Slope has to be a number
+		// - Intercept has to be a number
+		// - Note cannot be blank
+		const validConversion = !isNaN(state.sourceId)
```
huss: See create for some thoughts on this and the next check.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -297,7 +370,7 @@ export default function CreateConversionModalComponent() {
 						<FormattedMessage id="discard.changes" />
 					</Button>
 					{/* On click calls the function handleSaveChanges in this component */}
-					<Button color='primary' onClick={handleSubmit} disabled={!validConversion} >
+					<Button color='primary' onClick={handleSubmit} disabled={!validConversion || !canSave} >
```
huss, line 385: This is probably going but I think canSave already includes validConversion.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: I see you made this change to remove save and it is good.

I'm still seeing the same behavior about being able to leave, cancel & leave (item 2.). Does this relate to your comment on it not working cleanly in the code?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -102,126 +105,200 @@ export default function MetersCSVUploadComponent() {
...
+		if (blocker.state === 'blocked') {
+			handleToggle();
+		}
+	}, [blocker.state, hasUnsavedChanges]);
+
+	// Keeps canSave state up to date. Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks if the inputs for each field is of their appropriate types.
+		const validMeter = typeof meterData.gzip === 'boolean'
```
huss: For similar reasons to comment in create conversion, I'm not sure the valid is needed.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -189,386 +191,486 @@ export default function ReadingsCSVUploadComponent() {
...
+		if (blocker.state === 'blocked') {
+			handleToggle();
+		}
+	}, [blocker.state, hasUnsavedChanges]);
+
+	// Keeps canSave state up to date. Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks if the inputs for each field is of their appropriate types.
+		const validReading = typeof readingsData.cumulative	=== 'boolean'
```
huss: For similar reasons to comment in create conversion, I'm not sure the valid is needed.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -279,18 +298,79 @@ export default function CreateGroupModalComponent() {
 		// pik is needed since the compatible units is not correct until pik is available.
 	}, [graphicUnitsState.possibleGraphicUnits, state.deepMeters]);
 
+	// Keeps canSave state up to date. Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks if the inputs for each field is of their appropriate types.
+		const validGroup = !isNaN(state.id)
```
huss: For similar reasons to comment in create conversion, I'm not sure the valid is needed.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -279,18 +298,79 @@ export default function CreateGroupModalComponent() {
...
+		const editMade =
+			state.id !== defaultValues.id
+			|| state.name !== defaultValues.name
+			// These three always return true, which causes the unsaved
+			// warning to always appear despite having no changes made.
+			// For now, these have been commented out, but need to be addressed
+			// later on.
+			// These problems exist as there are more than one state that needs
+			// to be accounted for.
+			//|| state.childMeters !== groupChildrenState.childMeters
```
huss: I suspect the two states need a deep comparison to see if the arrays changed. I can try to help with this if you want. If it relates to multiple versions of state then I can also help but have not yet thought about that.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -403,6 +422,54 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 		// the deep meters of this group are properly updated.
 	}, [graphicUnitsState.possibleGraphicUnits, groupState.deepMeters, loggedInAsAdmin]);
 
+	// Keeps canSave state up to date. Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	// Reuses code from this file's handleSubmit()
+	useEffect(() => {
```
huss, line 428: See edit for related comments.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -195,6 +214,89 @@ export default function CreateMeterModalComponent(props: CreateMeterModalProps):
 		}
 	};
 
+	// Keeps canSave state up to date. Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks if the inputs for each field is of their appropriate types.
+		// (maxError is left out because it is of type 'any')
+		const validMeter =
```
huss: For similar reasons to comment in create conversion, I'm not sure the valid is needed.

src/client/app/components/meters/EditMeterModalComponent.tsx
```diff
@@ -66,6 +85,89 @@ export default function EditMeterModalComponent(props: EditMeterModalComponentPr
 	const [validMeter, setValidMeter] = useState(isValidMeter(localMeterEdits));
 
 	useEffect(() => { setValidMeter(isValidMeter(localMeterEdits)); }, [localMeterEdits]);
+
+	// Keeps canSave state up to date. Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks if the inputs for each field is of their appropriate types.
+		// (maxError is left out because it is of type 'any')
+		const validMeter =
```
huss: For similar reasons to comment in create conversion, I'm not sure the valid is needed.

src/client/app/components/unit/CreateUnitModalComponent.tsx
```diff
@@ -18,13 +18,30 @@ import { showSuccessNotification, showErrorNotification } from '../../utils/noti
 import { MIN_VAL, MAX_VAL } from '../../utils/input';
 import { LineGraphRates } from '../../types/redux/graph';
 import { customRateValid, isCustomRate } from '../../utils/unitInput';
+import { SimpleUnsavedWarningComponent } from '../SimpleUnsavedWarningComponent';
 
 /**
  * Defines the create unit modal form
  * @returns Unit create element
  */
 export default function CreateUnitModalComponent() {
```
huss, line 27: I was able to test this modal since the is valid was okay. I noticed that you can click save on the warning when you try to leave the page when it is disabled on the actual modal. This causes a failure message to show because it is an invalid save. Is it possible for your warning to accept a boolean about if save should be disabled and then use the same logic (or a variable) to track this? This probably applies to a number of modals.

huss (changes requested):
Thanks to @aduques for this update and your continued efforts on this. I'm sorry it sat for so long. Overall, I like the changes. I've made a few comments to consider. I'm also noting that I did not redo extensive testing since the comments may lead to some changes in when modals are valid and can be saved. I don't expect issues but will check after the current comments are resolved. Please let me know if you would like any help, clarification or you have thoughts.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: Yes, I am referring to that issue. When working on adding the suggestions, I wasn't able to find a fix, so it still applies. I believe that this small issue could be revisited by using useNavigator (or something similar), which would save the URL path of the destination whenever the user is blocked by the unsaved warning. Then, when the user presses 'Leave' it would fetch the saved URL path of the destination and redirect the user there. If desired, I could spend time after addressing the other suggestions to try again and implement this.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I agree with this suggestion. I implemented the following in this way as I was modeling it after the EditUnitModalComponent.tsx example that you have previously pointed to me. If these changes can potentially cause an error with the save, then it should go.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I have noted this. Now, the only checks made for editMade are sourceId and destinationId.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I have removed this, and tested by making a new conversion. I did not notice any problems as it was created successfully, but I will still go back and test all of functionality again after implementing the changes.

src/client/app/components/conversion/EditConversionModalComponent.tsx
aduques: I had removed the validConversion, but decided to keep the checks for editMade. My thinking was that since we are not considering the default values, and instead, considering the current saved values, we should make note if any of the fields have been changed. Therefore, the checks on editMade will be untouched. Please let me know if I have misunderstood the suggestions (same goes for CreateConversion), and I will make the necessary changes.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: Noted, I have removed the check for validMeter, and also the canSave as it is no longer used with validMeter gone.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
aduques: Noted, I have removed the check for validReading, and also the canSave as it is no longer used with validReading gone.

src/client/app/components/groups/CreateGroupModalComponent.tsx
aduques: Noted, I have removed the check for validGroup, and also the canSave as it is no longer used with validGroup gone.

src/client/app/components/groups/CreateGroupModalComponent.tsx
aduques: I believe that a similar issue was handled in EditGroupModalComponent.tsx, specifically in the handleSubmit(). I have tried to copy its implementation into the editMade check in EditGroupModalComponent.tsx. However, I did not thoroughly check if the way I implemented it successfully checks if changes were made in the related fields. 

I did not previously make the same change in CreateGroupModalComponent.tsx, but I can try this time to implement something similar.

As I am going off existing code, I am not too confident if this way will properly make the comparisons, so I will appreciate any insight on this.

The related code:
// Check children separately since lists.
const childMeterChanges = !isEqual(originalGroupState.childMeters, groupState.childMeters);
const childGroupChanges = !isEqual(originalGroupState.childGroups, groupState.childGroups);
const deepMeterChanges = !isEqual(originalGroupState.deepMeters, groupState.deepMeters);

src/client/app/components/groups/EditGroupModalComponent.tsx
aduques: Noted, I have removed the check for validGroup, and also the canSave as it is no longer used with validGroup gone.

src/client/app/components/meters/CreateMeterModalComponent.tsx
aduques: Noted, I have removed the check for validMeter, and also the canSave as it is no longer used with validMeter gone.

src/client/app/components/meters/EditMeterModalComponent.tsx
aduques: Noted, I have removed the check for validMeter, and also the canSave as it is no longer used with validMeter gone.

src/client/app/components/unit/CreateUnitModalComponent.tsx
aduques: I am looking into this issue. I have noticed this same issues on all of the other modals that I was working on the revisions.

aduques: As a note, while looking into this issue, I have noticed that some files do not have a failure message that pops up for invalid save.

Related files:
CreateUnitModalComponent.tsx
CreateGroupModalComponent.tsx

aduques: I have implemented the changes in:
CreateUnitModalComponent.tsx
CreateGroupModalComponent.tsx
CreateMeterModalComponent.tsx
CreateConversionModalComponent.tsx

I didn't implement the changes in the corresponding edit pages as those did not have the same requirements (required fields for a valid save).

I didn't implement the changes in either of the CSV pages for the same reason.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -136,13 +155,51 @@ export default function CreateConversionModalComponent() {
...
+			&& conversionState.destinationId !== defaultValues.destinationId;
+		// Compare the local changes to the default values
+		const editMade =
+			conversionState.sourceId !== defaultValues.sourceId
+			|| conversionState.destinationId !== defaultValues.destinationId;
+		setCanSave(validChange && editMade);
+		// Automatically checks for unsaved changes and addresses the issue
+		// of having to manually set the setHasUnsavedChanges
+		// If editMade is true, then hasUnsavedChanges will be set to true.
+		setHasUnsavedChanges(editMade);
```
huss, line 185: I know I asked for a change here but I'm seeing an issue. It now correctly controls the save. However, if you change everything except the source & destination then it thinks there are no changes so you can leave the page without warning. I think the issue is editMade is used for both can save and hasUnsavedChanges. The unsaved should check all fields but the former should only check what it now does. What do you think?

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -19,13 +19,32 @@ import { useTranslate } from '../../redux/componentHooks';
 import TooltipMarkerComponent from '../TooltipMarkerComponent';
 import ConfirmActionModalComponent from '../ConfirmActionModalComponent';
 import { UnitType } from '../../types/redux/units';
+import { SimpleUnsavedWarningComponent } from '../SimpleUnsavedWarningComponent';
 
 /**
```
huss, line 24: I have not looked into why but if you create a conversion with slope/intercept of zero and then click somewhere else to leave then the unsaved warning appears (as it should). If you choose to save then a warning pops up about the zero values. If you now click to save it does not happen. Let me know if you want me to look into this.

Note the same behavior happens on conversion edit.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -309,13 +325,49 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 		}
 	};
 
+	// Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// Compare the local changes to the default values
+		const editMade =
```
huss, line 334: I think my suggestion was confusing. Controlling being disabled on whether an edit was made is a good idea. If there are not changes then save is disabled. That functionality was removed. It is correct that it no longer depends on the valid check. I'm seeing a similar situation with edit meter.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -102,126 +105,190 @@ export default function MetersCSVUploadComponent() {
...
-											<div style={checkBox}>
-												<Input
-													type='checkbox'
-													id='gzip'
-													name='gzip'
-													onChange={handleCheckboxChange}
-												/>
-												<div className='ps-2'>
-													{translate('csv.common.param.gzip')}
+		<>
```
huss, line 163: Not sure this is the best place but putting this comment here. For meters and readings CSV upload, setting the file is not noticed as an unsaved change so you can leave without warning. Not a big deal if it isn't easy to fix.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -102,126 +105,190 @@ export default function MetersCSVUploadComponent() {
...
+	// since there is no Modal used here, instead uses useEffect()
+	// to prevent the user from navigating to a different page if
+	// there are unsaved changes
+	React.useEffect(() => {
+		if (blocker.state === 'blocked') {
+			handleToggle();
+		}
+	}, [blocker.state, hasUnsavedChanges]);
+
+	// Checks if valid and if edit made.
```
huss: Is the valid still done or should the comment be updated?

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
```diff
@@ -189,386 +191,463 @@ export default function ReadingsCSVUploadComponent() {
...
+	// since there is no Modal used here, instead uses useEffect()
+	// to prevent the user from navigating to a different page if
+	// there are unsaved changes
+	React.useEffect(() => {
+		if (blocker.state === 'blocked') {
+			handleToggle();
+		}
+	}, [blocker.state, hasUnsavedChanges]);
+
+	// Checks if valid and if edit made.
```
huss: valid per meter version

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -279,18 +299,76 @@ export default function CreateGroupModalComponent() {
...
+		const validChange = state.name !== defaultValues.name;
+		// Check children separately since lists.
+		const childMeterChanges = !isEqual(state.childMeters, defaultValues.childMeters);
+		const childGroupChanges = !isEqual(state.childGroups, defaultValues.childGroups);
+		const deepMeterChanges = !isEqual(state.deepMeters, defaultValues.deepMeters);
+		//Compare the local changes to the default values
+		const editMade =
+			state.id !== defaultValues.id
+			|| state.name !== defaultValues.name
+			// These three always return true, which causes the unsaved
```
huss: I think these comments/commented out lines can go.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -279,18 +299,76 @@ export default function CreateGroupModalComponent() {
...
 
+	// Checks if valid and if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// This checks if all of the required fields have been filled out.
+		const validChange = state.name !== defaultValues.name;
+		// Check children separately since lists.
+		const childMeterChanges = !isEqual(state.childMeters, defaultValues.childMeters);
+		const childGroupChanges = !isEqual(state.childGroups, defaultValues.childGroups);
+		const deepMeterChanges = !isEqual(state.deepMeters, defaultValues.deepMeters);
```
huss: Users cannot change deepMeters. This is automatically updated when the child meters or groups are changed. Thus, I would think this test is not needed.

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -403,6 +420,40 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
...
+		// Check for changes by comparing the original, global state to edited state.
+		// This is the unedited state of the group being edited to compare to for changes.
+		const originalGroupState = groupDataById[groupState.id];
+		// Check children separately since lists.
+		const childMeterChanges = !isEqual(originalGroupState.childMeters, groupState.childMeters);
+		const childGroupChanges = !isEqual(originalGroupState.childGroups, groupState.childGroups);
+		const deepMeterChanges = !isEqual(originalGroupState.deepMeters, groupState.deepMeters);
+
+		//Compare the local changes to the default values
+		const editMade =
```
huss, line 437: See comment in create about deep meters. Also, I think save needs to depend on editMade. Unlike create where a change must be made to make it valid, it starts valid in edit so save is shown even if there are no edits. The save state needs to go to the unsaved popup too.

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -35,13 +35,33 @@ import ListDisplayComponent from '../ListDisplayComponent';
...
 import TooltipHelpComponent from '../TooltipHelpComponent';
 import TooltipMarkerComponent from '../TooltipMarkerComponent';
+import { SimpleUnsavedWarningComponent } from '../SimpleUnsavedWarningComponent';
+import { isEqual } from 'lodash';
 
 /**
  * Defines the create group modal form
  * @returns Group create element
  */
 export default function CreateGroupModalComponent() {
```
huss, line 45: Noting here since not sure where I should/can. If you make a simple change and then try to leave, you still get the option to save even if not allowed and it does save.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: I appreciate you noting the issue about lacking a failure message. Would you like to create an issue for this so it gets resolved?

huss (changes requested):
Thanks to @aduques for addressing my recent comments. Review and testing found it generally works as desired. I've made a few additional comments, esp. given I tried to do real testing this time. Please let me know if anything is not clear, you have thoughts or would like help.

src/client/app/components/unit/CreateUnitModalComponent.tsx
aduques: Yes, after finishing up this issue I could create a new issue for this. But before I do, I think it would be better for me to go through all of the current features and take note of where a failure message is missing. In addition, it would be helpful if you could provide insight on where possible failure messages could be added, so that more can be asked in the new issue. I think creating a new issue just to add the failure message to those two files alone would require a quick solution (I believe it only takes a few lines to implement), so it would be a good opportunity to look at where more can be done.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: When making this change, I was having the same doubt as well. I thought it wouldn't make sense for other changes to be neglected while only source & destination were the only ones that mattered. I may have implemented it this way as I got confused with a previous suggestion. I will update editMade to keep track of all changes made in the upcoming edit.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I will look into this issue as well. I assume it is because of the unsaved warning component not interacting with the handleSubmit() properly. As in, the save all in unsaved warning does not properly handle the slope/intercept 0 values in the same way that the save all button on the modal itself handles it. I will update my progress on this.

aduques: So to summarize, when going from simple unsaved warning component to warning modal, the conversion data (conversionState) is not carried over as it gets cleared. I'm not too sure why just playing around with boolean values to control the submit did not work (I assume it's due to not handling state properly), but ultimately, just making a separate variable to store the conversion data works.

I did the following:
- Create a new variable to store the current conversion data (conversionState), so that it can be used when either handleSubmit() or handleWarningConfirm() is called.
- Essentially copied over the handleSubmit logic into onConfirm (small difference with the else. I implemented the way that I did as I was unsure of how to still include handleClose() in onConfirm).

I think the way that I implemented it has room to be optimized, such as just modifying handleSubmit so that there is no more repeated code.

src/client/app/components/conversion/EditConversionModalComponent.tsx
aduques: I have noted this suggestion. In the following files, I have re-added the canSave which gets updated based on the editMade. If no edit has been made, then the save button is disabled.

EditConversionModalComponent.tsx
EditGroupModalComponent.tsx
EditMeterModalComponent.tsx

In EditGroupModalComponent.tsx, I noticed that groups created with no default graphic unit or child groups/meters (In my test groups, I was able to create a group with only the group name filled out since it was the only required element) have the save button disabled on the modal itself, but is able to be saved through the unsaved warning component's save all button. I understand that this issue happens because the save button gets disabled when invalid group. To make sure that this issue does not happen, can we make it so that choosing a child meter/child group/default graphic unit is required?

To note, I have encountered this issue when trying to edit test groups I have made in the past (groups with no child meter/group/default graphic unit). This does not seem to happen with new groups I make (groups are now required a name and child meter to be created) or the default groups. This probably has been resolved and accounted for already, but I will appreciate if this could get looked into.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: I will look into updating this.

aduques: I was able to check for the file change simply by just including selectedFile as part of the conditions. However, this alone would only consider valid files as an edit made as invalid files are cleared/ not saved. I was not sure of the desired behavior, so I decided to add a new variable that also keeps track if an invalid file was attempted. This change can easily be reverted, but I wanted to know your thoughts on it.

The addition of the new variable required me to add the corresponding setter to handleFileChange(), handleSubmit() and handleClear().

I was not able to figure out how to submit successful CSV data and have it created. I was only able to test if the unsaved warning would appear if I submitted a valid and invalid CSV. I used the HappyPlace.png and the regAscEndonly.csv that was found in the OED documentation to make these tests.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: I have updated the comment.

src/client/app/components/csv/ReadingsCSVUploadComponent.tsx
aduques: I have updated the comment.

src/client/app/components/groups/CreateGroupModalComponent.tsx
aduques: I have removed the comments.

src/client/app/components/groups/CreateGroupModalComponent.tsx
aduques: Noted, I removed deepMeterChanges.

src/client/app/components/groups/EditGroupModalComponent.tsx
aduques: I removed deepMeterChanges from here as well.

I believe I recently addressed this issue when handling the suggestion made on EditConversionModalComponent.tsx. To summarize, I have updated the code so that editMade can modify setCanSave. If no edit is made, then the save button is disabled. I have also included some thoughts on EditGroup in the comment for EditConversion.

If this suggestion is related to a different issue regarding EditGroup, please let me know.

src/client/app/components/groups/CreateGroupModalComponent.tsx
aduques: I will look into this. I believe the issue is that I updated the disabled for only the unsaved warning component, but not the actual create group modal. I think I made this mistake on the other Create modals, so I will update those as well.

aduques: Changes that I have made:
- CreateGroupModalComponent.tsx: validGroup is now considered for setCanSave(). This change should make it so that the save button is disabled on both the unsaved warning component and the modal itself.
- CreateUnitModalComponent.tsx: no change
- Meter: Added !canSave as a condition to disable the save button on the modal itself.
- CreateConversionModalComponent.tsx: validConversion is now considered for setCanSave(). Previously, you could save a conversion that already exists when saving from the unsaved warning component, but not in the modal itself.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: If you want to create an issue to note it for this location and note it probably applies to other files would be okay. Someone in the future can check every page to see which ones need fixing. Also, issue #869 may relate as all routes do not return the errors. Does this help?

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: Ok. I'm wondering if an issue should be created for any outstanding issue(s) with the relevant information/discussion so someone can pick it up in the future. What do you think?

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: I'm not able to save a group if there are no children. I also don't see if as required. Do you still see the issue. If not, you can resolve/comment on this. If you think another change would be good or there is still an issue then let me know.

huss:
Thanks to @aduques for the updates and comments. I've left any open items as unresolved. Some probably only need a new issue to resolve them.

I only see one issue now. If you edit a meter and remove its name then it is invalid. The save button is disabled. If you leave the modal (without saving) then the unsaved warning does not have the save disabled.

src/client/app/components/conversion/EditConversionModalComponent.tsx
aduques: I believe that test group was created before I fixed the problem of when you could save an invalid group by using the save button on the unsaved warning modal. Since this problem is no longer possible (as far as I'm aware), I think that this can be resolved.

aduques:
> Thanks to @aduques for the updates and comments. I've left any open items as unresolved. Some probably only need a new issue to resolve them.
> 
> I only see one issue now. If you edit a meter and remove its name then it is invalid. The save button is disabled. If you leave the modal (without saving) then the unsaved warning does not have the save disabled.

I noticed that this problem is consistent for other edit pages, so I will go back and fix this issue. This issue occurred because I did not add the disabled condition to the edit pages, unlike I did for the create pages.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: I think that an issue could be created for this, but I think it would be a very small issue since it would only require working with the two files. If I can, this can be something that I can try and implement once this issue is closed as the related files are still fresh in my mind.

src/client/app/components/unit/CreateUnitModalComponent.tsx
aduques: Since I have added a disabled feature that disables the save button if changes are not valid, I'm not sure how the failure messages can be reachable. However, I did notice that the success messages were not implemented everywhere. I can create a issue that asks to implement both success and failure messages but unsure how the testing would be done for it.

Files that don't have a success/failure message present:
CreateConversionComponent.tsx
EditConversionComponent.tsx
CreateGroupModalComponent.tsx
EditGroupModalComponent.tsx

Could not check completely:
MetersCSVUploadComponent.tsx
ReadingsCSVUploadComponent.tsx

I think that issue #869 directly relates to the lack of failure/success messages (assuming from the mentions of showSuccessNotification/showErrorNotification). However, since I have not worked with these directly, I can't be sure without diving into the code.

That issue wants both success and failure messages to appear on create/edit pages when needed (or possibly other pages too that I'm not familiar with), so I think that issue could just be updated to include these additions.

On another note, that issue would be an issue that I am interested in working on once this issue is closed as it seems to build off what I am already familiar with.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: Thanks for creating issue #1517. Thus, I'm resolving this.

src/client/app/components/unit/CreateUnitModalComponent.tsx
huss: I'm fine with updating the existing issue with the new information. I tried to assign you #869 but could not. If you would leave a comment that you want to work on it then I'll do that. Or, if you post the new information you found that should allow me to do it. If you want help in working on it after this then let me know so we can talk.

Note there are a number of ways that a request can fail that fall outside the front-end testing of the request. Thus, it is good to catch any failure and display the message.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
huss: The new behavior is good. I was able to verify that a correct meter csv file was uploaded.

huss (approved):
Thanks to @aduques for addressing the new comments and your continued efforts. My review and testing found it works as expected. I think this is ready to merge. Can you please confirm you feel it is ready to be finished. Congratulations on another accepted contribution to OED.

aduques:
> Implement ‘Leave’ functionality on CSV Pages’ Unsaved Warning Component #1517

Yes, the changes are ready to merge if all of the features are working as expected.

src/client/app/components/unit/CreateUnitModalComponent.tsx
aduques: I can work on this issue as well, but I'm planning on working on the new issue before this issue, as it seems that I might need time to familiarize myself with any new files.

PR #1505, opened 2025-07-11
Protecct admin when delete conversion Unit-to-unit case

TheREALNguyenAnh:
# Description

This should add a warning, warning users of all the meters and groups affected when deleting a Unit of type unit to a Unit of type unit. 

Partially Addresses #905

## Type of change

## Checklist

## Limitations
Suffix are still not addressed.


src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -39,10 +39,12 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 	const translate = useTranslate();
 	const [editConversion] = conversionsApi.useEditConversionMutation();
 	const [deleteConversion] = conversionsApi.useDeleteConversionMutation();
+	const [simulateDeleteConversion] = conversionsApi.useSimulateDeleteConversionMutation();
 	const unitDataById = useAppSelector(selectUnitDataById);
 	const meterDataById = useAppSelector(selectMeterDataById);
 	const conversionDetails = useAppSelector(selectConversionsDetails);
```
huss, line 51: Please remove extra blank line.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -102,52 +104,46 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
-	// Performs checks to warn the admin of the impact deleting a conversion will have on meter units and possible graphing units.
-	const checkState = () => {
+	const checkState = async () => {
 		const source = unitDataById[state.sourceId];
 		const dest = unitDataById[state.destinationId];
 		let msg = '';
 		let cancel = false;
+
+		// Meter source orphan check
 		if (source.typeOfUnit === UnitType.meter) {
```
huss, line 138: A number of comments were removed from the following code. I'm wondering if there was a reason or if they should be present?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -161,25 +157,62 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 				msg += `${translate('conversion.delete.unit.orphan')} "${dest.name}".\n`;
 				cancel = true;
 			}
+		}
+
+		// Only run simulation if the previous orphan check passed and it's unit-to-unit
+		if (
```
huss: Generally OED has if on one line unless it is too long. I think this could be condensed.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -161,25 +157,62 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			dest.typeOfUnit === UnitType.unit &&
+			!cancel
+		) {
+			try {
+				const result = await simulateDeleteConversion({
+					sourceId: state.sourceId,
+					destinationId: state.destinationId
+				}).unwrap();
+
+				if (result.affectedMeters.length > 0) {
```
huss: I think some comments on the cases would be useful.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -161,25 +157,62 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+					msg += translate('conversion.delete.meter.affected') + ':\n';
+					result.affectedMeters.forEach(meter => {
+						msg += `• ${meter.meterName}\n`;
+						if (meter.lostUnits.length > 0) {
```
huss: Is the meter listed if there are no lost units? I'm try to understand the check on length.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -161,25 +157,62 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+							msg += '  ' + translate('conversion.delete.lost.units') + ': ';
+							msg += meter.lostUnits.map(id => unitDataById[id]?.name || id).join(', ') + '\n';
```
huss: I'm trying to understand the ? after the use of getting the unit id info. Can it be undefined for the item so the ? is needed?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -161,25 +157,62 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+						}
+					});
+				}
+
+				if (result.affectedMeters.length > 0 && result.affectedGroups && result.affectedGroups.length > 0) {
```
huss: I'm curious about checking for result.affectedGroups here but not needing to do it for meters. Also, would it be simpler to put this inside the if below and check if meters was > 0 to add the \n?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -161,25 +157,62 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+						msg += `• ${group.groupName}\n`;
+						if (group.lostUnits.length > 0) {
+							msg += '  ' + translate('conversion.delete.lost.units') + ': ';
+							msg += group.lostUnits.map(id => unitDataById[id]?.name || id).join(', ') + '\n';
+						}
+					});
+					cancel = true;
+				}
+			} catch (e) {
+				msg += translate('conversion.delete.simulation.error');
```
huss: Is this a normal safety check or do you expect it to happen? I'm just checking.

src/client/app/redux/api/conversionsApi.ts
```diff
@@ -94,6 +94,13 @@ export const conversionsApi = baseApi.injectEndpoints({
 					dispatch(setRefresingReadings(false));
 				}
 			}
+		}),
+		simulateDeleteConversion: builder.mutation<SimulateDeleteResult, Pick<ConversionData, 'sourceId' | 'destinationId'>>({
```
huss: I'm wondering why the simulation is a mutation and not a query?

src/client/app/translations/data.ts
```diff
@@ -75,9 +75,13 @@ const LocaleTranslationData = {
 		"conversion.create.source.destination.same": "The source and destination cannot be the same for a conversion",
 		"conversion.delete.conversion": "Delete Conversion",
 		"conversion.delete.meter.orphan": "Deleting this meter conversion will orphan (make ungraphable) meter unit",
+		"conversion.delete.lost.units": "Units this will lose access to after deletion",
+		"conversion.delete.meter.affected": "The following meters will be affected by this deletion",
+		"conversion.delete.group.affected": "The following groups will lose these units and be UNGRAPHABLE if you delete this conversion",
```
huss: Consider an alternative to uppercase to emphasize. I don't know if HTML bold might work. I'm open to ideas and not spending too much time on this.

src/client/app/types/redux/conversions.ts
```diff
@@ -19,3 +19,21 @@ export interface ConversionsState {
 	submitting: ConversionData[];
 	conversions: ConversionData[];
 }
+
+export interface SimulateDeleteResult {
+	affectedMeters: SimulateDeleteAffectedMeter[];
+	affectedGroups: SimulateDeleteAffectedGroup[];
+	// affectedUnits: number[];
```
huss: Status of commented out line?

src/server/routes/conversions.js
```diff
@@ -5,8 +5,13 @@
 const express = require('express');
 const { log } = require('../log');
 const { getConnection } = require('../db');
+const { createCikArray } = require('../services/graph/createConversionArrays');
 const Conversion = require('../models/Conversion');
+const Meter = require('../models/Meter');
+const Group = require('../models/Group');
+const Unit = require('../models/Unit');
 const { success, failure } = require('./response');
+const { getAllPaths, createConversionGraph, createConversionGraphFromArray } = require('../services/graph/createConversionGraph');
```
huss: getAllPaths does not seem to be used.

src/server/routes/conversions.js
```diff
@@ -185,5 +190,100 @@ router.post('/delete', async (req, res) => {
...
+
+		// 4. Get the current Cik array
+		const currentGraph = await createConversionGraph(conn);
+		const currentCik = await createCikArray(currentGraph, conn);
 
+		// 5. Precompute compatible units for each meter (current and simulated)
+		const meterIdToUnits_current = {};
+		const meterIdToUnits_sim = {};
+		for (const meter of allMeters) {
+			if (meter.unitId == -99) continue;
```
huss: OED uses { } around all if/for/etc. blocks so please use here. Also, I think doing an !== check to do the work of the loop and removing the continue would be cleaner. This applies below too.

src/server/routes/conversions.js
```diff
@@ -185,5 +190,100 @@ router.post('/delete', async (req, res) => {
...
+			meterIdToUnits_current[meter.id] = new Set(currentCik.filter(cik => cik.source === meter.unitId).map(cik => cik.destination));
```
huss: OED uses camelcase for code variables. I think it would be best to use that rather than _.

src/server/routes/conversions.js
```diff
@@ -185,5 +190,100 @@ router.post('/delete', async (req, res) => {
...
+		const groupIdToMeterIds = {};
+		await Promise.all(allGroups.map(async group => {
+			groupIdToMeterIds[group.id] = await Group.getDeepMetersByGroupID(group.id, conn);
+		}));
+
+		// 7. For each meter, compare compatible units before/after
+		const affectedMeters = [];
+		for (const meter of allMeters) {
+			if (meter.unitId == -99) continue;
+			const before = meterIdToUnits_current[meter.id] || new Set();
```
huss: This is fine but I was wondering if it would be easier to use an if/else above and put in an empty set if if meterID is -99 in the loop above rather than make the empty set here.

src/server/routes/conversions.js
```diff
@@ -185,5 +190,100 @@ router.post('/delete', async (req, res) => {
...
+			}
+		}
+
+		// 8. For each group, intersect the sets (using cached meter compatible units)
+		const affectedGroups = [];
+		for (const group of allGroups) {
+			const meterIds = groupIdToMeterIds[group.id];
+			if (!meterIds || meterIds.length === 0) continue;
+
+			const intersectSets = sets => sets.reduce((a, b) => new Set([...a].filter(x => b.has(x))));
```
huss: Is this similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection?

src/server/services/graph/createConversionArrays.js
```diff
@@ -62,6 +62,8 @@ async function createCikArray(graph, conn) {
 	// TODO: The table in the database for the logical Cik needs to be wiped and these values stored. This code 
 	// will be added once the database table for using it to get readings is set.
 	// At the moment, we just return the array.
+	// WARNING: Currently this function is used to simulate deletion of conversions in conversions.js, if at some point we do change the database when the function is called
```
huss: This line exceeds the OED limit of 150 characters.

src/server/services/graph/createConversionGraph.js
```diff
@@ -29,6 +29,21 @@ async function createConversionGraph(conn) {
 
 	return graph;
 }
+// Helper to build a graph from arrays (not from Database).
```
huss: Blank line between functions. Also, this should have JSDoc function comment.

src/server/routes/conversions.js
```diff
@@ -185,5 +190,100 @@ router.post('/delete', async (req, res) => {
...
+			destinationId: { type: 'number', minimum: 0 }
+		}
+	};
+	const validatorResult = validate(req.body, validConversion);
+	if (!validatorResult.valid) {
+		log.warn(`Got request to simulate deletion of conversions with invalid conversion data, errors: ${validatorResult.errors}`);
+		failure(res, 400, `Got request to delete conversions with invalid conversion data. Error(s): ${validatorResult.errors}`);
+		return;
+	}
+	try {
```
huss: There is a fair amount of logic here. I'm wondering if it should be in the model rather than the route. What do you think?

huss (changes requested):
Thanks to @TheREALNguyenAnh for doing this work. Overall it is good. I've made a few comments on specific parts of the code. Here are some general thoughts:

- I don't see the group warning. I created a group with the temperature Fahrenheit meter and that unit. I then tried deleting the conversion from the meter unit to Fahrenheit. It warned about the meters but not the group. Am I doing this wrong?
- As we discussed, the test for losing the default graphic meter for meters & groups needs to be added and the setting to no unit will be added.
- You agreed to now add a test for reduced units to groups and warn the admin.

Please let me know if anything is not clear or I can help.

src/server/routes/conversions.js
TheREALNguyenAnh: Yes, a bit. This takes in an array of sets, which then find the one common set amongst them all. If we used that intersection function then we would have something like this instead. 
const intersectSets = sets => sets.reduce((a, b) => a.intersection(b));

src/client/app/translations/data.ts
```diff
@@ -643,11 +651,18 @@ const LocaleTranslationData = {
 		"conversion.create.mixed.represent": "A conversion cannot mix units of quantity, flow and raw\u{26A1}",
 		"conversion.create.source.destination.not": "Source or destination not set\u{26A1}",
 		"conversion.create.source.destination.same": "The source and destination cannot be the same for a conversion\u{26A1}",
+		"conversion.default.graphic.unit": "Default Graphic Unit\u{26A1}",
 		"conversion.delete.conversion": "Delete Conversion\u{26A1}",
 		"conversion.delete.meter.orphan": "Deleting this meter conversion will orphan (make ungraphable) meter unit\u{26A1}",
+		"conversion.delete.lost.units": "Units this will lose access to after deletion\u{26A1}",
+		"conversion.delete.meter.affected": "The following meters will be affected by this deletion\u{26A1}",
+		"conversion.delete.group.affected": "The following groups will lose these units and be UNGRAPHABLE if you delete this conversion\u{26A1}",
```
huss: I think the text here differs from the en & es one.

src/server/routes/conversions.js
```diff
@@ -5,9 +5,15 @@
 const express = require('express');
 const { log } = require('../log');
 const { getConnection } = require('../db');
+const { createCikArray } = require('../services/graph/createConversionArrays');
 const Conversion = require('../models/Conversion');
+const Meter = require('../models/Meter');
+const Group = require('../models/Group');
+const Unit = require('../models/Unit');
 const { success, failure } = require('./response');
+const { createConversionGraph, createConversionGraphFromArray } = require('../services/graph/createConversionGraph');
```
huss: I cannot see where either of these imports are used.

src/client/app/translations/data.ts
```diff
@@ -73,11 +73,19 @@ const LocaleTranslationData = {
 		"conversion.create.mixed.represent": "A conversion cannot mix units of quantity, flow and raw",
 		"conversion.create.source.destination.not": "Source or destination not set",
 		"conversion.create.source.destination.same": "The source and destination cannot be the same for a conversion",
+		"conversion.default.graphic.unit": "Default Graphic Unit",
 		"conversion.delete.conversion": "Delete Conversion",
-		"conversion.delete.meter.orphan": "Deleting this meter conversion will orphan (make ungraphable) meter unit",
+		"conversion.delete.meter.orphan": "Deleting this conversion will orphan (make ungraphable) meter unit",
+		"conversion.delete.group.orphan": "Deleting this conversion will orphan (make ungraphable) group",
```
huss, line 81: I think this is missing in other languages.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -158,34 +197,167 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+				);
+				cancel = true;
 			}
 		}
 
-		if (msg === '') {
-			handleDeleteConfirmationModalOpen();
-		} else if (cancel) {
-			setDeleteConfirmationMessage(msg + `${translate('conversion.delete.restricted')}\n`);
+
```
huss: Extra blank line.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -158,34 +197,167 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+		// Only run simulation if the previous orphan check passed and it's unit-to-unit
+		if (source.typeOfUnit === UnitType.unit && dest.typeOfUnit === UnitType.unit && !cancel) {
+			try {
+				const result = await simulateDeleteConversion({
+					sourceId: state.sourceId,
+					destinationId: state.destinationId
+				}).unwrap();
+
+				// Orphaned groups check
+				const orphanedGroups = result.affectedGroups?.filter(group => group.orphaned) || [];
```
huss: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter, filter returns an empty array if nothing matches. I tried removing the ``|| []`` and it seemed to work fine. This applies to one other place below.

huss:
While testing I think I found a case that I did not cover in any previous discussion. I created these units/conversions:
Screenshot 2025-08-11 155439
I create meter 11m with meter unit 1t and meter 6m with meter unit 6u and then this group:
Screenshot 2025-08-11 155502
and then I tried deleting 6u->2t conversion. I would think that would orphan the group since there are no longer any compatible units (2t was the only one). However, I get:
Screenshot 2025-08-11 155539
What do you think?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -201,12 +373,40 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+		// Update meters
+		for (const meterId of metersWithLostDefault) {
+			const meterData = meterDataById[meterId];
+			await editMeter({ meterData: { ...meterData, defaultGraphicUnit: -99 }, shouldRefreshViews: false });
+		}
+
+		// Update groups
+		for (const groupId of groupsWithLostDefault) {
+			const groupData = groupDataById[groupId];
+			// The GroupsAPI expect only 10 properties, and does not expect deepMeters, so this is my way of removing it
```
huss: Some places OED uses omit to remove a key.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -201,12 +373,40 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
 	};
-	const handleDeleteConversion = () => {
+	const handleDeleteConversion = async () => {
 		// Closes the warning modal
 		// Do not call the handler function because we do not want to open the parent modal
 		setShowDeleteConfirmationModal(false);
 
-		// Delete the conversion using the state object, it should only require the source and destination ids set
+		// Update meters
+		for (const meterId of metersWithLostDefault) {
```
huss: This does each meter and group one at a time. A few thoughts:
- If there is a failure, for example, delete conversion, then meters/groups were changed but not the conversion. If this was done as a route to the server and it did all the steps as a transaction then this could not happen.
- Doing lots of routes is slower. Not a big deal but a single route would address that too.
If you want to discuss then just let me know.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -158,34 +197,167 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+								<br /><br />
+							</div>
+						);
+					});
 
-			if (msg === '') {
-				msg += `${translate('conversion.delete.unit')}\n`;
-				// TODO: Check after deleting the conversion to see if a change happens.
-				// 		 Notify the admin of any consequences caused by deleting the conversion.
+					// After you get result from simulateDeleteConversion check default graphic units
```
huss, line 307: The formatting on lost default graphic units for both meters & groups seems off:
Screenshot 2025-08-11 171208
Did you mean to put each on a separate line (only one here).

src/server/services/conversionSimulation.js
```diff
@@ -0,0 +1,96 @@
...
+	const currentCik = await createCikArray(currentGraph, conn);
+
+	// 5. Precompute compatible units for each meter (current and simulated)
+	const meterIdToUnitsCurrent = {};
+	const meterIdToUnitsSim = {};
+	for (const meter of allMeters) {
+		if (meter.unitId == -99) {
+			meterIdToUnitsCurrent[meter.id] = new Set();
+			meterIdToUnitsSim[meter.id] = new Set();
+			continue;
```
huss: I think it is better to remove this continue and put the code outside the if into an else clause. I think research shows it leads to fewer errors. I'm sorry if my previous comment was confusing.

src/server/services/conversionSimulation.js
```diff
@@ -0,0 +1,96 @@
...
+		}
+		meterIdToUnitsCurrent[meter.id] = new Set(currentCik.filter(cik => cik.source === meter.unitId).map(cik => cik.destination));
+		meterIdToUnitsSim[meter.id] = new Set(simulatedCik.filter(cik => cik.source === meter.unitId).map(cik => cik.destination));
+	}
+
+	// 6. Batch load all group-to-meter relationships
+	const groupIdToMeterIds = {};
+    await Promise.all(allGroups.map(async group => {
```
huss: Space indented here and a few other formatting to fix up.

src/server/services/conversionSimulation.js
```diff
@@ -0,0 +1,96 @@
...
+			});
+		}
+	}
+
+	// 8. For each group, intersect the sets (using cached meter compatible units)
+	const affectedGroups = [];
+	for (const group of allGroups) {
+		const meterIds = groupIdToMeterIds[group.id];
+        if (!meterIds || meterIds.length === 0) {
+			continue;
```
huss: See above about else.

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: Could you help me with this?

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: Can you just verify this is for unexpected issues?

huss (changes requested):
Thanks to @TheREALNguyenAnh for addressing the comment. Review and testing found it generally works as desired. I resolved all the original comments that are complete and put new comments on the ones I'm unsure about. I also made a few new comments to consider. I also noticed it now has a merge conflict. Please let me know if you need anything.

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: Unlikely but could happen in unusual circumstances (it seems). Done in other places so fine.

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: Discussed with @TheREALNguyenAnh and only if unexpected error.

src/server/services/compatibleUnits.js
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: I'm wondering if this file should be in src/server/util/?

huss:
> While testing I think I found a case that I did not cover in any previous discussion. I created these units/conversions: 
> ...

I'm noting this is resolved.

huss (changes requested):
Thanks to @TheREALNguyenAnh for the updated version. I've made one small comment to consider. Also, there are two previous comments that are not resolved and I think should be considered:
- ~In conversionsApi.ts about mutation~ (resolved)
- ~In EditConversionModalComponent.tsx about multiple route calls for meter/groups.~ (resolved)

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -219,13 +396,27 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+		//Edit the meters that lost default graphic units
+		const meterEdits = metersWithLostDefault.map(meterId => {
+			const meterData = meterDataById[meterId];
+			return editMeter({ meterData: { ...meterData, defaultGraphicUnit: -99 }, shouldRefreshViews: false });
+		});
+		//Edit the groups that lost default graphic units
+		const groupEdits = groupsWithLostDefault.map(groupId => {
+			const groupData = groupDataById[groupId];
+			const groupPayload = omit(groupData, ['deepMeters']);
+			return editGroup(groupPayload);
```
huss: I'm unclear on how the groups are having the default graphic unit set to -99. I see it above for meters.

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: I was thinking this would be done by routing all the meter and group ids that need their default graphic id set to -99 along with the id of the conversion to delete. Then the route would pass this to a server model function in conversions to make all the items happen in a transaction. I think the way you changed it to does do them all as one promise but if it fails then nothing is cleaned up in the database. I think the only clean way to do this is via a transaction on the server for the database. I'm sorry if my previous comment was not clear and let me know if you have any thoughts/questions.

Also, the promise used should show success or a failure message based on the result. This was not done in the old code but it really should be done.

huss (changes requested):
Thanks to @TheREALNguyenAnh for addressing the comments. Overall it is good. I made one new comment and added one comment to an old comment (the only one not resolved) that should be considered. They both apply to the one part of the code.

src/server/routes/conversions.js
```diff
@@ -150,41 +153,79 @@ router.post('/addConversion', adminAuthMiddleware('add conversions'), async (req
  * Route for POST, delete conversion.
  */
 router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, res) => {
-	// Only require a source and destination id
+	// Accept sourceId, destinationId, meterIds, groupIds
 	const validConversion = {
 		type: 'object',
-		required: ['sourceId', 'destinationId'],
+		required: ['sourceId', 'destinationId', 'meterIds', 'groupIds'],
 		properties: {
```
huss, line 160: I know you followed the current code but OED is currently improving its route parameter checking. Thus, I think all these should have a type of 'integer'. Also do ``additionalProperties: false``. The arrays should add ``uniqueItems: true, maxItems: 1000,``.

src/server/routes/conversions.js
```diff
@@ -150,41 +153,79 @@ router.post('/addConversion', adminAuthMiddleware('add conversions'), async (req
...
-	} else {
-		const conn = getConnection();
-		try {
-			// Don't worry about checking if the conversion already exists
-			// Just try to delete it to save the extra database call, since the database will return an error anyway if the row does not exist
-			await Conversion.delete(req.body.sourceId, req.body.destinationId, conn);
-		} catch (err) {
-			log.error(`Error while deleting conversion with error(s): ${err}`);
-			failure(res, 500, `Error while deleting conversion with errors(s): ${err}`);
+		return;
```
huss: I prefer the original else without the return. Could it be put back?

src/server/routes/conversions.js
```diff
@@ -150,41 +153,79 @@ router.post('/addConversion', adminAuthMiddleware('add conversions'), async (req
...
+	}
+	const { sourceId, destinationId, meterIds = [], groupIds = [] } = req.body;
+	const conn = getConnection();
+	try {
+		await conn.tx(async t => {
+			// Update meters if any
+			for (const meterId of meterIds) {
+				await t.none('UPDATE meters SET default_graphic_unit = NULL WHERE id = $1', [meterId]);
```
huss: Could this use the syntax in src/server/test/db/cyclicGroupTests.js that directly uses the variable name in the query? Same for group.

src/server/routes/conversions.js
```diff
@@ -150,41 +153,79 @@ router.post('/addConversion', adminAuthMiddleware('add conversions'), async (req
...
+			// Update groups if any
+			for (const groupId of groupIds) {
+				await t.none('UPDATE groups SET default_graphic_unit = NULL WHERE id = $1', [groupId]);
+			}
+			// Delete conversion
+			await Conversion.delete(sourceId, destinationId, t);
+		});
+		success(res, 'Successfully deleted conversion and updated meters/groups');
+	} catch (err) {
+		log.error(`Error while deleting conversion transaction: ${err}`);
```
huss: I think it is better to make the two messages the same. I know OED does not do this in many places but the new way is:
```
const msg = `Error while deleting conversion and updating meters/groups: ${err}`;
log.error(msg);
failure(res, 500, msg);
```

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -219,13 +394,23 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
-		// Delete the conversion using the state object, it should only require the source and destination ids set
-		deleteConversion({ sourceId: state.sourceId, destinationId: state.destinationId });
+		const payload = {
+			sourceId: state.sourceId,
+			destinationId: state.destinationId,
+			meterIds: metersWithLostDefault,
+			groupIds: groupsWithLostDefault
+		};
+		try {
+			await deleteConversion(payload).unwrap();
```
huss: The common OED usage is with a .then, .catch as in src/client/app/components/admin/users/EditUserModalComponent.tsx. Could this please be used so OED has common usage.

huss (changes requested):
Thanks to @TheREALNguyenAnh for addressing the previous comments. I've made a few new, minor ones to consider. ~Also, during testing I tried using the units/conversions in the earlier image. I then created a meter with unit 6u and default graphic unit of 8u. Next I tried to delete the conversion u -> 8u. That should make the DGU for the meter invalid so it should become no unit. It is not orphaned since the meter can still graphic in 2t. However, that does not happen. After the delete, when you try to graph the meter it sets the graphic unit to 8u but then grays out the meter since it cannot graph it which should not happen. I can show you in our next call if you like and let me know if you think the example is wrong. I also have a similar group but did not test yet since the meter was an issue.~ (resolved)

src/server/routes/conversions.js
huss: Thanks for removing the return. I still think the logic would be clearer and the same as other files if the rest of the function was in an else clause.

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss: I see the .then was added but the .unwrap was removed. Other uses in OED to the .unwrap to get to the error message from the server instead of the Toolkit. Do you think it is not needed here (and maybe elsewhere)?

src/server/routes/conversions.js
huss: Did you try additionalProperties?

src/server/routes/conversions.js
```diff
@@ -150,41 +153,82 @@
...
-			log.error(`Error while deleting conversion with error(s): ${err}`);
-			failure(res, 500, `Error while deleting conversion with errors(s): ${err}`);
+	}
+	const { sourceId, destinationId, meterIds = [], groupIds = [] } = req.body;
+	const conn = getConnection();
+	try {
+		await conn.tx(async t => {
+			// Update meters if any
+			for (const meterId of meterIds) {
+				await t.none(`UPDATE meters SET default_graphic_unit = NULL WHERE id = ${meterId}`);
```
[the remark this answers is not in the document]
huss: This is correct but there are several mitigating factors:

- You have to have an admin token to do this and admins are allowed to change lots of values in the DB. A bad actor should not have this token.
- This avoids rechecking the impacted meters/groups. It also guarantees that the values are what were shown to the admin when they decided to proceed are the ones done.
- If someone sent bad values the impact is not huge since it changes the default graphic unit and nothing else.

Given these facts, this is not going to be changed.

src/server/routes/conversions.js
```diff
@@ -150,41 +153,82 @@
...
+			}
+			// Update groups if any
+			for (const groupId of groupIds) {
+				await t.none(`UPDATE groups SET default_graphic_unit = NULL WHERE id = ${groupId}`);
```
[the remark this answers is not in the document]
huss: This is correct but there are several mitigating factors:

- You have to have an admin token to do this and admins are allowed to change lots of values in the DB. A bad actor should not have this token.
- This avoids rechecking the impacted meters/groups. It also guarantees that the values are what were shown to the admin when they decided to proceed are the ones done.
- If someone sent bad values the impact is not huge since it changes the default graphic unit and nothing else.

Given these facts, this is not going to be changed.

huss (changes requested):
Thanks to @TheREALNguyenAnh for the update. I've put an additional comment on some previous comments for you to consider.

huss (approved):
Thanks to @TheREALNguyenAnh for addressing all the comments. Review and testing found everything is good. This is ready to merge. Congratulations on another accepted contribution to OED.

Issue #896, opened 2023-04-22
conversions that vary with time


huss:
### Is your feature request related to a problem? Please describe.

OED v1.0 allows for general units. As part of the OED plan, we would like to generalize the conversions to allow ones that vary with time. The plan is outlines in the [conversion that vary with time design doc](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/unitVaryTime/conversionsVaryTime.md).

huss:
Issue #1303 discusses client-side conversions. If time varying conversions are successful then that should not be done.

PR #1514, opened 2025-08-05
Server-Side: Time-Varying Unit Conversions

tfrendl:
# Description

Implemented server-side functionality to allow admin users to create, manage, and apply time-varying unit conversions. Time-varying logic can be used at the conversion, day, and week levels.

Partly Addresses [#896]

## Type of change

## Checklist

## Limitations

There are 14 errors when running `npm run test`. I believe most of them are unrelated to the work in this branch. Three of them involve the Conversions table, which I think are caused by the tests not being updated to the new Conversions table schema.


docker-compose.yml
```diff
@@ -21,8 +21,8 @@ services:
             interval: 10s
             timeout: 10s
             retries: 3
-        # ports:
-        #    - "5432:5432"
+       # ports:
```
huss: Could you please remove the two spacing changes in this file so no changes show.

src/client/app/components/conversion/EditConversionModalComponent.tsx
huss, line 1: I think this is one of only two files that were part of other's work. Do you think it makes sense to remove the changes?

src/client/app/types/redux/conversions.ts
huss, line 1: I think this is one of only two files that were part of other's work. Do you think it makes sense to remove the changes?

huss:
This PR has 153 failing tests when I run it from a clean clone without loading in any test data. The description notes fewer. I looked at them an they involve readings where most return a null for the value. This is undoubtedly due to the fact that the conversion table no longer has the slope & intercept so converting readings does not work properly. Given time-varying conversions is ongoing work by multiple people this should resolve later and is not an issue for the PR.

src/server/migrations/1.0.0-2.0.0/sql/conversions/alter_conversions_table.sql
```diff
@@ -0,0 +1,8 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ALTER TABLE conversions
+    DROP COLUMN IF EXISTS slope,
```
huss: Nice if used tab indenting as in other files.

src/server/models/Conversion.js
```diff
@@ -67,12 +63,38 @@ class Conversion {
 	}
 
 	/**
-	 * Inserts a new conversion to the database.
+	 * Inserts a new conversion to the database, along with a conversion segment.
+	 * The default conversion segment spans from -inf to inf.
+	 * @param {*} weekPatternsId The id for a weekly pattern
+	 * @param {*} slope The slope for the conversion
```
huss: The slope and intercept are formally for the conversion segment and not the conversion.

src/server/routes/conversions.js
```diff
@@ -120,22 +125,28 @@ router.post('/addConversion', async (req, res) => {
 		}
 	};
 	const validatorResult = validate(req.body, validConversion);
+
```
huss: I would remove this new blank line as the code all relates to validation.

src/server/routes/conversions.js
```diff
@@ -120,22 +125,28 @@ router.post('/addConversion', async (req, res) => {
...
 	if (!validatorResult.valid) {
 		log.error(`Got request to insert conversion with invalid conversion data, errors: ${validatorResult.errors}`);
 		failure(res, 400, `Got request to insert conversion with invalid conversion data. Error(s): ${validatorResult.errors}`);
 	} else {
 		const conn = getConnection();
+
 		try {
 			await conn.tx(async t => {
```
huss, line 138: I'm wondering if the conn.tx should be in the model since that is where multiple DB calls are made. I don't know best practice and it may be that would be tricky because of the conn?

src/server/sql/conversion/update_conversion.sql
```diff
@@ -3,8 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 UPDATE conversions
-	SET bidirectional = ${bidirectional},
-        slope = ${slope},
-        intercept = ${intercept},
-        note = ${note}
+	SET
+	bidirectional = ${bidirectional},
```
huss: I'm okay with this on a separate line. I do think it might be easier to parse if all the items under SET are indented one more level.

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,241 @@
...
+		await conn.none(sqlFile('conversionSegment/delete_conversion_segment.sql'), {
+			sourceId: sourceId,
+			destinationId: destinationId,
+			startTime: startTime,
+			endTime: endTime
+		});
+	}
+}
+
+function formatTimestampValue(value) {
```
huss: The same function seems to be used/recreated in the route. If so, I think it should be put in the util directory for reuse.

src/server/sql/conversionSegment/get_by_source_destination.sql
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+SELECT 
+	source_id,
+	destination_id,
+	week_patterns_id,
+	slope,
+	intercept,
+	start_time::TEXT AS start_time,
```
huss, line 12: Other places in OED (logmsg, meter) do not convert datetime stamps to text as done here. Is this needed? I'm partly asking because otherwise this could select * which is simpler. This would apply to several queries. I'm happy to discuss to decide what is best.

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,241 @@
...
+	/**
+	 * Returns a promise to get the conversion segment associated with the given source id, destination id, startTime, and endTime from the database. 
+	 * @param {*} sourceId The source meter's id.
+	 * @param {*} destinationId The destination meter's id.
+	 * @param {*} startTime The start time of the conversion segment.
+	 * @param {*} endTime The end time of the conversion segment.
+	 * @param {*} conn The connection to use.
+	 * @returns {Promise.<ConversionSegment>}
+	 */
+
```
huss: No blank lines between JSDoc and function.

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,241 @@
...
+		// check that -infinity and infinity aren't being updated
+		if ((startChanged && (originalStartTime === '-infinity')) || endChanged && (originalEndTime === 'infinity')) {
+			const errMsg = `Cannot update starting time of -infinity or ending time of infinity`;
+			log.error(errMsg);
+			throw new Error(errMsg);
+		}
+
+		// update the previous segment's end time to the updated start time
+		if (startChanged) {
+				await conn.none(sqlFile('conversionSegment/update_prev_seg_end_to_new_start.sql'), conversionSegment);
```
huss: extra tab indent

src/server/sql/conversionSegment/update_prev_seg_end_to_new_start.sql
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+UPDATE conversion_segments
+	SET end_time = ${startTime}::TIMESTAMP
```
huss: Commenting here but applies to a number of files. In other OED files it seems to use the time directly without the ::TIMESTAMP. I'm wonder if there is an easy way to do that in the new queries to avoid these. This may relate to the other comment for ::TEXT. I have not thought about this and am happy to discuss.

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,241 @@
...
+
+	/**
+	 * Deletes the conversion associated with the source, destination, start time, and end time from the database.
+	 * @param {*} sourceId The source meter's id.
+	 * @param {*} destinationId The destination meter's id.
+	 * @param {*} startTime The start time of the conversion segment.
+	 * @param {*} endTime The end time of the conversion segment.
+	 * @param {*} conn The connection to use.
+	 */
+	static async delete(sourceId, destinationId, startTime, endTime, conn) {
```
huss: I'm nervous about this ability (route, model but not DB SQL). If this is done then it would seem the guarantee that all segments span all time would be violated. Since there is a delete earlier and later should this go? Is it actually used in the front-end?

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,241 @@
...
+	 */
+	static async deleteEarlier(sourceId, destinationId, startTime, endTime, conn) {
+		if (startTime === '-infinity') {
+			const errMsg = `There is no earlier segment to update in order to delete this segment.`;
+			log.error(errMsg);
+			throw new Error(errMsg);
+		}
+
+		// update the end time of the previous segment
+		await conn.none(sqlFile('conversionSegment/update_prev_seg_end_to_curr_end.sql'), {
```
huss: It wouldn't be bad if these multi-request DB calls were in a transaction. I realize OED doesn't always do that but it is probably the right thing to do.

src/server/sql/conversionSegment/update_next_seg_start_to_curr_start.sql
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+UPDATE conversion_segments
+	SET start_time = ${startTime}::TIMESTAMP
+	WHERE source_id = ${sourceId}
```
huss: Not a big deal but a number of the SQL files have extra spaces at the end of lines.

src/server/routes/conversionSegments.js
```diff
@@ -0,0 +1,500 @@
...
+ 
+const express = require('express');
+const { log } = require('../log');
+const { getConnection } = require('../db');
+const ConversionSegment = require('../models/ConversionSegment');
+const { success, failure } = require('./response');
+const validate = require('jsonschema').validate;
+const { adminAuthMiddleware } = require('./authenticator');
+const moment = require('moment');
+const { format } = require('path');
```
huss: This does not seem to be used.

src/server/routes/conversionSegments.js
```diff
@@ -0,0 +1,500 @@
...
+		startTime: item.startTime, 
+		endTime: item.endTime, 
+		note: item.note
+	};
+}
+
+/**
+ * GET all conversion segments.
+ */
+router.get('/', adminAuthMiddleware('get all conversion segments'), async (req, res) => {
```
huss: This seems fine but I wonder if it is needed. I'm thinking about this since returning all the segments for all conversions might be a lot of items. I cannot think of a place where the UI would do this. If it isn't used, should this be removed? I'm less certain about the model since it might be used by the cik update that is still in development. See other comment on if the current card setup needs this.

src/server/routes/conversionSegments.js
```diff
@@ -0,0 +1,500 @@
...
+	
+	const validatorResult = validate(req.body, validConversionSegment);
+	if (!validatorResult.valid) {
+		const errMsg = `Got request to add a conversion segment with invalid conversion segment data, error(s): ${validatorResult.errors}`;
+		log.warn(errMsg);
+		failure(res, 400, errMsg);
+	} else {
+		const conn = getConnection();
+		try {
+			await conn.tx(async t => {
```
huss: Probably not an issue but I'm not clear on why this needs to be a transaction,

src/server/routes/conversionSegments.js
```diff
@@ -0,0 +1,500 @@
...
+			note: {
+				oneOf: [
+					{type: 'string'},
+					{type: 'null'}
+				]
+			},
+			originalStartTime: {
+				oneOf: [
+					{type: 'string'},
+					{type: 'null'}
```
huss: I'm unclear on why the start/end original can be null. I'm thinking the code below would fail in this case.

src/server/routes/weeks.js
```diff
@@ -0,0 +1,285 @@
...
+ * GET week by id
+ */
+router.get('/:id', adminAuthMiddleware('get week by id'), async(req, res) => {
+	const validParams = {
+		type: 'object',
+		maxProperties: 1,
+		required: ['id'],
+		properties: {
+			id: {
+				type: 'string',
```
huss, line 53: I know OED does not use everywhere but I think there is an integer type that might be better and allow for min value of 1. I don't know if all of these are an issue due to it  passing as part of the URI but the check seems the same.

src/server/routes/weeks.js
```diff
@@ -0,0 +1,285 @@
...
+
+	const validatorResult = validate(req.body, validWeek);
+	if (!validatorResult.valid) {
+		const errMsg = `Got request to insert a week with invalid week data, error(s): ${validatorResult.errors}`;
+		log.warn(errMsg);
+		failure(res, 400, errMsg);
+	} else {
+		const conn = getConnection();
+		try {
+			await conn.tx(async t => {
```
huss: Again, unsure if tx is needed for single DB item. I see edit below does not.

src/server/routes/weeks.js
```diff
@@ -0,0 +1,285 @@
...
+ * @param {number} friday The id for the day pattern used for friday.
+ * @param {number} saturday The id for the day pattern used for saturday.
+ */
+router.post('/addWeek', adminAuthMiddleware('add week'), async (req, res) => {
+	const validWeek= {
+		type: 'object',
+		maxProperties: 9,
+		required: ['name', 'sunday','monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday'],
+		properties: {
+			name: {
```
huss, line 93: In other items OED makes sure the name is at least 1 character. I think the same applies to week names. The DB is doing this. Applies to edit too.

src/server/routes/weeks.js
```diff
@@ -0,0 +1,285 @@
...
+				type: 'string',
+			},
+			note: {
+				oneOf: [
+					{ type: 'string' },
+					{ type: 'null' }
+				]
+			},
+			sunday: {
```
huss, line 103: Other places verify the id is at least 0 so I think all these routes should do it here.

src/server/routes/weeks.js
```diff
@@ -0,0 +1,285 @@
...
+		const errMsg = `Got request to delete a week with invalid week data, error(s): ${validatorResult.errors}`;
+		log.warn(errMsg);
+		failure(res, 400, errMsg);
+	} else {
+		const conn = getConnection();
+		try {
+			// Don't worry about checking if the week already exists
+			// Just try to delete it to save the extra database call, since the database will return an error anyway if the row does not exist
+			await Week.delete(
+				req.body.id,
```
huss: Extra space at end of line.

src/server/models/Day.js
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	async insert(slope, intercept, segmentNote, conn) {
+		const day = this;
+
+		// insert new day
+		const dayData = {
+			name: day.name,
+			note: day.note
+		};
+		const resp = await conn.one(sqlFile('day/insert_new_day_pattern.sql'), dayData);
```
huss: This might be good to have in a transaction as it is two DB calls.

src/server/sql/day/get_by_day_name.sql
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+SELECT * FROM day_patterns
```
huss: I don't see where this is used and it does not seem to be in the day model. Could it be removed?

src/server/routes/days.js
```diff
@@ -0,0 +1,223 @@
...
+ * GET day by id.
+ */
+router.get('/:dayId', adminAuthMiddleware('get day by id'), async (req, res) => {
+	const validParams = {
+		type: 'object',
+		maxProperties: 1,
+		required: ['dayId'],
+		properties: {
+			dayId: {
+				type: 'string',
```
huss, line 46: Consider integer per other comment.

src/server/routes/days.js
```diff
@@ -0,0 +1,223 @@
...
+ * @param {string} segmentNote The notes for the default day segment.
+ */
+router.post('/addDay', adminAuthMiddleware('add day'), async (req, res) => {
+	const validDay = {
+		type: 'object',
+		maxProperties: 5,
+		required: ['name', 'slope', 'intercept'],
+		properties: {
+			name: {
+				type: 'string'
```
huss: min size of 1.

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,201 @@
...
+			row.note
+		);
+	}
+
+	/**
+	 * Get all DaySegment objects
+	 * @param {*} conn The database connection to use.
+	 * @returns all DaySegment objects.
+	 */
+	static async getAll(conn) {
```
huss: As with conversion segments, I'm wondering if this is needed. Thinking about it more, maybe the current front-end needs to get them all rather than as each conversion/week is edited. However, it would be nice not to have to do this.

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,201 @@
...
+
+		// check that 0 and 24 aren't being updated
+		if ((this.startHour !== originalStartHour && originalStartHour === 0) || (this.endHour !== originalEndHour && originalEndHour === 24)) {
+			const errMsg = `Cannot update starting hour of 0 or ending hour of 24`;
+			log.error(errMsg);
+			throw new Error(errMsg);
+		}
+
+		// Check and update previous segment's end time to updated start time
+		await conn.none(sqlFile('daySegment/update_prev_seg_end_to_new_start.sql'), daySegment);
```
huss: In conversionSegment it checks if there is a change and only updates in this case. That makes sense to me. Is there a reason this is different?

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,201 @@
...
+
+
+		// update the current segment
+		await conn.none(sqlFile('daySegment/update_day_segment.sql'), daySegment);
+	}
+
+	/**
+	 * Delete the day segment associated with the id
+	 * @param {*} id The day segment id.
+	 * @param {*} conn The connection to use.
```
huss: JSDoc is missing start/end hour.

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,201 @@
...
+	 */
+	static async delete(id, startHour, endHour, conn) {
```
huss: As with conversion segments, should there be a way to delete one without the fixup?

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,201 @@
...
+	 */
+	static async deleteEarlier(dayId, startHour, endHour, conn) {
+		if (startHour === 0) {
+			const errMsg = `There is no earlier segment to update in order to delete this segment.`;
+			log.error(errMsg);
+			throw new Error(errMsg);
+		}
+		// update the end time of the previous segment
+		await conn.none(sqlFile('daySegment/update_prev_seg_end_to_curr_end.sql'), {
+			startHour: startHour,
```
huss: I feel this needs the dayId. Otherwise it seems you could have two different days with the same hour range and it would do both. SQL is the same change. delete later too.

src/server/routes/daySegments.js
```diff
@@ -0,0 +1,442 @@
...
+ * GET day segment by id
+ */
+router.get('/:id', adminAuthMiddleware('get day segment by id'), async(req, res) => {
+	const validParams = {
+		type: 'object',
+		maxProperties: 1,
+		required: ['id'],
+		properties: {
+			id: {
+				type: 'string',
```
huss: Maybe integer

src/server/routes/daySegments.js
```diff
@@ -0,0 +1,442 @@
...
+			const errMsg = `Error while editing day segment with error(s): ${err}`;
+			log.error(errMsg);
+			failure(res, 500, errMsg);
+		}
+	}
+});
+
+/**
+ * POST delete day segment.
+ * @param {integer} id The id for the day segment to be deleted.
```
huss: I think start/end are also params but this may be removed.

src/server/routes/daySegments.js
```diff
@@ -0,0 +1,442 @@
...
+		maxProperties: 3,
+		required: ['dayId', 'startHour', 'endHour'],
+		properties: {
+			dayId: {
+				type: 'integer', 
+				minimum: 0
+			},
+			startHour: {
+				type: 'number',
+				minimum: 1,		// if it was 0, there would be no previous segment
```
huss: Could the comment please go above the line as OED prefers.

src/server/routes/daySegments.js
```diff
@@ -0,0 +1,442 @@
...
+			},
+			startHour : {
+				type: 'number',
+				minimum: 0,
+				maximum: 23
+			},
+			endHour : {
+				type: 'number',
+				minimum: 1,
+				maximum: 23	// if it was 24, there would be no following segment
```
huss: comment above

src/server/services/createDB.js
```diff
@@ -11,10 +11,26 @@ const { redoCik } = require('../services/graph/redoCik');
...
+		} catch(err) {
+			log.error('insert standart units failed', err);
+		}
+		try {
+			await insertStandardConversions(conn);
+		} catch (err) {
+			log.error('insertconversions failed', err);
+		}
+		try {
 		await redoCik(conn);
```
huss: indenting off

src/server/services/createDB.js
```diff
@@ -11,10 +11,26 @@ const { redoCik } = require('../services/graph/redoCik');
 (async function createSchemaWrapper() {
 	const conn = getConnection();
 	try {
-		await createSchema(conn);
-		await insertStandardUnits(conn);
-		await insertStandardConversions(conn);
+		try {
```
huss: I see the merit in logging these errors and how it could help developers. What I'm thinking about is that the exitCode is now 1 instead of 1 as it was before. Maybe the exception should be rethrown so the wrapper try/catch sees it? I'm assuming the error was not specific enough to understand what happened.

huss (changes requested):
Thanks to @tfrendl for their first contribution and one involving so many changes. Overall it seems good. I've made number of comments to consider. Note I did not test this code as I did not want to create independent tests and the UI is not yet ready. The testing can occur once the UI is done. I know you told me you did testing so I'm fairly confident it should work. This is safe since it isn't being merged into development. Please let me know if anything is not clear, you have thoughts or wish to discuss anything more.

src/server/models/ConversionSegment.js
tfrendl: I consulted the frontend team, and they are currently using this function to handle splitting.

src/server/models/ConversionSegment.js
tfrendl: What I had been doing was having the transaction surrounding the call from the route to the model. Should I be having the transaction in the model instead of the route?

src/server/models/ConversionSegment.js
huss: I believe I discussed this with the front-end team and they agreed to switch to using the server code. This could mirror the delete that is already done on the server. This will require new routes/models that have some similarity to delete and update. I'm not sure who will do this but once complete then this function should go.

src/server/models/ConversionSegment.js
huss: I was thinking that since the model is at a lower level and can be used without the route, it is a better and safer place to do this. It would also include the route that uses the model.

src/server/sql/day/get_by_day_name.sql
tfrendl: I reached out to the frontend team, and they are currently using this functionality to create weeks.

huss: If they really need it then it should be in the route & model. I did not see it. Am I missing something or is that in their work?

src/server/models/DaySegment.js
tfrendl: The frontend team is currently using this functionality to handle splitting. Should I leave this function in place until a function is built to handle splitting?

src/server/sql/conversionSegment/get_by_source_destination.sql
tfrendl: Before adding `::TEXT` casting to SQL queries, requests were returning "null" for "-infinity" and "-infinity" values. When researching this issue, I read that JavaScript doesn't handle PostgreSQL's "-infinity" and "infinity" timestamp values. Casting the timestamps to text avoids this issue.

src/server/routes/weeks.js
tfrendl: I originally had id as an integer, but it passes as a string in the url. Would it be better to leave as is, or change it to a post method and pass it in the body of the request as an integer?

src/server/models/DaySegment.js
huss: Yes or the current front-end code will break. Hopefully the change will happen soon. I would like it to happen before PR #1516 is merged.

src/server/sql/conversionSegment/get_by_source_destination.sql
huss: Okay, let's leave it for now. OED could convert the value in a function but that is more work. Maybe a comment in the code about why this is done so others will understand.

src/server/routes/weeks.js
huss: I've never really liked how it is sent but it is a more general issue. OED might change it everywhere at some point but not now. I say leave it for now.

src/server/sql/conversionSegment/update_prev_seg_end_to_new_start.sql
tfrendl: Timestamp casting in SQL files proved unnecessary (except in instances where it was used for returning sorted results), so they have been removed.

src/server/services/createDB.js
tfrendl: Thank you for commenting on this. I had meant to remove these after hunting down a bug!

src/server/sql/day/get_by_day_name.sql
tfrendl: I apologize, my mistake, I thought this was for the get all functionality. You're correct, we don't need this. I am removing it now.

src/server/models/DaySegment.js
tfrendl: An update has been pushed, implementing splitting for day segment. Delete can be removed once the frontend team is no longer using it.

src/server/models/ConversionSegment.js
tfrendl: An update has been pushed, implementing splitting for conversion segment. Delete can be removed once the frontend team is no longer using it.

src/server/routes/conversionSegments.js
huss: I'm noting the model getAll() was also removed so anyone wanting it can easily get it back.

src/server/routes/weeks.js
```diff
@@ -0,0 +1,300 @@
...
+			note: {
+				oneOf: [
+					{ type: 'string' },
+					{ type: 'null' }
+				]
+			},
+			sunday: {
+				type: 'integer',
+				minimum: 0
+
```
huss: extra empty line

src/server/routes/daySegments.js
```diff
@@ -0,0 +1,551 @@
...
+		required: ['dayId', 'startHour', 'endHour'],
+		properties: {
+			dayId: {
+				type: 'integer', 
+				minimum: 0
+			},
+			startHour: {
+				type: 'number',
+				// if it was 0, there would be no previous segment
+				minimum: 1,
```
huss: extra whitespace at end of line

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,317 @@
...
+		if (this.startTime !== '-infinity' || this.endTime !== 'infinity') {
+			const errMsg = `Only time ranges spanning from -infinity to infinity are allowed for insertion.`;
+			log.error(errMsg);
+			throw new Error(errMsg);
+		}
+
+		// check it doesn't exist in the database
+		const row = await conn.any(sqlFile('conversionSegment/get_by_source_destination.sql'), conversionSegment);
+		if (row.length > 0) {
+			const errMsg = `Segment(s) exist for this conversion.`;
```
huss: Really small point but I think the (s) can be removed. If there is more than 1 then something went very wrong given the new check.

src/server/routes/conversionSegments.js
```diff
@@ -0,0 +1,645 @@
...
+	
+	const validatorResult = validate(req.body, validConversionSegment);
+	if (!validatorResult.valid) {
+		const errMsg = `Got request to split a conversion segment earlier with invalid conversion segment data, error(s): ${validatorResult.errors}`;
+		log.warn(errMsg);
+		failure(res, 400, errMsg);
+	} else {
+		const conn = getConnection();
+		try {
+			const earlierSegment = new ConversionSegment(
```
huss: This is okay here but I'm wondering why not move it to the model so the logic is there. That way the model (and therefore the route) cannot make a mistake. Also, the model only needs the current conversion and the split time to work. Do you agree? If so, applies to later too. It seems daySegments does this so it seems to differ and would be okay with my suggestion.

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,326 @@
...
+			const laterSegment = {
+				dayId: originalSegment.day_id,
+				startHour: splitTime,
+				endHour: originalSegment.end_hour,
+				slope: newSlope,
+				intercept: newIntercept,
+				note: newNote
+			};
+
+			// console.log(earlierSegment);
```
huss: remove

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,326 @@
...
+		if (this.startHour !== 0 || this.endHour !== 24) {
+			const errMsg = `Only time ranges spanning from 0 to 24 are allowed for insertion.`;
+			log.error(errMsg);
+			throw new Error(errMsg);
+		}
+
+		// check it doesn't exist in the database
+		const row = await conn.any(sqlFile('daySegment/get_by_dayId.sql'), daySegment);
+		if (row.length > 0) {
+			const errMsg = `Segment(s) exist for this day.`;
```
huss: Per other comment remove (s)

src/server/models/DaySegment.js
```diff
@@ -0,0 +1,326 @@
...
+				intercept: newIntercept,
+				note: newNote
+			};
+
+			// console.log(earlierSegment);
+			await t.none(sqlFile('daySegment/insert_new_day_segment.sql'), earlierSegment);
+
+			// later segment - update start time
+			await t.none(sqlFile('daySegment/update_day_segment.sql'), {
+				id: id,
```
huss: This directly uses the values but above it creates a variable. Either okay but consistent is nice. I have a slight preference for the variable since debugging is easier. Same for later.

src/server/models/ConversionSegment.js
```diff
@@ -0,0 +1,317 @@
...
+			const originalSegment = await t.one(sqlFile('conversionSegment/get_by_source_destination_start_end.sql'), {
+				sourceId: this.sourceId,
+				destinationId: this.destinationId,
+				startTime: startTime,
+				endTime: endTime
+			});
+
+			// later segment - update start time
+			await t.none(sqlFile('conversionSegment/update_conversion_segment.sql'), {
+				sourceId: this.sourceId,
```
huss: See other comment on variable use

src/server/routes/daySegments.js
```diff
@@ -0,0 +1,551 @@
...
+			startHour : {
+				type: 'number',
+				minimum: 0,
+				maximum: 23
+			},
+			endHour : {
+				type: 'number',
+				minimum: 1,
+				// if it was 24, there would be no following segment
+				maximum: 23
```
huss: end of line whitespace

src/server/sql/conversionSegment/update_conversion_segment.sql
```diff
@@ -0,0 +1,16 @@
...
+UPDATE conversion_segments
+	SET week_patterns_id = ${weekPatternsId},
+		slope = ${slope},
+		intercept = ${intercept},
+		start_time = ${startTime},
+		end_time = ${endTime},
+		note = ${note}
+	WHERE source_id = ${sourceId}
+		AND destination_id = ${destinationId}
+		AND start_time = ${originalStartTime}::TIMESTAMP
```
huss: I want to check if the TIMESTAMP casting is needed.

src/server/routes/conversions.js
huss: Did it turn out to be hard to do with without passing a transaction?

huss (changes requested):
Thanks to @tfrendl for addressing the comments. Review found it is good. I've made a few comments to consider. I think this PR is very close to being done. I plan to wait until the front-end team has a chance to use the new functions and then have the undesired ones removed from this PR before it is merged. Please let me know if you need anything or have thoughts.

src/server/routes/conversions.js
tfrendl: It ended up being quite simple to move the transaction to the model.

src/server/routes/conversionSegments.js
tfrendl: I agree. This has been updated.

src/server/models/DaySegment.js
tfrendl: Verified no longer used by the frontend team - removed.

src/server/models/ConversionSegment.js
tfrendl: Verified the frontend team is no longer using - removed.

huss (approved):
Thanks to @tfrendl for addressing the comments. Review found everything seems fine. Use by the FE team will also test the code. Congratulations on having this work accepted.

Issue #1517, opened 2025-08-08
Implement ‘Leave’ functionality on CSV Pages’ Unsaved Warning Component


aduques:
### Is your feature request related to a problem? Please describe.
This feature request is related to the unsaved warning component implementation done in Issue #1191. Issue #1191 involves implementing the unsaved warning component on the CSV Readings and CSV Meters pages, but the functionality is not complete. The functionality of the unsaved warning component is complete on the other Edit/Create pages, as they are designed differently than the CSV pages (create/edit pages being modals). Therefore, the functionality could not be simply carried over to the CSV pages.

When clicking on the Unsaved Warning Component’s ‘Leave’ button, instead of rerouting the user to the page that they attempted to go to before being stopped by the unsaved warning, the warning is closed, and the user has to attempt to go to the page that they were previously attempting to go to.

### Describe the solution you'd like

Instead of having the user needing to perform the action again (make changes -> attempt to go to another page -> unsaved warning pop up -> press ‘Leave’ -> closes out of unsaved warning -> attempt to go to another page -> successfully went to different page), the user should be redirected to the desired page upon pressing ‘Leave.’

### Describe alternatives you've considered

useNavigator() was considered to store the URL path of the attempted page. When the user presses ‘Leave,’ the saved URL path is supposed to be fetched, and that is where the user will be rerouted immediately.

In other words, pressing ‘Leave’ is supposed to go to the page of the saved URL path.

### Additional context

Previous discussions on attempted solutions to this issue can be found on Issue #1191 and Pull Request #1504. See the commit history of #1504, for commented-out attempts of using useNavigator() as a possible solution.

huss:
@aduques I assigned you this issue since I think you wanted to try it. Let me know if you don't want this.

aduques:
After familiarizing myself further with the useBlocker() documentation, it seems that this issue can be solved relatively quickly. My previous attempts at using useNavigate() were going in the right direction, as useBlocker() and useNavigate() work well together. However, the reason it was not working was because of the way that I was changing the blocker.state.

This code in ReadingsCSVUploadComponent.tsx (also found in other related files worked on), seems to be solely responsible for updating the blocker state.
	const blocker = useBlocker(hasUnsavedChanges);

[...]

PR #1522, opened 2025-08-14
Updated Unsaved Warning Implementation

aduques:
# Description

These changes address Issue #1517. These changes aim to build off the efforts done in Issue #1191. Issue #1191 and Pull Request #1504 implement the unsaved warning feature on multiple pages, but the functionality was left incomplete for the CSV Readings and CSV Meters pages. The problem with the functionality was the 'Leave' button not properly redirecting users to their desired pages immediately upon pressing the button. Pull Request #1504 implemented a workaround where the user would have to attempt their request again for it to work properly. In addition, the user page was accidentally left out from the issue, so this issue implements the unsaved warning feature for the user page.

In this pull request:
- useNavigate() was used along with the existing useBlocker() implementation to handle the navigation of users between pages
- implemented unsaved warning feature in EditUserModalComponent.tsx and CreateUserModalComponent.tsx
- added a resetState() in EditUserModalComponent.tsx as it previously did not have a way for the user details of the selected user to be reverted back to (Unsaved changes would remain even after closing out of the edit page. Unsaved changes would revert if the page was refreshed).
- To keep track of changes made in the password fields, a boolean is used (This is done as passwords aren't loaded in when entering the edit page, and this change would be counted as a edit otherwise).

Partly Addresses #1191

## Type of change

## Checklist

## Limitations

Personal testing was done for the implementation of the unsaved warning feature in the user pages. This involved creating test accounts and making sure that the save button on both the modal and unsaved warning were only enabled when appropriate. When reviewing this pull request, further testing on this implementation is appreciated as the implementation involved slight differences in implementation compared to other pages (ex. resetState()).

In addition, the way that implementation for handling password edit changes may need to be reviewed as other ways might be desired.

These notes are why this issue is marked as partially addressing.


aduques:
I was working on a new small commit, so that I can retest the run-checks-test again. However, in the process, I noticed a bug in which retyping the same existing password in either of the edit password or confirm password field will result in the changes not being considered as an edit made. In the next commit, I will address this.

huss:
> I was working on a new small commit, so that I can retest the run-checks-test again. However, in the process, I noticed a bug in which retyping the same existing password in either of the edit password or confirm password field will result in the changes not being considered as an edit made. In the next commit, I will address this.

If you change something and put it back to the original state then OED does not consider it a change (normally) and disables save. In the case of passwords, OED may not know the original value since we don't store them. So, I think you need to see if the admin typed anything to know. You may already know this. I hope this makes sense and is correct.

aduques:
> > I was working on a new small commit, so that I can retest the run-checks-test again. However, in the process, I noticed a bug in which retyping the same existing password in either of the edit password or confirm password field will result in the changes not being considered as an edit made. In the next commit, I will address this.
> 
> If you change something and put it back to the original state then OED does not consider it a change (normally) and disables save. In the case of passwords, OED may not know the original value since we don't store them. So, I think you need to see if the admin typed anything to know. You may already know this. I hope this makes sense and is correct.

I'm experiencing difficulty in trying to implement the editMade check for passwords. At the moment, I have only figured out how to make any changes made to the password fields be considered as an edit made (by comparing userDetails.password.length > 0). However, the user is still able to successfully make an edit by typing in the existing password.

As you mentioned that the original value is not stored, this means that it would not be possible to make the comparison.
(ex. If the password is 'password,' the admin can edit and change the password to be 'password').

If the behavior is acceptable, I can leave it as is, or I can continue on working on it in a future commit.

Edit: In EditUserModalComponent.tsx, line 223 and 224, were supposed to be commented out/removed. It is a redundant function call to isFormValid(), which is already properly called elsewhere. I will include this change in the next commit, as the changes alone may be too small for a commit.

huss:
Yes, seeing if anything is input seems fine and I cannot think of a good alternative. The field should only be typed into if a password is being entered so it must be a change. Thanks for realizing this and making it work.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -15,13 +15,32 @@ import { useTranslate } from '../../../redux/componentHooks';
...
+import { SimpleUnsavedWarningComponent } from '../../SimpleUnsavedWarningComponent';
 
 /**
  * Defines the create user modal form
  * @returns CreateUserModal component
  */
 export default function CreateUserModal() {
 	const translate = useTranslate();
+
+	// boolean that updates if any change is made to any meter modal
```
huss: I think meter should be user.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -15,13 +15,32 @@ import { useTranslate } from '../../../redux/componentHooks';
...
+	const [canSave, setCanSave] = useState(false);
+
+	// displays the unsaved warning component whenever there's unsaved
+	// changes, otherwise closes out of the modal
+	const handleToggle = () => {
+		if (hasUnsavedChanges) {
+			setShowUnsavedWarning(true);
+		}
+		else {
+			handleCloseModal(); // Proceed to close the modal
```
huss: OED prefers comments on the line above not at the end of the line.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -30,6 +31,32 @@ interface EditUserModalComponentProps {
...
+	// boolean that updates if any change is made to any meter modal
+	const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
+	const [showUnsavedWarning, setShowUnsavedWarning] = useState(false);
+	// If there are no changes, then save is disabled
+	const [canSave, setCanSave] = useState(false);
+	// When opening the edit page, the password is not filled in for security reasons.
+	// Therefore, this boolean value is used to keep track if there are any changes
+	// made to either the password field or the confirm password field.
+	const [passwordModified, setPasswordModified] = useState(false);
+	// Compare the local changes made to the password fields.
```
huss: Should these commented out lines be removed?

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -30,6 +31,32 @@ interface EditUserModalComponentProps {
...
+	// const [samePassword, setSamePassword] = useState(false);
+
+	// displays the unsaved warning component whenever there's unsaved
+	// changes, otherwise closes out of the modal
+	const handleToggle = () => {
+		if (hasUnsavedChanges) {
+			setShowUnsavedWarning(true);
+		}
+		else {
+			props.handleClose(); // Proceed to close the modal
```
huss: OED prefers comments on the line above not at the end of the line.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -41,12 +68,31 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
 		disableDelete: props.user.username === currentLoggedInUser.username
 	});
 
+	// Store the details of the currently selected user
+	// This variable is used to reset back to the initial user details.
+	// Otherwise, the unsaved changes will remain when reopening the
+	// edit page.
+	const initialUserDetails = {
```
huss, line 72: This behavior is fine but I noticed that if you click the cancel button it does not clear state. It might be nice if it did it there too.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -41,12 +68,31 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
...
+		...userDefaults,
```
huss, line 73: I'm unclear on why the userDefaults are needed here. The props.user is provided and comes from an API endpoint that seems to provide all the user values. Do you agree?

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -41,12 +68,31 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
...
 	const [submitUserEdits] = userApi.useEditUserMutation();
 	const [submitDeleteUser] = userApi.useDeleteUsersMutation();
 
 	// check if passwords match and if password length is at least 8
 	useEffect(() => {
+		// If any character is added in either field, it will count as password
+		// being modified. This will actively update the passwordModified
+		// boolean value when any change is made.
+		const passwordFieldChanged = userDetails.password.length > 0 || userDetails.confirmPassword.length > 0;
+		if (passwordFieldChanged) {
```
huss: I think it would be cleaner to ``setPasswordModified(passwordFieldChanged);`` rather than an if/else.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -61,6 +107,7 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
 			userDetails.passwordLength &&
 			userDetails.role &&
 			userDetails.username.length > 2;
+	// 		!samePassword;
```
huss: Should this commented out line be removed?

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -160,21 +217,72 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
 			});
 	};
 
+	// Checks if edit made.
+	// References the original implementation in EditUnitModalComponent.tsx
+	useEffect(() => {
+		// Call isFormValid()
+		isFormValid();
+		// Compare the local changes made to the password fields.
```
huss: I'm unclear on this commented out code.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -160,21 +217,72 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
...
+		// && initialUserDetails.confirmPassword === userDetails.confirmPassword
+		// && passwordModified) {
+		// 	setSamePassword(true);
+		// }
+		// else {
+		// 	setSamePassword(false);
+		// }
+		// Compare the local changes to the default values
+		const editMade =
+			initialUserDetails.id !== userDetails.id
```
huss: The id cannot be changed so I don't think this is needed.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
```diff
@@ -118,8 +118,40 @@ export default function MetersCSVUploadComponent() {
...
+	// If useBlocker() is true, the user is unable to navigate to other pages
+	// unless the blocker state is changed.
+	// This line of code is called on every time hasUnsavedChanges is updated.
 	const blocker = useBlocker(hasUnsavedChanges);
 
+	// Stores the URL of the page that the user tries to go to before
+	// being blocked by the unsaved warning.
+	const [attemptedDestinationURL, setAttemptedDestinationURL] = useState<string | null>(null);
+	// When blocker.state is unblocked, useNavigate() is used to navigate
+	// to the attempted destination URL.
```
huss: I might move this to the start of the function where the same is done for translate. OED tends to put these items at the top closer to the import involved. Same for edit.

huss (changes requested):
Thanks to @aduques for working on this improvement. I found it works well so I like what you did (since you asked). I've made a few comments to consider. Please let me know if anything is not clear or you have thoughts.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
aduques: Noted, this comment was copy pasted for multiple files, so I will also look if this comment needs to be updated elsewhere.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
aduques: Noted, this comment was copy pasted for multiple files, so I will also look if this comment needs to be updated elsewhere.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I left the commented out code to keep a reference of my attempt at working on handling passwords in editMade. I will remove these commented out code in the next commit.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: Noted, I have made this change.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I will look into this. It seems that the initial design of handleCloseModal only reset the password fields. I will update the 'Cancel' button on the modal itself to reset all fields as well.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: Line 60 from EditUserModalComponent.tsx:
	// user edit form state and use the defaults plus the user's data
	const [userDetails, setUserDetails] = useState({
		...userDefaults,
		...props.user,
		// if editing current logged in user, do not allow user to delete their own account
		disableDelete: props.user.username === currentLoggedInUser.username
	});
	
I was following the design of how the details of the current user is fetched. It seems that the fields for userDefaults and props.user are different. From my understanding, userDefaults contains all of the fields, whereas props.user fills in userDefaults with its specific parameters.

In other words, using props.user alone will result in an error as it does not contain all of the expected parameters. I did not explore the code enough to see how it would be possible to just rely on props.user, but I assume it would require for more features to be modified.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I will make this change.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: Yes, this commented out line has been removed.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I've left the commented out code, so that I could leave a reference of how I attempted to fix the passwords behavior. I ultimately could not get it to work as the passwords are not stored anywhere in OED, making it difficult to compare for unlike the other fields. As mentioned before, these commented out lines will be removed.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I have made this change.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I've left the commented out code, so that I could leave a reference of how I attempted to fix the passwords behavior. I ultimately could not get it to work as the passwords are not stored anywhere in OED, making it difficult to compare for unlike the other fields. As mentioned before, these commented out lines will be removed.

src/client/app/components/csv/MetersCSVUploadComponent.tsx
aduques: I will make this change. I think that other files that I've worked on also had this design, so I will look to see if I need to move the variables to the top elsewhere.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: My recent commit addressing this issue has actually caused a different issue to appear (saving from the unsaved warning does not work). I'm working on a quick commit that will address the new problem.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -30,6 +31,29 @@ interface EditUserModalComponentProps {
  */
 export default function EditUserModalComponent(props: EditUserModalComponentProps) {
 	const translate = useTranslate();
+
+	// boolean that updates if any change is made to user meter modal
```
huss: Should the edit have been "any user modal" since this is editing users?

src/client/app/components/admin/users/EditUserModalComponent.tsx
huss: My testing did not find an issue. Do you think it is resolved?

src/client/app/components/admin/users/EditUserModalComponent.tsx
huss: You are correct. I missed this in my review. The defaults have fields for validation so it is needed.

huss (changes requested):
Thanks to @aduques for addressing the comments. I want to apologize for this sitting so long as somehow I missed that this was ready for me to look at it again. I made one small new comment to consider and also one to verify that a comment is resolved. My review and testing found this works very well and is almost ready to merge.

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: I have done some testing on this comment and don't see any more issues. I believe shortly after I commented this, I was able to push out a commit that fixed the issue.

Commit I added: https://github.com/OpenEnergyDashboard/OED/pull/1522/commits/36f113e2893e937254b248eea77728ca72a86eff

src/client/app/components/admin/users/EditUserModalComponent.tsx
aduques: Yes, that is correct. I must have overlooked this detail as I pasted the comment onto the relevant files.

aduques:
@huss I apologize for the delay, but I have finished reviewing the recent suggestions made. In addition, I went again and briefly went over the features that I've implemented in this pull request as it has been a few weeks since I've looked at it. I did not notice any faulty behaviors, so I think that everything that I wanted to implement in this issue has been done, and the changes can be merged again if all features are desirable and final tests are good.

huss (approved):
Thanks to @aduques for the recent changes. Review and testing found that this works as desired. I appreciate you seeing this through to completion. Congratulations on another accepted contribution to OED.

Issue #457, opened 2020-03-26
Extend chart link to support from latest readings


huss:
#456 will allow date ranges in the URL. It is a use case that someone wants to show others the last day/week/month of data where it is always the current time frame. To support this, the chart link needs to give the user the option (button?) to use the same amount of time as the slider represents but make it so the recreated graph starts from the latest time reading. We want to allow the URL to do what it currently does in case the link is intended to reproduce exactly the dates currently show. It might be best if the user option only appears if the end time point is the latest time reading on the current graph. Note doing it this way will allow the user to show arbitrary time frames not just day/week/month. Having it fixed to the earliest time reading is not a normal use case and it seems very unlikely that earlier reading would be added later so the current URL is fine and we are not going to do that case.

huss:
There is now a [design document](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/latestChartlink/latestChartlink.md) to work on this issue. It should be used to get information on what to do.

w-b-d:
We are working on this (team Rian, Ann, Will, Nicolas)

anntreasajojo:
Hello,

My team (@w-b-d @rianhassett @nickhitos) and I working on implementing the UI for #456 (allowing relative date ranges in ChartLinks), and I wanted to verify a detail regarding the current UI to ensure our enhancements remain cohesive with the existing design.

I'm modifying the "More Options" popup in the Graph view, specifically the ChartLinks section, to support the new changes for the “Keep Current” option. We've set up OED locally via Docker following the Developers First Steps Documentation. However, I noticed that the date range picker appears inside the More Options modal in my development environment UI, compared to the design doc image, where the date range picker appears in the sidebar, not in the modal.
We’re unsure which layout we should work off of to modify the UI. I have attached reference images below comparing the UI in the design docs versus my development environment UI.

<img width="1225" height="680" alt="Image" src="https://github.com/user-attachments/assets/dbbbe576-120e-4294-93f3-31037fe49eed" />

<img width="1348" height="629" alt="Image" src="https://github.com/user-attachments/assets/5222b5fe-8961-49d7-a362-540fdb74b1ba" />

We believe the in-modal version is the current official layout in OED v1.0.0 and the design doc is outdated. Based on this, we are planning to 
- Add a new checkbox (tentatively labeled “Keep Current”) under Chart Link in the More Options modal.
[...]

PR #1530, opened 2025-08-25
Extended chart link to support from latest readings

w-b-d:
# Description

Extend chart link to support from latest readings. Users can now select "keep current" and the current timespan within the slider range will be saved as a parameter in the URL. If the right is bounded, the "keep current" checkbox is greyed out, as well as if the whole graph is unbounded the checkbox is greyed out. When a user changes the slider range, the graph must be refreshed for the "keep current" functionality to work for the updated slider range. 

Partly Addresses #457

- Added "Chart Link Options:" section title.

- Moved "Hide Options When Using This Link" checkbox outside of the button (UI only, logic unchanged).

- Added "Keep Chart Current" checkbox with full logic implementation.

- Implemented tooltip content for “Keep Chart Current” checkbox.

- All hard coded strings moved into data.ts file for consistency and to support multiple languages.

- Updated UI so that the "Keep Chart Current" checkbox remains visible but is disabled and greyed out when requirements are not met.

- Preserved original behavior for the hide options feature.

- Added isKeepCurrent variable in appStateSlice.ts

- Added "&timeSpan=" to URL in uiSelectors.ts if keep current checkbox is checked

- added timeCreated in graphSlice.ts and is updated within PlotNavComponent.tsx

- added case for timeSpan within the processGraphLink function in graphSlice.ts to handle the logic of the "keep current" functionality. 

Documentation for the Chart LInk should be updated so users will be able to understand the functionality of the new "keep current" checkbox.

## Type of change

## Checklist

- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement]

## Limitations
Tests have not been created using Mocha/Chai, which is why this is labeled as a partial address. 

Contributors: @w-b-d, @anntreasajojo, @rianhassett, @nickhitos


.vscode/launch.json
```diff
@@ -1,7 +1,9 @@
 {
 	"version": "0.2.0",
 	"configurations": [
+
```
huss, line 4: This file only has new blank lines so please revert changes so none in file.

as
```diff
@@ -0,0 +1,7 @@
+  [31morigin/HEAD[m -> origin/development
```
huss, line 1: I'm unsure what this file is for. Should it be removed?

d
```diff
@@ -0,0 +1,7 @@
+  [31morigin/HEAD[m -> origin/development
```
huss, line 1: I'm unsure what this file is for. Should it be removed?

.gitignore
```diff
@@ -53,7 +53,7 @@ nohup.out
 
 # Visual Studio Code settings file
 .vscode/settings.json
-
+.vscode/launch.json
```
huss, line 56: This file is used by OED to define the debugging setup, etc. I'm not sure it should be in .gitignore.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -140,7 +141,10 @@ export const graphSlice = createSlice({
 			if (state.current.threeD.meterOrGroupID !== action.payload) {
 				state.current.threeD.meterOrGroupID = action.payload;
 			}
+		},updateTimeCreated: state => {
+			state.current.timeCreated = new TimeInterval(moment(), moment());
 		},
+
```
huss, line 147: Please remove blank line.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -140,7 +141,10 @@ export const graphSlice = createSlice({
 			if (state.current.threeD.meterOrGroupID !== action.payload) {
 				state.current.threeD.meterOrGroupID = action.payload;
 			}
+		},updateTimeCreated: state => {
+			state.current.timeCreated = new TimeInterval(moment(), moment());
```
huss, line 145: This is a small point but if you want the start and end of the TimeInterval to be the same then I think you need to make a variable to store the moment() value to use twice. Otherwise the two constructors may return different values. I don't know if it is important here.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -349,11 +360,12 @@ export const {
 	selectSelectedGroups, selectQueryTimeInterval,
 	selectThreeDMeterOrGroup, selectCompareTimeInterval,
 	selectThreeDMeterOrGroupID, selectThreeDReadingInterval,
-	selectLastMeterOrGroup, selectGraphAreaNormalization,
+	selectLastMeterOrGroup,
 	selectSliderRangeInterval, selectDefaultGraphState,
 	selectHistoryIsDirty, selectPlotlySliderMax,
 	selectPlotlySliderMin, selectShiftAmount,
-	selectShiftTimeInterval, selectInitialXAxisRange
+	selectShiftTimeInterval, selectInitialXAxisRange,selectGraphAreaNormalization,selectTimeCreated
```
huss, line 367: Three thoughts:

- I'm unclear on why selectGraphAreaNormalization position was moved. Should it be put back?
- Every other one has two items per line so maybe the new one on a new line since this line already has two.
- The formatting of spaces is a little off in several places. If you are using VSC you can use the command "Format Document" to fix it up. If you need help then let me know.

src/client/app/redux/slices/appStateSlice.ts
```diff
@@ -132,7 +137,9 @@ export const appStateSlice = createThunkSlice({
 		selectOptionsVisibility: state => state.optionsVisibility,
 		selectSelectedLanguage: state => state.selectedLanguage,
 		selectChartLinkHideOptions: state => state.chartLinkHideOptions,
-		selectRefreshingReadings: state => state.refreshingReadings
+		selectRefreshingReadings: state => state.refreshingReadings,
+		selectIsKeepCurrent: state => state.isKeepCurrent
+
```
huss, line 142: Please remove extra blank line.

src/client/app/redux/selectors/uiSelectors.ts
```diff
@@ -453,9 +453,10 @@ export const selectChartLink = createAppSelector(
 		selectGraphState,
 		selectChartLinkHideOptions,
 		selectSliderRangeInterval,
+		selectIsKeepCurrent,
 		state => state.maps.selectedMap
 	],
-	(current, chartLinkHideOptions, rangeSliderInterval, selectedMap) => {
+	(current, chartLinkHideOptions, rangeSliderInterval, isKeepCurrent,selectedMap) => {
```
huss, line 459: Spacing off - see other comment.

src/client/app/redux/selectors/uiSelectors.ts
```diff
@@ -514,6 +515,12 @@ export const selectChartLink = createAppSelector(
 		if (chartLinkHideOptions) {
 			linkText += '&optionsVisibility=false';
 		}
+		if (isKeepCurrent) {
+			const timeCreatedEnd = current.timeCreated.getEndTimestamp();
+			const sliderStart = current.rangeSliderInterval.getStartTimestamp();
+			const diffDays = timeCreatedEnd.diff(sliderStart, 'days', true);
+			linkText += `&timeSpan=${diffDays.toFixed(2)}`;
```
huss, line 522: I'm curious about the ``toFix(2)``. Was it to limit the size of the link? My quick calculation indicates this would be accurate to about 15 minutes. If it was 5 decimal places then it would be about 1 second. I don't think it will matter too often but if the user set a precise time I'm wondering if there is a reason to reduce the accuracy this much. Thoughts?

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -5,15 +5,28 @@
 import * as React from 'react';
 import { toast } from 'react-toastify';
 import ReactTooltip from 'react-tooltip';
-import { Button, ButtonGroup, Input } from 'reactstrap';
+import { Button, ButtonGroup,Input} from 'reactstrap';
```
huss, line 8: space after comma.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -5,15 +5,28 @@
...
+import {
+	showErrorNotification,
+	showInfoNotification
+} from '../utils/notifications';
 import { useTranslate } from '../redux/componentHooks';
 import TooltipMarkerComponent from './TooltipMarkerComponent';
-import { wellStyle, rowFlexStart } from '../styles/modalStyle';
+import { wellStyle, rowFlexStart, labelStyle } from '../styles/modalStyle';
+import { checkboxStyle } from '../styles/modalStyle';
+
```
huss, line 28: Please remove new blank lines so only one.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -26,59 +39,134 @@ export default function ChartLinkComponent() {
 	const linkHideOptions = useAppSelector(selectChartLinkHideOptions);
 	const selectedMeters = useAppSelector(selectSelectedMeters);
 	const selectedGroups = useAppSelector(selectSelectedGroups);
+	const queryTimeInterval = useAppSelector(selectQueryTimeInterval);
+	const isKeepCurrent = useAppSelector(selectIsKeepCurrent);
 	const ref = React.useRef<HTMLDivElement>(null);
+
+	// THIS react.UseMemo ONLY RETURNS TRUE WHEN THE CONDITIONS ARE MET FOR USING KEEP CURRENT (left is is bounded and right is unbounded)
```
huss, line 46: Is the uppercase needed? I think you are trying to emphasize the point but a regular comment seems fine as there are lots of comments on the code throughout OED. Some people find uppercase to be shouting so I try to avoid.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -26,59 +39,134 @@ export default function ChartLinkComponent() {
...
+	const shouldShowKeepCurrentCheckbox = React.useMemo(() => {
+		if (!queryTimeInterval) return false;
```
huss, line 48: I'm wondering if the logic could be made simpler and clearer since you only want to return true in one case. What if you set a variable to false. Then check if the left is bounded and right unbounded (protecting against undefined which is also false) and set variable to true. (A trinary as a single statement to do this is also fine.) Then return the variable value. I note this might avoid Multiple returns that, I believe, research has shown can lead to errors.

It may not matter after the changes but OED wants braces around all if/else statements, even if a single one for clarity.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -26,59 +39,134 @@ export default function ChartLinkComponent() {
...
+			!queryTimeInterval.getIsBounded()
+		)
+			return false;
+		return true;
+	}, [queryTimeInterval]);
+
 	const handleButtonClick = () => {
 		// First attempt to write directly to user's clipboard.
-		navigator.clipboard.writeText(linkText)
+		navigator.clipboard
```
huss, line 60: I preferred the old formatting without the extra lines to shorten the code.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -26,59 +39,134 @@ export default function ChartLinkComponent() {
...
+					toast.POSITION.TOP_RIGHT,
+					1000
+				);
 				setLinkTextVisible(true);
 			});
 	};
 	if (selectedMeters.length > 0 || selectedGroups.length > 0) {
 		return (
 			<div>
+				{/* inputting new "keep chart current" feature */}
```
huss, line 82: I think this comment should just say what the code does or be removed if you think it is not necessary.

src/client/app/components/PlotNavComponent.tsx
```diff
@@ -12,7 +12,7 @@ import {
 	changeSliderRange, selectChartToRender, selectHistoryIsDirty,
 	selectSelectedGroups, selectSelectedMeters,
 	selectSliderRangeInterval, selectInitialXAxisRange,
-	selectQueryTimeInterval, updateTimeIntervalAndSliderRange
+	selectQueryTimeInterval, updateTimeIntervalAndSliderRange,updateTimeCreated
```
huss, line 15: space after comma.

src/client/app/components/ChartLinkComponent.tsx
```diff
@@ -26,59 +39,134 @@ export default function ChartLinkComponent() {
...
-										onMouseLeave={() => {
-											ref.current && ReactTooltip.hide(ref.current);
-										}}
-									/>
-								</div>
+
 							</div>
 						</Button>
-						<Button outline onClick={() => setLinkTextVisible(visible => !visible)}>
+						<Button
```
huss, line 157: Can the original formatting please be put back since shorter and will remove these lines as having changes.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -289,6 +293,12 @@ export const graphSlice = createSlice({
 							case 'shiftTimeInterval':
 								current.shiftTimeInterval = TimeInterval.fromString(value);
 								break;
+							case 'timeSpan':{
+								const days = parseFloat(value);
+								const now = moment();
```
huss, line 298: This is a subtle point but this uses moment to get the current time (as was done in the link creation). This will use the time zone of the current system to get the time. If it differs from the time zone where the link was created then it will differ by the shift amount between the time zones. This means that if a user creates a chart link and quickly sends it to someone else in a different time zone that uses it quickly, the graph will be slightly different. I think this is a small concern but here is a proposal to fix it up:

OED tries to avoid this issue with readings by storing them all in UTC. The time help page discusses this. I propose that the time used in all the calculations be UTC for the current time. Thus, the time at a given point is the same everywhere in the world in UTC and moment() will give this value. It also means that the time sent to the server for readings will relate to the time zone used for readings.

I'm also happy to help make this happen by either working with you or doing it as an addition to this work. Having done a fair amount with time zones in moment I can say it can be a little tricky to use the right code (and why there are so many comment about it in the code). I welcome your thoughts on doing this and how it will get done.

huss (changes requested):
Thanks to @w-b-d, @anntreasajojo, @rianhassett & @nickhitos for this contribution, esp. since I think it is your first one to OED. Overall, it works well. I have made a few comments within files to consider. Here are some global ones:

~The box indicating everyone signed the CLA is checked but the OED records do not show that @rianhassett & @nickhitos have signed. Could they please sign or let me know if you think our records are off.~

I tried using a chart link from a graphic that had the slider range set so the left side differed from the data range of the graphic. If I don't use the new option then it reproduces the slider range. If I use the new option then it does not. I also tried to test the new option by doing: ``const now = moment('2025-08-28 10:35:00');`` in timeSpan in src/client/app/redux/slices/graphSlice.ts. This shifted the current time forward to convince the code the chart link was used in the future. It seems to do that and changes leftBound but it does not change what the graphic shows. I have not investigated the cause. I looked at the design document and it talks about two values that needed to be fixed (in the general case):  1) the data fetched from the server (server range); 2) the points displayed on the graphic (slider range). The code in src/client/app/redux/slices/graphSlice.ts only seems to adjust ``current.rangeSliderInterval`` and I think it might relate to that. I appreciate this is subtle so please ask questions or seek my help (I can do video calls) as needed.

rianhassett:
> Thanks to @w-b-d, @anntreasajojo, @rianhassett & @nickhitos for this contribution, esp. since I think it is your first one to OED. Overall, it works well. I have made a few comments within files to consider. Here are some global ones:
> 
> The box indicating everyone signed the CLA is checked but the OED records do not show that @rianhassett & @nickhitos have signed. Could they please sign or let me know if you think our records are off.
> 
> I tried using a chart link from a graphic that had the slider range set so the left side differed from the data range of the graphic. If I don't use the new option then it reproduces the slider range. If I use the new option then it does not. I also tried to test the new option by doing: `const now = moment('2025-08-28 10:35:00');` in timeSpan in src/client/app/redux/slices/graphSlice.ts. This shifted the current time forward to convince the code the chart link was used in the future. It seems to do that and changes leftBound but it does not change what the graphic shows. I have not investigated the cause. I looked at the design document and it talks about two values that needed to be fixed (in the general case): 1) the data fetched from the server (server range); 2) the points displayed on the graphic (slider range). The code in src/client/app/redux/slices/graphSlice.ts only seems to adjust `current.rangeSliderInterval` and I think it might relate to that. I appreciate this is subtle so please ask questions or seek my help (I can do video calls) as needed.

Hey @huss! Thanks for the feedback, currently our classes just started but I do want to continue with this, I will try fixing these issues when I can.

huss:
> Hey @huss! Thanks for the feedback, currently our classes just started but I do want to continue with this, I will try fixing these iss

@rianhassett Thanks for letting me know. This work is not on a tight clock so as long as it moves forward at a reasonable pace that is fine. I understand that you have classes. Anyone else on the team is welcome to continue. Also, I have your (and all the CLAs) now.

huss:
> > Hey @huss! Thanks for the feedback, currently our classes just started but I do want to continue with this, I will try fixing these iss
> 
> @rianhassett Thanks for letting me know. This work is not on a tight clock so as long as it moves forward at a reasonable pace that is fine. I understand that you have classes. Anyone else on the team is welcome to continue. Also, I have your (and all the CLAs) now.

@w-b-d, @anntreasajojo, @rianhassett & @nickhitos I was looking at all the open PRs and wanted to touch base to verify you still want to work on this. Thanks for any information.

src/client/app/redux/slices/graphSlice.ts
rianhassett: Hello @huss
@w-b-d and I are currently working on this. I wanted to ask for your help on making this UTC friendly. I did some research into the npm moment package and I found how to use the offset for UTC to get the respective time with moment().utcOffset()
Can we just use this and store the correct time for UTC inside const now after calculating the correct UTC time?

huss: @rianhassett Yes, moment supports UTC shifts. OED does this type of calculation in a number of places in the code. Some examples are in src/client/app/components/LineChartComponent.tsx & src/client/app/components/CompareBarComponent.tsx but there are many others. You probably found the .utc() function used in these components. moment can shift to another timezone but you have to be careful to know if you want it to use the same time or the equivalent time in UTC. Also, be aware that a number of moment functions modify the item and do not create a new object so clone is used in the OED code. I'm happy to work with you on this if that would help.

rianhassett: When you say "you have to be careful to know if you want it to use the same time or the equivalent time in UTC", do we not want to always use UTC time for the chart link to address your use case of "a _user creates a chart link and quickly sends it to someone else in a different time zone that uses it quickly, the graph will be slightly different_".
Also yes we'd love to work on you with this. Would it be easier for you if we communicated through the discord?

huss: @rianhassett I have been thinking about this and rereading the design doc. I think there might be some subtle considerations here that are not reflected in the design and I need to think about them. Pleas give me a little more time to think them through and I should get back to you soon.

As for how to communicate. If it is just text then GitHub is fine (and keeps a record here) unless it is something that should not be here. Having said that, I'm also okay with Discord. What I was actually thinking about is if you wanted to talk so using Discord (or another service you prefer) would let us do that to work out these details in a way that more readily supports a real-time discussion. So, let me know if you would like to talk and we can set it up.

rianhassett: Hey @huss sounds good! If you need help with any doc work let us know I'd be happy to help. And discord voice call works aswell, should we work it out after the doc work is updated? Or do you think we can still move on regardless.

huss: @rianhassett Sorry for the delay in responding. I've only given this a little thought but I'm starting to think some modification in how the current time used is determined may be necessary. I'm happy to talk about this at any point as getting input from others usually helps. If you would like to chat, why don't you send me a DM on Discord about what might work for you and then we can settle a time.

huss: @rianhassett As I feared, fixing for the edge cases was tricky. I've updated the [design doc](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/latestChartlink/latestChartlink.md#260208-update) so it should now be correct. I note this is a non-trivial change so I wanted to see if you wished to pursue this or make it a future task?

w-b-d:
Hi @huss, sorry for the delay, I made some changes to our original implementation after reading the updated design doc. 

The main changes:

Replaced the timeSpan parameter with currentTime, which includes the time zone offset of the browser creating the link, and implemented the steps from the 260208 update when the link is used. 

Fixed the issue where only rangeSliderInterval was adjusted — the link now shifts both queryTimeInterval and the slider range together.

timeCreated is now updated whenever the graphic is recreated (date picker, refresh, or hotlink), not just on refresh.

Would you mind taking a look to see if this matches what the design doc is requesting?

One question: in this implementation, when a keep-current link is used the resulting server range ends bounded at the computed current time (from my understanding of step 5 of the 260208 update). This means the person opening the link can't immediately re-share it as a keep-current link, since the checkbox is disabled when the right side is bounded. If you'd prefer the used link to keep the right side unbounded, I can update it.

Thank you!

huss:
@w-b-d Just a quick note. What you did was fine but I'm going to ask that you not resolve any more comments. I am going to use unresolved ones to know what I still need to look at again. You can either reply or add a thumbs up (if you think it is done) and then I'll look at it the next time. Thanks.

as
huss: I still see this file. It probably should be deleted. I can do it/help if you want as it isn't hard with git if you know the process.

d
huss: I still see this file. It probably should be deleted. I can do it/help if you want as it isn't hard with git if you know the process.

.vscode/launch.json
```diff
@@ -2,6 +2,7 @@
 	"version": "0.2.0",
 	"configurations": [
 		{
+
```
huss, line 5: Thanks for removing the blank I commented on. I think I missed this on my previous review. This is another added blank line.

.vscode/launch.json
```diff
@@ -44,6 +45,7 @@
 				"Attach to Node in Docker",
 				"Launch Firefox against localhost"
 			]
+
```
huss, line 48: Another added blank line I missed before.

.gitignore
```diff
@@ -53,9 +53,10 @@ nohup.out
 
 # Visual Studio Code settings file
 .vscode/settings.json
-
 # tmp file
 src/server/tmp/*
+as
```
huss, line 58: I think I understand what you might have tried. Adding these two lines to the .gitignore should stop them from being saved going forward. However, they won't remove them if they were already committed (as done). Thus, delete those files with git (see comments in those files). Then you can you remove these two lines here since they won't exist.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -349,11 +376,12 @@ export const {
...
 	selectThreeDMeterOrGroup, selectCompareTimeInterval,
 	selectThreeDMeterOrGroupID, selectThreeDReadingInterval,
-	selectLastMeterOrGroup, selectGraphAreaNormalization,
+	selectLastMeterOrGroup,
 	selectSliderRangeInterval, selectDefaultGraphState,
 	selectHistoryIsDirty, selectPlotlySliderMax,
 	selectPlotlySliderMin, selectShiftAmount,
-	selectShiftTimeInterval, selectInitialXAxisRange
+	selectShiftTimeInterval, selectInitialXAxisRange,
+	selectGraphAreaNormalization
```
huss, line 384: Sorry to be picky about this but this used to be a few lines above as:

```
selectLastMeterOrGroup, selectGraphAreaNormalization,
```

This makes it two for each line and keeps the order the same. Could you please do this.

src/client/app/components/ChartLinkComponent.tsx
huss: A commit seems to remove a blank line but there are still two of them. Could you please fix this up unless I'm wrong.

.gitignore
```diff
@@ -53,9 +53,10 @@ nohup.out
 
 # Visual Studio Code settings file
 .vscode/settings.json
```
huss, line 55: I appreciate you removing the ignore for launch.json. Could you add a blank line after this line so there is a space before the next comment as for other items.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -289,8 +294,30 @@ export const graphSlice = createSlice({
 							case 'shiftTimeInterval':
 								current.shiftTimeInterval = TimeInterval.fromString(value);
 								break;
+							case 'currentTime':
+								keepCurrentTime = value;
+								break;
 						}
 					});
+					if (keepCurrentTime !== undefined) {
```
huss, line 302: Some commenting throughout this block of code would help or saying to look at the design doc for the info.

src/client/app/components/ChartLinkComponent.tsx
huss: No longer an item given the other changes.

src/client/app/redux/slices/graphSlice.ts
huss: My testing indicates that this issue is gone so the new design and implementation should cover it.

huss:
> One question: in this implementation, when a keep-current link is used the resulting server range ends bounded at the computed current time (from my understanding of step 5 of the 260208 update). This means the person opening the link can't immediately re-share it as a keep-current link, since the checkbox is disabled when the right side is bounded. If you'd prefer the used link to keep the right side unbounded, I can update it.

Good catch. This value is needed in future steps so I think it must be computed. However, I don't think the server end time should be bounded on the right. Here is my reasoning:

- It could not have been when the graphic was created since the keep current option was used. The parallel would be to do the same when used.
- There used to be issues with the range of time shown. Those are now resolved with the careful application of time zone (at least I hope so :-). Thus, leaving it unbounded would show the same range of time assuming there is data in the full range requested.
- The cases when the meter data does not exist or go beyond current time have always lead to seeing more/less data in OED. It would be hard to fix (at best) so this seems fine here and consistent.

So, if you can change so the code does not use this value for the actual end server time and leaves it unbounded then I'm hoping it will all be good.

huss (changes requested):
@w-b-d Thank you for updating the code. I have now reviewed and tested. There are a modest number of hopefully straightforward comments to address. I resolved any that were done so only current comments should show by default. I assume you will see my comment about not resolving comments so I can track the work. Also, I tried to break the code but have not found any issues so it seems the implementation is good. Please let me know if anything is not clear, you have thoughts/questions or I can help. I think this should be done soon after the next update.

huss:
@w-b-d, @anntreasajojo, @rianhassett & @nickhitos I'm sweeping PRs to figure out their status. Could you let me know if you have or plan to address the open comments? Thanks.

w-b-d:
@huss Sorry for the delay, been busy with work. I do have a plan to address the open comments, I will have them done hopefully by the end of the week!

PR #1532, opened 2025-09-01
Updating cik for time varying

TheREALNguyenAnh:
# Description

This create a new cikVary table. This also implement a redoCikVary that is similar to redoCik. redoCikVary use conversionSegments in the database, and make the cikVary Table.

## Type of change

## Checklist

## Limitations

~I have not put in suffix units and handle them.~ This was done.


package.json
```diff
@@ -45,7 +45,9 @@
 		"generateTestingData": "node -e 'require(\"./src/server/data/automatedTestingData\").generateTestingData()'",
 		"testData": "node -e 'require(\"./src/server/data/automatedTestingData.js\").insertSpecialUnitsConversionsMetersGroups()'",
 		"webData": "node -e 'require(\"./src/server/data/websiteData.js\").insertWebsiteData()'",
-		"addLogMsg": "node -e 'require(\"./src/server/services/addLogMsg.js\").addLogMsgToDB()'"
+		"addLogMsg": "node -e 'require(\"./src/server/services/addLogMsg.js\").addLogMsgToDB()'",
+		"testCikVary": "node ./src/server/tmp/uploads/createCikVaryTest.js",
```
huss: While these two npm commands were useful during testing, I think they should be removed in the final version to merge.

src/server/sql/cik_vary/get_cik_vary.sql
```diff
@@ -0,0 +1,9 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+-- Get all cik_vary entries through joining cik_vary and units tables.
```
huss: This existed in cik before but it does not do what it claims. It seems to get all cik but rename two columns. Maybe the comment in both files could be updated. I also wonder if the name should be get_all_cik_vary and get_all_cik.

src/server/routes/cikVary.js
```diff
@@ -0,0 +1,58 @@
...
+  return {
+    meterUnitId: item.meterUnitId,
+    nonMeterUnitId: item.nonMeterUnitId,
+    startTime: item.startTime,
+    endTime: item.endTime,
+    slope: item.slope,
+    intercept: item.intercept
+  };
+}
+
```
huss: Unlike cik that is used in the client, I'm not clear where these routes would be used. If only used on the server then a model would be sufficient. Did you have an intended usage or could they go? If they need to stay then the authentication needs to be put in place (likely just admin).

src/server/sql/cik_vary/get_all_conversions.sql
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
huss: I cannot find where this is used nor the old one in cik so I'm wondering if both should be removed? I think get_cik_vary is actually used.

src/server/sql/cik_vary/delete_all_conversions.sql
```diff
@@ -0,0 +1,8 @@
+/*
```
huss, line 1: I think the name could be much better as delete_all_cik_vary. This is the same name as cik so it really should be delete_all_cik.

src/server/sql/cik_vary/get_conversion.sql
```diff
@@ -0,0 +1,6 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss, line 1: I think this would be better if named get_ci_vary_by_source_destination_start_end since it would be similar to the new conversion segment. OED is now inconsistent about including the name of the item in these files so I say just leave it here for now.

src/server/sql/cik_vary/insert_new_conversion.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss, line 1: For similar reasons, I think the name insert_new_cik_vary is better. Similarly, insert_new_cik for cik one.

src/server/models/CikVary.js
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	/**
+	* Create a new CikVary object from row's data.
+	* @param {*} row The row from which CikVary will be created.
+	* @returns the created CikVary object
+	*/
+	static mapRow(row) {
+		return new CikVary(
+			row.meter_unit_id || row.source_id,
```
huss, line 46: I suspect this || can go away once the SQL files are changed as proposed. Seeing this makes me wonder if just returning the actual row name and not renaming in the SQL is better and it would be done here.

src/server/models/CikVary.js
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	* Get CikVary conversions valid at a specific time.
+	* @param {*} conn The database connection to use.
+	* @param {*} sourceId Source unit id.
+	* @param {*} destinationId Destination unit id.
+	* @param {*} queryTime Timestamp to check validity.
+	* @returns Matching CikVary objects
+	*/
+	static async getConversionAtTime(conn, sourceId, destinationId, queryTime) {
```
huss: This name does not mirror src/server/models/ConversionSegment.js but I think it should. I know the name would be longer but I think getBySourceDestinationStartEnd is better. This also removes the model name but OED seems to do that fairly consistently.

src/server/models/CikVary.js
```diff
@@ -0,0 +1,106 @@
...
+	* @param {*} conn The database connection to use.
+	* @returns all CikVary objects
+	*/
+	static async getAll(conn) {
+		const rows = await conn.any(sqlFile('cik_vary/get_cik_vary.sql'));
+		return rows.map(CikVary.mapRow);
+	}
+
+	/**
+	* Get CikVary conversions valid at a specific time.
```
huss: I'm unsure why valid and validity is in the comments. It might be how it is used but this seems to just return the cik vary of interest.

src/server/models/CikVary.js
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	* Inserts each element of the array with an actual time-varying conversion into the cik_vary table.
+	* The current values in the table are removed first.
+	* @param {*} cikVaryArr Array of time-varying conversions.
+	* @param {*} conn The database connection to use.
+	*/
+	static async insert(cikVaryArr, conn) {
+		// Remove all the current values in the table.
+		await conn.none(sqlFile('cik_vary/delete_all_conversions.sql'));
```
huss: I think this entire process should be in a transaction in case some step fails.

src/server/models/ConversionSegment.js
```diff
@@ -326,6 +326,21 @@ class ConversionSegment {
...
 	}
+
+	/**
+	 * Get all conversion segments for a specific edge (sourceId, destinationId).
+	 * @param {*} conn The database connection to use.
+	 * @param {*} sourceId Source unit id.
+	 * @param {*} destinationId Destination unit id.
+	 * @returns {Promise.<Array.<ConversionSegment>>}
+	 */
+	static async getAllForEdge(conn, sourceId, destinationId) {
```
huss: Is this the same as getBySourceDestination that already existed? If so, then use that one. If this does need to stay then conn is normally the last argument.

src/server/services/graph/createConversionArrays.js
```diff
@@ -65,6 +67,52 @@ async function createCikArray(graph, conn) {
...
+/**
+ * Returns the CikVary array: all time-varying conversions between each meter and unit.
+ * Each entry is { source, destination, start_time, end_time, slope, intercept }
+ * @param {*} graph The conversion graph.
+ * @param {*} conn The connection to use.
+ * @returns Array of time-varying conversion segments.
+ */
+async function createCikVaryArray(graph, conn) {
+	const sources = await Unit.getTypeMeter(conn);
+	//console.log('Sources:', sources.map(u => ({ id: u.id, name: u.name, type: u.type })));
```
huss: Please remove all commented out console.log.

src/server/services/graph/createConversionArrays.js
```diff
@@ -65,6 +67,52 @@ async function createCikArray(graph, conn) {
...
-	createCikArray
+/**
+ * Returns the CikVary array: all time-varying conversions between each meter and unit.
+ * Each entry is { source, destination, start_time, end_time, slope, intercept }
+ * @param {*} graph The conversion graph.
+ * @param {*} conn The connection to use.
+ * @returns Array of time-varying conversion segments.
+ */
+async function createCikVaryArray(graph, conn) {
+	const sources = await Unit.getTypeMeter(conn);
```
huss, line 77: I'm just noting that when createCikArray is fused with this then the comments from that need to be updated and included here.

src/server/services/graph/createConversionArrays.js
```diff
@@ -65,6 +67,52 @@ async function createCikArray(graph, conn) {
...
+		return toReturn;
+	}
+	for (const source of sources) {
+		for (const destination of destinations) {
+			const sourceId = source.id;
+			const destinationId = destination.id;
+			const path = getPath(graph, sourceId, destinationId);
+			//console.log(`Path from ${sourceId} to ${destinationId}:`, path);
+			if (path !== null) {
+				const segments = await timeVaryingPathConversion(path, conn, getEdgeConversions);
```
huss: I think an overview comment on what this is doing would be good.

src/server/services/graph/createConversionGraph.js
```diff
@@ -30,6 +31,34 @@ async function createConversionGraph(conn) {
...
+ * @returns {Object}
+ */
+async function createConversionSegmentGraph(conn) {
+	const graph = createGraph();
+	const units = await Unit.getAll(conn);
+	for (const unit of units) {
+		graph.addNode(unit.id, unit.name);
+	}
+
+	// Get all conversion segments (may be multiple per edge, but we only care about existence)
```
huss: I'm wondering about this process. If a conversion exists then there must be one or more conversion segments with the same source/destination. Thus, I'm wondering why this separate process is needed. I might be missing something. If it is needed then maybe the DB could return just the unique source/destination to simplify the logic.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -4,6 +4,7 @@
 
 const Unit = require('../../models/Unit');
 const { pathConversion } = require('./pathConversion');
+const { timeVaryingPathConversion } = require('./timeVaryingPathConversion');
```
huss: I don't see where this import is used. If not, the file seems unchanged.

src/server/services/graph/redoCik.js
```diff
@@ -2,10 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const { createConversionGraph } = require('./createConversionGraph');
-const { createCikArray } = require('./createConversionArrays');
+const { createConversionGraph, createConversionSegmentGraph } = require('./createConversionGraph');
```
huss: I don't see where createConversionSegmentGraph is used.

src/server/services/graph/redoCik.js
```diff
@@ -34,7 +35,36 @@ async function updateCikAndViews() {
...
+/**
+ * Creates CikVary based on units and conversion segments and then inserts these values
+ * in the cik_vary table in the database.
+ */
+async function redoCikVary(conn) {
+	// Create graph based on units and conversion segments.
+	const graph = await createConversionGraph(conn);
+	
+	// Processes suffix units to update graph and database (not used for now).
+	// await handleSuffixUnits(graph, conn);
```
huss: I thought we had decided that this was okay and should be done. Do I remember incorrectly? Is there an issue I'm not thinking about?

src/server/services/graph/createConversionArrays.js
```diff
@@ -65,6 +67,52 @@ async function createCikArray(graph, conn) {
...
+ * @returns Array of time-varying conversion segments.
+ */
+async function createCikVaryArray(graph, conn) {
+	const sources = await Unit.getTypeMeter(conn);
+	//console.log('Sources:', sources.map(u => ({ id: u.id, name: u.name, type: u.type })));
+	const destinations = (await Unit.getTypeUnit(conn)).concat(await Unit.getTypeSuffix(conn));
+	//console.log('Destinations:', destinations.map(u => ({ id: u.id, name: u.name, type: u.type })));
+	const c = [];
+	// Helper to fetch all segments for an edge
+	async function getEdgeConversions(sourceId, destinationId, conn) {
```
huss: Do you think this function should be created here and passed or declared in src/server/services/graph/timeVaryingPathConversion.js since it appears to be the only place it is used?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+ * @param {*} getEdgeConversions Function to fetch all time-varying conversions for an edge.
+ * @returns Array of {source, destination, startTime, endTime, slope, intercept}
+ */
+async function timeVaryingPathConversion(path, conn, getEdgeConversions) {
+    // Fetch all time-varying conversions for each edge in the path
+    const edgeSegments = [];
+    for (let i = 0; i < path.length - 1; ++i) {
+        const sourceId = path[i].id;
+        const destinationId = path[i + 1].id;
+        // getEdgeConversions should return sorted array of {start_time, end_time, slope, intercept}
```
huss: Just to be precise, it sorts by start_time.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+function parsePostgresDate(val) {
+
+    if (val === 'infinity') {
+        return Number.POSITIVE_INFINITY;
+    }
+    if (val === '-infinity') {
+        return Number.NEGATIVE_INFINITY;
+    }
+    if (typeof val === 'string') {
+        return new Date(val).getTime();
```
huss: Any chance that moment could be used? I feel we talked about this but cannot remember the decision.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+ * Chains time-varying conversions along a path, producing combined segments for cik_vary.
+ * Each edge in the path may have multiple time segments (start_time, end_time, slope, intercept).
+ * The algorithm aligns all segments and combines them for each time range.
+ * @param {*} path Array of units (nodes) from source to destination.
+ * @param {*} conn Database connection.
+ * @param {*} getEdgeConversions Function to fetch all time-varying conversions for an edge.
+ * @returns Array of {source, destination, startTime, endTime, slope, intercept}
+ */
+async function timeVaryingPathConversion(path, conn, getEdgeConversions) {
+    // Fetch all time-varying conversions for each edge in the path
```
huss: File uses space instead of tab indenting.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+                seg = segments.find(s =>
+                    parsePostgresDate(s.startTime) <= startTime &&
+                    parsePostgresDate(s.endTime) >= endTime &&
+                    s.sourceId === path[edgeIndex + 1].id &&
+                    s.destinationId === path[edgeIndex].id
+                );
+                if (seg) {
+                    // Check bidirectional in conversions table
+                    const bidirectional = await isBidirectional(s.sourceId, s.destinationId, conn);
+                    if (bidirectional) invert = true;
```
huss: OED likes braces around all if/loop statements even if only one. Also, is there a need for both bidirectional and invert?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+
+            let segSlope = seg.slope;
+            let segIntercept = seg.intercept;
+            if (invert) {
+                [segSlope, segIntercept] = invertConversion(segSlope, segIntercept);
+            }
+
+            [slope, intercept] = updatedConversion(slope, intercept, segSlope, segIntercept);
+        }
+        if (valid) {
```
huss: Should not valid be an error or some sort? Also, can it happen if the segments are correct?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+ * Each edge in the path may have multiple time segments (start_time, end_time, slope, intercept).
+ * The algorithm aligns all segments and combines them for each time range.
+ * @param {*} path Array of units (nodes) from source to destination.
+ * @param {*} conn Database connection.
+ * @param {*} getEdgeConversions Function to fetch all time-varying conversions for an edge.
+ * @returns Array of {source, destination, startTime, endTime, slope, intercept}
+ */
+async function timeVaryingPathConversion(path, conn, getEdgeConversions) {
+    // Fetch all time-varying conversions for each edge in the path
+    const edgeSegments = [];
```
huss: This algorithm differs from the one in the design doc. Could you explain to me the advantages?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+    if (val === Infinity) {
+        return 'infinity';
+    }
+    if (val === -Infinity) {
+        return '-infinity';
+    }
+    return new Date(val);
+}
+
+function invertConversion(slope, intercept) {
```
huss: This conversion existed in src/server/services/graph/pathConversion.js. Is there a reason not to use that one?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,144 @@
...
+                // Try to find segment in reverse direction
+                seg = segments.find(s =>
+                    parsePostgresDate(s.startTime) <= startTime &&
+                    parsePostgresDate(s.endTime) >= endTime &&
+                    s.sourceId === path[edgeIndex + 1].id &&
+                    s.destinationId === path[edgeIndex].id
+                );
+                if (seg) {
+                    // Check bidirectional in conversions table
+                    const bidirectional = await isBidirectional(s.sourceId, s.destinationId, conn);
```
huss: I would think a given source/destination (the outer loop) would decided if the conversion is bidirectional so only it only needs to be checked once.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,150 @@
...
+		'-infinity', 'infinity', '2020-01-01 00:00:00', conn
+	);
+
+	// Split at 2020-06-01 00:00:00 (on the segment starting at 2020-01-01)
+	await ConversionSegment.splitLater(
+		unit10Id, unit1Id, null, 3, 0, null,
+		'2020-01-01 00:00:00', 'infinity', '2020-06-01 00:00:00', conn
+	);
+
+	// Split at 2021-01-01 00:00:00 (on the segment starting at 2020-06-01)
```
huss: Status of commented out code.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,150 @@
...
+	const conversionPreInsert2 = new Conversion(unit1Id, unit2Id, false, 'note');
+	await Promise.all([
+		conversionPreInsert1.insert(null, 1, 0, "", conn),
+		conversionPreInsert2.insert(null, 1, 0, "notes", conn)
+	]);
+}
+
+mocha.describe('CIK Vary Chaining', () => {
+
+	mocha.describe('Insert test unit', function () {
```
huss: I'm not sure the second describe is needed. I tried commenting it out and it worked fine with a good description. Applies to other tests too.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,150 @@
...
+		mocha.it('should insert units without error', async function () {
+			const conn = testDB.getConnection();
+			await insertTestUnits(conn);
+			const units = await conn.any('SELECT * FROM units');
```
huss: Is there a reason you use a direct DB query rather than the model to get values? Here you do units but you do it in other places too.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,150 @@
...
+			await insertTestUnits(conn);
+			await insertTestConversions(conn);
+			await insertTestSegments(conn);
+			const segments = await conn.any('SELECT * FROM conversion_segments');
+			expect(segments).to.have.lengthOf(7);
+		});
+	});
+
+	mocha.describe('Update cik_vary and views', function () {
+		mocha.it('should update cik_vary with the correct amount of segments', async function () {
```
huss: I can see the value of having separate tests while these were developed. However, in many OED tests, it first checks the length and then tests each item. This means these two tests could be combined. It would also mean that the current second test would also test length before items which is a stronger test.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,150 @@
...
+}
+
+mocha.describe('CIK Vary Chaining', () => {
+
+	mocha.describe('Insert test unit', function () {
+		mocha.it('should insert units without error', async function () {
+			const conn = testDB.getConnection();
+			await insertTestUnits(conn);
+			const units = await conn.any('SELECT * FROM units');
+			expect(units).to.have.lengthOf(3);
```
huss: See below about combining tests. Since it is unlikely that the lengths would be wrong it seems okay to also combine these into one test to reduce the number.

huss (changes requested):
Thanks to @TheREALNguyenAnh for another contribution. I've made a number of comments to consider. Overall, this seems to work fine. It is harder to test since all the pieces are not yet in place for time-varying conversion. Thus, careful testing will happen later. Having your new test code helps with this for now.

src/server/models/CikVary.js
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	* Get CikVary conversions rate at a specific time.
+	* @param {*} conn The database connection to use.
+	* @param {*} sourceId Source unit id.
+	* @param {*} destinationId Destination unit id.
+	* @param {*} queryTime Timestamp to check validity.
+	* @returns Matching CikVary objects
+	*/
+	static async getBySourceDestinationTime(conn, sourceId, destinationId, queryTime) {
```
huss: I don't have an issue with the name except it does not match src/server/models/ConversionSegment.js which has getBySourceDestinationStartEnd. Given that the other one is used by the front-end team, I'm inclined to leave it unchanged which would mean this one has to change to do StartEnd instead of Time. It also matches the sql file name. If you think time is better then please create an issue to change it consistently across OED.

src/server/sql/cik_vary/get_cik_vary.sql
huss: The update comment is fine. How do you feel about adding "all" to the name?

src/server/sql/cik_vary/insert_new_conversion.sql
huss: Would you like to update the sql/cik/insert_new_conversion name or should I?

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,151 @@
...
+
+	mocha.describe('Update cik_vary and views', function () {
+		mocha.it('should update cik_vary with the correct amount of segments', async function () {
+			const conn = testDB.getConnection();
+			await insertTestUnits(conn);
+			await insertTestConversions(conn);
+			await insertTestSegments(conn);
+			await redoCikVary(conn);
+			const results = await conn.any('SELECT * FROM cik_vary ORDER BY source_id, destination_id, start_time');
+
```
huss: I'm not sure this empty line is needed but if it stays then the tabs at start should go away.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,106 @@
...
+ */
+async function timeVaryingPathConversion(path, conn) {
+
+	// 1. Fetch and sort segments for each edge
+	const edgeSegments = [];
+	for (let i = 0; i < path.length - 1; ++i) {
+		const sourceId = path[i].id;
+		const destinationId = path[i + 1].id;
+		let segments = await ConversionSegment.getBySourceDestination(sourceId, destinationId, conn);
+		segments = segments.sort((a, b) => parsePostgresDate(a.startTime) - parsePostgresDate(b.startTime));
```
huss: I think this uses src/server/sql/conversionSegment/get_by_source_destination.sql that has ``ORDER BY start_time::TIMESTAMP ASC;``. Is the local sorting needed?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,106 @@
...
+		let done = false;
+		for (let i = 0; i < pointers.length; ++i) {
+			if (parsePostgresDate(currentSegments[i].endTime) === currentEnd) {
+				pointers[i]++;
+				if (pointers[i] >= edgeSegments[i].length) {
+					done = true;
+				}
+			}
+		}
+		if (done || currentEnd === Number.POSITIVE_INFINITY) {
```
huss: Could this be a do/until with this as the condition rather than a break?

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,106 @@
...
+		edgeSegments.push(segments);
+	}
+
+	// 2. Initialize pointers for each edge
+	const pointers = Array(path.length - 1).fill(0);
+
+	// 3. Main loop
+	let currentStart = Number.NEGATIVE_INFINITY;
+	const results = [];
+	let loopCount = 0;
```
huss: I don't see where this is used.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,106 @@
...
+			source: path[0].id,
+			destination: path[path.length - 1].id,
+			start_time: toPostgresTimestamp(currentStart),
+			end_time: toPostgresTimestamp(currentEnd),
+			slope,
+			intercept
+		});
+
+		// Advance pointers for segments ending at currentEnd
+		let done = false;
```
huss: I'm wondering about checking if the currentEnd is infinity here to avoid the loop. I also think it avoids the need to check about going past the end of the array in the loop. I have not tested this but something like:

```js
let done = true;
if (currentEnd !== Number.POSITIVE_INFINITY) {
    // Have not included all paths to end so continue.
    let done = false;
    for (let i = 0; i < pointers.length; ++i) {
        if (parsePostgresDate(currentSegments[i].endTime) === currentEnd) {
            // This segment at currentEnd so advance.
            pointers[i]++;
        }
    }
    // Advance the currentStart to the previous end for next round.
    currentStart = currentEnd;
}
// The loop exists if done is true without break if do/while.
```

Now maybe there is a case I am not thinking about that would be an issue.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,106 @@
...
+		loopCount++;
+		// Find current segments for each edge
+		const currentSegments = edgeSegments.map((segments, idx) => segments[pointers[idx]]);
+		// Find minimum end time among current segments
+		let currentEnd = Math.min(...currentSegments.map(seg => parsePostgresDate(seg.endTime)));
+
+		// Combine conversions for the path
+		let slope = 1, intercept = 0;
+		for (const seg of currentSegments) {
+			[slope, intercept] = updatedConversion(slope, intercept, seg.slope, seg.intercept);
```
huss, line 66: I previously made the comment:

| I would think a given source/destination (the outer loop) would decided if the conversion is bidirectional so only it only needs to be checked once.

and I talked with you about bidirectional. I think what I have said was off. Each segment is part of a conversion along the path. If the source/destination are backward from the path then it needs to be inverted so it is in the correct direction. This process needs to take that into account. For any given conversion in the path this will be the same for each segment used for that conversion so it can either be determined before this loop and saved or it is also fine to do each time.

Also, a previous comment that I resolved had:

| This conversion existed in src/server/services/graph/pathConversion.js. Is there a reason not to use that one?

I think that still applies here and would give both the inversion and update function to use.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	return results;
+}
+
+/**
+ * Chains two conversions: (slope1, intercept1) and (slope2, intercept2)
+ * Returns [slope, intercept] for the combined conversion.
+ */
+function updatedConversion(origSlope, origIntercept, newSlope, newIntercept) {
```
huss: See above about need for this function.

huss (changes requested):
Thanks to @TheREALNguyenAnh for another update to the code. I made some new comments based on the changes and there are still unresolved comments from before (mostly in cikVaryTests.js).

~The description of the PR has "I have not put in suffix units and handle them.". I think that may now be done so please update if you agree.~ I changed the text.

Finally, I have not yet tested the code. I'm waiting for the final algorithm to do that but it looks fine to me.

src/server/services/graph/timeVaryingPathConversion.js
```diff
@@ -0,0 +1,118 @@
...
+			start_time: toPostgresTimestamp(currentStart),
+			end_time: toPostgresTimestamp(currentEnd),
+			slope,
+			intercept
+		});
+		
+		if (currentEnd === Number.POSITIVE_INFINITY) {
+			done = true;
+		}
+		// Advance pointers for segments ending at currentEnd
```
huss: Could the rest of this function be an else of the above if? It avoids doing the work if it is done already. Also, maybe the check on pointers length would not be needed?

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,83 @@
...
+		unit1Id, unit2Id, null, 8, 0, null,
+		'2020-06-01 00:00:00', 'infinity', '2021-01-01 00:00:00', conn
+	);
+}
+
+mocha.describe('redoCikVary integration', function () {
+	let conn;
+	mocha.beforeEach(function () {
+		conn = testDB.getConnection();
+	});
```
huss: I'm unclear on why two beforeEach are needed. Could they be combined?

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,83 @@
...
+		await redoCikVary(conn);
+		const results = await CikVary.getAll(conn);
+		const expectedSlopes = {
+			'1->2': [1, 2, 3],
+			'1->3': [1, 12, 21, 24],
+		};
+		const grouped = {};
+		results.forEach(row => {
+			const key = `${row.source_id}->${row.destination_id}`;
+			if (!grouped[key]) grouped[key] = [];
```
huss: Please put {} around statement in if.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+const { mocha, expect, testDB } = require('../common');
+const { redoCikVary } = require('../../services/graph/redoCik');
+const Conversion = require('../../models/Conversion');
+const Unit = require('../../models/Unit');
+const ConversionSegment = require('../../models/ConversionSegment');
+const CikVary = require('../../models/CikVary');
+
+async function setupTestData(conn) {
```
huss, line 14: I worked out the test setup and what cik vary should be but it took a little time. Maybe a comment with the original and cik values would be valuable? Some simple version of how the design doc discussed its example. I can help if you want.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,83 @@
...
+		await redoCikVary(conn);
+		const results = await CikVary.getAll(conn);
+		expect(results).to.be.an('array').that.is.not.empty;
+		expect(results).to.have.lengthOf(7);
+	});
+
+	mocha.it('should have correct slopes for each cik_vary segment', async function () {
+		await redoCikVary(conn);
+		const results = await CikVary.getAll(conn);
+		const expectedSlopes = {
```
huss, line 100: Does this assume unit10 has id 1, unit 1 has id 2 and unit 2 has id 3? While this is likely it is not guaranteed. I think it might be better to use the unit10Id, unit1Id & unit2Id by making them available here. This would parallel their use in split above.

Another option that would be a little more general is to do this in the beforeEach. Then the expected length is the sum of two key value array lengths in expectedSlopes (3 + 4 = 7).

Also, it seems this assumes the segments come back in a given order which may not happen or that the order does not matter. If you want to put in a comment that is okay. I think the right way is to also look at the start/end datetime stamp for the compare. If they were sorted by that time then they should line up. There may be an issue with the infinity values and moment to fix up with this.

src/server/models/CikVary.js
huss: I'm going to resolve this but some day OED should be consistent on which one of these is used.

src/server/test/db/cikVaryTests.js
```diff
@@ -0,0 +1,83 @@
...
+			'1->2': [1, 2, 3],
+			'1->3': [1, 12, 21, 24],
+		};
+		const grouped = {};
+		results.forEach(row => {
+			const key = `${row.source_id}->${row.destination_id}`;
```
huss: This test is failing for me. I looked into it and it seems the source_id is actually meterUnitId with a similar issue on destination. It worked after I changed these.

huss (changes requested):
Thanks to @TheREALNguyenAnh for the update. Overall, it addressed the comments. I made a few new ones about the updated code to consider.

huss (approved):
Thank to @TheREALNguyenAnh for the updated code. Review found no issues. Complete testing will need to wait until all the pieces of time-varying conversions are in place. This contribution is ready to merge.

PR #1536, opened 2025-09-15
Test Case LR7

HienXuanPham:
# Description

Test case LR7 tests range with partial days/hours for daily gives only full days.

Partly Addresses [#962](https://github.com/OpenEnergyDashboard/OED/issues/962)

## Type of change

Test case

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [ ] I have removed text in ([/testing/testing.md](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/testing/testing.md)) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations

None

PR #1539, opened 2025-09-17
Test Case LR6

tinisha-davis:
# Description

Test Case LR6
- Added LR6 test for raw points without min/max 
- Ran full test suite with `npm test`: all tests passing
- Verified LR6 specifically with `--grep "LR6"`

Fixes #1536 

## Type of change
Test Case

## Checklist


src/server/test/web/readingsLineMeterRangeQuantity.js
```diff
@@ -11,20 +11,22 @@ const { chai, mocha, app } = require('../common');
...
-	unitDatakWh,
-	conversionDatakWh,
-	meterDatakWh } = require('../../util/readingsUtils');
-
-mocha.describe('readings API', () => {
-	mocha.describe('readings test, test if data returned by API is as expected', () => {
-		mocha.describe('for line charts', () => {
-			mocha.describe('for range (min/max)', () => {
-				mocha.describe('for quantity meters', () => {
+		expectRangeToEqualExpected,
```
Chocopepero: Lines 14-29 were modified unnecessarily. Please revert these change or use an auto formatter

src/server/test/web/readingsLineMeterRangeQuantity.js
```diff
@@ -90,7 +92,42 @@ mocha.describe('readings API', () => {
...
+						// Send API request with LR6 date range and graphic unit
+						const res = await chai.request(app).get(`/api/unitReadings/line/meters/${METER_ID}`).query({
+							timeInterval: createTimeString('2022-09-21', '00:00:00', '2022-10-05', '00:00:00'),
+							graphicUnitId: unitId
+						});
+
+						// Extract rows from API response
+						const rows = res.body[METER_ID] || [];
+						// Loop through each row if API response and compare to corresponding expected row
+						rows.forEach((row, i) => {
```
Chocopepero: Instead of using a for loop to iterate through and verifying each value manually, please use the function expectRangeToEqualExpected(Response, Expected).

Chocopepero:
Thank you for your contribution to OED. I've verified that you've signed the CLA and have made some suggestions for changes.

huss:
@tinisha-davis It has been a week since comments were made on this PR. I wanted to see if you needed anything.

tinisha-davis:
Hey,

Thanks for checking in. The main reason it’s been a week is because I got cut with glass on my left hand, hand stitches put in and taken out and I’m having surgery on it this upcoming Monday.

When you left the comment, I started working on reformatting and using the helper method right away. I did run into an issue with the helper method, the LR6 test has null values and the helper method doesn’t accept null values. That’s what prevented me from being able to resubmit the same day. Paired with my hand issues and it’s taking me longer than I expected to finish the task! I’m hoping I can get it working before Mondays surgery.

 In all fairness I haven’t dedicated much time to it due to doctor appointments but I have today and tomorrow off and so I’d like to get it finished this week.

If you have any advice about the helper method and null values I’d appreciate it!

Thank you for being a rad maintainer!
Tinisha Davis

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

huss:
@tinisha-davis I'm sorry to hear about your injury and wish you the best in getting through it. The delay is not an issue - I just wanted to know if we could help.

I'm not sure why the test case is returning null values. Any chance you could commit/push your update even though it does not work? I could then look at it to try to figure out what is going on. If you prefer to work together to figure this out then we can arrange a video call. When you are up to it, let me know what you think.

tinisha-davis:
A video call would be awesome! What’s your availability today or tomorrow? The helper method doesn’t return null values correct but the time period the LR6 testing is using includes null values and so those are being passed into the helper function as parameters and it’s not accepting null value parameters. Which does make sense. That’s why I have the hard coded loop instead of the helper method. I’m sure there’s a way to make it work I just haven’t fiddled with it enough yet. This is my first open source project so I feel nervous about changing anything other than LR6 test code.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

huss:
> A video call would be awesome! What’s your availability today or tomorrow?

@tinisha-davis  Today would be easier as I have a lot of appointments tomorrow. Do you use Discord? If so, why don't you message me about when - the OED website has the link to our Discord server and I'm pinned/easy to find. If not, then we can figure something else out.

huss:
I want to document that @tinisha-davis found a bug in our test code for range. The testing function OED provided does not correctly work for raw/meter data and that is what this test is doing. When this is the case, the min/max values are all null so OED can detect that no min/max should be shown. (It was done this way to be easy instead of not sending back all the null values.) This is why the test had issues since the test function does not consider this. I very much appreciate the efforts of @tinisha-davis in working so hard to change the test code for this case so it worked and reaching out about the issue as it will fix an issue with OED code.  Since @tinisha-davis found the issue, they are going to work to update our testing function to work properly and then use the standard, updated method to test the result.

src/server/test/web/readingsLineMeterRangeQuantity.js
tinisha-davis: Lines 14-29 have now been properly formatted, thank you for pointing that out.

src/server/test/web/readingsLineMeterRangeQuantity.js
tinisha-davis: Initially I was using a for loop here because there was a bug in expectRangeToEqualExpected(response, expected) that would not allow null parameters, but LR6 requires using null as a parameter. This issue has been resolved, the for loop has been deleted, and now expectRangeToEqualExpected(response, expected) works as intended with LR6.

huss (approved):
Thanks to @tinisha-davis for the updated code and for helping to find an issue with the test code. Everything now looks good. Congratulations on your first accepted contribution to OED.

PR #1542, opened 2025-10-02
Add Group Default Graphic Unit Error Handling for Unit Deletion

### LINKED ISSUE #1513: Unit protection for default graphic unit
Currently, if you delete a unit that is the Default Graphic Unit of a Group. You get a Database error about foreign key.

I would like to have custom message about how I can not delete the Default Graphic Unit. And also mention the Groups that it is the default graphic unit of. Similar to how meter is done

Image

Or you can just delete the default graphic unit anyway with a warning, and then set the default graphic unit of the group to be null

### Additional context

None

hmhngx:
# Description

This PR implements comprehensive error handling for unit deletion following mentor feedback to match the patterns and logic from PR #1505 (conversion deletion). The implementation provides immediate feedback to users and prevents accidental deletion of units that are still in use.

Fixes #1513 

## Type of change

**Enhanced Error Handling & User Experience:**
- **Immediate Error Detection**: Moved dependency checking to occur when "Delete Unit" button is clicked, before confirmation modal
- **Improved Message Formatting**: Updated to use React elements matching conversion deletion style
- **Priority-Based Error Display**: Reordered checks (conversions first, then default graphic units)
- **Proper Modal Navigation**: Fixed error notification modal to return to main units page when dismissed

**Comprehensive Dependency Support:**
- **Group Support**: Added detection for units used as group default graphic units
- **Enhanced Error Messages**: Shows group dependencies alongside meter and conversion dependencies
- **Server-Side Transaction Handling**: Added database transaction to update dependent meters/groups default graphic units to NULL during deletion

**Internationalization:**
- **Translations**: Added error message translations for English, French, and Spanish

## Checklist

## Limitations


huss (changes requested):
Thanks to @hmhngx for another contribution. Overall, it works as written. The linked issue #1513 discussed PR #1505 and its changes. I note the following:

- ~Your PR is over 3 weeks behind development. This means the new code from that PR is not reflected in your code.~
- ~The look/formatting of the messages in that PR is cleaner and nicer. While units did not use it before, I think it should be used here.~
- ~The order and logic is different for PR 1505. I like that it has default graphic units later and its ordering. The next bullet on default graphic units should probably not be done if conversions are impacted so only those messages are shown. I think the changes to conversions also limit later messages if earlier ones are present. This is needed since changing conversions is not allowed and impacts default graphic units.~
- ~It checks for impacted default graphic units, warns the user and then fixes them as needed. In the case of conversions it is done on the server since it involved a graph algorithm. In the case of units, the checks on if it is used in a conversion will stop these types of changes. Thus, the client can check for impacted default graphic units (as done) and then warn the admin as in the updated conversions that the default graphic unit will become no unit. As noted in the updated issue, this should be done for units as with conversions. It does say it could be done later but maybe now would be good (we can discuss). Note it is important that the actual delete and changes for default graphic units happens on the server as a transaction to avoid inconsistencies in the DB.~

I think these should be done and we can discuss if you have questions or thoughts.

src/client/app/translations/data.ts
```diff
@@ -33,7 +33,12 @@ const LocaleTranslationData = {
 		"AreaUnitType.meters": "sq. meters",
 		"AreaUnitType.none": "no unit",
 		"as.meter.defaultgraphicunit": "as meter default graphic unit",
+		"as.group.defaultgraphicunit": "as group default graphic unit",
 		"as.meter.unit": "as meter unit",
+		"unit.delete.meter.default.lost": "The following meters will lose their default graphic unit:",
```
huss: Note: Not all keys are likely to stay given the comment on default graphic units.

These keys are not in the correct alphabetical location.

src/client/app/translations/data.ts
```diff
@@ -33,7 +33,12 @@ const LocaleTranslationData = {
...
+		"unit.delete.group.default.lost": "The following groups will lose their default graphic unit:",
+		"unit.default.graphic.unit": "Default Graphic Unit",
```
huss: Note: This may be obsolete given the comment on default graphic units testing.

The key defaultGraphicUnit (with colon at end) and conversion.default.graphic.unit (without colon) already exist. I would use one of these and not create a new key. The later would reuse referring to conversion but that is okay (or it could be globally refactored).

src/server/routes/units.js
```diff
@@ -231,12 +231,17 @@ router.post('/delete', adminAuthMiddleware('delete units'), async (req, res) =>
 	} else {
 		const conn = getConnection();
 		try {
-			// Don't worry about checking if the unit already exists
-			// Just try to delete it to save the extra database call, since the database will return an error anyway if the row does not exist
-			await Unit.delete(req.body.id, conn);
-			success(res, 'Successfully deleted conversion');
+			await conn.tx(async t => {
+				// Update meters that use this unit as default graphic unit
```
huss: Note: Another comment may mean this code is going away so the comment is not important then. That also may resolve the GitHub security warning.

Two thoughts. First, you reuse ``req.body.id`` several times. It might be better to create a const and use that. Second, you are directly calling the SQL in the route as was done in src/server/routes/conversions.js. While it was done there, it would be better to do it in the meter/group model. Also, it duplicates the code (why having in the model is better). Thus, I think a new function in both meter and group to update based on the provided default graphic unit should be created and used in both files.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -382,6 +477,13 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 					disabled={!canSave}
 				/>
 			)}
+			<ConfirmActionModalComponent
+				show={showCancelModal}
+				actionConfirmMessage={deleteConfirmationMessage}
+				handleClose={handleCancelModalClose}
+				actionFunction={handleCancelModalClose}
+				actionConfirmText={translate('ok')}
+				actionRejectText={translate('cancel')} />
```
huss: src/client/app/components/conversion/EditConversionModalComponent.tsx has ``forceCancel={true}``. This means that it only shows cancel and not ok. I think that is better when you must cancel.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -156,62 +160,153 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
 		for (let i = 0; i < conversionData.length; i++) {
 			if (conversionData[i].sourceId === state.id) {
 				// This unit is the source of a conversion so cannot be deleted.
-				error_message += ` ${translate('conversion')} ${unitDataByID[conversionData[i].sourceId].name}` +
-					`${conversionArrow(conversionData[i].bidirectional)}` +
-					`${unitDataByID[conversionData[i].destinationId].name} ${translate('uses')} ${translate('unit')}` +
-					` "${state.name}" ${translate('unit.source.error')};`;
+				msgElements.push(
+					<div key={`conversion-source-${i}`}>
+						<span className="bold">{translate('unit.failed.to.delete.unit')}: </span>
```
huss: This message and similar ones:
- In conversion the bold message that the delete cannot be done is shown once and then all the impacted items are listed on separate lines below. I think I prefer that look.
- There is no space between unit and the quoted unit name.

huss:
> * The order and logic is different for PR 1505. I like that it has default graphic units later and its ordering. The next bullet on default graphic units should probably not be done if conversions are impacted so only those messages are shown. I think the changes to conversions also limit later messages if earlier ones are present. This is needed since changing conversions is not allowed and impacts default graphic units.
> 
> * It checks for impacted default graphic units, warns the user and then fixes them as needed. In the case of conversions it is done on the server since it involved a graph algorithm. In the case of units, the checks on if it is used in a conversion will stop these types of changes. Thus, the client can check for impacted default graphic units (as done) and then warn the admin as in the updated conversions that the default graphic unit will become no unit. As noted in the updated issue, this should be done for units as with conversions. It does say it could be done later but maybe now would be good (we can discuss). Note it is important that the actual delete and changes for default graphic units happens on the server as a transaction to avoid inconsistencies in the DB.

Now that the code is working and I tried to test it, I'm thinking the thoughts above were not needed and not thought out enough. When I tried to test the catching of losing a default graphic unit it did not show up. I thought about it and I now think that it cannot happen for units. To set the default graphic unit for a meter/group, the unit must be compatible with a compatible meter unit which implies there is a chained conversion from the meter unit to the default graphic unit. At a minimum, this means the default graphic unit is the destination of a conversion (it might also be a source). The first tests check for this and stop the deletion if the unit is either a source or destination. It skips the default graphic unit test in this case so it is never seen. If the admin removes the conversion involving the default graphic unit then the warning for that delete will show that the default graphic unit must be changed to no unit and it will happen if the admin proceeds with the delete.

Overall, this means, if I am correct, that the second set of tests for the default graphic unit will never be true so they are not needed. It also means the route can have the setting of the default graphic unit to no unit removed. A comment in the route file notes this too (coming soon).

As a side note, conversions differ since the chaining means the conversion removed is not sufficient to know if a default graphic unit is impacted. It requires running a graph path algorithm. With a unit this is not the case.

This was done and is complete.

huss (changes requested):
Thanks to @hmhngx for addressing the previous comments. I'm sorry this took so long while my time was turned elsewhere. I've made one big picture comment on the default graphic unit check that I asked for last time. If you agree, it will change the code needed. Several comments note the impact of this and there are a couple of other comments to consider. Please let me know if anything is not clear or you have thoughts.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -382,6 +447,14 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 					disabled={!canSave}
 				/>
 			)}
+			<ConfirmActionModalComponent
+				show={showCancelModal}
+				actionConfirmMessage={deleteConfirmationMessage}
+				handleClose={handleCancelModalClose}
+				actionFunction={handleCancelModalClose}
+				actionConfirmText={translate('ok')}
```
huss: This mirrors the code in src/client/app/components/conversion/EditConversionModalComponent.tsx. I just noticed that you should not need a confirm msg if forceCancel is true. I tried commenting it out and did not have an issue. If you agree, could it be removed here and in conversion for the similar case.

.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
```
huss: I think this change should be removed as it is a local user item.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -156,62 +158,125 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
 				// This unit is the source of a conversion so cannot be deleted.
-				error_message += ` ${translate('conversion')} ${unitDataByID[conversionData[i].sourceId].name}` +
-					`${conversionArrow(conversionData[i].bidirectional)}` +
-					`${unitDataByID[conversionData[i].destinationId].name} ${translate('uses')} ${translate('unit')}` +
-					` "${state.name}" ${translate('unit.source.error')};`;
+				conversionSources.push(
+					<div key={`conversion-source-${i}`}>
+						{translate('conversion')} {unitDataByID[conversionData[i].sourceId].name}
+						{conversionArrow(conversionData[i].bidirectional)}
+						{unitDataByID[conversionData[i].destinationId].name} {translate('uses')} {translate('unit')}"{state.name}" {translate('unit.source.error')}
```
huss: There should be a space between Unit and the unit name in quotes.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -156,62 +158,125 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
 				// This unit is the destination of a conversion so cannot be deleted.
-				error_message += ` ${translate('conversion')} ${unitDataByID[conversionData[i].sourceId].name}` +
-					`${conversionArrow(conversionData[i].bidirectional)}` +
-					`${unitDataByID[conversionData[i].destinationId].name} ${translate('uses')} ${translate('unit')}` +
-					` "${state.name}" ${translate('unit.destination.error')};`;
+				conversionDestinations.push(
+					<div key={`conversion-dest-${i}`}>
+						{translate('conversion')} {unitDataByID[conversionData[i].sourceId].name}
+						{conversionArrow(conversionData[i].bidirectional)}
+						{unitDataByID[conversionData[i].destinationId].name} {translate('uses')} {translate('unit')}"{state.name}" {translate('unit.destination.error')}
```
huss: Same space issue as above.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -156,62 +158,125 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+			}
+
+			if (conversionDestinations.length > 0) {
+				msgElements.push(...conversionDestinations);
+			}
+
+			if (metersUsingUnit.length > 0) {
+				msgElements.push(
+					<div key="meters-using-unit">
+						{translate('meter')} {translate('uses')} {translate('unit')}"{state.name}" {translate('as.meter.unit')}:
```
huss: same space issue as above.

Would this be clearer if it said "The following meters use....." since the list of meters follow?

The conversion messages have analogous lines as bold so I think this should be also. I'm less certain about the other messages so those are up to you.

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -156,62 +158,125 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+				cancel = true;
+			}
+		}
+
+		// Check meters using this unit
+		const metersUsingUnit = Object.values(meterDataByID).filter(meter => meter.unitId === state.id);
+
+		if (cancel || metersUsingUnit.length > 0) {
+			if (conversionSources.length > 0 || conversionDestinations.length > 0 || metersUsingUnit.length > 0) {
+				msgElements.push(
```
huss: I was thinking that creating conversionSources, conversionDestinations & metersUsingUnit to then push into msgElements seems like extra work and makes the logic more complex. Would it be possible to check if any happened (not empty) and then add this message as a new index 0 in msgElements array (maybe with splice) after all msgs are pushed?

src/client/app/components/unit/EditUnitModalComponent.tsx
```diff
@@ -156,62 +158,125 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
...
+					</div>
+				);
+				cancel = true;
+			}
+		}
+
+		// Check meters using this unit
+		const metersUsingUnit = Object.values(meterDataByID).filter(meter => meter.unitId === state.id);
+
+		if (cancel || metersUsingUnit.length > 0) {
```
huss: The comment on adding the first msg below may make this unnecessary. If it stays, could you add a comment on what it is doing.

huss (changes requested):
Thanks to @hmhngx for the update for previous comments. Overall, this is looking good. I've made some comments to consider. Also, may be title and description of the PR should be updated given the changes in what the code does.

src/client/app/components/unit/EditUnitModalComponent.tsx
huss: I removed it since there was no comment about it.

huss (approved):
Thanks to @hmhngx for the updated code. Review and testing found it is all good. Congratulations on another accepted contribution to OED.

PR #1543, opened 2025-10-06
1187 horizontal shift when error bars

### LINKED ISSUE #1187: horizontal shift when show error bars

MatthewTran22:
# Description

Looked into how Plotly works and confirmed that the offset is being caused by the Plotly autorange feature. The change now uses the information in the response data to set the x range. I was able to address the issues that were noted in PR #1409. I changed the logic to check for both enoughData and data[0].x. I added logic to adjust the x-axis min and max values to autofit the line chart. Note that this led to the SliderRangeInterval component being unused in this file

Matthew Tran - @MatthewTran22 

Fixes #1187 

## Type of change

## Checklist

### DIFF
--- a/src/client/app/components/LineChartComponent.tsx
+++ b/src/client/app/components/LineChartComponent.tsx
@@ -128,9 +128,12 @@ export default function LineChartComponent() {
 			}
 		}
 		// Tries to get the range from the slider range interval, undefined if not bounded
-		const sliderRange: [string, string] | undefined = sliderRangeInterval?.getIsBounded() ? 
-		[sliderRangeInterval.getStartTimestamp()!.utc().toISOString(), 
-		sliderRangeInterval.getEndTimestamp()!.utc().toISOString()] : undefined;
+		const sliderRange: [string, string] | undefined = sliderRangeInterval?.getIsBounded()
+			? [
+				sliderRangeInterval.getStartTimestamp()!.utc().toISOString(),
+				sliderRangeInterval.getEndTimestamp()!.utc().toISOString()
+			]
+			: undefined;
 		// Either sets the xRange to the minDate maxDate or the saved slider range. This keeps the range from resetting when we toggle error bars.
 		const xRange: [string, string] = sliderRange ?? [minDate, maxDate];
 		return (
--- a/src/client/app/components/LineChartComponent.tsx
+++ b/src/client/app/components/LineChartComponent.tsx
@@ -19,7 +19,7 @@ import { selectSelectedLanguage } from '../redux/slices/appStateSlice';
 import Locales from '../types/locales';
 import { useTranslate } from '../redux/componentHooks';
 import SpinnerComponent from './SpinnerComponent';
-import { setInitialXAxisRange, selectSliderRangeInterval } from '../redux/slices/graphSlice';
+import { setInitialXAxisRange } from '../redux/slices/graphSlice';
 import { fullSizeContainer } from '../styles/modalStyle';
 
 /**
@@ -34,7 +34,6 @@ export default function LineChartComponent() {
 	const { meterDeps, groupDeps } = useAppSelector(selectLineChartDeps);
 	const locale = useAppSelector(selectSelectedLanguage);
 	// initial slider range
-	// const sliderRangeInterval = useAppSelector(selectSliderRangeInterval);
 
 	// Fetch data, and derive plotly points
 	const { data: meterPlotlyData, isFetching: meterIsFetching } = readingsApi.useLineQuery(meterArgs,
@@ -147,7 +146,7 @@ export default function LineChartComponent() {
 						showgrid: true,
 						gridcolor: '#ddd'
 					},
-					uirevision: 'static'
+					
 				}}
 				config={{
 					responsive: true,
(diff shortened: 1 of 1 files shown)


src/client/app/components/LineChartComponent.tsx
```diff
@@ -108,9 +107,30 @@ export default function LineChartComponent() {
 	// See https://community.plotly.com/t/replacing-an-empty-graph-with-a-message/31497 for showing text not plot.
 	if (data.length === 0) {
 		return <h1>{`${translate('select.meter.group')}`}	</h1>;
-	} else if (!enoughData) {
+	} else if (!enoughData || !data[0].x) {
```
huss: Above checks ``data.x!.length > 1`` for enough data. Could you help me understand how this differs and the case(s) it is trying to protect against? If needed, could it be rolled into enoughData?

src/client/app/components/LineChartComponent.tsx
```diff
@@ -108,9 +107,30 @@ export default function LineChartComponent() {
...
 		return <h1>{`${translate('no.data.in.range')}`}</h1>;
 	} else {
+		// Adjust the min and max values for the x axis
+		// This goes through the list of groups/meters that have data to find the earliest and latest dates.
+		console.log(data);
```
huss: The console.log should be removed.

src/client/app/components/LineChartComponent.tsx
```diff
@@ -108,9 +107,30 @@ export default function LineChartComponent() {
...
+		let minDate = '';
+		let maxDate = '';
+		for (const trace of data) {
+			if (trace.x && trace.x.length > 0) {
+				const traceMin = trace.x[0] as string;
+				const traceMax = trace.x[trace.x.length - 1] as string;
+				// Update minX if this is the first trace or has an earlier date
+				if (minDate === '' || traceMin < minDate) {
```
huss: Can you help me understand why the string comparison is safe so < and > give the correct answer? Often OED uses moment for comparisons.

src/client/app/components/LineChartComponent.tsx
```diff
@@ -123,8 +143,7 @@ export default function LineChartComponent() {
 					// 'fixedrange' on the yAxis means that dragging is only allowed on the xAxis which we utilize for selecting dateRanges
 					xaxis: {
 						rangeslider: { visible: true },
-						range: [sliderRangeInterval.getStartTimestamp()?.toISOString(),
-							sliderRangeInterval.getEndTimestamp()?.toISOString()],
+						range: [minDate, maxDate],
```
huss: I see how this helped with the issue. A few thoughts:
- The slider shows gray space on the left/right ends. This isn't a big deal but does need to be documented if this is the final solution. I'm noting this for the record. Also, do you know why this happens?
- Whenever show error bars is changed (checked or unchecked) then it scales back to full size. Thus, the user slider range is undone. I'm not sure that is a desirable behavior. I think it is due to removing the sliderRangeInterval from the Plotly xaxis info. I have not tried to address this but wondered if using that in the minData/maxDate calculation would help.

huss (changes requested):
Thanks to @MatthewTran22 for another contribution to OED. I'm sorry this review took so long as my time was spent elsewhere and I lost track that this was ready for a few days. Overall it is fine. I've made a few comments to consider where the loss of the user selected slider range seems the hardest. Please let me know if anything is not clear, you have thoughts or need anything.

src/client/app/components/LineChartComponent.tsx
MatthewTran22: The data[0].x check was there when I was testing how to fix the issue from the previous PR, the data[0].x check isn't needed I can take that out

src/client/app/components/LineChartComponent.tsx
MatthewTran22: Looking at the response data, I saw the x ranges were in ISO String Format and correctly ordered. Making comparisons between two ISO strings are generally safe as long as they are in the same timezone. I can still switch it to use moment for comparison to keep everything consistent.

MatthewTran22:
@huss addressed the comments above.

- Switched line 110 to only check if there is enough data, the second condition was unnecessary
- Removed the console log from the Line chart component
- Switched the date comparisons to using moment
- Fixed the range reset bug when toggling error bars
- Specified the min and max range for the slider

The slider you mentioned had the same issue as the main line chart. For that logic, we just needed to specify the min and max date range, just like the main line chart

As for the toggling bug you detailed, I found that we still did need to use the sliderRangeInterval that was originally there to save that custom range set by the users. The main difference now is that if we do not have a custom range set by the user, we use the minDate and maxDate as the default range.

Let me know if there are any other improvements I can make in my implementation

huss (approved):
Thanks to @MatthewTran22 for addressing the comments. I want to apologize that it took so long for this review. I made a mistake and my tracking system missed a couple of PRs that were ready. The changes look good and the testing showed it works as desired. Congratulations on another accepted contribution to OED.

PR #1545, opened 2025-10-16
Disable meter unit #1521

### LINKED ISSUE #1521: edit unit to create conversion with meter that is invalid
### Describe the bug

If one creates a conversion that involves two units that are of type unit and then edits the unit to change its type to meter then illegal conversions can be created. OED should not allow this.

### To Reproduce

1. Create two units of type quantity (flow/raw also okay). In the figures this was 1ux and 2u. This shows the 1ux unit but 2u is similar:

Image

2. Create a conversion between these two units and make it bidirectional. This shows its setup after creation is done:
Image

3. Edit the source unit (1ux here) to make it of type meter as shown:

Image

4. Go back and look at the conversion. It will show that it is bidirectional but make it red since this is not allowed:

Image

Another issue is created by this:

- In a step 2 make 1ux the destination unit of the conversion. In the figure below, Euro was the source and is a quantity unit. If you look at this conversion in step 4 then you see you have a meter as the destination which is not allowed:

Image

### Expected behavior

Admins should not be allowed to have such conversions. OED should not allow an admin to change the type of unit to meter if that unit is used in any conversion (as source or destination). A nice way to do this would be to disable that choice from the menu options in this case. 

### Additional context

Note OED does not allow an admin to change the type of unit if it is in a conversion. For example, if the conversion is of type quantity you cannot change a unit to flow since that would also violate a rule. It may show ways to deal with this case.

HienXuanPham:
# Description

Disable meter unit from the menu options if the unit is used in any conversion as a source or destination.

Fixes #1521

disable meter

## Type of change

## Checklist

## Limitations

None

### DIFF
--- a/src/client/app/components/unit/EditUnitModalComponent.tsx
+++ b/src/client/app/components/unit/EditUnitModalComponent.tsx
@@ -447,7 +447,7 @@ export default function EditUnitModalComponent(props: EditUnitModalComponentProp
 									>
 										{Object.keys(UnitType).map(key => {
 											const isMeter = key === UnitType.meter;
-											const disableMeter = isMeter && inConversions()
+											const disableMeter = isMeter && inConversions();
 											return (
 												<option
 													value={key}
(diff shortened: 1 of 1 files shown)


huss (approved):
Thanks to @HienXuanPham for another contribution to OED. I'm sorry it took so long to review this PR while my time was turned elsewhere. Review and testing found it works as desired. It is ready to merge.

PR #1550, opened 2025-10-30
3D: Auto-adjust date ranges and remove 1-year limit; add data range API

### LINKED ISSUE #1185: show 3D graphic even if more than 1 year of data in range

For efficiency reasons, a 3D graphic is limited to one year of max time range. Should OED display some time range when the range is greater than one year? Note this applies to unbounded time ranges even if the actual data is less than one year.

This is open to discussion but here are some thoughts:

- Start from a given final date and graph a certain number of days backward. The length of time is open (6 months, 1 year, other)?
- Looking at the actual readings for the chosen meter/group might be good. An alternative might be the max date on the Plotly line graphic (rounded to the full day up/down) where nothing is shown if this meter/group only has data well before the max date on the graphic. An advantage of this method is it will not change the date as you select a different meter/group. I like this later one (right now) if it is easy to do. Either way should deal with unbounded date ranges that currently work.
- Should the date range picker be set to this range? That makes changing it easier but also impact others graphics. It also makes it easier to see the exact range. Need to figure this out where another mechanism for automatically set might be used that does not impact other graphics.

Have the current system where the user must select the time range.

### Additional context

None

hmhngx:
# Description

This PR enhances the 3D visualization experience and supporting APIs:
- Adds backend data range API endpoints to support 3D auto-adjustment.
- Implements client-side utilities to auto-adjust 3D graphics based on available data.
- Enables 3D date ranges greater than one year by removing the previous one-year limit.
- Minor cleanup and comment updates in related UI components.

Fixes #1185 

## Type of change

## Checklist

## Limitations
- API and UI changes should be validated across large datasets for performance.
- Additional integration tests may be needed for edge cases around very sparse or extremely long time ranges.
- Documentation for the new data range endpoints and 3D auto-adjustment behavior should be added in a follow-up if not covered by existing docs.


huss:
~The type check failed in the build and should be addressed.~

src/client/app/redux/api/readingsApi.ts
```diff
@@ -150,6 +150,10 @@ export const readingsApi = baseApi.injectEndpoints({
 				return error ? { error } : { data: data as LineReadings };
 			},
 			providesTags: ['Readings']
+		}),
+		dataRange: builder.query<{ minDate: string | null, maxDate: string | null }, { id: number, meterOrGroup: 'meters' | 'groups' }>({
```
huss: I'm not sure but can the type of meterOrGroup be set by the enum MeterOrGroup in src/client/app/types/redux/graph.ts?

src/server/routes/unitReadings.js
```diff
@@ -494,25 +486,80 @@ function createRouter() {
...
+
+	// Route for getting data range for meter/group (for 3D auto-adjustment)
+	router.get('/dataRange/meters/:meter_ids', optionalAuthMiddleware, async (req, res) => {
+		if (!validateMeterLineReadingsParams(req.params)) {
+			res.sendStatus(400);
+		} else {
+			const meterIDs = req.params.meter_ids.split(',').map(idStr => Number(idStr));
+			const conn = getConnection();
+			try {
+				const result = await conn.oneOrNone(`
```
huss: The SQL query should probably be in the Meter model and the SQL in its own file as with most queries. Same for groups.

src/server/routes/unitReadings.js
```diff
@@ -494,25 +486,80 @@ function createRouter() {
...
-				const duration = moment.duration(timeInterval.endTimestamp.diff(timeInterval.startTimestamp));
-				// Gets 0 unless one day beyond a year but that okay since don't do partial days.
-				const durationInYears = duration.years();
-				if (durationInYears >= 1) {
-					// Limit 3D to one year of data.
-					res.sendStatus(400);
+				const groupID = req.params.group_id;
+				const graphicUnitID = req.query.graphicUnitId;
+				const readingInterval = req.query.readingInterval;
+				const forJson = await groupThreeDReadings(groupID, graphicUnitID, timeInterval, readingInterval);
```
huss: This is an open route (login not needed). Thus, a bad actor could request arbitrary readingInterval and return a lot of data. Before it limited it to 1 year but that is no longer done. The range needs to be limited to the desired value. Applies to meter too.

src/server/routes/unitReadings.js
```diff
@@ -494,25 +486,80 @@ function createRouter() {
...
+		} else {
+			const meterIDs = req.params.meter_ids.split(',').map(idStr => Number(idStr));
+			const conn = getConnection();
+			try {
+				const result = await conn.oneOrNone(`
+					SELECT 
+						MIN(start_timestamp) as min_date,
+						MAX(end_timestamp) as max_date
+					FROM readings 
+					WHERE meter_id = ANY($1)
```
huss: I was thinking that the limit would be for each meter and now across all meters (or groups) since you can only graph one at a time. It makes me wonder if this route and related code should only be for a single data source. I'm not sure how that would impact the code.

src/server/routes/unitReadings.js
```diff
@@ -494,25 +486,80 @@ function createRouter() {
...
+				`, [meterIDs]);
+				
+				if (result && result.min_date && result.max_date) {
```
huss: Does this only happen if there are no readings for the data source? Just wondering.

src/server/routes/unitReadings.js
```diff
@@ -494,25 +486,80 @@ function createRouter() {
...
+					res.json({
+						minDate: result.min_date,
+						maxDate: result.max_date
+					});
+				} else {
+					res.json({ minDate: null, maxDate: null });
+				}
+			} catch (error) {
+				console.error('Error getting meter data range:', error);
```
huss: All errors must be logged and not to the console. Better yet, use the success/failure methods found in src/server/services/csvPipeline/success.js.

src/client/app/components/ThreeDComponent.tsx
```diff
@@ -46,6 +49,68 @@ export default function ThreeDComponent() {
...
+		}
+
+		// Show warning to user if result.shouldWarn is true (only once per change)
+		if (result.shouldWarn && !prevWarningRef.current) {
+			showWarnNotification(translate('threeD.date.range.exceeds.max'));
+			prevWarningRef.current = true;
+		} else if (!result.shouldWarn) {
+			prevWarningRef.current = false;
+		}
+	}, [meterOrGroupID, dataRange, queryTimeInterval, numDays, dispatch, threeDMeterOrGroup, translate]);
```
huss: I'm unclear on how dispatch or translate change.

src/client/app/components/ThreeDComponent.tsx
```diff
@@ -46,6 +49,68 @@ export default function ThreeDComponent() {
...
+	// Get numDays from Redux state
+	const numDaysRedux = useAppSelector(selectThreeDNumDays);
+	const numDays = getEffectiveNumDays(numDaysRedux);
+
+	// Track previous warning state and calculated interval to avoid duplicate notifications/dispatches
+	const prevWarningRef = React.useRef<boolean>(false);
+	const prevCalculatedIntervalRef = React.useRef<string | null>(null);
+
+	// Calculate 3D date range based on the 4-case logic
+	React.useEffect(() => {
```
huss, line 82: In the big picture, I'm wondering about the use of a Redux selector vs having the logic here.

At an even higher level, I'm thinking if all the logic should be on the server. My thinking is the user selects the date range and the span and then the result is based on data on the server. This is why I'm thinking it may be easier and better to do the calculation on the server. This might address other comments about bad routes, the useEffect, etc. I'm open to thoughts.

src/client/app/components/ThreeDComponent.tsx
```diff
@@ -8,22 +8,24 @@ import { selectGroupDataById } from '../redux/api/groupsApi';
 import { selectMeterDataById } from '../redux/api/metersApi';
 import { readingsApi } from '../redux/api/readingsApi';
 import { selectUnitDataById } from '../redux/api/unitsApi';
-import { useAppSelector } from '../redux/reduxHooks';
+import { useAppSelector, useAppDispatch } from '../redux/reduxHooks';
 import { selectThreeDQueryArgs } from '../redux/selectors/chartQuerySelectors';
 import { selectThreeDComponentInfo } from '../redux/selectors/threeDSelectors';
 import { selectScalingFromEntity } from '../redux/selectors/entitySelectors';
 import { selectGraphState } from '../redux/slices/graphSlice';
+import { updateTimeInterval, updateThreeDInterval, selectThreeDNumDays, selectQueryTimeInterval, selectThreeDMeterOrGroup } from '../redux/slices/graphSlice';
```
huss: There are a couple of lines longer than 150 characters. ESLint can show you where they are.

src/client/app/components/ThreeDComponent.tsx
```diff
@@ -46,6 +49,68 @@ export default function ThreeDComponent() {
...
+			// Store the calculated 3D interval in Redux
+			if (result.shouldShowGraph && result.threeDInterval.getIsBounded()) {
+				dispatch(updateThreeDInterval(result.threeDInterval));
+			} else {
+				dispatch(updateThreeDInterval(undefined));
+			}
+			prevCalculatedIntervalRef.current = intervalString;
+		}
+
+		// Show warning to user if result.shouldWarn is true (only once per change)
```
huss, line 117: I have not looked into the details but having to check for pervious messages makes me think there may be a bigger picture issue in the logic. This may go away so I'm not thinking about it too much.

src/client/app/utils/dateRangeCompatibility.ts
```diff
@@ -100,10 +100,133 @@ export function roundTimeIntervalForFetch(timeInterval: TimeInterval): TimeInter
 }
 
 /**
```
huss, line 102: I'm not going to review these changes until the issue about if it is client or server code is resolved.

src/client/app/components/ThreeDIntervalControlsComponent.tsx
```diff
@@ -0,0 +1,187 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss, line 1: This file seems to have a high overlap with src/client/app/components/IntervalControlsComponent.tsx. Given that file already has logic for different chart types, I'm hoping it would be better to use the other component where it is modified to deal with the new 3D cases.

src/client/app/components/ThreeDIntervalControlsComponent.tsx
```diff
@@ -0,0 +1,187 @@
...
+			dispatch(graphSlice.actions.updateThreeDNumDays(value));
+		}
+	};
+
+	const numDaysTranslations: Record<keyof typeof ThreeDNumDays, string> = {
+		OneMonth: '1.month',
+		SixMonths: '6.months',
+		OneYear: '1.year',
+		TwoYears: '2.years',
+		Max: 'max',
```
huss: I think from a UI perspective and to be consistent with other menus, this should show '3 years'. I realize it means that changes might mean editing but I'm also hoping that doing the translation as in src/client/app/components/IntervalControlsComponent.tsx will help mitigate that. Also, bar does the translation in a somewhat different way.

src/client/app/components/ThreeDIntervalControlsComponent.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { FormFeedback, FormGroup, Input, Label } from 'reactstrap';
+import { useAppDispatch, useAppSelector } from '../redux/reduxHooks';
+import { graphSlice, selectThreeDNumDays } from '../redux/slices/graphSlice';
+import { useTranslate } from '../redux/componentHooks';
+import TooltipMarkerComponent from './TooltipMarkerComponent';
+import { labelStyle, divTopBottomPadding } from '../styles/modalStyle';
+import { MAX_3D_DAYS, DEFAULT_3D_DAYS, getEffectiveNumDays } from '../utils/dateRangeCompatibility';
```
huss: DEFAULT_3D_DAYS ia not used. This and a couple of other TS issues are seen in the checks that are failing on the PR.

src/client/app/redux/selectors/chartQuerySelectors.ts
```diff
@@ -194,17 +194,21 @@ export const selectMapChartQueryArgs = createSelector(
...
 		const args: ThreeDReadingApiArgs = {
 			id: threeD.meterOrGroupID!,
-			timeInterval: roundTimeIntervalForFetch(queryTimeInterval).toString(),
+			timeInterval: roundTimeIntervalForFetch(intervalToUse).toString(),
 			graphicUnitId: selectedUnit,
 			readingInterval: threeD.readingInterval,
 			meterOrGroup: threeD.meterOrGroup!
 		};
-		const shouldSkipQuery = !threeD.meterOrGroupID || !queryTimeInterval.getIsBounded();
+		// Skip query if no meter/group selected or if interval is not bounded
```
huss: I think the issue specified that the 3D graph happens if unbounded as from the last reading time backward. This seems to stop 3D graphics in this case.

src/client/app/redux/slices/graphSlice.ts
```diff
@@ -37,7 +38,8 @@ const defaultState: GraphState = {
 	threeD: {
 		meterOrGroupID: undefined,
 		meterOrGroup: undefined,
-		readingInterval: ReadingInterval.Hourly
+		readingInterval: ReadingInterval.Hourly,
+		numDays: undefined // Default to undefined, will use 1 year (365 days) if not set
```
huss: OED really prefers comments on the line above the code.

src/client/app/translations/data.ts
```diff
@@ -500,12 +500,16 @@ const LocaleTranslationData = {
...
 		"threeD.date": "Date",
+		"threeD.date.range.exceeds.max": "3D date range exceeds maximum allowed days. Please reduce the date range or numDays setting.",
 		"threeD.date.range.too.long": 'Date Range Must be a year or less',
 		"threeD.incompatible": "Not Compatible with 3D",
+		"threeD.numDays": "3D Date Range Span",
 		"threeD.rendering": "Rendering",
 		"threeD.time": "Time",
 		"threeD.x.axis.label": "Hours of Day",
 		"threeD.y.axis.label": "Days of Calendar Year",
+		"6.months": "6 months",
```
huss: This is in the wrong place alphabetically. See 2.months for location.

src/client/app/types/redux/graph.ts
```diff
@@ -54,6 +54,7 @@ export interface ThreeDState {
 	meterOrGroupID: MeterOrGroupID | undefined;
 	meterOrGroup: MeterOrGroup | undefined;
 	readingInterval: ReadingInterval;
+	numDays: number | undefined; // Number of days for 3D graphic span (undefined means use default: 1 year)
```
huss, line 57: I'm curious about using undefined vs setting the default value to 1 year in src/client/app/redux/slices/graphSlice.ts.

huss (changes requested):
Thanks to @hmhngx for another contribution. I have made some comments to consider. I also have some thoughts that were not easy to put in a specific place so I'm putting them here.

- The custom range (and maybe the non-custom ones but not checked) seems to include one too many days. For example, choosing 3 days seems to show 4 on the graphic.
  - Update: I'm having problems testing this as the date range is often not honored. This applies to several other ones in the comment.
- Setting the date range outside the range of readings caused a server error and it never graphed (``ERROR:  range lower bound must be less than or equal to range upper bound``) and it then went to the client.
  - Update: This now says rendering and also pops up a warning about the range.
- Some actions are causing a client message of exceeding the max days when it seems valid but this might naturally go away with other items.
- If set the upper bound on the date range that is above the actual readings available then get client message about exceeding days but still graphs the result.
- ~Maybe change ``		"threeD.y.axis.label": "Days of Calendar Year",`` on the 3D graphic label for the axis to "Date" since no longer just one year.~
- I'm thinking about whether the behavior should be changed from the issue if have only DRS is set. In this case, it would then do DRS to DRS + numDays. If there is not enough data in readings then the graphic would end early. This allows a user to easily get data from a given date forward. This also seems parallel to only have the DRE. What do you think?

As always, I'm happy to clarify and discuss as desired.

src/client/app/components/DateRangeComponent.tsx
```diff
@@ -35,6 +37,14 @@ export default function DateRangeComponent() {
 		dispatch(changeSliderRange(dateRangeToTimeInterval(value)));
 	};
 
+	// Provide a default date range if the interval is unbounded to prevent error display
+	// Default to last 30 days when unbounded
+	const displayInterval = queryTimeInterval.getIsBounded()
+		? queryTimeInterval
+		: new TimeInterval(
+			moment().subtract(30, 'days').startOf('day'),
```
huss: This seems to be causing several undesirable behaviors. 

1. The date range is shown and set to the thirty days back from the current day. However, the actual 3D data goes back 1 year from the last day with data. Thus, they do not correspond to each other.
2. In src/client/app/components/ThreeDIntervalControlsComponent.tsx there is DEFAULT_3D_DAYS. How does this relate to the 30 days here? This relates to item 1.
3. It is directly setting the data range shown and not setting queryTimeInterval. As a result, the date range is set but not shown on several other graphic pages. It also is not honored.
4. I'm seeing warning, it not honoring the time interval & resetting the time interval in 3D.

My current thinking is the date range should not be modified and the 3D data is gotten from the server by going back by DEFAULT_3D_DAYS (currently 1 year). Thus, date range remains unbounded but 3D gets the correct data. I think it may be possible that this was causing an error given your comment. If so, maybe we should discuss options.

src/client/app/redux/selectors/chartQuerySelectors.ts
```diff
@@ -194,17 +194,25 @@ export const selectMapChartQueryArgs = createSelector(
 
 export const selectThreeDQueryArgs = createSelector(
 	selectQueryTimeInterval,
+	selectThreeDInterval,
 	selectSelectedUnit,
 	selectThreeDState,
-	(queryTimeInterval, selectedUnit, threeD) => {
+	(queryTimeInterval, threeDInterval, selectedUnit, threeD) => {
```
huss: queryTimeInterval seems unused.

src/client/app/translations/data.ts
```diff
@@ -12,6 +12,9 @@ const LocaleTranslationData = {
 		"1.week": "1 week",
 		"1.year": "1 year",
 		"2.months": "2 months",
+		"2.years": "2 years",
```
huss, line 15: Moving the items does not appear in other languages.

src/client/app/translations/data.ts
```diff
@@ -500,12 +503,14 @@ const LocaleTranslationData = {
...
 		"threeD.date": "Date",
+		"threeD.date.range.exceeds.max": "3D date range exceeds maximum allowed days. Please reduce the date range or numDays setting.",
 		"threeD.date.range.too.long": 'Date Range Must be a year or less',
 		"threeD.incompatible": "Not Compatible with 3D",
+		"threeD.numDays": "3D Date Range Span",
 		"threeD.rendering": "Rendering",
 		"threeD.time": "Time",
 		"threeD.x.axis.label": "Hours of Day",
-		"threeD.y.axis.label": "Days of Calendar Year",
+		"threeD.y.axis.label": "Date",
```
huss, line 513: The text change does not appear in other languages.

src/client/app/utils/dateRangeCompatibility.ts
huss: I'm still unsure that the calculation should happen on the client. I think we should discuss to resolve. Then I can carefully review the code.

src/server/models/Reading.js
```diff
@@ -475,6 +475,55 @@ class Reading {
...
 
+	/**
+	 * Gets the data range (min and max timestamps) for a single meter
+	 * Used for 3D auto-adjustment to determine available data range
+	 * @param meterID The meter ID to get data range for
+	 * @param conn the connection to use
+	 * @returns {Promise<{minDate: Moment | null, maxDate: Moment | null}>}
+	 */
+	static async getMeterDataRange(meterID, conn) {
+		const result = await conn.oneOrNone(`
```
huss: OED normally prefers to have an SQL file for queries. Unless there is a reason not to do this then it should be changed. Also, I think using $1 is less desirable and not done in our SQL files.

This also applies to group below.

src/server/models/Reading.js
```diff
@@ -475,6 +475,55 @@ class Reading {
...
+			SELECT 
+				MIN(start_timestamp) as min_date,
+				MAX(end_timestamp) as max_date
+			FROM readings 
+			WHERE meter_id = $1
+		`, [meterID]);
+		if (result && result.min_date && result.max_date) {
```
huss: My testing found (as I would expect) that if there are not readings so no min/max then you get:

result:  { min_date: null, max_date: null }

Thus, I don't think you need the extra logic for this.

src/server/routes/unitReadings.js
```diff
@@ -466,22 +477,24 @@ function createRouter() {
...
-				// Gets 0 unless one day beyond a year but that okay since don't do partial days.
-				const durationInYears = duration.years();
-				if (durationInYears >= 1) {
-					// Limit 3D to one year of data.
+				// Calculate calendar days (inclusive of both start and end day) to match client calculation
+				// Client uses: threeDEndDate.diff(threeDStartDate, 'days') + 1
+				const durationInDays = timeInterval.endTimestamp.diff(timeInterval.startTimestamp, 'days') + 1;
+				// Security: Limit 3D to MAX_3D_DAYS (1095 days = 3 years) to prevent abuse from unauthenticated users.
+				// Frontend auto-adjustment provides UX, but backend must enforce security limits.
+				if (durationInDays > 1095) {
```
huss: As was done on the client, it is nice to make limits a const though it will be a little different due to JS not TS. I think that should be done here, esp. since the value is used more than once.

src/client/app/components/ThreeDIntervalControlsComponent.tsx
```diff
@@ -0,0 +1,177 @@
...
+		Max: '3.years',
+		Custom: 'custom.value'
+	};
+
+	return (
+		<div>
+			<div style={divTopBottomPadding}>
+				<p style={labelStyle}>
+					{translate('threeD.numDays')}:
+					<TooltipMarkerComponent page='home' helpTextId='help.home.threeD.numDays.tip' />
```
huss, line 124: There should be a placeholder in src/client/app/components/TooltipHelpComponent.tsx for this popup help.

huss (changes requested):
Thanks to @hmhngx for the updated code. There are comments from the past that I think are unresolved & I made some new comments based on the changes. As indicated, it might be good to discuss the big picture ones to decide how they will be resolved. Also, there are three failing tests due to changed in 3D that need to be fixed so they pass.

PR #1554, opened 2025-11-11
Fix Insecure Default Configuration

Zachary-Squires:
# Description

Changes installOED.sh and docker-compose.yml to automatically generate POSTGRES_PASSWORD and OED_TOKEN_SECRET when OED is launched in production mode. This should remediate an issue where OED_TOKEN_SECRET could be forged to allow a login as an admin. Will also make the postgres database more secure as the password will be much more robust by default. Additionally, installOED.sh now specifically checks if OED_PRODUCTION is no rather than just checking if it is not yes. 

Developed and implemented by:
Zachary Squires - https://github.com/Zachary-Squires
Zachary Bates - https://github.com/Zach-O-Bates

## Type of change

## Checklist

## Limitations

In order to fully remediate the insecure default configuration we will need to use a separate docker file when launching in development and production, this will be done in a separate PR since it will be intrusive to every OED developer.


docker-compose.yml
```diff
@@ -13,7 +13,8 @@ services:
         environment:
             # Custom PGDATA per recommendations from official Docker page
             - PGDATA=/var/lib/postgresql/data/pgdata
-            - POSTGRES_PASSWORD=pleaseChange # default postgres password that should be changed for security.
+            - POSTGRES_PASSWORD=${POSTGRES_PASSWORD:-pleaseChange} # In development password may (and should be) be manually changed.
```
huss: OED prefers comments above the line so could all the new ones please be changed to do that.

docker-compose.yml
```diff
@@ -111,4 +112,4 @@ services:
         /bin/sh -c "
         rm -f /tmp/.X99-lock &&
         Xvfb :99 -screen 0 1024x768x16"
-        
\ No newline at end of file
+
```
huss: Could this line be removed since it has spaces at the start and there are two empty lines in a row.

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
...
+			OED_TOKEN_SECRET=$(openssl rand -hex 32)
+			export OED_TOKEN_SECRET
+
+			printf "Generated OED_TOKEN_SECRET: %s\n" "$OED_TOKEN_SECRET"
+
+			# Save to .env for future runs
+			if [ -f ".env" ]; then
+				if grep -q "^OED_TOKEN_SECRET=" .env; then
+					sed -i "s/^OED_TOKEN_SECRET=.*/OED_TOKEN_SECRET=$OED_TOKEN_SECRET/" .env
+				else
```
huss, line 235: This comment applies in a general way but put here. I'm thinking about the fact that there are settings in .env and docker-compose.yml that apply to the same case. I think originally it was designed to allow flexibility in how the values were given and to override them. I think the reality is that people don't use this ability. Also, security has become more important so I'm thinking having it in two places is less desirable since people might miss one. Finally, I think OED will be adopting new ways to set values that is more secure and easier so getting to one way now might help for that future effort. First: What do you think about getting rid of one of the ways to set these values? Second: If yes, then should we only use the docker-compose.yml file?

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
...
+		fi
+		#If the user is in production and their token secret has been left default, generating a random one
+		if [ -z "$OED_TOKEN_SECRET" ] || [ "$OED_TOKEN_SECRET" = "?" ]; then
+			printf "\nNo valid OED_TOKEN_SECRET detected. Generating a secure random secret...\n"
+	
+			# Generate 32 bytes of random data and convert to 64-character hex
+			OED_TOKEN_SECRET=$(openssl rand -hex 32)
+			export OED_TOKEN_SECRET
+
+			printf "Generated OED_TOKEN_SECRET: %s\n" "$OED_TOKEN_SECRET"
```
huss, line 227: Curious why here it does not tell them to save but for PG password it does.

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
...
+		if [ -z "$POSTGRES_PASSWORD" ] || [ "$POSTGRES_PASSWORD" = "pleaseChange" ]; then
+			printf "\n********************************************************************************\n"
+			printf "* WARNING: YOU ARE USING OED IN DEVELOPMENT MODE WITH THE DEFAULT POSTGRESQL PASSWORD SET IN docker-compose.yml IF THIS IS NOT INTENTIONAL GO THERE TO CHANGE IT. *\n"
+			printf "********************************************************************************\n\n"
+		fi
+		printf "%s\n" "Starting OED in development mode."
 		./src/scripts/devstart.sh
-	fi
+	else
+		printf "\nFailure: Invalid or missing enviroment configuration."
```
huss: enviroment misspelled

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
 if [ "$dostart" == "yes" ]; then
 	if [ "$production" == "yes" ] || [ "$OED_PRODUCTION" == "yes" ]; then
 		printf "%s\n" "Starting OED in production mode"
+		#Checking if the user has set a mail method and left one of the mailing environment variables default, warning if so
```
huss: Could you please add a space after # as done elsewhere in the file.

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
...
+			[ "$OED_MAIL_TO" = "someone@example.com" ] || \
+			[ "$OED_MAIL_ORG" = "My Organization Name" ]; then
+				printf "\n********************************************************************************\n"
+				printf "* WARNING: You have set your mail method but one or more of the mail environment variables are still set to the default value!*\n"
+				printf "********************************************************************************\n\n"
+			fi
+		fi
+		#If the user is in production and their token secret has been left default, generating a random one
+		if [ -z "$OED_TOKEN_SECRET" ] || [ "$OED_TOKEN_SECRET" = "?" ]; then
+			printf "\nNo valid OED_TOKEN_SECRET detected. Generating a secure random secret...\n"
```
huss, line 220: I think this and other cases should have a banner to highlight the change as was done for the mail warning.

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
 if [ "$dostart" == "yes" ]; then
 	if [ "$production" == "yes" ] || [ "$OED_PRODUCTION" == "yes" ]; then
```
huss: This is only one of several places in this script that checks for the type of installation. Different ones are slightly different in the check. For security reasons and to abstract the logic, I am thinking it would be best to decide near the start of the script and set a separate variable to indicates what case is needed. Then that variable can be consistently used for all logic. It also means the check below for an unknown value can happen at the start to quickly stop the install.

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
...
+				else
+					echo "OED_TOKEN_SECRET=$OED_TOKEN_SECRET" >> .env
+				fi
+			else
+				echo "OED_TOKEN_SECRET=$OED_TOKEN_SECRET" > .env
+			fi
+		fi
+		#If the user is in production and their postgres password has been left default, generating a random one
+		if [ -z "$POSTGRES_PASSWORD" ] || [ "$POSTGRES_PASSWORD" = "pleaseChange" ]; then
+			printf "\n No valid PostgreSQL password detected. Generating a secure random password...\n"
```
huss: The messages elsewhere don't have a space at the start so I would remove it here.

src/scripts/installOED.sh
```diff
@@ -188,11 +188,80 @@ printf "%s\n" "OED install finished"
...
+				if grep -q "^OED_TOKEN_SECRET=" .env; then
+					sed -i "s/^OED_TOKEN_SECRET=.*/OED_TOKEN_SECRET=$OED_TOKEN_SECRET/" .env
+				else
+					echo "OED_TOKEN_SECRET=$OED_TOKEN_SECRET" >> .env
+				fi
+			else
+				echo "OED_TOKEN_SECRET=$OED_TOKEN_SECRET" > .env
+			fi
+		fi
+		#If the user is in production and their postgres password has been left default, generating a random one
```
huss: I'm not sure this is actually changing the password for Postgres. I'm not an expert but from what I understand, the POSTGRES_PASSWORD environment variable is used when the DB is initialized. This happens at an earlier stage so the original value is used. I'm also unsure if it changes it if done on a later startup. I tried checking with login and it was not clear but seems it might be an issue. We should discuss this.

I don't know how the token is impacted or if that is okay.

huss (changes requested):
Thanks to @Zachary-Squires & @Zach-O-Bates for this contribution. I've made some comments to consider. I'm happy to discuss them. I did some testing but not complete as I'm going to wait for any changes to do that.

src/scripts/installOED.sh
Zachary-Squires: Yeah I did some testing and it seems like you're exactly right, for the first time since the database is created before this is run it uses the default password but after that it uses the new one. I'm not sure if there's anything I can do about that but I'm open to ideas.

src/scripts/installOED.sh
Zachary-Squires: I added the banners on the part where they're told to save or change the value, that seemed like the most important part.

src/scripts/installOED.sh
Zachary-Squires: Added that, forgot to.

src/scripts/installOED.sh
Zachary-Squires: Fixed

src/scripts/installOED.sh
Zachary-Squires: Fixed

src/scripts/installOED.sh
Zachary-Squires: I went in and did that, let me know if it's implemented correctly and if I missed manual checks anywhere.

src/scripts/installOED.sh
Zachary-Squires: Fixed

docker-compose.yml
Zachary-Squires: Pretty sure I got them all.

docker-compose.yml
Zachary-Squires: Should be gone I think.

src/scripts/installOED.sh
```diff
@@ -186,13 +201,84 @@ printf "%s\n" "OED install finished"
...
+					echo "POSTGRES_PASSWORD=$POSTGRES_PASSWORD" >> .env
+				fi
+			else
+				echo "POSTGRES_PASSWORD=$POSTGRES_PASSWORD" > .env
+			fi
+		fi
 		npm run start
-	else
-		printf "%s\n" "Starting OED in development mode"
+	elif [ "$INSTALL_MODE" = "development" ]; then
```
huss: Given the checks above, can this now be an else? I'm not sure it is that risky to do development if something went wrong (which it should not).

src/scripts/installOED.sh
```diff
@@ -186,13 +201,84 @@ printf "%s\n" "OED install finished"
...
+		# Warning the user if they've left their token or postgres password default, we don't randomly generate it in dev mode 
+		if [ -z "$OED_TOKEN_SECRET" ] || [ "$OED_TOKEN_SECRET" = "?" ]; then
+			printf "\n********************************************************************************\n"
```
huss: I'm thinking about the banners here. I think most developers will simply leave the default values as they are generally safe. Given this, do you think there should be banners? If not, how much do you think should be all uppercase?

src/scripts/installOED.sh
```diff
@@ -58,6 +58,21 @@ if [ -f ".env" ]; then
 	source .env
 fi
 
+# Creating a centralized variable to keep track of the type of installation. 
+INSTALL_MODE="invalid"
```
huss: I'm wondering about not initializing to a bad value and changing the second if to be an else on the first one. I think it has the same impact in a clean way in the code. What do you think?

src/scripts/installOED.sh
```diff
@@ -186,13 +199,74 @@ printf "%s\n" "OED install finished"
...
+				echo "OED_TOKEN_SECRET=$OED_TOKEN_SECRET" > .env
+			fi
+		fi
+		# If the user is in production and their postgres password has been left default, generating a random one
+		if [ -z "$POSTGRES_PASSWORD" ] || [ "$POSTGRES_PASSWORD" = "pleaseChange" ]; then
+			printf "\nNo valid PostgreSQL password detected. Generating a secure random password...\n"
+			POSTGRES_PASSWORD=$(openssl rand -base64 12)
+			node ./src/scripts/changePass.js "$POSTGRES_PASSWORD"
+			printf "\n********************************************************************************\n"
+			printf "Generated POSTGRES_PASSWORD: %s\n" "$POSTGRES_PASSWORD"
```
huss: I'm thinking that it might be best to change the "Generated POSTGRES_PASSWORD" to "Generated a secure PostgreSQL password and set PostgreSQL to use it". This will mostly be seen by production sites so the internal name is not important.

src/scripts/changePass.js
```diff
@@ -0,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { Client } = require('pg');
+
+const newPassword = process.argv[2];
```
huss: I'm wondering why this line and the next one are outside the function ``changePassword``. Is that necessary?

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,57 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const { Client } = require('pg');
+
+const newPassword = process.argv[2];
+
+console.log('Attempting to change postgres user password...');
+
+function escapePassword(password) {
```
huss: I think some comments on what each function is supposed to do and any for difficult to understand blocks of code would be nice.

src/scripts/changePass.js
```diff
@@ -0,0 +1,57 @@
...
+		await client.connect();
+		
+		const sql = `ALTER USER postgres WITH PASSWORD '${escapePassword(newPassword)}'`;
+		await client.query(sql);
+		
+		console.log('Password changed successfully');
+		await client.end();
+		
+	} catch (error) {
+		console.error('Error:', error.message);
```
huss: I think it would be nice if the error messages indicated that this was happening during a password change as this may be happening during another process (install).

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,57 @@
...
+		} else if (error.message.includes('password authentication')) {
+			console.error('Authentication failed. Current password may be incorrect.');
+		}
+		
+		throw error;
+	}
+}
+
+// Wait for database to be ready
+setTimeout(async () => {
```
huss: I don't see where this is used in this file and it is not exported.

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: I'm thinking the name of this file should include Postgres so it is clear it changes that password.

I'm wondering if there should be a node script command in package.json to call this script so it would be much easier for a site to update the Postgres password if they wanted. I'm not sure if one can easily pass the new password. If not, this isn't going to work.

src/scripts/installOED.sh
huss: Do you have thoughts on setting it this way? Do you think it should be left for now?

src/scripts/installOED.sh
```diff
@@ -186,13 +199,74 @@ printf "%s\n" "OED install finished"
...
+			printf "\n Make sure to save or change this value"
+			printf "\n********************************************************************************\n\n"
+			if grep -q "^POSTGRES_PASSWORD=" .env; then
+				# Don't want to store the password in plain text in the .env file, but need to let the code know the password has been changed
+    			sed -i.bak 's/^POSTGRES_PASSWORD=.*/POSTGRES_PASSWORD=autoSet/' .env
```
huss: I'm unclear on how the autoSet value is used within the code. Can you help me understand?

src/scripts/installOED.sh
huss: I thought this had been addressed but my testing did not find it working. I did the following:

- Without OED running, I removed the postgres-data/ directory. This forces complete initialization of the DB on the next startup of OED.
- I edited src/scripts/installOED.sh so it had:
```
if [ "$dostart" == "yes" ]; then
	if [ "$INSTALL_MODE" = "development" ]; then
```
so it used my development environment to force the password initialization to a script generated one.
- Started OED and it said it set the password.
- Shut down OED.
- Edited postgres_data/pg_hba.conf to change trust to password in all local connections to force PG to ask for a password.
- Restart OED.
- Open a shell on the OED Docker database container.
  - Do ``psql -U oed``. It asks for the password. If I give an invalid one it fails. If I use the one output by the first run of the install script then it fails. If I use the default OED PG password then it works. This is why I think it is not changing the password.

Do you have thoughts or ideas? I'm happy to talk with you about this.

huss (changes requested):
@Zachary-Squires Thank you for the updated version. A number of comments are now resolved. I added a new comment on a couple of old ones and added a few new ones. Please let me know if you need anything.

src/scripts/installOED.sh
Zachary-Squires: I'm not currently randomizing the password for the oed user, just the postgres superuser. I can also do it for oed if you want.

src/scripts/installOED.sh
Zachary-Squires: I agree, I'll change it.

src/scripts/installOED.sh
Zachary-Squires: When the password has been set automatically set the POSTGRES_PASSWORD variable in the .env file to autoSet so the script can detect that it's already been automatically changed and not re-randomize it. I didn't really want to store the plaintext password in the file or anything. The problem is that without this I can't really access the database to see if a randomization is needed since I need to have the correct password beforehand to access the database from changePass.js

src/scripts/changePostgresPass.js
Zachary-Squires: I will change the file name, as for the second part I thought we talked about it before Christmas break and decided that if someone wanted to change the password after the randomization they could just do it manually. Not a problem if that's changed, I can look into doing that.

src/scripts/changePostgresPass.js
Zachary-Squires: It's just used to make the script wait before trying to connect, before I had it I was getting ECONNREFUSED errors.

src/scripts/changePostgresPass.js
Zachary-Squires: Will add

src/scripts/changePass.js
Zachary-Squires: Will add

src/scripts/changePass.js
Zachary-Squires: I don't think so, pretty sure they were just left out there on accident. I'll move them in.

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,58 @@
...
+		await client.query(sql);
+		
+		console.log('Password changed successfully');
+		await client.end();
+		
+	} catch (error) {
+		console.error('Error during Postgres password change:', error.message);
+		
+		if (error.message.includes('ECONNREFUSED')) {
+			console.error('Database is not accepting connections yet.');
```
huss: Thanks for clarifying what the timeout was for. In src/scripts/installOED.sh, it also checks for ECONNREFUSED and then does the wait if it happens. It also makes sure it only tries a max number of times (10) and then errors out. I think that might be better than always calling the setTimeout. Thus, I propose a loop in changePassword and wait if there is a connection error with a max number of loops before it reports and error. The wait could then be reduced to 3 seconds as in installOED.sh so it runs faster in most cases. What do you think about that?

src/scripts/installOED.sh
huss: Just noting we talked about this and you are working to make that happen.

huss (changes requested):
@Zachary-Squires I'm trying to get all PRs up-to-date so I looked at this one even though you are still working on it. I resolved some comments and left one to think about along with any other open ones.

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,96 @@
...
+
+// Escape single quotes in password for SQL
+function escapePassword(password) {
+    return password.replace(/'/g, "''");
+}
+
+// Update .env with PostgreSQL and OED user passwords
+function updateEnvFile(postgresPassword, oedPassword) {
+    let env = '';
+    if (fs.existsSync(ENV_PATH)) {
```
huss: Adding some comments about what the logic is doing in this function would be helpful.

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,96 @@
...
+const { Client } = require('pg');
+const crypto = require('crypto');
+const fs = require('fs');
+const path = require('path');
+
+const ENV_PATH = path.resolve('.env');
+
+// Generate a secure random password
+function generatePassword() {
+    return crypto.randomBytes(18).toString('base64');
```
huss: The tab indenting has become space indenting. Could it please be changed back. Let me know if you need any help.

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,96 @@
...
+
+        process.exit(0);
+
+    } catch (error) {
+        console.error('Error changing PostgreSQL or OED password:', error.message);
+
+        if (error.message.includes('ECONNREFUSED')) {
+            console.error('Database is not accepting connections yet.');
+        } else if (error.message.includes('password authentication')) {
+            console.error('Authentication failed — default password may already be changed.');
```
huss: Now that it is written to .env, does this still happen regularly or is it a rare event?

I saw that if I tried to change the password twice via manually doing the npm command then it failed this way. I know you told me you were still trying to get it to work after the initial startup and I wonder if this relates to that. What I mean is maybe the process.env.POSTGRES_PASSWORD is not updated on change but I'm very uncertain about this. Whatever it is, it would be nice to track down both the initial setting and future changes. I note it seemed to work fine on the first change of the password via the npm command.

docker-compose.yml
```diff
@@ -13,7 +13,10 @@ services:
         environment:
             # Custom PGDATA per recommendations from official Docker page
             - PGDATA=/var/lib/postgresql/data/pgdata
-            - POSTGRES_PASSWORD=pleaseChange # default postgres password that should be changed for security.
+            # The postgres database will only pull from this value on the first
+            # startup of OED. If a change is desired after this, one must manually
```
huss: Does one still have to use SQL or will the npm command now do it?

src/scripts/installOED.sh
huss: I've decided it could be a separate effort and just opened issue #1569 about doing this.

src/scripts/changePostgresPass.js
huss: What is the status of this?

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,96 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
```
huss: I wondering about the location of this file. All the other files in this directory are scripts of .sh, .bash, etc. Given this, would it make sense and could this be move the src/server/util/?

package.json
```diff
@@ -45,7 +45,8 @@
 		"generateTestingData": "node -e 'require(\"./src/server/data/automatedTestingData\").generateTestingData()'",
 		"testData": "node -e 'require(\"./src/server/data/automatedTestingData.js\").insertSpecialUnitsConversionsMetersGroups()'",
 		"webData": "node -e 'require(\"./src/server/data/websiteData.js\").insertWebsiteData()'",
-		"addLogMsg": "node -e 'require(\"./src/server/services/addLogMsg.js\").addLogMsgToDB()'"
+		"addLogMsg": "node -e 'require(\"./src/server/services/addLogMsg.js\").addLogMsgToDB()'",
+		"changePostgresPassword": "node ./src/scripts/changePostgresPass.js"
```
huss: Should this directly call changePasswords() as some npm commands do above?

src/scripts/changePostgresPass.js
```diff
@@ -0,0 +1,96 @@
...
+        env = env.trimEnd() + `\nOED_DB_PASSWORD=${oedPassword}\n`;
+    }
+
+    fs.writeFileSync(ENV_PATH, env, { mode: 0o600 });
+    console.log('.env updated with new PostgreSQL and OED passwords');
+}
+
+async function changePasswords() {
+    const postgresPassword = process.argv[2] || generatePassword();
+	const oedPassword = process.argv[2] || generatePassword();
```
huss: Not a big deal but should both passwords have to be the same? Maybe use [2] and [3] so can be different?

src/scripts/installOED.sh
huss: Both passwords are now set but, as you told me, you have to restart OED for it to work. I know you are working on this.

huss (changes requested):
Thanks to @Zachary-Squires for addressing many previous comments. I've made a few more to consider and commented on a few older ones. I think this is getting close.

src/scripts/installOED.sh
```diff
@@ -186,13 +199,63 @@ printf "%s\n" "OED install finished"
...
+		if [ -z "$OED_TOKEN_SECRET" ] || [ "$OED_TOKEN_SECRET" = "?" ]; then
+			printf "\nNo valid OED_TOKEN_SECRET detected. Generating a secure random secret...\n"
+	
+			# Generate 32 bytes of random data and convert to 64-character hex
+			OED_TOKEN_SECRET=$(openssl rand -hex 32)
+			export OED_TOKEN_SECRET
+			
+			printf "\n********************************************************************************\n"
+			printf "Generated OED_TOKEN_SECRET: %s\n" "$OED_TOKEN_SECRET"
+			printf "\n Make sure to save or change this value"
```
huss: Could the leading space be removed so msgs have consistent indenting.

src/scripts/installOED.sh
huss: Could you provide info on the need to restart OED after the change and if any progress was made to avoid this.

src/scripts/changePostgresPass.js
huss: The location where this happens has moved but I don't see any updates. Could you please comment on what you are thinking.

src/scripts/changePostgresPass.js
huss: Could you provide info on what you think about this.

huss (changes requested):
@Zachary-Squires Thank you for the updated version. I'm sorry it took me a while to get back to this. I have gone through all the comments. I added a new comment on any one that is outstanding and these should be the only unresolved ones. Can you please comment on each one so I know what you are thinking. I also added one comment that relates to a previous one but only part of it so I thought this would be easier (new one not resolved but older one is resolved). I also hid the duplicate comments from before to make knowing what is active easier. It may be the case that the new ones get duplicated but I don't yet know. I am waiting on a version that is hopefully the final one to retest everything. Please let me know if anything is not clear or I can help.

src/scripts/changePostgresPass.js
Zachary-Squires: The reason you can't change it manually twice via the npm command is the same reason you have to restart OED once when it first gets changed, I can't by any method I've yet seen know what the password is once it's changed so the script can't access the database. This comes back to the same issue I've been having which is that env variables are initialized when the containers are first run and then at least as far as I know can't be reinitialized. I could add a third parameter that lets the user provide the current database password to the script in case they want to run it twice without restarting.

src/scripts/installOED.sh
Zachary-Squires: Yes I will do that.

src/scripts/changePostgresPass.js
Zachary-Squires: Yeah I'd agree that your method would probably be better, I kept having an error where postgres wouldn't be ready to take queries by the time the script ran so I just put the timeout in there as a stop gap. I'll change it.

src/scripts/installOED.sh
Zachary-Squires: I don't think there's any way to avoid having to restart OED after the change because, to my understanding, the config.js file is initialized at the start and then the config is used in several other files which actually connect to the database. The issue is that the config can't be changed to reflect the new password after it's been sent out, at least as far as I could tell after multiple attempts.

Zachary-Squires: If you have any ideas for me to try obviously I am very open to suggestions.

huss: > If you have any ideas for me to try obviously I am very open to suggestions.

I'm no expert here so take everything with a grain of salt but here are some thoughts:

- The docker-compose.yml file is using ``comand:`` to start the installOED.sh script after the containers are running. Is there any way to run a command before it starts the containers? Then maybe the new password(s) could be set and files edited before the process starts.
- If there is no known way to start OED with the new passwords available then could it automatically be shut down and restarted to reflect the new passwords? This isn't the ideal solution but it would mean that once the process is done then it is running with the new passwords. Trying to limit this to only when the passwords are changed would be desirable.

There may be other options that I'm not thinking of and I'm open to any ideas from anyone. I have not researched how others may have dealt with similar situations.

I hope this helps a little and I'm happy to discuss.

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,158 @@
...
+	if (!fs.existsSync(envPath)) {
+		return {};
+	}
+
+	return dotenv.parse(fs.readFileSync(envPath, 'utf8'));
+}
+
+// Generate a secure random password
+function generatePassword() {
+	return crypto.randomBytes(18).toString('base64');
```
huss: This number (18) is decent but I think 32 (256 bits) is considered very secure. Is there any reason not to use a larger value? I'm not certain it makes a longer password but I doubt anyone is going to memorize the password so it must be stored anyway.

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,158 @@
...
+function updateEnvFile(postgresPassword, oedPassword) {
+	let env = '';
+
+	// Reading current .env contents if it exists
+	if (fs.existsSync(ENV_PATH)) {
+		env = fs.readFileSync(ENV_PATH, 'utf8');
+	}
+
+	// Updating the postgres password if it exists, else appending it instead
+	if (/^POSTGRES_PASSWORD=/m.test(env)) {
```
huss, line 51: I think I'm being slow here. test is a RegEx item (I think) from TS. I'm unclear on where "m" is coming from and why not RegEx.test. Could you help me out?

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,158 @@
...
+	fs.writeFileSync(ENV_PATH, env, { mode: 0o600 });
+	console.log('.env updated with new PostgreSQL and OED passwords');
+}
+
+// Pause execution
+function sleep(ms) {
+	return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+// Detect retryable Postgres errors
```
huss: Is it possible to enumerate the errors that are retried in the comment (esp. since some are hex code)? I realize they are well documented online but it would make it easier. Finally, why these specific ones?

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	// Updating the oed user password if it exists, else appending it instead
+	if (/^OED_DB_PASSWORD=/m.test(env)) {
+		env = env.replace(/^OED_DB_PASSWORD=.*/m, `OED_DB_PASSWORD=${oedPassword}`);
+	} else {
+		env = env.trimEnd() + `\nOED_DB_PASSWORD=${oedPassword}\n`;
+	}
+
+	// Writing new passwords to .env or creating it if it doesn't exist yet
```
huss: Maybe add that only allowing the current user to read & write since I think this is for security.

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,158 @@
...
+			await client.query(sqlOED);
+
+			await client.end();
+
+			updateEnvFile(newPostgresPassword, newOedPassword);
+
+			console.log('********************************************************************************');
+			console.log('Generated a secure PostgreSQL and OED password and applied them successfully.');
+			console.log('The passwords have been stored in ".env" for reference.');
+			console.log('If this was run manually, you will need to restart OED for the changes to take effect.');
```
huss: I tested this and if you try to refresh the web page for OED it expires your admin token and then hangs up with console errors trying to load OED. Given this, I think the message should be very clear that OED is currently not working for everyone already using OED until the restart.

Also, could changePostgresPass have another arg that indicates whether this is via the install (so no restart needed) vs. script in package.json (restart needed)? Then the script should know if a restart is needed and highlight it specially in that case. The value could automatically be set by OED so there is no issue (might need to be the first one to avoid missing args from user - maybe a reason to have consts for the index into args so they are easier to change.) What do you think? If this is done, maybe a warning about stopping all current users until restart before it starts and seeking confirmation if doing manually so no one accidentally does this at a bad time.

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,158 @@
...
+				console.error(`Transient Postgres error on attempt ${attempt}: ${error.message}`);
+				console.error('Retrying password update...');
+				await sleep(1000 * attempt);
+				continue;
+			}
+
+			console.error('Error changing PostgreSQL or OED password:', error.message);
+
+			if (error.message.includes('password authentication')) {
+				console.error('Authentication failed: default password may already be changed.');
```
huss: Maybe add "or password used is incorrect"?

src/scripts/installOED.sh
huss: You found another methods and it resolved the issue.

huss (changes requested):
First, I want to thank @Zachary-Squires for the ongoing efforts to make this as good as possible. I tried a reasonable level of testing and it seems to work correctly. As a result, I resolved all the previous comments. I did make a few new comments to consider but I'm hopeful they are not anything of significance and finishing this up should not be difficult.

Here are some overall comments to consider:

- I think you did the merge of development to resolve the merge conflict which was nice. However, per the Discord developer channel message that I recently posted and pinned, this breaks OED. To get around this I reset head to the commit before that merge and it worked fine. The other issue needs to be resolved before merging this PR but whatever changes are made hopefully that should not impact this PR.
- For manual reset of password:
  - The short-cut version to start OED without initializing the DB works. (Just for information to keep in the record).
  - It allows any password including very short ones. Should OED require the provided password to be at least some number of characters (say 12)? This should not impact the automatically generated ones which are very long.
- If you stop OED, manually delete the postgres_data/ directory & remove the Docker oed database container then it fails (either with production yes or no). Maybe it is trying to use the new passwords but since it is initializing it starts with the default ones? Whatever the reason, I tried deleting the two Postgres lines in .env (POSTGRES_PASSWORD=... & OED_DB_PASSWORD=...) and starting OED and that did work. This is unusual but sometimes done by developers to fix issues or during testing. However, it is uncommon for developers to change the default passwords and it isn't required in your new system. Unless you have another idea, I say it is fine and we add to the developer docs that this extra step is needed if you changed the passwords.
- What is the status of the limitation listed in the PR description?

src/server/util/changePostgresPass.js
Zachary-Squires: I upped it to 32.

src/server/util/changePostgresPass.js
Zachary-Squires: The m just puts it into multiline mode so that the POSTGRES_PASSWORD is checked for in every line until they find it. The test just tells it whether or not POSTGRES_PASSWORD was found in the .env file.

src/server/util/changePostgresPass.js
Zachary-Squires: I added them, it's these specific ones because I was getting the tuple concurrently updated error and after I looked it up it only happens with those two and is fixed by a retry usually.

src/server/util/changePostgresPass.js
Zachary-Squires: Added.

src/server/util/changePostgresPass.js
Zachary-Squires: I changed the message and added the new arg.

src/server/util/changePostgresPass.js
Zachary-Squires: Added.

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,180 @@
...
+// Change the database passwords for both the default postgres user and the OED user
+// if this is done after the initial setup, OED must be restarted to get a connection with the server
+async function changePasswords() {
+	const fileEnv = parseEnvFile(ENV_PATH);
+
+	// Determine context: 'install' (automatic, no restart needed) or 'manual' (script, restart needed)
+	const isManual = process.argv[4] !== 'install';
+
+	// Warn if manual invocation that all OED users will lose access until restart
+	if (isManual) {
```
huss, line 94: I need to test this. I cannot get my version of OED to start. I think it may be due to previous testing where I was manually playing around with files to try to break. Unfortunately, I cannot wipe the DB and restart due to the issue with development. I tried to manually overcome this but it didn't readily work. Thus, I'm noting that testing will be delay until the issue with development is resolved.

I'm also inclined to ask the user if they want to continue and stop if they do not input yes. Would you like to add that or should I? This also probably applies to a similar message below but I need to verify that once I run it.

src/server/util/changePostgresPass.js
huss: I'm resolving given another message about testing this.

huss (changes requested):
@Zachary-Squires Overall, this looks good. I made one comment about my testing status that suggested a small change. Let me know if you want to do it or not (or have thoughts). Except for being able to fully test this looks except for the comment.

The only other open item was my question about this limitation in the PR description:

> In order to fully remediate the insecure default configuration we will need to use a separate docker file when launching in development and production, this will be done in a separate PR since it will be intrusive to every OED developer.

Could you please comment on the status of this? I want to figure out if there is a plan or an issue needs to be opened.

Please let me know if anything is unclear, have thoughts or need anything.

src/server/util/changePostgresPass.js
Zachary-Squires: I went ahead and added it.

src/server/util/changePostgresPass.js
```diff
@@ -0,0 +1,202 @@
...
+
+			const sqlOED = `ALTER USER oed WITH PASSWORD '${escapePassword(newOedPassword)}'`;
+			await client.query(sqlOED);
+
+			await client.end();
+
+			updateEnvFile(newPostgresPassword, newOedPassword);
+
+			console.log('********************************************************************************');
+			console.log('Generated a secure PostgreSQL and OED password and applied them successfully.');
```
huss, line 164: I've managed to get the DB passwords reset so I could test some more. I noticed that this message is always displayed even if the user provided the password. Could it be changed so it says this when generated and, if possible, that it used the user provided ones if not? I do see that it is possible of for only one of the two passwords to be provided to the script so that may make it necessary to do a little tracking.

huss (changes requested):
@Zachary-Squires Thank you for the update. I believe the PR is down to this:

- I made one new comment to consider.
- There is still one comment (not to a line of code) that needs a response on the status.
- This will have to wait for final testing when the development install issue is resolved. That isn't something this PR needs to do but I am noting.

Please let me know if you have any thoughts or need anything.

src/server/util/changePostgresPass.js
Zachary-Squires: I looked at it and thought that maybe it just telling the user what passwords were used would accomplish the same purpose, that way you can both verify which method of generation was used and for which password and it won't say "generated" if it was manual. Let me know if the change works.

huss: @Zachary-Squires What you did was fine but I am always a little nervous about dumping a password to a terminal. The person might not notice and leave it so others can see it. I don't think it is a big deal but putting it in a file seems a little better. Of course, if they use the args to set the new password then it is in the history but hopefully they are aware they did this. I'm open to other thoughts.

huss:
@Zachary-Squires  I'm trying to be sure the status of comments. I think the only older one that I wanted your input is:

> It allows any password including very short ones. Should OED require the provided password to be at least some number of characters (say 12)? This should not impact the automatically generated ones which are very long.

huss:
I've been working on this for several days. I need some more time to get it to a great place given the changes I decided to make. I will push my changes at times to be sure they are not lost but it isn't ready/done until I comment about it.

huss:
This is to note that this is probably close. I have tried to read carefully and test weird cases. The biggest remaining item is that the common environment variables in the docker-compose.yml file are not doing what is hoped so they are duplicated. It works now but you can cause issues if you don't change both the same way.

@Zachary-Squires I have made non-trivial changes to get it to be how I thought was best (no reflection on you as I changed my mind while looking this over). I know your semester is over but would you be able to look my changes over? Also, line 223 and the following lines in changePostgresPasswords.js have two TODO comments about if it is needed. I was unsure if they were needed to update the values for other places or if it was a hold over from before the later lines set the value involved. Any insight from you would be appreciated.

huss:
> Also, line 223 and the following lines in changePostgresPasswords.js have two TODO comments about if it is needed. I was unsure if they were needed to update the values for other places or if it was a hold over from before the later lines set the value involved. Any insight from you would be appreciated.

I tested commenting it out and it seems to work. Given this, I pushed a commit to do that. Commented out code remains in case something goes wrong along with a comment.

huss:
> The biggest remaining item is that the common environment variables in the docker-compose.yml file are not doing what is hoped so they are duplicated. It works now but you can cause issues if you don't change both the same way.

I've created issue #1654 to fix this after this code is merged. I seems more important to get this in to make OED more secure.

huss:
Given this helps secure OED, I've decided to merge it even though the original developer has not indicated they review my recent changes. I've tried to be careful but anyone with future thoughts or concerns should let OED know. Since OED is still a while away from a new release, it is hoped that time will allow others to try this out to make sure nothing goes wrong.

PR #1559, opened 2025-12-03
Fix: resolve security vulnerabilities identified by Semgrep SAST analysis

ell13-c:
# Description

This PR addresses 13 security vulnerabilities identified through static application security testing using Semgrep. The vulnerabilities were found and fixed as part of a computer security assignment. All fixes have been validated and no longer trigger Semgrep findings. 

**Vulnerabilities Fixed:** 
1. Missing subresource integrity (SRI) on CDN resources - Added integrity checks to Font Awesome CDN link
2. Docker container security hardening - Added 'read_only' and 'no-new-privileges' to cypress service
3. Remote property injection - Added validation for user-controlled object property access
4. Unsafe format strings in logging (3 instances) - Parameterized console.log statements using util.format
5. Path traversal vulnerability - Sanitized and validated file paths before resolution
6. XSS false positives (2 instances) - Documented validated database responses with nosemgrep annotations

**Author:** Eleanor Colvin (ell13-c)
(Note: This PR does not fix a specific existing issue but addresses security vulnerabilities discovered through SAST analysis)

## Type of change

## Checklist

## Limitations
None. All identified vulnerabilities have been addressed an verified with Semgrep scanning.


huss:
@ell13-c Thank you very much for the contribution to OED. ~First, the box for having filled out the CLA is checked in the description but I cannot find a record of your submission of the CLA. Could you please either fill it out or let me know that our records are off.~ Second, there is a team of students who are working on security aspects in OED. I think it would be good for the team to look this over. I'm not sure how soon that will happen as it is the end of the semester. Please let me know if this will cause you any inconvenience/issues.

I did look this over and, overall, it seems good. A final review is still needed and waiting on the items above.

ell13-c:
Hi! Thank you so much for reviewing my PR. I apologize - I just submitted the CLA form. Please let me know if you need anything else from me. 

I'm happy to wait for the security team's review. This was part of a semester project and grading isn't dependent on review, so there's no rush on my end. I'm available to make any changes or answer questions if needed.

huss:
Great. I may try to look at this between semesters. I talked to the security team this morning. They are not likely to start looking at this until mid-January when their next semester starts. Please let me know if you have any thoughts or if you need something sooner.

docker-compose.yml
```diff
@@ -111,4 +111,7 @@ services:
         /bin/sh -c "
         rm -f /tmp/.X99-lock &&
         Xvfb :99 -screen 0 1024x768x16"
-        
\ No newline at end of file
+      read_only: true
```
huss, line 114: This seems fine. I'm wondering if either applies to other services. The database service definitely needs to be able to write back to the file system. I'm less sure about the web service and if being in development mode matters. Do you have any thoughts on this since you tested it out and put it in? OED could make this a new issue to look at if that is appropriate.

src/server/middleware/paramsProcessing.js
```diff
@@ -12,17 +12,28 @@
  * A middleware to lowercase all params, including those passed by form/multipart
  */
 function lowercaseAllParamNames(req, res, next) {
-	for (const key of Object.entries(req.query)) {
-		req.query[key[0].toLowerCase()] = key[1];
+
+    const newQuery = Object.create(null);
```
huss, line 17: Am I correct this is the code doing "Remote property injection - Added validation for user-controlled object property access"?

I see the change to access by key, value rather than an index in key. Could you explain why this helps and maybe put in a comment about why since it should be preserved for security reasons?

src/server/models/database.js
```diff
@@ -56,9 +56,15 @@ const loadedSqlFiles = {};
  * @returns {pgPromise.QueryFile}
  */
 function sqlFile(filePath) {
-	const sqlFilePath = path.join(sqlFilesDir, filePath);
+	const sanitizedPath = filePath.replace(/\.\./g, ''); 
+    const resolvedPath = path.resolve(sqlFilesDir, sanitizedPath);
+	if (!resolvedPath.startsWith(path.resolve(sqlFilesDir))) {
+		throw new Error('Invalid file path: path traversal detected');
```
huss, line 64: I think logging this event with the standard OED logging might be a good idea for debugging and because it indicates an issue. I think it should be in the error category. I think it would help to include the two paths checked in the statement above to provide more information for the logging.

src/server/models/database.js
```diff
@@ -56,9 +56,15 @@ const loadedSqlFiles = {};
  * @returns {pgPromise.QueryFile}
  */
 function sqlFile(filePath) {
-	const sqlFilePath = path.join(sqlFilesDir, filePath);
+	const sanitizedPath = filePath.replace(/\.\./g, '');
```
huss, line 60: I don't have a problem with this change but I don't think it should ever happen. As far as I know, all the filePath values passed are hardcoded into the software so users cannot cause an issue. Thus, I think it would mean bad/incorrect/malicious code was added to the software. Please let me know if you agree or think otherwise.

src/server/util/insertData.js
```diff
@@ -27,7 +28,7 @@ async function insertUnits(unitsToInsert, update = false, conn) {
 			let ok = true;
 			requiredKeys.forEach(key => {
 				if (!unitData.hasOwnProperty(key)) {
-					console.log(`********key "${key}" is required but missing so unit number ${index} not processed with values:`, unitData);
+					console.log(util.format('********key "%s" is required but missing so unit number %s not processed with values:', key, index), unitData);
```
huss, line 31: As in another place, I don't have a problem with this but I believe it is only used by internal testing code so I don't think issues would arise. However, it does protect against future use. Let me know if you think otherwise.

huss (changes requested):
Thanks to @ell13-c for this contribution. I'm sorry this took a while to get through this process. Testing and review found it seems to work well. I made a few comments to consider. Please let me know if anything is not clear or you need anything.

Also, this PR was out of date with development. This was not your fault but because it sat for so long. I wanted to  verify there were no conflicts and it still worked correctly. I decided to push the updates to the PR since it was done.

Update on 30 March 2026: I ran all the tests again on my machine and did not see a failure. Also, it has never showed up on GitHub. Given this, I'm leaving the comment in case someone needs to find it but it is not an issue with this PR.

I am seeing occasional failures of the test cases on my machine. The only one so far that has happened both when doing the full tests and a single test file (multiple times) has this error:

```text
1) Log Routes
       Insert and verify log functionality
         should handle date range filtering:
     AssertionError: expected undefined not to be undefined
      at Context.<anonymous> (src/server/test/routes/logsRouteTests.js:322:31)
```

I'm unsure why this may be happening. It is also unclear to me if it relates to the PR, my machine, previous work, etc. I am noting in case it fails for others or during CI testing on GitHub. If anyone has seen this or has any ideas then I'd be pleased to hear.

docker-compose.yml
ell13-c: I only applied read_only and no-new-privileges to the cypress service since that's what Semgrep flagged. Agreed that the database service can't be read_only since it needs to write to disk. 

For the web service, it could be a candidate in a production environment, but I'm not sure if in development mode Node.js might need to write compiled assets or cache files, so I didn't want to risk breaking the dev workflow without more testing. This could be worth opening as a separate issue to investigate properly.

src/server/middleware/paramsProcessing.js
ell13-c: Yes, this is the remote property injection fix. The main change is building a new null-prototype object with Object.create(null) instead of mutating req.query/params/body in place to prevent prototype pollution from user-controlled keys. The destructuring to [key, value] is just more of a minor cleanup on top of that. 

I'll add a comment explaining this so the intent is clear for future contributors.

src/server/models/database.js
ell13-c: Good suggestion. I'll update this to log an error with the standard OED logging before throwing, and include both the resolved path and the expected base directory in the message so it's easier to diagnose.

src/server/models/database.js
ell13-c: I agree on this one. Since all the filePath values passed to sqlFile() are hardcoded internally, this shouldn't be reachable under normal circumstances. I kept the check as a defensive measure in case a future code change accidentally passes a user-controlled value, but I understand if you'd prefer to remove it instead.

src/server/util/insertData.js
ell13-c: Agreed, this is definitely low risk in practice since it's only used by internal testing code. I made the change as a consistent defense habit and to guard against any future use where the input might be less controlled. Happy to revert it if you'd prefer to keep the diff minimal.

src/server/middleware/paramsProcessing.js
ell13-c: done in f1d78939

src/server/models/database.js
ell13-c: done in 10b93adf

docker-compose.yml
huss: There is an ongoing effort around hardening the Docker containers and I've let them know about this. I'll leave this to them but am not resolving so it is easier for them to see this.

src/server/models/database.js
huss: It is fine to leave. It adds security at a low cost.

src/server/util/insertData.js
huss: As with another one, this adds security at low cost so let's leave it.

huss (approved):
@ell13-c Thank you for addressing all of my previous comments. Review and testing found everything is working fine. I would like to thank you for your patience on moving this PR forward and your continued engagement over many month. This work is now ready to merge. Congratulations on your first accepted PR to OED.

PR #1560, opened 2025-12-05
Fix unit route so not overwrite unspecified values - Team B

### LINKED ISSUE #1534: Fix unit route so not overwrite unspecified values

src/server/routes/units.js in the /edit route sets all the unit values but does not require them all as parameters. As a result, it might inappropriately set values.

The current values should be merged with any new ones provided. This is what is done in src/server/routes/meters.js.

OED could require all the values but that would change the functionality.

### Additional context

None

AndrewTr0612:
# Description

(This PR updates the /edit route in units.js so that unit fields are only updated when those parameters are specifically provided. Previously, the route set all unit values even when some were not passed in the request body, which could result in unintended changes.

Contributors

Thien An Truong, @AndrewTr0612 email: [andrewtruong0612@gmail.com](mailto:andrewtruong0612@gmail.com)
Anthony Duenez Ramirez, @anthonyduenez email: [duenezanthony67@gmail.com](mailto:duenezanthony67@gmail.com)
Nathan Espina, @Nespina24, email: [nespina1@toromail.csudh.edu](mailto:nespina1@toromail.csudh.edu)
Princeton Nguyen, @princetonn, email: [princeton.nguyen13@gmail.com](mailto:princeton.nguyen13@gmail.com)
Fixes #1534 

## Type of change

## Checklist

## Limitations

This change addresses only the behavior of the /edit route in units.js. No additional validation are included.

### DIFF
--- a/src/server/routes/units.js
+++ b/src/server/routes/units.js
@@ -100,8 +100,8 @@ router.post('/edit', adminAuthMiddleware('edit units'), async (req, res) => {
 				// Suffix changes so some conversions and units need to be removed.
 				await removeAdditionalConversionsAndUnits(unit, conn);
 			}
-			for (const key of Object.keys(unit)){
-				if (Object.prototype.hasOwnProperty.call(req.body, key) && !['id'].includes(key)){
+			for (const key of Object.keys(unit)) {
+				if (Object.hasOwn(req.body, key)) {
 					unit[key] = req.body[key]
 				}
 			}
--- a/src/server/routes/units.js
+++ b/src/server/routes/units.js
@@ -100,18 +100,11 @@ router.post('/edit', adminAuthMiddleware('edit units'), async (req, res) => {
 				// Suffix changes so some conversions and units need to be removed.
 				await removeAdditionalConversionsAndUnits(unit, conn);
 			}
-			unit.name = req.body.name;
-			unit.displayable = req.body.displayable;
-			unit.identifier = req.body.identifier;
-			unit.unitRepresent = req.body.unitRepresent;
-			unit.typeOfUnit = req.body.typeOfUnit;
-			unit.preferredDisplay = req.body.preferredDisplay;
-			unit.secInRate = req.body.secInRate;
-			unit.suffix = req.body.suffix;
-			unit.note = req.body.note;
-			unit.minVal = req.body.minVal;
-			unit.maxVal = req.body.maxVal;
-			unit.disableChecks = req.body.disableChecks;
+			for (const key of Object.keys(unit)){
+				if (Object.prototype.hasOwnProperty.call(req.body, key) && !['id'].includes(key)){
+					unit[key] = req.body[key]
+				}
+			}
 			// TODO Consider if this might be a better way.
 			// Object.assign(unit, req.body);
 			await unit.update(conn);
--- a/src/server/test/routes/unitsRouteTests.js
+++ b/src/server/test/routes/unitsRouteTests.js
@@ -22,14 +22,16 @@ mocha.describe('Units Route', () => {
 			const unit = new Unit(undefined, 'Unit', 'Unit Id', Unit.unitRepresentType.QUANTITY,
 				1000, Unit.unitType.UNIT, 'Suffix', Unit.displayableType.ALL, true, 'Note');
 			await unit.insert(conn);
+			const beforeSuffix = unit.suffix;
 			const res = await chai.request(app).post('/api/units/edit').set('token', token).send({
-				...unit,
 				id: unit.id,
 				name: 'New name',
+				identifier: unit.identifier,
 			});
 			expect(res).to.have.status(200);
 			const updatedUnit = await Unit.getById(unit.id, conn);
 			expect(updatedUnit.name).to.equal('New name');
+			expect(updatedUnit.suffix).to.equal(beforeSuffix);
 		});
 	
 	});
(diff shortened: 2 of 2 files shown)


danielbachhuber:
I created a test you can extend to cover this new behavior: https://github.com/AndrewTr0612/OED-Testcase-CG9-and-1534/pull/4

src/server/test/routes/unitsRouteTests.js
```diff
@@ -0,0 +1,37 @@
...
+
+	mocha.describe('Edit endpoint', () => {
+
+		mocha.it('returns 200 and updates the database for valid request', async () => {
+			const conn = testDB.getConnection();
+			const unit = new Unit(undefined, 'Unit', 'Unit Id', Unit.unitRepresentType.QUANTITY,
+				1000, Unit.unitType.UNIT, 'Suffix', Unit.displayableType.ALL, true, 'Note');
+			await unit.insert(conn);
+			const res = await chai.request(app).post('/api/units/edit').set('token', token).send({
+				...unit,
```
danielbachhuber: `...unit` can now be removed, but you'll need to add the `identifier` property to the request because it's required.

src/server/test/routes/unitsRouteTests.js
```diff
@@ -0,0 +1,37 @@
...
+				id: unit.id,
+				name: 'New name',
+			});
+			expect(res).to.have.status(200);
+			const updatedUnit = await Unit.getById(unit.id, conn);
+			expect(updatedUnit.name).to.equal('New name');
```
danielbachhuber, line 33: Can you add one more assertion that verifies the properties of unit aren't included in the request aren't being updated in the database?

danielbachhuber (changes requested):

huss:
@AndrewTr0612 & @danielbachhuber I wanted to know if this PR is ready for my review or if I should hold off given the comments made. Thanks.

danielbachhuber:
> I wanted to know if this PR is ready for my review or if I should hold off given the comments made.

@huss As of my conversation with them on Friday, they're going to try to address my comments first.

anthonyduenez:
@huss The PR is ready for review, I just finished addressing the comments.

src/server/routes/units.js
```diff
@@ -100,18 +100,11 @@ router.post('/edit', adminAuthMiddleware('edit units'), async (req, res) => {
...
-			unit.unitRepresent = req.body.unitRepresent;
-			unit.typeOfUnit = req.body.typeOfUnit;
-			unit.preferredDisplay = req.body.preferredDisplay;
-			unit.secInRate = req.body.secInRate;
-			unit.suffix = req.body.suffix;
-			unit.note = req.body.note;
-			unit.minVal = req.body.minVal;
-			unit.maxVal = req.body.maxVal;
-			unit.disableChecks = req.body.disableChecks;
+			for (const key of Object.keys(unit)){
```
huss: OED likes a space between ``){``. If you are using VSC then the "format document" command will fix this. There is at least one more formatting issue in this file. The other file has a tab on an otherwise blank line (line 36) that should also be fixed.

src/server/routes/units.js
```diff
@@ -100,18 +100,11 @@ router.post('/edit', adminAuthMiddleware('edit units'), async (req, res) => {
...
+				if (Object.prototype.hasOwnProperty.call(req.body, key) && !['id'].includes(key)){
```
huss: I started looking at this since the syntax for using ``hasOwnProperty`` differs from other places in OED. I then noticed that MDN recommends using [hasOwn()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) over [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) if it is supported. Since this is on the server I think (not certain) that is depends on our node version which is past the minimum of 16.9. OED has not made this change anywhere else but maybe we should start here. What do you think? If not making this change, could you comment/think about the way OED has done before which would involve ``req.body.hasOwnProperty(key)``.

I'm also wondering about the ``&& !['id'].includes(key)``. I think the code above guarantees that the id must be the one provided or unit would not be defined (would lead to an exception/catch in the code). If I'm correct then overwriting the id would only replace it with the same value. Not a big deal but it would simplify the code if you think removing the check makes sense. If not, I think a comment above the code explaining what it is doing would be nice.

src/server/test/routes/unitsRouteTests.js
```diff
@@ -0,0 +1,39 @@
...
+	});
+
+	mocha.describe('Edit endpoint', () => {
+
+		mocha.it('returns 200 and updates the database for valid request', async () => {
+			const conn = testDB.getConnection();
+			const unit = new Unit(undefined, 'Unit', 'Unit Id', Unit.unitRepresentType.QUANTITY,
+				1000, Unit.unitType.UNIT, 'Suffix', Unit.displayableType.ALL, true, 'Note');
+			await unit.insert(conn);
+			const beforeSuffix = unit.suffix;
```
huss: There is nothing wrong with testing the suffix. However, they do different actions in OED so I might be inclined to get the note and test that if you are okay with that change.

huss (changes requested):
Thanks to @AndrewTr0612, @anthonyduenez, @Nespina24 & @princetonn for their first contribution to OED. Overall, this looks good. I made a few comments to consider. Please let me know if you have any thoughts, something is not clear or I can help.

anthonyduenez:
@huss The PR is ready for review. I loved your feedback, I did all the changes you recommended!

src/server/test/routes/unitsRouteTests.js
huss: I think this is done.

huss (approved):
Thanks for addressing my comments. Review and testing found it works as desired. Congratulations to @AndrewTr0612, @anthonyduenez, @Nespina24 & @princetonn on having your work merged into OED.

Issue #1291, opened 2024-07-07
Weather data


huss:
### Is your feature request related to a problem? Please describe.

Some types of resource usage relates to weather data, most notably temperature.

### Describe the solution you'd like

OED should acquire and display temperature data (and maybe other if desired). It may also analyze the data to model usage.

### Describe alternatives you've considered

OED does not have this feature.

### Additional context

See the [weather design document](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/weather/weather.md) for details and status.

Jinanmeotaku:
My team and I are working on this issue.

huss:
@Jinanmeotaku Thank you for noting this on the issue and I've assigned you the issue. Please let OED know if they can help.

huss:
@Jinanmeotaku As I'm planing work for the fall and it has been a while, I wanted to touch base to see where this is at and if you are still working on this issue. Thanks.

PR #1562, opened 2025-12-18
Weather

danielshid:
# Description

Refactored and added to weather data and weather location code according to the "Getting temperature data" section of the weather data design document. 

Partially Addresses #1291

## Type of change

## Checklist

## Limitations

The data that my code gets from Openmateo does not match what Openmateo outputs directly.


package.json
```diff
@@ -84,6 +85,7 @@
 		"ngraph.graph": "~20.0.0",
 		"ngraph.path": "~1.5.0",
 		"nodemailer": "~6.9.13",
+		"openmeteo": "~1.1.3",
```
huss: openmeteo is now at version 1.2.3.

src/server/test/routes/weatherLocationRouteTests.js
```diff
@@ -0,0 +1,52 @@
...
+const app = require('../../app');
+const { expect } = require('chai');
+
+mocha.describe('Weather Location Routes', () => {
+	mocha.describe('GET all weather locations', () => {
+		mocha.it('correct call', async () => {
+			const response = await request(app).get('/api/weatherLocation');
+			// Check for successful response
+			expect(response.status).to.equal(200);
+			expect(response.body).to.be.an('array');
```
huss: This would be a much stronger test if it added a location first and verified it could retrieve that location. Please either do or add a TODO.

The other two test are also weak in not checking the returned values.

src/server/test/routes/weatherLocationRouteTests.js
```diff
@@ -0,0 +1,52 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const { mocha } = require('../common');
+const request = require('supertest');
+const app = require('../../app');
+const { expect } = require('chai');
+
+mocha.describe('Weather Location Routes', () => {
```
huss, line 10: I think all the tests should be under this one describe.

src/server/test/routes/weatherLocationRouteTests.js
```diff
@@ -0,0 +1,52 @@
...
+	mocha.describe('GET all weather locations', () => {
+		mocha.it('correct call', async () => {
+			const response = await request(app).get('/api/weatherLocation');
+			// Check for successful response
+			expect(response.status).to.equal(200);
+			expect(response.body).to.be.an('array');
+		});
+	});
+
+	mocha.describe('POST for adding a weather location', () => {
```
huss: This and the next test fails because the route now (correctly) checks for admin status. Thus, it needs to login in first to get the token as in other admin route tests.

src/server/test/db/enumTests.js
```diff
@@ -134,6 +134,35 @@ mocha.describe('Enums JS to DB', () => {
 			)
 	})
 
+	//Unit.temperatureUnitType JS object to temperature_unit_type SQL ENUM
+	mocha.it('can equate Unit.temperatureUnitType to SQL enum', async () => {
+		const conn = DB.getConnection();
+		let serverEnum = [];
+		let jsObject = [];
+		//SQL query returning area_unit_type ENUM
```
huss: area_unit_type -> temperature_unit_type

src/server/sql/group/create_groups_tables.sql
```diff
@@ -13,6 +13,10 @@ CREATE TABLE IF NOT EXISTS groups (
 	area_unit area_unit_type NOT NULL DEFAULT 'none'
 );
 
+-- TODO: Add weather_location_id column to groups table
```
huss, line 16: The TODO here and in meters is fine but I note that it will also require adding on the client to allow input/route.

src/server/sql/meter/create_meters_table.sql
```diff
@@ -32,6 +32,7 @@ unit_id: The foreign key to the unit table. The meter receives data and points t
 default_graphic_unit: The foreign key to the unit table represents the preferred unit to display this meter
 area_unit: The area unit enum
 reading_frequency: The time between readings
+weather_location_id: A foreign key, that indicates which weather_location_id is associated with the meter
```
huss, line 35: Maybe this should be moved to the TODO below since it isn't yet in the SQL?

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -358,6 +391,16 @@ export default function PreferencesComponent() {
 					onChange={e => makeLocalChanges('defaultHelpUrl', e.target.value)}
 				/>
 			</div>
+			<div>
+				<p className='mt-2' style={titleStyle}>
+					<FormattedMessage id='default.weather.location' />:
+				</p>
+				<Input
+					type='text'
```
huss: I think this should be a drop down menu with a list of all the known weather locations for this site (or empty if none). This is similar to the timezone above or on a number of other pages.

src/server/models/database.js
```diff
@@ -87,6 +92,8 @@ async function createSchema(conn) {
 	const Cik = require('./Cik');
 
 	/* eslint-enable global-require */
+	await WeatherLocation.createTable(conn);
+	await WeatherData.createTable(conn);
```
huss: Below needs Unit.createTemperatureUnitTypesEnum. It would also be good to add to the migration to document.

src/server/routes/preferences.js
```diff
@@ -80,6 +80,9 @@ router.post('/', adminAuthMiddleware('edit site preferences'), async (req, res)
 				},
 				defaultHelpUrl: {
 					type: 'string'
+				},
+				defaultWeatherLocation: {
+					type: 'string'
```
huss: I think with the other changes this will become an integer as the id of the location.

Above, I don't see the default temperature unit.

src/server/models/Reading.js
```diff
@@ -210,6 +211,20 @@ class Reading {
...
 	}
 
+	/**
+	 * Returns a promise to get the earliest start timestamp from all readings.
+	 * @param conn the database connection to use
+	 * @returns {Promise<Moment>} the earliest start timestamp as a Moment object
+	 */
+	static async getEarliestTimeStamp(conn) {
+		try {
+			return moment( await conn.one(sqlFile('reading/get_earliest_timestamp.sql')));
```
huss: Extra space after (.

src/server/models/database.js
```diff
@@ -99,6 +106,7 @@ async function createSchema(conn) {
 	// This sql code creates a function to check meter's timezone.
 	// It needs to be called before meter table is created.
 	await conn.none(sqlFile('meter/check_timezone.sql'));
+	// await WeatherLocation.createTable(conn);
```
huss: Remove?

src/server/sql/weather_data/create_weather_data_table.sql
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+CREATE TABLE IF NOT EXISTS weather_data (
+    weather_location_id INTEGER NOT NULL,
+    start_time TIMESTAMP NOT NULL,
```
huss: src/server/sql/reading/create_readings_table.sql has the end_time and reading (temperature) as not null. It avoids issues and maybe makes sense here.  It also has ``	CHECK (start_timestamp < readings.end_timestamp),`` which also makes sense to me.

src/server/sql/weather_data/get_weather_data_by_id_and_date_range.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss, line 1: I could not find where this is used except in the matching model where it is called without any of the needed values. I'm thinking it should probably get by a start & end time as done for readings since that is what is probably needed for graphing.

src/server/models/WeatherData.js
```diff
@@ -0,0 +1,86 @@
...
+
+const database = require('./database');
+const { log } = require('../log');
+const moment = require('moment');
+
+const sqlFile = database.sqlFile;
+
+class WeatherData {
+	/**
+	 * Constructor without the id sine it's not in the schema
```
huss: sine -> since if stays. However, I'm unclear about this schema comment since id is in the table.

src/server/models/WeatherData.js
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	/**
+	  * Returns a promise to retrieve the weather data by the primary key
+	  * @param {*} conn The connection to use.
+	  * @returns {Promise.<WeatherData>}
+	  */
+	// TODO: needs model and db get_weather_data
+	static async getWeatherData(conn) {
+		const rows = await conn.one(sqlFile('weather_data/get_weather_data_by_id_and_start_time.sql'));
```
huss: This isn't passing the needed values. See the sql for thoughts on how this should work.

src/server/models/WeatherData.js
```diff
@@ -0,0 +1,86 @@
...
+		return rows.map(WeatherData.mapRow);
+	}
+
+	/**
+	 * Returns a promise to get the latest end timestamp for a given weather data location.
+	 * @param id
```
huss: Maybe say what id is?

src/server/sql/weather_location/delete_weather_location.sql
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+DELETE
```
huss, line 5: I'm noting that you cannot delete a location if there is any weather data since here is no way to remove those. I think we should discuss how to deal with this. Unlike readings you can recreate weather data so removing a location is not as serious. Thus, the readings can also be removed but this would require a cascade. However, when locations get tied to meters/groups then that is another constraint where I think the admin needs to be warned and remove them manually. This is how units, etc. work.

src/server/models/WeatherLocation.js
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	/**
+	 * Returns a promise to retrieve the weather location with the given coordinates from the database.
+	 * @param longitude the longitude to look up
+	 * @param latitude the latitude to look up
+	 * @param conn the connection to use.
+	 * @returns {Promise.<WeatherLocation>} either the weather_location object with info or null if does not exist.
+	 */
+	static async getByCoordinates(longitude, latitude, conn) {
```
huss: This is fine but I do want to understand where it will be used.

src/server/models/WeatherLocation.js
```diff
@@ -0,0 +1,119 @@
...
+
+const database = require('./database');
+
+const sqlFile = database.sqlFile;
+
+class WeatherLocation {
+	/**
+	 * @param id This weather location's ID.
+	 * @param identifier This weather location's identifier
+	 * @param gps location in format of GIS coordinates
```
huss: OED normally says GPS not GIS.

src/server/models/WeatherLocation.js
```diff
@@ -0,0 +1,119 @@
...
+	 * @returns {Promise.<>}
+	 */
+	static createTable(conn) {
+		return conn.none(sqlFile('weather_location/create_weather_location_table.sql'));
+	}
+
+	/**
+	 * Returns a promise to retrieve the weather location with the given id from the database.
+	 * @param conn is the connection to use.
+	 * @param id
```
huss: Maybe add what id is.

src/server/models/WeatherLocation.js
```diff
@@ -0,0 +1,119 @@
...
+	 * @returns {Promise.<WeatherLocation>}
+	 */
+	static async getByID(id, conn) {
+		const row = await conn.one(sqlFile('weather_location/get_weather_location_by_id.sql'), { id: id });
+		return new WeatherLocation(row.id, row.identifier, row.longitude, row.latitude, row.note);
```
huss, line 38: OED normally has a mapRow function to do this. It also applies to other functions in this class.

src/server/models/WeatherLocation.js
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	static async getAll(conn) {
+		const rows = await conn.any(sqlFile('weather_location/get_all_weather_locations.sql'));
+		return rows.map(row => new WeatherLocation(row.id, row.identifier, row.gps, row.note));
+	}
+
+	/**
+	 * Returns a promise to insert this weather location into the database
+	 * @param conn is the connection to use.
+	 * @returns {Promise.<>}
```
huss, line 65: Maybe should indicate that updates the id for this.

src/server/models/WeatherLocation.js
```diff
@@ -0,0 +1,119 @@
...
+		this.id = resp.id;
+	}
+
+	/**
+	 * Returns a promise to insert all of the given weather locations into the database (as a transaction)
+	 * @param {array<WeatherLocation>} weatherLocations the weather locations to insert
+	 * @param conn is the connection to use.
+	 * @returns {Promise.<>}
+	 */
+	static insertAll(weatherLocations, conn) {
```
huss: I'm trying to think about where this might be used. Can you help me?

src/server/routes/weatherLocation.js
```diff
@@ -0,0 +1,219 @@
...
+				oneOf: [
+					{
+						type: 'object',
+						required: ['latitude', 'longitude'],
+						properties: {
+							latitude: { type: 'number', minimum: '-90', maximum: '90' },
+							longitude: { type: 'number', minimum: '-180', maximum: '180' }
+						}
+					},
+					{ type: 'null' }
```
huss: I think the sql has that gps cannot be null. I'm unclear on when that would be useful. Should this be allowed here?

src/server/routes/weatherLocation.js
```diff
@@ -0,0 +1,219 @@
...
+	if (earliestDate === null) {
+		earliestDate = await Reading.getEarliestTimeStamp(conn);
+		if (earliestDate === null) {
+			log.warn('addWeatherDataForLocation(): Cannot find earliest time from WeatherData or Reading.');
+			return;
+		}
+	}
+
+	// this only gets the data for the current day(?)
+	// earliestDate = earliestDate.format('YYYY-MM-DD');
```
huss, line 99: This seems like testing code. I think it should be fixed up for the PR.

src/server/routes/weatherLocation.js
```diff
@@ -0,0 +1,219 @@
...
+	const weatherData = await fetchWeatherData(
+		location.gps.latitude,
+		location.gps.longitude,
+		earliestDate,
+		latestDate
+	);
+
+	// Insert weather data records
+	if (weatherData) {
+		await conn.tx(async t => {
```
huss, line 113: Should this be done via a new model function in weatherData so it is generally available?

src/server/routes/weatherLocation.js
```diff
@@ -0,0 +1,219 @@
...
+		failure(res, 400, `Got request to edit weather location with invalid weather location data, errors:${validatorResult.errors}`);
+	} else {
+		const conn = getConnection();
+		try {
+			await conn.tx(async t => {
+				const location = await WeatherLocation.getByID(req.body.id, t);
+				const newGPS = (req.body.gps) ? new Point(req.body.gps.longitude, req.body.gps.latitude) : null;
+				location.identifier = req.body.identifier;
+				location.gps = newGPS,
+					location.note = req.body.note;
```
huss, line 206: extra indenting tab

The note is not required in the route. If it is not provided then it will wipe out the current note. I think either make it required or make them all optional and then merge only the ones provided. There are method to do the merge to make it easier.

src/server/services/weather/fetchData.js
```diff
@@ -0,0 +1,111 @@
...
+		"end_date": endDate,
+		"hourly": "temperature_2m",
+		"temperature_unit": "fahrenheit",
+	};
+	const url = "https://archive-api.open-meteo.com/v1/archive";
+
+	return fetchWeatherApi(url, params).then(responses => {
+		const response = responses[0];
+		const utcOffsetSeconds = response.utcOffsetSeconds();
+		// console.log(utcOffsetSeconds);
```
huss: The debug statements need to go.

src/server/services/weather/fetchData.js
```diff
@@ -0,0 +1,111 @@
...
+		const timezone = response.timezone();
+		// console.log(timezone);
+		const timezoneAbbreviation = response.timezoneAbbreviation();
```
huss: This is not used except in the commented out console.log. Same for latitude and longitude.

src/server/services/weather/fetchData.js
```diff
@@ -0,0 +1,111 @@
...
+		const latitude = response.latitude();
+		const longitude = response.longitude();
+
+		const hourly = response.hourly();
+
+		// Helper function to form time ranges
+		const range = (start, stop, step) =>
+			Array.from({ length: (stop - start) / step }, (_, i) => start + i * step);
+
+		const weatherData = {
```
huss, line 38: I think some commenting would help with this block of code.

src/server/services/weather/fetchData.js
```diff
@@ -0,0 +1,111 @@
...
+
+		// for (let i = 0; i < weatherData.hourly.time.length; i++) {
+		//     console.log(
+		//         weatherData.hourly.time[i].toISOString(),
+		//         weatherData.hourly.temperature2m[i]
+		//     );
+		// }
+		// Instead of logging, return the formatted weather data
+		return weatherData.hourly.time.map((time, index) => ({
+			// TODO Lock to correct timezone.
```
huss, line 57: This needs to be resolved before the PR is merged.

src/server/routes/weatherLocation.js
```diff
@@ -0,0 +1,219 @@
...
+	return { valid: paramsValidationResult.valid, errors: paramsValidationResult.errors };
+}
+
+/**
+ * Fetches and inserts weather data for a newly created weather location.
+ * If the location has no existing weather data, it retrieves data starting from the earliest reading timestamp.
+ * @param location - Instance of WeatherLocation.
+ * @param conn - The database connection.
+ */
+async function addWeatherDataForLocation(location, conn) {
```
huss, line 85: When OED gets meter data it logs the start and then the end for the record. I think the same should happen here.

src/client/app/utils/getTemperatureUnitConversion.ts
```diff
@@ -0,0 +1,27 @@
...
+	fahrenheit = 'fahrenheit'
+}
+
+/**
+ * Gets the conversion between two temperature units
+ * @param fromUnit unit to convert FROM
+ * @param toUnit unit to convert TO
+ * @returns conversion multiplier, or zero if conversion to none
+ */
+export function getTemperatureUnitConversion(fromUnit: TemperatureUnitType, toUnit: TemperatureUnitType): number {
```
huss, line 16: I could not find where this is used. OED can convert between units so I'm wondering if this is needed or if OED should always add a conversion for this if it is needed.

src/client/app/types/redux/weather.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface WeatherLocationEditData {
+	id: number;
+	identifier: string;
+	gps: GPSPoint | null;
+	note: string;
+}
+
+export interface WeatherLocationById extends Record<number, WeatherLocationData> { }
+
+export interface WeatherState {
```
huss, line 22: I'm not seeing where this is used at this point. It did make me look and think about if it will be in src/client/app/types/redux/state.ts to be in state.

src/client/app/translations/data.ts
```diff
@@ -170,13 +172,16 @@ const LocaleTranslationData = {
...
+		"default.temperature.unit": "Default Temperature Unit",
 		"default.time.zone": "Default Time Zone",
 		"default.warning.file.size": "Default Warning File Size",
+		"default.weather.location" : "Default Weather Location",
 		"defaultGraphicUnit": "Default Graphic Unit:",
 		"delete.group": "Delete Group",
 		"delete.map": "Delete Map",
 		"delete.self": "Cannot delete your own Account.",
 		"delete.user": "Delete User",
+		"delete.weather-location": "Delete Weather Location",
```
huss: OED has not used a hyphen in the translation strings elsewhere and it isn't used in edit.weather.location so I say replace the hyphen with a period.

src/client/app/translations/data.ts
```diff
@@ -65,6 +65,7 @@ const LocaleTranslationData = {
 		"compare.period": "Compare Period",
 		"compare.raw": "Cannot create comparison graph on raw units such as temperature",
 		"confirm.action": "Confirm Action",
+		"confirm.delete.weather-location": "Are sure you want to delete this weather location?",
```
huss: See below about hyphen in key.

src/client/app/translations/data.ts
```diff
@@ -588,6 +599,18 @@ const LocaleTranslationData = {
...
 		"visual.input.units.graphic": "Input Units Visual Graphic",
 		"visual.unit": "Units Visual Graphics",
+		"weather": "Weather",
+		"weather.delete.location": "Delete Location",
+		"weather.failed.to.create.location": "Failed to create a weather location.",
+		"weather.failed.to.delete.location": "Failed to delete weather location.",
+		"weather.failed.to.edit.location": "Failed to edit weather location.",
+		"weather.input.error": "Input invalid so weather location not created or edited.",
+		"weather.latitude": "Latitude:",
+		"weather.location.no.changes": "", // WIP
```
huss: Can you help me understand what the WIP is for?

src/client/app/redux/slices/appStateSlice.ts
```diff
@@ -104,6 +105,7 @@ export const appStateSlice = createThunkSlice({
 				// Request meter/group/details post-auth
 				dispatch(metersApi.endpoints.getMeters.initiate());
 				dispatch(groupsApi.endpoints.getGroups.initiate());
+				dispatch(weatherLocationApi.endpoints.getWeatherLocationDetails.initiate());
```
huss: I'm wondering about getting the weather data on startup. The other ones (meters, groups, conversions, ...) are needed by everyone to use OED. The weather locations are only needed if you are working on them as an admin (and for meters/groups in the future). Normally a useAppSelectoris used in each component where they are needed. Please let me know if I'm missing something here.

src/client/app/redux/slices/adminSlice.ts
```diff
@@ -27,12 +28,14 @@ export const defaultAdminState: AdminState = {
...
 	defaultAreaUnit: AreaUnitType.none,
+	defaultTemperatureUnit: TemperatureUnitType.celsius,
 	defaultMeterReadingFrequency: '00:15:00',
 	defaultMeterMinimumDate: moment(0).utc().format('YYYY-MM-DD HH:mm:ssZ'),
 	defaultMeterMaximumDate: moment(0).utc().add(5000, 'years').format('YYYY-MM-DD HH:mm:ssZ'),
 	defaultMeterReadingGap: 0,
 	defaultMeterMaximumErrors: 75,
-	defaultHelpUrl: ''
+	defaultHelpUrl: '',
+	defaultWeatherLocation: '',
```
huss: No comma after last item.

src/client/app/redux/slices/adminSlice.ts
```diff
@@ -59,6 +62,10 @@ export const adminSlice = createSlice({
 			state.defaultAreaUnit = action.payload;
 			state.submitted = false;
 		},
+		updateDefaultTemperatureUnit: (state, action: PayloadAction<TemperatureUnitType>) => {
```
huss: While looking over the new code I notice that many (maybe all but did not check) of the updates for preferences are not used and some are not exported. Thus, I'm wondering if they are needed or if I'm overlooking something. I'm wondering if they are left from when admin pages updated global state before sending to the server on save. Now it is all done in local state. I did not check carefully but I think preferences are all updated as a group on the server and then in Redux state. This is only partly about this PR but if you have thoughts that would be nice. If they are not needed then OED can open an issue to remove them.

src/client/app/redux/api/weatherLocationApi.ts
```diff
@@ -0,0 +1,61 @@
...
+export const selectWeatherLocationDataResult = weatherLocationApi.endpoints.getWeatherLocationDetails.select();
+export const {
+	selectAll: selectAllWeatherLocations,
+	selectById: selectWeatherLocationById,
+	selectTotal: selectWeatherLocationTotal,
+	selectIds: selectWeatherLocationIds,
+	selectEntities: selectWeatherLocationDataById
+} = weatherLocationAdapter.getSelectors((state: RootState) => selectWeatherLocationDataResult(state).data ?? weatherLocationInitialState);
+
+export const stableEmptyWeatherLocationById: WeatherLocationById = {};
```
huss, line 61: This isn't currently used but may be if you get the values in the component.

src/client/app/components/TooltipHelpComponent.tsx
```diff
@@ -44,6 +44,9 @@ export default function TooltipHelpComponent(props: TooltipHelpProps) {
 		'help.admin.unitedit': { link: `${helpUrl}/adminUnitEditing/` },
 		'help.admin.unitview': { link: `${helpUrl}/adminUnitViewing/` },
 		'help.admin.users': { link: `${helpUrl}/adminUser/` },
+		'help.admin.weathercreate': { link: `${helpUrl}/` }, // NEEDS HELP LINK
```
huss: I suggest that the URL be put in so it mirrors other ones by name/location. OED will add the help pages as part of the release.

src/client/app/components/weather/CreateWeatherModalComponent.tsx
```diff
@@ -0,0 +1,184 @@
...
+				</ModalBody>
+				<ModalFooter>
+					<Button color='secondary' onClick={handleClose}>
+						<FormattedMessage id="discard.changes" />
+					</Button>
+					<Button color='primary' onClick={handleSubmit} disabled={!validUnit}>
+						<FormattedMessage id="save.all" />
+					</Button>
+				</ModalFooter>
+				<TooltipMarkerComponent page='weather-create' helpTextId={tooltipStyle.tooltipCreateWeatherView} />
```
huss: OED puts the tooltip right after the title at the top.

src/client/app/components/weather/WeatherDetailComponent.tsx
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Defines the weather page card view
+ * @returns Weather page element
+ */
+export default function WeatherDetailComponent() {
+	const { status } = useAppSelector(selectWeatherLocationDataResult);
+	const weatherData = useAppSelector(selectAllWeatherLocations);
+	return (
+		<div className='flexGrowOne'>
+			{status === QueryStatus.pending ? (
```
huss: This spinner is needed on some pages because the server does complex/lengthy ops. I don't think this page does it and if it does then it isn't doing cik/DB views. For example, users does not do this. I think it can be removed but if not then the message needs to be changed.

src/client/app/components/weather/WeatherViewComponent.tsx
```diff
@@ -0,0 +1,64 @@
...
+			</div>
+			<div className="item-container">
+				{/* Only show first 30 characters so card does not get too big. Should limit to one line. Protect against null from DB in note. */}
+				<b><FormattedMessage id="note" /></b> {props.weather.note ? props.weather.note.slice(0, 29) : ''}
+			</div>
+			<div className="edit-btn">
+				<Button color='secondary' onClick={handleShow}>
+					<FormattedMessage id="edit.weather.location" />
+				</Button>
+				{/* Creates a child UnitModalEditComponent */}
```
huss: Should refer to the weather component not unit.

src/client/app/components/weather/CreateWeatherModalComponent.tsx
```diff
@@ -0,0 +1,184 @@
...
+		setState({ ...state, [e.target.name]: e.target.value });
+	};
+
+
+	/* Create Weather Location Validation:
+		Identifier cannot be blank
+		Gps cannot be blank
+	*/
+
+	const [validUnit, setValidUnit] = useState(false);
```
huss: The state name should be refactored to be weather not unit.

src/client/app/components/weather/CreateWeatherModalComponent.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { tooltipBaseStyle } from '../../styles/modalStyle';
+import { useTranslate } from '../../redux/componentHooks';
+import { showSuccessNotification, showErrorNotification } from '../../utils/notifications';
+import { GPSPoint, isValidGPSInput } from '../../utils/calibration';
+
+/**
+ * Defines the create weather modal form
+ * @returns Weather create element
+ */
+export default function CreateWeatherModalComponent() {
```
huss, line 23: This page (and edit) do not have the warning about leaving when unsaved changes such as unit has.

src/client/app/components/weather/EditWeatherModalComponent.tsx
```diff
@@ -0,0 +1,280 @@
...
+	const resetState = () => {
+		setState(values);
+	};
+
+	const handleClose = () => {
+		props.handleClose();
+		resetState();
+	};
+
+	const compareLocations = (loc1: { [x: string]: any; }, loc2: { [x: string]: any; }) => {
```
huss: A few thoughts on this. First, I think it is trying to see if edits were made. This is not how many other pages do it. They check each item and some use a useEffect so the save is disabled if no change was made. I'm okay with the new way as it is more general but OED should consider switching to it across pages if it is going to be used. I also think the disable of the save is better and should be used here.

src/client/app/components/weather/EditWeatherModalComponent.tsx
```diff
@@ -0,0 +1,280 @@
...
+		setOpenDeleteConfirmation(true);
+	};
+
+	/* Edit WeatherLocation Validation:
+		Identifier cannot be blank
+		GPS cannot be blank
+	*/
+	const [validLocation, setValidLocation] = useState(false);
+	useEffect(() => {
+		setValidLocation(state.identifier !== '' && state.gps !== '');
```
huss: This is a weak check to disable save. See below on making it better.

src/client/app/components/weather/EditWeatherModalComponent.tsx
```diff
@@ -0,0 +1,280 @@
...
+								id='gps'
+								name='gps'
+								type='text'
+								autoComplete='on'
+								onChange={e => handleStringChange(e)}
+								value={state.gps}
+								placeholder='gps'
+								disabled={true}
+							/>
+							<FormFeedback>
```
huss: Since it cannot be changed (disabled) then I don't think form feedback is needed. The autocomplete, etc also seem unneeded (if can remove).

src/server/routes/weatherLocation.js
```diff
@@ -0,0 +1,219 @@
...
+		const conn = getConnection();
+		try {
+			const newLocation = new WeatherLocation(
+				undefined, // id
+				req.body.identifier,
+				req.body.gps ? new Point(req.body.gps.longitude, req.body.gps.latitude) : null,
+				req.body.note
+			);
+			await newLocation.insert(conn);
+			await addWeatherDataForLocation(newLocation, conn);
```
huss, line 147: When you add a new weather location I get a failure notice but the location is added. However, I don't think any weather data is added. I believe the issue is that the insert happens on the line above but the id is not captured and set in the newLocation so getLatestTimeStamp does not find it and it fails. However, maybe it should have been done via this and it is something else.

src/client/app/components/weather/EditWeatherModalComponent.tsx
```diff
@@ -0,0 +1,280 @@
...
+						<FormGroup>
+							<Label for='identifier'>{translate('identifier')}</Label>
+							<Input
+								id='identifier'
+								name='identifier'
+								type='text'
+								autoComplete='on'
+								onChange={e => handleStringChange(e)}
+								value={state.identifier}
+								placeholder='Identifier'
```
huss: OED does not use placeholders as there is a title above each input. All of these should be removed.

huss (changes requested):
Thanks to @danielshid for another contribution to OED. Not unexpectedly given the size of this PR, I made a number of comments within files. Here are some thoughts that don't fit nicely within files:

- The check failed because src/client/app/components/weather/EditWeatherModalComponent.tsx does not have an MPL header.
- There needs to be a migration for the changes to the preferences table including making the correct default values.
- src/server/sql/preferences/insert_default_row.sql needs values for the new preferences columns. I propose the temperature unit be celsius since area is meter.
- There is a model and sql for weather data but no route to get the data. This is going to be needed for the graphics.
- I tried to do a careful reading of all the code. I still need to sweep the TODO items within the code to decide what to do. I also need to check the old PR comments  and design doc to make sure all is good but I don't think that will show much at this point.

If anything is not clear or you have thoughts then please let me know. I'm also here to help as needed.

src/client/app/types/redux/weather.ts
danielshid: Needs to be looked into.

src/client/app/utils/getTemperatureUnitConversion.ts
danielshid: Area unit conversion is done in a similar way.

PR #1565, opened 2026-01-06
Create generate Rrule Function

davanmur:
# Description

Created a backend function that could take in information from the weekly and daily conversion patterns, and created an rrule (recurrence rule) so that conversions patterns could easily and efficiently be applied over a length of time. Integrated that function in timeVaryingPathConversion.js.

Partly Addresses #896 

## Type of change

## Checklist

## Limitations

- Based on my manual testing generateRrule.js works and segments are correctly added to edgeSegments timeVaryingPathConversion.js
- The is an issue with memory that seems to be related to a potentially infinite while in  timeVaryingPathConversion.js


huss:
I have made a sequence of commits lately. At this point the status is:

- The code works to generate the rule, apply to conversions & put into the DB (cik, cik_vary). This includes general patterns, suffix units, etc.
- The code passes all tests.
- The code works for the standard developer test data (not really time-varying).
- I have read the code carefully to make changes but need to do a final pass that it is ok.
- More test cases are needed but it does seem to run fine.
- There are a few open TODO items that can be addressed either before or after this is merged.
- An open question is when this will merge. There is ongoing work on adding holidays and getting close to a final DB implementation for time-varying. This work includes the unmerged time-varying front-end (still has issues) so that really should be merged first. The other work on DB may get merged first and become the point for moving forward so this would merge on top of that too. It is a version that allows for graphing of data so certain manual tests are much easier. The bottom line is I am okay with it merging when the time is right (after resolving merge conflicts and retesting) as long as it will undergo more testing in the future as will have to be done with the entire time-vary system.

Thanks to @davanmur for this contribution. I look forward to its formal inclusion in OED.

PR #1579, opened 2026-02-15
Issue 1217 - CSV Upload Validation Checks

kyle-parker-1500:
# Description

Partly Addresses #1217

## Type of change

Added validation checks for the following 6 fields within the `uploadMeters.js` file according to the note left on [PR1403](https://github.com/OpenEnergyDashboard/OED/pull/1403#pullrequestreview-2921682555):
1. Max Error: Max Error must be within the range of 0 to 75.
2. Area Unit & Value: If area unit is none then area cannot be specified.
3. Min & Max Dates: The max date cannot be before the min date & validate if they are dates.
4. Duplicates: Duplicates within the range of 1 to 9.
5. Reading Gap: Should be greater than or equal to 0.
6. Reading Variation: Should be greater than or equal to 0.

## Checklist

## Limitations

The branch which these changes were pushed on is around 300 commits behind the development branch. Also not all meters have checks being done on them, only those listed in [PR1403](https://github.com/OpenEnergyDashboard/OED/pull/1403) and those listed here. Destiny <destinyveloz@msn.com>, Oscar <oaviles-saldana@csumb.edu>, and I <kylerparker1500@gmail.com> are open to continue working on this issue.

Thank you to @SageMar and @cmatthews444 for writing the previous checks, much of our code was derived from what you wrote.


huss:
@kyle-parker-1500 Thank you for this PR. I see the CLA box is checked in the description. Given this, could you help with these:
- I could not find an entry for you in our CLA records. Could you either do it or let me know if you think our records are wrong.
- Our CLA records are by GitHub ID. I found @DestinyVeloz & @OscarAvilSal who I think might be the two other people listed. Could you or the other people verify that I have the correct GitHub IDs.

Please let me know if anything is not clear or you have thoughts.

kyle-parker-1500:
Hi @huss ,
That may have been my fault. I use a very similar username that I may have typed in by accident. I resubmitted the CLA with my current GitHub ID so I'm hoping that works out. The other IDs are correct. Thank you!

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -7,18 +7,21 @@ const { CSVPipelineError } = require('./CustomErrors');
 const Meter = require('../../models/Meter');
 const readCsv = require('../pipeline-in-progress/readCsv');
 const Unit = require('../../models/Unit');
-const { normalizeBoolean } = require('./validateCsvUploadParams');
+const { normalizeBoolean, MeterTimeSortTypesJS, MeterTimeSortTypesJS } = require('./validateCsvUploadParams');
```
huss, line 10: Is ``MeterTimeSortTypesJS`` included twice?

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+	let msg = '';
+
+	//if its a number, validate its range
+	if (!isNaN(maxErrorValue)) {
+		if (maxErrorValue < 0 || maxErrorValue > 75) {	
+			msg = `Invalid max error value in row ${rowIndex + 1}: maxError="${meter[31]}". ` + `MaxError must be a number larger than 0, and less than 75.`;
+			return { maxErrorMsg: msg, value: false };
+		}
+	}
+	return { maxErrorMsg: '', value: true };
```
huss, line 444: I'm wondering about returning this as ``value: true`` in the case that it is a NaN above. It should be a number so should this case also fail? I think this applies to other new tests.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+ * @returns 
+ */
+function validateArea(meter, rowIndex) {
+	const areaValue = Number(meter[9]);
+	const areaUnit = meter[25];
+	let msg = '';
+
+	if (areaUnit && areaUnit.toLowerCase() === 'none') {
+		if(!isNaN(areaValue) && areaValue !== 0) {
+      msg = `Invalid area value in row ${rowIndex + 1}: area="${meter[9]}". ` + `Area must be empty when area unit is 'none'.`;
```
huss, line 460: The indenting seems off and is using spaces instead of tabs as OED wants.

I looks like the test is for a non-zero area value so I wonder if the msg should say that too.

huss:
@kyle-parker-1500, @DestinyVeloz & @OscarAvilSal: First I want to apologize for how long my response has taken during an usual time. Second, I have started to review the work but I need more time to do it well. Please let me know if these are an issue.

kyle-parker-1500:
@huss don't worry about it and take your time with the review! My team and I are ready and able to fix any issues you find.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -7,18 +7,21 @@ const { CSVPipelineError } = require('./CustomErrors');
...
+const { normalizeBoolean, MeterTimeSortTypesJS, MeterTimeSortTypesJS } = require('./validateCsvUploadParams');
+const moment = require('moment-timezone');
 
 /**
  * Middleware that uploads meters via the pipeline. This should be the final stage of the CSV Pipeline.
- * @param {express.Request} req 
- * @param {express.Response} res 
+ * @param {express.Request} req
+ * @param {express.Response} res
+ * @param {express.Request} req
```
huss, line 17: I'm wondering about listing req & res twice as params since they are once in the function. I don't think you made this change but is from the original PR. Thoughts?

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -165,6 +258,83 @@ function switchGPS(gpsString) {
...
+}
+
+/**
+ * Checks if the time sort value provided is accurate (should be increasing or decreasing)
+ * @param timeSortValue the provided time sort
+ * @returns true or false
+ */
+function isValidTimeSort(timeSortValue) {
+	const validTimes = Object.values(MeterTimeSortTypesJS);
+	// must be one of the three values
```
huss, line 302: (old code) This is not correct as there are currently two values. I would put "must be one of the allowed enum values."

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -165,6 +258,83 @@ function switchGPS(gpsString) {
...
+}
+
+/**
+ * Checks if the area unit provided is an option
+ * @param areaUnit the provided area for the meter
+ * @returns true or false
+ */
+function isValidAreaUnit(areaUnit) { 
+    const validTypes = Object.values(Unit.areaUnitType); 
+    // must be one of the three values
```
huss, line 287: (old code) See isValidTimeSort for similar comment and should also be changed.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -165,6 +258,83 @@ function switchGPS(gpsString) {
...
+	// must be one of the three values
+	if (validTimes.includes(timeSortValue)) {
+		return true;
+	} else {
+		return false;
+	}
+}
+
+/**
+ * Checks if the meter type provided is one of the 5 options allowed when creating a meter.
```
huss, line 311: Similar to isValidTimeSort, change "5 options allowed" to "allowed options"

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+			throw new CSVPipelineError(
+				`Invalid input for '${name}' in row ${rowIndex + 1}: "${meter[index]}". Expected 'true' or 'false'.`,
+				undefined,
+				500
+			);
+		}
+	}
+}
+
+function validateMinMaxValues(meter, rowIndex) {
```
huss, line 403: Looking at this PR has made me think about the min/max value limits. First, I'm not sure why the meter web pages limit the value to the min/max integer when this can be a floating point number. Second, I'm unsure about checking if it is below/above a given value. I would think if the string is a valid number then it could be used. I also think -infinity/infinity should be fine to indicate not practical limits. If checking if it is a number can do this then that seems file. Note this impacts the meter create/edit web pages.

Another issue is how the test is being done. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN discusses issues with this global version such as true returning false. Igt discusses using ``typeof x === "number"`` and other ideas. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_coercion is also useful about what is a number. If the number status was checked in advance and then ``Number.isNaN()`` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) is used this may be better. Anyone doing this should verify the best way. I did a global search and found around 20 uses of isNaN() that should be checked along with any other number checks. This seems beyond the scope of this PR but wanted to document to see if others have thoughts. It probably should become an issue to address.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+	const minValue = Number(meter[27]);
+	const maxValue = Number(meter[28]);
+
+	if (isNaN(minValue) || minValue < -9007199254740991 || (!isNaN(maxValue) && minValue > maxValue)) {
```
huss, line 407: (old code) I think this code would be clearer if it used the following logic.

1. Test if minValue is a number and NaN and then if it is too small (but see comment above on possibly removing this).
2. Similar test for maxValue.
3. Convert  min/maxValue to a Number since that is now safe.
4. Test if min > max.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+}
+
+/**
+ * Validates the area value for a given meter row.
+ * @param {Number} meter 
+ * @param {Number} rowIndex 
+ * @returns 
+ */
+function validateArea(meter, rowIndex) {
+	const areaValue = Number(meter[9]);
```
huss, line 454: See other comment with info on issues with converting strings to Number without checking. I think this could cause issues such as true returns 1. This could be in other places.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+			`Min or/and max must be a number larger than -9007199254740991, and less then 9007199254740991, and min must be less than max.`,
+			undefined,
+			500
+		);
+	}
+}
+
+/**
+ * Validates the max error value for a given meter row.
+ * @param {Number} meter
```
huss, line 429: I think meter is an Array as done with validateBooleanFields. This applies to other new functions.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,22 +44,124 @@ async function uploadMeters(req, res, filepath, conn) {
...
+				
+				//checks to make sure if area unit is none then area value must be empty or 0, and if area unit is not none then area value must be a number greater than 0.
+				const areaCheck = validateArea(meter, i);
+                if (!areaCheck.value) {
+                    throw new CSVPipelineError(areaCheck.areaMsg, undefined, 500);
+                }
+			}
+
+			//checks to make sure max error value is a number between 0 and 75 if it is provided.
+			const maxErrorCheck = validateMaxError(meter, i);
```
huss, line 132: Other checks grab the value from the meter array by index and then passes that to the validation function. I think that makes the check more general and isolates the determination of the index to this main part of the code. Thus, I prefer that. This applies to more checks.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,22 +44,124 @@ async function uploadMeters(req, res, filepath, conn) {
...
+			if (!maxErrorCheck.value) {
```
huss, line 133: I think I prefer this way of having the check generate the message as then it stays with the check and is consistent. Also, throwing the error here seems better. I did notice other tests do not consistently do this. Would you be willing to modify these to be consistent with your way?

If this change is going to be made then a new function to handle an invalid input could be made. It would construct the final message with the  row and throw the error. That would be nice encapsulation. Also, the check functions would probably not need the row index as it would be taken care of here.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,22 +44,124 @@ async function uploadMeters(req, res, filepath, conn) {
...
+
+			// Verify area unit provided
+			const areaUnitString = meter[25];
+			if (areaUnitString) {
+				if (!isValidAreaUnit(areaUnitString)) {
+					let msg = `For meter ${meter[0]} the area unit of ${areaUnitString} is invalid.`;
+					throw new CSVPipelineError(msg, undefined, 500);
+				}
+				
+				//checks to make sure if area unit is none then area value must be empty or 0, and if area unit is not none then area value must be a number greater than 0.
```
huss, line 124: I think the second part of this comment is not done here but above via isValidArea.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+	let msg = '';
+	
+	// get correctly formatted dates and check if they're correctly formatted
+	const correctMinFormat = correctDateTimeFormat(minDate);
+	const correctMaxFormat = correctDateTimeFormat(maxDate);
+	
+	// validate that minDate was formatted correctly
+	let formattedMinDate, formattedMaxDate;
+	if (!correctMinFormat.value || !correctMaxFormat.value) {
+		// add error messages to the overall error message
```
huss, line 487: Below the msg gets a \n even if one is empty. Maybe check min first and add to msg as needed, then check max and add as needed. I think trying to avoid extra \n values is going to be a pain so maybe just make each one a separate sentence or something like that.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+    // check if date is valid
+    if (!mDate.isValid()) {
+        return { msg: `\nError: Inputted date not valid.`, value: false };
+    }
+	
+	// check if time was included in the input
+	const hasTime = date.includes(':') || date.split(/[\s-]/).length > 3;
+
+	if (hasTime) {
+		return { msg: mDate.format('YYYY-MM-DD HH:mm:ss'), value: true };
```
huss, line 608: Not sure if this will stay but I would want to discuss why it is going back to a string and why must if have this format.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+
+	if (!minMoment.isValid() || !maxMoment.isValid()) {
+		msg += `\nError: Either Min Date ${minDate} or Max Date ${maxDate} is invalid (or both!).`;
+		return { msg: msg, value: false };
+	} else if (minMoment.isValid() && maxMoment.isValid()) {
+		bothValid = true;
+	}
+	
+    // dates validated now check if minDate is == maxDate
+    if (minMoment.isBefore(maxMoment) && bothValid) {
```
huss, line 523: isSameOrBefore could do both checks and I think that is fine with an appropriate msg.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+		// add error messages to the overall error message
+		msg += correctMinFormat.msg + '\n' + correctMaxFormat.msg;
+		return { msg: msg, value: false };
+	} else {
+		// set the formatted dates
+		formattedMinDate = correctMinFormat.msg;
+		formattedMaxDate = correctMaxFormat.msg;
+	}
+	
+    // validate that years are within range 0001 to the current date
```
huss, line 496: Aligning with the other comment, I don't think checking the year or length is needed. I don't think inputting a very early/late date is a mistake as it may mean all dates are valid according to the site. I think this also aligns with the create meter web page if this is done that way rather than CSV upload.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+
+/**
+ * In the validateGap function we take in the readings for the Gap and
+ * verify that it’s greater than or equal to 0.
+ * @param {Number} meter 
+ * @param {Number} rowIndex 
+ */
+function validateGap(meter, rowIndex){
+	const gapValue = Number(meter[14]);
+	if(!isNaN(gapValue)){
```
huss, line 634: formatting off

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+		if(gapValue < 0){
+		throw new CSVPipelineError(
```
huss, line 636: This is throwing rather than returning error as most of the newer checks. When changed, make sure values that are not numbers don't return true.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,22 +44,124 @@ async function uploadMeters(req, res, filepath, conn) {
...
+								egauge, mamac, metasys, obvius, and other. `;
+					throw new CSVPipelineError(msg, undefined, 500);
+				}
+			}
+
+			// verify the Gap input
+			const gapInput = meter[14];
+			if (gapInput) {
+				if (!validateGap(gapInput)) {
+					let msg = `For meter ${meter[0]} the Gap entry of ${gapInput} is invalid. Gap must be a number greater than 0.`;
```
huss, line 151: See function with comment that it is not returning a value for same here.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+ * @param {Number} meter 
+ * @param {Number} rowIndex 
+ */
+function validateGap(meter, rowIndex){
+	const gapValue = Number(meter[14]);
+	if(!isNaN(gapValue)){
+		if(gapValue < 0){
+		throw new CSVPipelineError(
+			`Invalid gap value in row ${rowIndex + 1}: GapValue="${meter[14]}". ` +
+			`Gap must be a number larger than 0.`,
```
huss, line 638: Formally this is the reading gap so it should probably say that.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+/**
+ * In the validateVariation function we take in the readings for the variation and
+ * verify that it’s greater than or equal to 0.
+ * @param {Number} meter 
+ * @param {Number} rowIndex 
+ */
+function validateVariation(meter, rowIndex){
+	const variationValue = Number(meter[15]);
+	if(!isNaN(variationValue)){
+		if(variationValue < 0)
```
huss, line 655: Please format so { is on same line as if statement.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+		{
+			throw new CSVPipelineError(
```
huss, line 657: Similar to other comment, this throws error inside check.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
+			`Invalid variation value in row ${rowIndex + 1}: VariationValue="${meter[15]}". ` +
+			`Variation Value must be a number larger than 0.`,
```
huss, line 659: Formally reading variation so should say that.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -41,22 +44,124 @@ async function uploadMeters(req, res, filepath, conn) {
...
+					let msg = `For meter ${meter[0]} the Gap entry of ${gapInput} is invalid. Gap must be a number greater than 0.`;
+					throw new CSVPipelineError(msg, undefined, 500);
+				}
+			}
+
+			// verify the Variation input
+			const variationInput = meter[15];
+			if (variationInput) {
+				if (!validateVariation(variationInput)) {
+					let msg = `For meter ${meter[0]} the Gap entry of ${variationInput} is invalid. Gap must be a number greater than 0.`;
```
huss, line 160: See other function about not returning value. Also, the msg should not be referring to gap.

src/server/services/csvPipeline/uploadMeters.js
```diff
@@ -183,4 +353,315 @@ async function getUnitId(unitName, expectedUnitType, conn) {
...
-module.exports = uploadMeters;
+
+/**
+ * Validates all boolean-like fields for a given meter row.
+ * @param {Array} meter - A single row from the CSV file.
+ * @param {number} rowIndex - The current row index for error reporting.
+ */
+function validateBooleanFields(meter, rowIndex) {
+	// all inputs that involve a true or false all being validated together.
+	const booleanFields = {
```
huss, line 364: I like defining the index via a key. I'm thinking it would be better to define them all at the start to isolate this ugly fact and to make it clearer what is being accessed since it would be by kay name. Then all the rest of the code gets rid of the index value and uses the key. This is a new change so let me know your thoughts and if you want to do this.

huss (changes requested):
Thanks to @kyle-parker-1500, @DestinyVeloz & @OscarAvilSal for this contribution. Again, I'm sorry my comments took so long. I've provided a number of comments to think about. Some are about code from earlier work as I reviewed this PR. Some are more general changes. Given this, if you don't want to address any of those you can leave a comment and OED can decide how to deal with them. Finally, I have yet to do careful testing of uploads (waiting for more finalized code) nor check that every index into the meter array is correct (didn't want you to wait while I do all of those and the method may change so doing that later may be bettert). Please let me know if anything is not clear, you have thoughts or I can help in any way.

huss:
This has not changed in a while so I'm marking it for reasignment on the new OED PR project board. If the original developers want to work on this then please add a comment here. For anyone else, contact OED to get started and be assigned this work.

PR #1587, opened 2026-03-23
Improve CSV gzip decompression safety and error reporting

Quillen09:
## Description

This pull request improves protection against gzip-based denial‑of‑service attacks
during CSV uploads by making the decompression limit explicit in the failure message.

Previously, oversized gzip uploads were rejected with a generic error that did not
indicate what limit was exceeded. This change updates the gunzip utility to include
the configured maximum decompressed size (in bytes) in the error message and adds
unit-level test coverage to prevent regressions.

## Type of change

## Checklist

- [x] You acknowledge that every person contributing to this work has signed the
      OED Contributing License Agreement and each author is listed in the Description section.

## Limitations

This change focuses on error clarity and regression coverage only. The configured
decompression limit value is unchanged, and no additional rate‑limiting or
throttling behavior has been introduced.

### DIFF
--- a/src/server/util/gunzipToFileWithLimit.js
+++ b/src/server/util/gunzipToFileWithLimit.js
@@ -1,0 +1,54 @@
+const zlib = require("zlib");
+const fs = require("fs");
+const fsp = require("fs").promises;
+const path = require("path");
+
+/**
+ * Stream gunzip -> file, while enforcing a hard cap on decompressed output bytes.
+ * Prevents gzip "zip bombs" from exhausting memory/disk.
+ *
+ * @param {string} inputPath - Path to uploaded gzip file
+ * @param {string} outputDir - Directory to write decompressed csv
+ * @param {string} outputFilename - Filename to write
+ * @param {number} maxBytes - Max allowed decompressed output size in bytes
+ * @returns {Promise<string>} - Full path of decompressed csv file
+ */
+async function gunzipToFileWithLimit(inputPath, outputDir, outputFilename, maxBytes) {
+  await fsp.mkdir(outputDir, { recursive: true });
+
+  const outPath = path.join(outputDir, outputFilename);
+
+  return new Promise((resolve, reject) => {
+    const source = fs.createReadStream(inputPath);
+    const gunzip = zlib.createGunzip();
+    const dest = fs.createWriteStream(outPath, { flags: "wx" }); // fail if file exists
+
+    let total = 0;
+
+    gunzip.on("data", (chunk) => {
+      total += chunk.length;
+      if (total > maxBytes) {
+        // Stop everything immediately
+        const err = new Error(`Decompressed size exceeds limit (max ${maxBytes} bytes)`);
+        source.destroy();
+        dest.destroy();
+        gunzip.destroy(err);
+      }
+    });
+
+    const cleanup = async (err) => {
+      try { await fsp.unlink(outPath); } catch (_) {}
+      reject(err);
+    };
+
+    source.on("error", cleanup);
+    gunzip.on("error", cleanup);
+    dest.on("error", cleanup);
+
+    dest.on("finish", () => resolve(outPath));
+
+    source.pipe(gunzip).pipe(dest);
+  });
+}
+
+module.exports = gunzipToFileWithLimit;


src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+ * Prevents gzip "zip bombs" from exhausting memory/disk.
+ *
+ * @param {string} inputPath - Path to uploaded gzip file
+ * @param {string} outputDir - Directory to write decompressed csv
+ * @param {string} outputFilename - Filename to write
+ * @param {number} maxBytes - Max allowed decompressed output size in bytes
+ * @returns {Promise<string>} - Full path of decompressed csv file
+ */
+async function gunzipToFileWithLimit(inputPath, outputDir, outputFilename, maxBytes) {
+  await fsp.mkdir(outputDir, { recursive: true });
```
huss, line 17: I'm unclear if OED creates new directories when it uploads files. If not, it would probably be more secure not to do this. Also, I'm unclear, given this seems to be a hard coded output directory, that it would not allow writing onto inappropriate directories.

Also, I wonder if the size of the zipped file should be checked. I think it may be safe to assume that if it is larger than the desired output size then the unzipped file would be larger than desired. That would stop the saving of the file. This may happen earlier but enforcing here seems a good idea.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const outPath = path.join(outputDir, outputFilename);
+
+  return new Promise((resolve, reject) => {
+    const source = fs.createReadStream(inputPath);
+    const gunzip = zlib.createGunzip();
+    const dest = fs.createWriteStream(outPath, { flags: "wx" }); // fail if file exists
```
huss, line 24: OED prefers comments on the line above code and not the same line.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+        gunzip.destroy(err);
+      }
+    });
+
+    const cleanup = async (err) => {
+      try { await fsp.unlink(outPath); } catch (_) {}
+      reject(err);
+    };
+
+    source.on("error", cleanup);
```
huss, line 44: I'm unclear on how the err is begin passed to cleanup by these. I may be missing something and happy for you to explain.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+        const err = new Error(`Decompressed size exceeds limit (max ${maxBytes} bytes)`);
+        source.destroy();
+        dest.destroy();
+        gunzip.destroy(err);
+      }
+    });
+
+    const cleanup = async (err) => {
+      try { await fsp.unlink(outPath); } catch (_) {}
+      reject(err);
```
huss, line 41: I'm wondering if this should be in a finally block. If so, would the catch be needed?

I'm not an expert here but if this code issues a reject on the first call then would the other cleanup calls necessarily get executed/completed?

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
+const zlib = require("zlib");
```
huss, line 1: This file lacks the MPL comment in all files.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+const fs = require("fs");
+const fsp = require("fs").promises;
+const path = require("path");
+
+/**
```
huss, line 6: The code is space indented but OED wants tabs.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+      reject(err);
+    };
+
+    source.on("error", cleanup);
+    gunzip.on("error", cleanup);
+    dest.on("error", cleanup);
+
+    dest.on("finish", () => resolve(outPath));
+
+    source.pipe(gunzip).pipe(dest);
```
huss, line 50: This file lacks comments inside the function and I think they should be added. This is definitely an example of code that should be explained.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+const path = require("path");
+
+/**
+ * Stream gunzip -> file, while enforcing a hard cap on decompressed output bytes.
+ * Prevents gzip "zip bombs" from exhausting memory/disk.
+ *
+ * @param {string} inputPath - Path to uploaded gzip file
+ * @param {string} outputDir - Directory to write decompressed csv
+ * @param {string} outputFilename - Filename to write
+ * @param {number} maxBytes - Max allowed decompressed output size in bytes
```
huss, line 13: I think maxBytes should have a default value that enforces a reasonable limit. I also think this function should put some max limit on the size that can be passed.

src/server/util/gunzipToFileWithLimit.js
```diff
@@ -0,0 +1,54 @@
...
+ * @returns {Promise<string>} - Full path of decompressed csv file
+ */
+async function gunzipToFileWithLimit(inputPath, outputDir, outputFilename, maxBytes) {
```
huss, line 16: I don't see this function used anywhere so I'm unclear on how it is actually used to enforce limits. For example, how does it relate to the gunzip usage in src/server/routes/csv.js and how it is done there.

huss (changes requested):
@Quillen09 Thank you for your first contribution to OED. I have made some comments to consider. Also, the description discussed unit tests but I don't see any tests. Could you explain or add them, esp. given this is security related code. Finally, can you explain your statement in the description that OED already enforced file limits but did not return a clear error message. I think it should be part of the record and an understanding of how this relates to that. Please let me know if anything is not clear, you have thoughts or would like any assistance.

huss:
Sorry to add something after I submitted the review. I believe there are several other aspects of potential malicious behavior of gzip files to address. I suspect some are fairly straightforward to implement. You note some in the description. Given this, I waned to know if you planned to address those. If not, I will have to decide if this should be merged or if it should wait until further enhancements are made.

huss:
@Quillen09 Thank you for your work on this issue. I know it is after the time you originally intended to work on OED. Given OED wants to finish this up & it has been several months since my last comment, I'm going to get someone else to complete this work unless you indicate by 6/15/26 that you intend to work on it and it is completed in the near future.

PR #1588, opened 2026-03-27
Issue 1439 Fix

### LINKED ISSUE #1439: enhance meter test code

src/server/test/web/meters.js has the comment:

```
// TODO These tests are not as good as they should be now that information on
// meters is returned to all users. They should be updated.
```

src/server/test/db/meterTests.js has the comment:

```
//  TODO add 2 new unit parameters for meters.
```

src/server/test/web/csvPipelineTest.js has the comment:

```
// TODO It would be nice to make this use the code in src/server/test/db/meterTests.js and make
// all meter tests use one common function fo meter comparison.
```

These are somewhat correct but more work is needed.

This following tests should be added to src/server/test/web/meters.js:

- [ ] src/server/models/Meter.js in the constructor shows there are a number of fields/values that are not checked. They need to be added to ``expectMetersToBeEquivalent`` and as values in the tests that specify meter data.
- [ ] For a few values, ``Meter.convertUnitValue`` is called to convert the value between null and -99 for JS vs database value. A test should be created to verify that the -99 value is correctly returned to know the conversion works both ways.
- [ ] ``insert`` and ``update`` use ``Meter.convertUnitValue`` to make sure the values are correct. Tests should be created where the changed values are different from the desired values and the returned values are the correct.
- [ ] The tests in meterTests.js should be rolled into meters.js and updated for new value/test as noted in items above this one. meters.js tests the route which then uses the model and database. OED has been doing tests via the route since any failure on the server side is found. The negative is you don't know where on the server the failure occurs but since they are rare this seems fine given it simplifies the test code and the source can be located once a test fails. Once complete, meterTests.js can be removed so its TODO is indirectly resolved.
- [ ] The test code checks ``insert`` and getting a single meter id. It would be good if it checked ``update`` and the other get functions that were not added when meterTests.js was rolled into this file.
- [ ] src/server/test/web/csvPipelineTest.js should be modified to remove ``compareMeters`` and import ``expectMetersToBeEquivalent`` from src/server/test/web/meters.js and use that instead. This addresses the TODO in this file but does it by using a different, common test function.

Once all are complete the TODO comments can be removed and the TODO design document updated to reflect it is complete.

None

### Additional context

None

tazmeenahhmed:
# Description

Implemented changes described in issue #1439 which are the following:

- Removed function compareMeters from file csvPipelineTest.js and replaced all instances of compareMeters with an imported expectMetersToBeEquivalent function from meters.js
- Added tests in meters.js that check if Meter.js functions getByIdentifier and getAll work properly
- Rolled tests from meterTests.js into meters.js and deleted meterTests.js
- Added tests in meters.js to check that Meter.js functions insert() and update() return correct values when Meter.convertUnitValue is called within them
- Added a test in meters.js that checks if Meter.convertUnitValue returns -99 when passed with null
- Added areaUnit and readingFrequency to meters.js function expectMetersToBeEquivalent to match mocha tests

Fixes #1439 

## Type of change

## Checklist

- [X] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [X] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.


src/server/models/Meter.js
```diff
@@ -9,7 +9,7 @@ const Unit = require('./Unit');
 const { log } = require('../log');
 const sqlFile = database.sqlFile;
 
-class Meter {
+class Meter {
```
huss, line 12: There is a space added to the end of this line. OED prefers not to do that and this is the only change in the file. Could you undo this so no changes are shown in the file.

src/server/test/web/meters.js
```diff
@@ -90,6 +89,9 @@ function expectMetersToBeEquivalent(meters, length, isAdmin, unit) {
 			expect(meter).to.have.property('startTimestamp', null);
 			expect(meter).to.have.property('endTimestamp', null);
 			expect(meter).to.have.property('previousEnd', null);
+			expect(meter).to.have.property('areaUnit', Unit.areaUnitType.METERS);
+			expect(meter).to.have.property('readingFrequency', null);
+
```
huss, line 94: Please remove the blank line.

src/server/test/web/meters.js
```diff
@@ -71,6 +68,8 @@ function expectMetersToBeEquivalent(meters, length, isAdmin, unit) {
 			expect(meter).to.have.property('startTimestamp', '0001-01-01 23:59:59');
 			expect(meter).to.have.property('endTimestamp', '2020-07-02 01:00:10');
 			expect(meter).to.have.property('previousEnd', '2020-03-05T13:15:13.000Z');
+			expect(meter).to.have.property('areaUnit', Unit.areaUnitType.METERS);
+			expect(meter).to.have.property('readingFrequency', 'PT13H57M19S');
 		} else {
```
huss, line 73: src/server/models/Meter.js has these params that I don't see being checked:

```
	 * @param minVal inclusive minimum acceptable reading value (won't be rejected)
	 * @param maxVal inclusive maximum acceptable reading value (won't be rejected)
	 * @param minDate inclusive earliest acceptable date (won't be rejected)
	 * @param maxDate inclusive latest acceptable date (won't be rejected)
	 * @param maxError maximum number of errors to be reported, ignore the rest
	 * @param disableChecks disable checks on meter for conditionSet 
```

I think they are all admin only so null for non-admins.

src/server/test/web/meters.js
```diff
@@ -261,3 +263,215 @@ mocha.describe('meters API', () => {
 		expect(res).to.have.status(500);
 	});
 });
+
+mocha.describe('Meter model', () => {
+	let unitId;
```
huss, line 268: It seems only the last test uses the unitId set in the beforeEach. Given this, I'm wondering about moving this code to that test.

src/server/test/web/meters.js
```diff
@@ -261,3 +263,215 @@ mocha.describe('meters API', () => {
...
+		actualMeters.sort((a, b) => a.id - b.id);
+		expectedMeters.sort((a, b) => a.id - b.id);
+
+		expect(expectedMeters.length).to.be.equal(actualMeters.length);
+		for (let i = 0; i < expectedMeters.length; ++i) {
+			expectMetersToBeEquivalent(expectedMeters[i], actualMeters[i]);
+		}
+	});
+
+	mocha.it('can get meter by identifier', async () => {
```
huss, line 435: I like this and the next new tests. Two thoughts about these and they may apply to other tests:

- A few values are null/undefined. I wonder if some tests should use other values to verify that works.
- The other values discussed in another comment that are part of a Meter do not have explicit values for the meter being created. That should be done for the tests.

src/server/test/web/meters.js
```diff
@@ -261,3 +263,215 @@ mocha.describe('meters API', () => {
...
+				'MeterB', 'notes 1', 35.0, true, true, '01:01:25', '00:00:00', 5, 0, 1, 'increasing', false,
+				1.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 02:12:00', unitA.id, unitA.id,
+				Unit.areaUnitType.METERS, undefined);
+			const meterC= new Meter(undefined, 'MeterC', null, true, true, Meter.type.MAMAC, null, gps, 
+				'MeterC', 'notes 1', 35.0, true, true, '01:01:25', '00:00:00', 5, 0, 1, 'increasing', false,
+				1.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 02:12:00', unitA.id, unitA.id,
+				Unit.areaUnitType.METERS, undefined);
+
+			await Promise.all([meterA, meterB, meterC].map(meter => meter.insert(conn)));
+			const allExpectedMeters = await Meter.getAll(conn);
```
huss, line 463: I think Expected and Actual should be the other way around. Normally expected is what the test knows is correct and actual is what you get back. Maybe this is based on another test but I would still like them flipped (if you agree).

src/server/test/web/meters.js
```diff
@@ -261,3 +263,215 @@ mocha.describe('meters API', () => {
...
+	mocha.it('can get all meter where unitId is not null', async () => {
+		const meterA = new Meter(undefined, 'MeterA', null, true, true, Meter.type.MAMAC, null, gps,
+			'MeterA', 'notes 1', 35.0, true, true, '01:01:25', '00:00:00', 5, 0, 1, 'increasing', false,
+			1.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 02:12:00', unitA.id, unitA.id,
+			Unit.areaUnitType.METERS, undefined);
+		const meterB = new Meter(undefined, 'MeterB', null, true, true, Meter.type.MAMAC, null, gps,
+			'MeterB', 'notes 2', 35.0, true, true, '01:01:25', '00:00:00', 5, 0, 1, 'increasing', false,
+			1.5, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 02:12:00', unitB.id, unitB.id,
+			Unit.areaUnitType.METERS, undefined);
+		const meterC = new Meter(undefined, 'Meter C', null, true, true, Meter.type.MAMAC, null);
```
huss, line 421: This is a good test for the default value of -99 for the unit. It might be good to add another meter that explicitly sets the unit to -99 to verify that is the same.

src/server/test/web/meters.js
```diff
@@ -261,3 +263,215 @@ mocha.describe('meters API', () => {
...
+			const allActualMeters = [meterA, meterB, meterC];
+			allExpectedMeters.sort((a,b) => a.id - b.id);
+			allActualMeters.sort((a,b) => a.id - b.id);
+
+			expect(allExpectedMeters.length).to.be.equal(allActualMeters.length);
+			for (let i = 0; i < allExpectedMeters.length; ++i) {
+				expectMetersToBeEquivalent(allExpectedMeters[i], allActualMeters[i]);
+			}
+	});
+});
```
huss, line 473: Overall, the test coverage is good for all functions in Meter. I noticed two that are not tested and it would be good to do those so this is complete:

- existsByName
- makeMeterDataValid (do other one of this type)

Also, the pipeline tests do check about admin vs not (I think) but I think that having a test in this file that adds the info and gets as a non-admin to be sure the correct values are nulled out would be a good idea since it is directly related to the Meter interface.

huss (changes requested):
@tazmeenahhmed Thank you for another contribution and addressing this issue. Overall, the changes look good. I've made a few comments to consider. Please let me know if anything is not clear, you have thoughts or would like my help.

huss:
@tazmeenahhmed Were you done with updates? I noticed the tests failed but that might relate to issue #1630. I will check your changes out and see if it is not a real issue with your code if you let me know you think it is ready.

tazmeenahhmed:
I finished all the updates, and the code is ready for checking! I am unsure if this helps, but when I ran the tests specifically for the file I was working on (meters.js), it seemed to pass with no issues.

src/server/test/web/meters.js
```diff
@@ -255,10 +335,276 @@ mocha.describe('meters API', () => {
...
+
+		const meterPostInsert = await Meter.getByID(meterPreInsert.id, conn);
+		expect(meterPostInsert.unitId).to.equal(-99);
+		expect(meterPostInsert.defaultGraphicUnit).to.equal(-99);
+	});
+
+	mocha.it('returns -99 when unitID & defaultGraphicUnit is updated to -99', async() => {
+		let unitId;
+		const conn = testDB.getConnection();
+		mocha.beforeEach(async () => {
```
huss: I wondered about having a beforeEach() inside an individual test. I also noticed that 4 tests were consistently failing on my machine when I ran the full suite. It took a little while to realize that these were related. The bottom line is removing the beforeEach made the issue go away. So, I think this entire beforeEach should go but the code inside should stay. Thus, the creation of the unit will be regular code inside the mocha.it. It would be nice to remove the ``let unitId;`` and put a ``const unitId`` where it is set.

src/server/test/web/meters.js
```diff
@@ -231,17 +261,67 @@ mocha.describe('meters API', () => {
 		}
 	});
 
+	mocha.it('certain parameters are correctly null on non-admin access', async () => {
+		const conn = testDB.getConnection();
+		const password = 'password';
```
huss: I realize this is done in other places but using 'password' is not the best since it matches the default admin test user. Doing this can lead to an incorrect login if the code is wrong. I've started to advocate for using something along the lines of 'exportpassword'.

src/server/test/web/meters.js
```diff
@@ -231,17 +261,67 @@ mocha.describe('meters API', () => {
...
+		const hashedPassword = await bcrypt.hash(password, 10);
+		const nonAdmin = new User(undefined, 'export@example.com', hashedPassword, User.role.EXPORT);
+		await nonAdmin.insert(conn);
+		nonAdmin.password = password;
+
+		let userres = await chai.request(app).post('/api/login')
```
huss: It would be better if this was camelcase (userRes).

src/server/test/web/meters.js
```diff
@@ -231,17 +261,67 @@ mocha.describe('meters API', () => {
...
+		await new Meter(undefined, 'Meter 1', '1.1.1.1', true, true, Meter.type.MAMAC, '+02', gps,
+			'Identified 2', 'notes 1', 20.0, true, true, '01:01:25', '05:05:05', 5.1, 7.3, 1, 'increasing', false,
+			1.0, '0001-01-01 23:59:59', '2020-07-02 01:00:10', '2020-03-05 13:15:13', unitId, unitId,
+			Unit.areaUnitType.METERS, '13:57:19', Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER, '1970-01-01T00:00:00.000Z', 
+			'6970-01-01T00:00:00.000Z', 75, Unit.disableChecksType.REJECT_ALL).insert(conn);
+		
+		const res = await chai.request(app).get('/api/meters').set('token', token);
+		const meter = res.body[0];
+
+		expect(meter.name).to.equal(null);
```
huss: I think it might be better to use expectMetersToBeEquivalent where isAdmin is false to check these values rather than duplicate the checks. Or is there an issue with doing that? This would also check that the other values are correct.

src/server/test/web/meters.js
```diff
@@ -138,6 +159,7 @@ mocha.describe('meters API', () => {
 		expectMetersToBeEquivalent(res.body, 4, false, unitId);
 	});
```
huss, line 157: Extra blank line.

src/server/test/web/meters.js
```diff
@@ -13,6 +13,7 @@ const Point = require('../../models/Point');
 const moment = require('moment-timezone');
 const gps = new Point(90, 45);
 const Unit = require('../../models/Unit');
+const { meterTimezone } = require('../../services/meterTimezone');
```
huss: I'm unclear where this import is used. I see ``meterPreInsert.meterTimezone = 'GMT';`` below but I don't think that is the same thing. Can you help me understand?

huss:
> I finished all the updates, and the code is ready for checking! I am unsure if this helps, but when I ran the tests specifically for the file I was working on (meters.js), it seemed to pass with no issues.

See another comment on the failed tests. It describes how the new code was causing 4 test failures when the full suite was run.

huss (changes requested):
@tazmeenahhmed Thank you for your continued work on this issue. Based on your recent work I resolved all the old comments. I have made a handful of new ones based on the changes. Please let me know if anything is not clear, you have thoughts/questions or I can help.

src/server/test/web/meters.js
tazmeenahhmed: I believe this was added when I was changing null/undefined values for meters and was in the process of figuring out what values were needed to fill out the parameter 'meterTimezone'. The import ended up not being used (& meterPreInsert.meterTimezone is indeed not the same thing), but I forgot to remove it & will implement that in my changes.

src/server/test/web/meters.js
tazmeenahhmed: Just some thoughts, but using expectMetersToBeEquivalent in this test makes it more similar to the 'Export role and Obvius role' test as both tests to my understanding are now checking admin data is not accessible/visible to non-admin users. Then, would it be necessary to have two tests with the same purpose or is there a reason to keep the test 'certain parameters are correctly null on non-admin access' around?

huss: I looked at this and I think you are right. The other test seems to verify that the values are null. So, I guess this test can go. Sorry you had to do it and then remove it.

tazmeenahhmed:
Hi @huss, it's been some time since I finished making my changes; would it be possible to get a review on them?

huss:
> Hi @huss, it's been some time since I finished making my changes; would it be possible to get a review on them?

@tazmeenahhmed Sorry. Somehow I missed that it was ready to be reviewed. Thanks for letting me know. I'll try to get to this soon.

huss (approved):
@tazmeenahhmed Thank you for the updated code. Review and testing found this works as desired. Congratulations on another accepted contribution.

PR #1593, opened 2026-03-31
Implement server-side session invalidation using token_invalid_before

Oykunle:
Description

This PR implements server-side session invalidation for JWT authentication using a token_invalid_before timestamp stored on the users table.

Previously, issued JWTs remained valid until expiration, even after logout. This meant that a token could still be reused after the user had logged out.

This change adds a server-side invalidation mechanism by comparing the token’s iat (issued-at timestamp) against the user’s token_invalid_before value. Any token issued before that timestamp is rejected.

---

## Type of change

---

### Database
- Added `token_invalid_before` column to the `users` table
- Implemented migration under `1.0.0 → 2.0.0` as requested

### Authentication / Security
- Updated authentication logic to reject tokens where:

```
token.iat < token_invalid_before
```

- Ensures invalidated tokens cannot be reused
- Applies validation consistently across middleware and verification routes

### Logout
- Refactored logout into `/api/login/logout`
- Logout updates `token_invalid_before` to invalidate all existing tokens
- Ensures idempotent behavior (repeated logout is safe)

### Code Improvements
- Improved consistency of authentication checks across routes
- Reduced duplication in token validation logic

---

## Tests

Added automated tests to validate session invalidation behavior:

- Token is valid before logout
- Token becomes invalid after logout
- Invalidated token is rejected on protected routes
- Repeated logout calls succeed safely
- Logout validation rejects malformed requests

All tests pass successfully in the test environment.

---

## Security Impact

- Prevents reuse of old or stolen JWT tokens after logout
- Adds server-side control over token validity
- Improves session security without changing authentication flow

---

## Limitations

- Invalidates all tokens globally per user (not per-device/session)
- Requires database migration for existing deployments


src/server/routes/logout.js
```diff
@@ -0,0 +1,25 @@
...
+
+const router = express.Router();
+
+/**
+ * Invalidates previously issued tokens for the authenticated user.
+ */
+router.post('/', authMiddleware, async (req, res) => {
+	const conn = getConnection();
+	try {
+		await User.invalidateTokensBeforeNow(req.decoded.data, conn);
```
huss: I'm concerned about the authentication of this route for the following reasons:

- It uses authMiddleware that has a comment stating it is only used in src/server/routes/authenticator.js and not elsewhere in the the code. I'm not sure that is validating the user. I'm wondering if a variant of isTokenAuthorized (without role) in src/server/util/userRoles.js is along the lines of what you are looking for.
- This takes the req and accepts the id provided. Given it is unclear the user is validated, it may be possible for someone to maliciously or incorrectly logout any user. This may not be needed but the user name could be sent and the id looked up (assuming that is easy info to get).

I also would like to understand what happens if a user's token is already invalidated and then they try to log out. Does it cleanly fail or some error? I think a unit test for this would be good.

Every route should validate any values used. This also needs to be added to route authentication per PR #1594.

src/server/routes/logout.js
```diff
@@ -0,0 +1,25 @@
...
+		res.status(200).json({ success: true, message: 'Logged out successfully.' });
+	} catch (err) {
+		res.status(500).json({ success: false, message: 'Internal Server Error' });
```
huss: Telling where the error occurred seems like a good idea and the message should probably only say the logout failed. Also, errors should be logged with the actual error and what it was doing. Finally, you should not be sending the response codes but using the standard success/failure methods.

src/server/routes/logout.js
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: OED generally has a single route file for each model/area. I'm thinking the login.js route and this route could be combined into a single route (not sure the best name). This may mean the default route ('/') should be replaced with named ones ('/login', 'logout') and all uses updated.

src/server/models/User.js
```diff
@@ -54,7 +63,14 @@ class User {
 	 */
 	static async getByUsername(username, conn) {
 		const row = await conn.oneOrNone(sqlFile('user/get_user_by_username.sql'), { username: username });
-		return row === null ? null : new User(row.id, row.username, row.password_hash, row.role, row.note);
+		return row === null ? null : new User(
```
huss: Seeing the need to update the returned row in two places and slight differences between them made me realize that this model does not have a mapRow function as in other cases. I think it should be added.

src/server/routes/verification.js
```diff
@@ -30,13 +32,29 @@ router.post('/', (req, res) => {
...
 		const token = req.body.token;
-		jwt.verify(token, secretToken, err => {
-			if (err) {
-				res.status(401).json({ success: false, message: 'Failed to authenticate token.' });
-			} else {
-				res.status(200).json({ success: true });
-			}
-		});
+		jwt.verify(token, secretToken, async (err, decoded) => {
+        	if (err) {
```
huss: This has a mix of space and tab indents. OED wants tabs.

src/server/routes/verification.js
```diff
@@ -30,13 +32,29 @@ router.post('/', (req, res) => {
...
+        			const conn = getConnection();
+        			const user = await User.getByID(decoded.data, conn);
+
+        			const tokenIssuedAt = decoded.iat;
+        			const invalidBefore = user.tokenInvalidBefore
+        				? Math.floor(new Date(user.tokenInvalidBefore).getTime() / 1000)
+        				: 0;
+
+        			if (tokenIssuedAt < invalidBefore) {
+        				return res.status(401).json({ success: false, message: 'Token invalidated.' });
```
huss: See other comment on how to return messages, the actual message and error logging. I realize it was done this way earlier but should be cleaned up unless there is a reason not to.

src/server/routes/verification.js
```diff
@@ -30,13 +32,29 @@ router.post('/', (req, res) => {
...
-			} else {
-				res.status(200).json({ success: true });
-			}
-		});
+		jwt.verify(token, secretToken, async (err, decoded) => {
+        	if (err) {
+        		res.status(401).json({ success: false, message: 'Failed to authenticate token.' });
+        	} else {
+        		try {
+        			const conn = getConnection();
```
huss: Esp. as security code, it should have careful commenting.

src/server/routes/verification.js
```diff
@@ -30,13 +32,29 @@ router.post('/', (req, res) => {
...
+        			const user = await User.getByID(decoded.data, conn);
+
+        			const tokenIssuedAt = decoded.iat;
+        			const invalidBefore = user.tokenInvalidBefore
+        				? Math.floor(new Date(user.tokenInvalidBefore).getTime() / 1000)
+        				: 0;
+
+        			if (tokenIssuedAt < invalidBefore) {
```
huss: Is this a safe way to compare dates? I'm also thinking about the zero value if the takenInvalidBefore has an issue.

src/server/routes/verification.js
```diff
@@ -30,13 +32,29 @@ router.post('/', (req, res) => {
...
-		});
+		jwt.verify(token, secretToken, async (err, decoded) => {
+        	if (err) {
+        		res.status(401).json({ success: false, message: 'Failed to authenticate token.' });
+        	} else {
+        		try {
+        			const conn = getConnection();
+        			const user = await User.getByID(decoded.data, conn);
+
+        			const tokenIssuedAt = decoded.iat;
```
huss: It seems very similar logic is used here and authenticator.js. Should it be centralized, esp. given it is critical it be consistent and correct for security?

huss (changes requested):
@Oykunle Thank you for working on this. I've made some comments within the code to consider. Here are some comments that cannot easily be placed in code comments:

- ~The PR description incorrectly links to PR 11. I think this whole sentence should be removed.~
- ~The PR mentions migrations. This needs to be done before this can be merged. Could you either do them or let me know you want them done by someone else.~
- I'm unclear how this relates to src/client/app/redux/api/authApi.ts and the current login/logout done by users. Could you help me understand?
  - I need to check this over carefully but I see the use in the file now.
- ~This is a security feature. As such, there should be automated tests to validate all aspects to verify it works correctly.~

I'm going to need to review this again after these comments are addressed. Please let me know if anything is not clear, you have thoughts or I can help.

Oykunle:
@huss,

Thank you for the detailed review. I’m going to address the requested changes in the code first and then update the PR for re-review.

I believe I understand the main concerns you raised. If I run into anything unclear while making the changes, I’ll follow up with a specific question.

Oykunle:
@huss 

I’ve pushed an updated revision addressing the requested changes.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,124 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const { expect } = require('chai');
+const { chai, mocha, app, testUser } = require('../common');
+const { HTTP_CODE } = require('../../util/readingsUtils');
```
huss, line 9: As noted, this will need to be integrated with the upcoming PR on http codes. This PR will probably wait until that can be done before it is merged.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+
+mocha.describe('Session Invalidation Security', () => {
```
huss, line 15: I appreciate the testing. This is a note that this new route needs to be added to the route testing in PR #1593 and the parameter testing of routes in this folder (some are done here but a comprehensive look/validation is needed). It can be done by others after this PR but I wanted to document.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+	const LOGIN_ENDPOINT = '/api/login';
+	const LOGOUT_ENDPOINT = '/api/login/logout';
+	const VERIFY_ENDPOINT = '/api/verification';
+	const PROTECTED_ENDPOINT = '/api/users';
+
+	async function loginAndGetToken() {
```
huss: This appears to mix a function to get the login token with a test. I suggest that the getting of the token be done with a beforeEach and the test be separate within a mocha.it.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+
+	mocha.it('should verify a valid token before logout', async () => {
+		const token = await loginAndGetToken();
+
+		const verifyRes = await chai.request(app)
+			.post(VERIFY_ENDPOINT)
+			.send({ token });
+
+		expect(verifyRes).to.have.status(HTTP_CODE.OK);
+		expect(verifyRes.body).to.deep.equal({ success: true });
```
huss: Is there a more precise way to do this? What I'm thinking about is if anything inside the response has a JSON success then the test will pass.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+	});
+
+	mocha.it('should invalidate a token after logout', async () => {
+		const token = await loginAndGetToken();
+
+		const logoutRes = await chai.request(app)
```
huss, line 58: While not likely to fail, it would be a little stronger if you validated the token works before the logout.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,124 @@
...
+	const LOGIN_ENDPOINT = '/api/login';
+	const LOGOUT_ENDPOINT = '/api/login/logout';
+	const VERIFY_ENDPOINT = '/api/verification';
+	const PROTECTED_ENDPOINT = '/api/users';
+
+	async function loginAndGetToken() {
+		const res = await chai.request(app)
+			.post(LOGIN_ENDPOINT)
+			.send({
+				username: testUser.username,
```
huss: I like that most tests are with the admin user. I'm thinking one test with another role to make sure token invalidation works would be good. An even better option done in some other places is to loop over all user roles to try each one for invalidation (not all the other tests). This would also do admin.

src/server/routes/authenticator.js
```diff
@@ -16,40 +16,39 @@ const { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, TOKEN_MAX_LENGTH, USERNAME_MIN
 
 /**
  * Middleware function to force a route to require authentication
- * Verifies the request's token against the server's secret token
- * It is used within this file but not by other parts of OED.
+ * Verifies the request's token against the server's secret token.
  */
```
huss, line 22: I think you removed the comment about it not being used elsewhere but now I cannot find any uses outside this file. Am I missing something?

src/server/routes/authenticator.js
```diff
@@ -103,10 +102,50 @@ async function verifyCredentials(username, password, returnUser = false) {
...
+			if (err) {
+				reject(err);
+			} else {
+				resolve(payload);
+			}
+		});
+	});
+
+	const conn = getConnection();
+	const user = await User.getByID(decoded.data, conn);
```
huss, line 128: There used to be a comment associated with this code about why it ran. Should it be added back?

Can the user not exist or does the jwt.verify guarantee it isn't forged?

src/server/routes/logout.js
huss: I see the new subroute but the file name is the same and the default route for login is still there. Can you explain why?

src/server/routes/verification.js
```diff
@@ -26,17 +29,26 @@ router.post('/', (req, res) => {
...
+
+		verifyActiveTokenAndGetUser(token)
+			.then(() => {
+				res.json({ success: true });
+			})
+			.catch(error => {
+				if (error.code === 'TOKEN_INVALIDATED') {
+					res.status(401).json({ success: false, message: 'Token invalidated.' });
+				} else if (error.message === 'No data returned from the query.') {
+					res.status(401).json({ success: false, message: 'User does not exist in database.' });
```
huss: I'm a little worried about this message. OED is trying hard not to allow someone to find out if a user is valid to avoid attacks. This seems it might give a way to do that. What do you think?

If this stays, I'm curious about why this is not logged but it is below.

src/server/routes/verification.js
```diff
@@ -26,17 +29,26 @@ router.post('/', (req, res) => {
...
-			}
-		});
+
+		verifyActiveTokenAndGetUser(token)
+			.then(() => {
+				res.json({ success: true });
+			})
+			.catch(error => {
+				if (error.code === 'TOKEN_INVALIDATED') {
+					res.status(401).json({ success: false, message: 'Token invalidated.' });
```
huss, line 44: 1. Hardcoded response code.
2. This is still using a direct status and success/failure functions. Could you explain the reason it is needed?

src/server/routes/verification.js
huss: I think this moved to authenticator.js but I don't see commenting with it.

src/server/routes/verification.js
huss: I think this moved to authenticator.js but I don't see any addressing of this comment. Can you help me out with this?

src/server/routes/verification.js
huss: Relates to another new comment on same idea so leaving until resolved.

src/server/models/User.js
huss: I see the new mapping but getAll does not seem to use it. Thoughts?

src/server/models/User.js
```diff
@@ -54,7 +72,7 @@ class User {
 	 */
 	static async getByUsername(username, conn) {
 		const row = await conn.oneOrNone(sqlFile('user/get_user_by_username.sql'), { username: username });
-		return row === null ? null : new User(row.id, row.username, row.password_hash, row.role, row.note);
+		return row === null ? null : User.mapRow(row);
```
huss: I think it would be better to add the null logic to your new mapRow and just mapRow them here without a conditional.

src/server/routes/logout.js
huss: I need to check this out, esp. after any other changes.

src/server/migrations/1.0.0-2.0.0/sql/users/add_token_invalid_before.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+ADD COLUMN IF NOT EXISTS token_invalid_before TIMESTAMP;
```
huss: Seeing this made me think about if this needed a default value and the related question of what happens with a new user that has never logged in. Could you help me understand the impact of having no entry for some users? Thanks, that way I don't have to figure it out.

huss (changes requested):
@Oykunle Thank you for the updated code. I tried to resolve/comment on every previous comment. I also made some new ones to consider. Here are some other comments:
- Have you done any testing to see that this all works with the auto timeout of tokens by OED?
- There are parts that I still need to carefully review given this is security related. I wanted to get this back to you so I'm delaying that, esp. if more changes may be made.
- I did not run the new test cases since I need to isolate and add the new DB column. I'll do that later.

As always, let me know if anything is not clear or your have thoughts.

src/server/routes/verification.js
Oykunle: The route currently following the existing response pattern already used here. Since you mentioned it will likely need to align with the HTTP codes work, I held off changing it for now so it can be updated consistently once that is ready.

Oykunle:
Hi Steve,

I checked the existing timeout behavior in the code. Tokens are currently issued with expiresIn: 86400 in src/server/routes/login.js, so the normal JWT auto-timeout is still handled by jwt.verify() before the new invalidation check runs.

That means expired tokens should still fail through the normal JWT path, while non-expired tokens then go through the additional token_invalid_before check.

I have not yet added a separate test for that interaction. If you would like, I can add a targeted test using a short-lived token to verify the timeout and invalidation behaviors remain compatible.

Thanks,

Oye

{ lots of quoted text removed by @huss }

huss:
> I checked the existing timeout behavior in the code. Tokens are currently issued with expiresIn: 86400 in src/server/routes/login.js, so the normal JWT auto-timeout is still handled by jwt.verify() before the new invalidation check runs. That means expired tokens should still fail through the normal JWT path, while non-expired tokens then go through the additional token_invalid_before check. 

That is what I hoped would happen but wanted verification. The important item is it is checking the token for verify before the expiration check. This should be fine.

> I have not yet added a separate test for that interaction. If you would like, I can add a targeted test using a short-lived token to verify the timeout and invalidation behaviors remain compatible.

In an ideal world that would be nice. You don't have to add it but if you do please make the timeout very short so the test will run quickly.

Also, please let me know when all the comments are addressed so I know to look at this again.

Oykunle:
Hi @huss,
I’ve addressed the review comments and pushed an updated revision.

Updates include:
	•	Fixed race condition using TIMESTAMP 'epoch' baseline
	•	Refactored authentication flow and user mapping
	•	Added test for JWT expiration alongside session invalidation

All tests are passing locally.

Please let me know if anything still needs adjustment.

src/server/routes/authenticator.js
```diff
@@ -15,41 +15,45 @@ const { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, TOKEN_MAX_LENGTH, USERNAME_MIN
...
- * Verifies the request's token against the server's secret token
- * It is used within this file but not by other parts of OED.
+ * Middleware function to require authentication on protected routes.
+ * Verifies the request's token, ensures the user exists, and checks that
+ * the token has not been invalidated.
+ *
+ * This middleware is primarily used within this file and applied to
+ * routes that require authentication.
  */
+
```
huss: There should not be a blank line between the JSDoc and the function.

src/server/routes/login.js
```diff
@@ -69,4 +69,42 @@ router.post('/', credentialsRequestValidationMiddleware, async (req, res) => {
 	}
 });
 
+/**
+ * Logs out the authenticated user by invalidating previously issued tokens.
+ */
+router.post('/logout', async (req, res) => {
```
huss, line 75: This route has no middleware for who can do it. OED wants something for every route now or a very clear comment and reason on why not. Many routes use optionalAuthMiddleware so please look into that to see if it would work.

I previously put:

> This takes the req and accepts the id provided. Given it is unclear the user is validated, it may be possible for someone to maliciously or incorrectly logout any user. This may not be needed but the user name could be sent and the id looked up (assuming that is easy info to get).

It may be possible that verifyActiveTokenAndGetUser helps with this but I would like you to explicitly explain why. Comments in the code about this are desirable.

Also, if this is merged before PR #1594 then a comment needs to be added there that this route needs to be added/tested. If after, then it needs to be addressed in this PR. This replaces part of a previous comment.

src/server/routes/login.js
```diff
@@ -69,4 +69,42 @@ router.post('/', credentialsRequestValidationMiddleware, async (req, res) => {
 	}
```
huss, line 69: I cannot readily add a comment on the unchanged lines above so doing here:

1. The default route still exists. A previous comment had:

> This may mean the default route ('/') should be replaced with named ones ('/login', 'logout') and all uses updated.

I still think this is a good idea since I see no reason the login would be the obvious default and OED is moving away from them. Doing /login would be parallel to the current /logout. If you feel otherwise then please comment.

2. I previously commented that a good name might be needed for this file. I think the current one of login.js is a misnomer. Maybe loginLogout.js but I'm open to other ideas.

src/server/routes/authenticator.js
```diff
@@ -15,41 +15,45 @@ const { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, TOKEN_MAX_LENGTH, USERNAME_MIN
...
 
 /**
- * Middleware function to force a route to require authentication
- * Verifies the request's token against the server's secret token
- * It is used within this file but not by other parts of OED.
+ * Middleware function to require authentication on protected routes.
+ * Verifies the request's token, ensures the user exists, and checks that
+ * the token has not been invalidated.
+ *
+ * This middleware is primarily used within this file and applied to
```
huss: I looked and could not find any usage outside this file. Should it go back to saying it is only used in this file or do you think otherwise?

src/server/migrations/1.0.0-2.0.0/sql/users/add_token_invalid_before.sql
huss: In the SQL to create the table it has ``NOT NULL DEFAULT TIMESTAMP 'epoch'``. I think that needs to be here so current sites have the same DB setup.

src/server/routes/logout.js
huss: These are in new comments as needed so resolving.

src/server/routes/logout.js
huss: Replaced with a new comment so resolving.

src/server/routes/verification.js
```diff
@@ -26,17 +29,27 @@ router.post('/', (req, res) => {
...
-			if (err) {
-				res.status(401).json({ success: false, message: 'Failed to authenticate token.' });
-			} else {
-				res.status(200).json({ success: true });
-			}
-		});
+
+		verifyActiveTokenAndGetUser(token)
+			.then(() => {
+				res.json({ success: true });
```
huss, line 40: I previously commented:

``See other comment on how to return messages, the actual message and error logging. I realize it was done this way earlier but should be cleaned up unless there is a reason not to.``

I don't see any comment or action on this so opening a new comment.

src/server/routes/verification.js
huss: resolved for new comment.

src/server/routes/authenticator.js
```diff
@@ -103,10 +107,53 @@ async function verifyCredentials(username, password, returnUser = false) {
...
+		const parsedDate = new Date(user.tokenInvalidBefore);
+		if (!isNaN(parsedDate.getTime())) {
+			invalidBefore = Math.floor(parsedDate.getTime() / 1000);
+		} else {
+			log.error(`Invalid tokenInvalidBefore value for user ${user.id}`);
+		}
+	}
+
+	const tokenIssuedAt = decoded.iat;
+	if (tokenIssuedAt <= invalidBefore) {
```
huss, line 144: I previously commented:

> I think this moved to authenticator.js but I don't see any addressing of this comment. Can you help me out with this?
>> I think this moved to authenticator.js but I don't see any addressing of this comment. Can you help me out with this?

I need you to comment on this to explain.

src/server/routes/authenticator.js
```diff
@@ -103,10 +107,53 @@ async function verifyCredentials(username, password, returnUser = false) {
...
+		});
+	});
+
+	// jwt.verify confirms the token signature is valid, but it does not guarantee
+	// the referenced user still exists in the database. The user may have been
+	// deleted after the token was issued, so OED must still verify the user record.
+	const conn = getConnection();
+	const user = await User.getByID(decoded.data, conn);
+
+	let invalidBefore = 0;
```
huss, line 131: Interesting here you use 0 but in the DB you default to the Unix epoch start. Thoughts on this? It might relate to the comment about how the < is used below.

src/server/routes/verification.js
huss: I'm not seeing how this was addressed so I need you to let me know.

src/server/routes/verification.js
huss: Not seeing how this was addressed.

src/server/routes/authenticator.js
huss: resolved for new comment

src/server/test/routes/sessionInvalidationTest.js
huss: I'm not seeing how this was addressed.

src/server/test/routes/sessionInvalidationTest.js
huss: I'm not seeing how this was addressed.

huss (changes requested):
@Oykunle Thank you for the updated code. I have resolved some comments, asked the status on a number of them (I hope I'm not missing something) and added some new ones. I am trying to get this back to you fairly quickly so I have not double checked every comment I made. I am also waiting to do full testing given the number of items that are open. I would like to request that you comment on every open comment to indicate it was done, why not or your thoughts. Please let me know when any updates are done and the comments are added (do not resolve any comments) so I know to look at this again. If anything is not clear, you have thoughts or questions then please let me know.

src/server/routes/authenticator.js
Oykunle: I kept the fallback as 0 to align with the database default of Unix epoch start, and I added a clarifying comment so that relationship is explicit in the code.

I added a clarifying comment here. The <= comparison is intentional so that tokens issued at or before token_invalid_before are rejected, which avoids edge cases where issuance and invalidation occur in the same second.

src/server/test/routes/sessionInvalidationTest.js
Oykunle: This is addressed now. I added a separate test that exercises the invalidation flow with a non-admin user instead of only the admin path. I didn’t expand it to loop across every role in this PR, but there is now explicit coverage beyond the admin user.

src/server/routes/login.js
Oykunle: I didn’t rename the default login route in this PR since that would touch multiple call sites and widen the scope quite a bit, but I agree that explicit routes like /login and /logout would be cleaner and more consistent.

src/server/test/common.js
```diff
@@ -70,13 +70,23 @@ function connectTestDB() {
 const testUser = new User(undefined, 'test@example.invalid', bcrypt.hashSync('password', 10), User.role.ADMIN);
 testUser.password = 'password';
 
+// Non-admin user (for role testing)
```
huss: While I'm an advocate for abstracting and centralizing items, I have a few thoughts about this one:

1. This is only used in one file in one place. Given that and the fact that non-admin users are created in a number of test files directly there, I think it might be best to just do this in your new test file and remove it here.
2. I don't like this name. Wherever two variables differ in small ways, it is prone to errors. Second, this is not a descriptive name. I'll agree that testUser doesn't indicate it is an admin but it was the only one so it isn't as bad.
3. I think reusing the same password as the testUser for admin is not a good idea. If one accidentally uses the wrong user name then it will still work.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,173 @@
...
+		const res = await chai.request(app)
+			.post(VERIFY_ENDPOINT)
+			.send({ token: expiredToken });
+
+		expect(res).to.have.status(HTTP_CODE.UNAUTHORIZED);
+		expect(res.body).to.have.property('success', false);
+		expect(res.body).to.have.property('message', 'Failed to authenticate token.');
+	});
+
+	mocha.it('should invalidate token for a non-admin user', async () => {
```
huss: This test is inside the same mocha.describe with a beforeEach so the admin user is also created. If you change the password on the non-admin (see common.js comment) then this is less of an issue. However, it really is not needed and could be accidentally used. I think it would be easy to put all the admin ones under one mocha and this under another with appropriate descriptions.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,173 @@
...
+
+mocha.describe('Session Invalidation Security', () => {
+	const LOGIN_ENDPOINT = '/api/login';
+	const LOGOUT_ENDPOINT = '/api/login/logout';
+	const VERIFY_ENDPOINT = '/api/verification';
+	const PROTECTED_ENDPOINT = '/api/users';
+
+	let token;
+
+	async function loginAndGetToken(user) {
```
huss: It is always nice to have JSDOC on a function.

src/server/routes/loginLogout.js
```diff
@@ -69,4 +69,42 @@ router.post('/', credentialsRequestValidationMiddleware, async (req, res) => {
 	}
```
huss: On 4/21 I still don't see these addressed.

src/server/routes/loginLogout.js
```diff
@@ -69,4 +69,42 @@ router.post('/', credentialsRequestValidationMiddleware, async (req, res) => {
...
 });
 
+/**
+ * Logs out the authenticated user by invalidating previously issued tokens.
+ */
+router.post('/logout', async (req, res) => {
```
huss: On 4/21: I see the comment in the code but I think an explicit comment at the start of this route (here) on why there is no auth middleware would help.

src/server/routes/authenticator.js
```diff
@@ -15,41 +15,42 @@ const { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, TOKEN_MAX_LENGTH, USERNAME_MIN
 	= require('../util/validationConstants');
 
 /**
- * Middleware function to force a route to require authentication
- * Verifies the request's token against the server's secret token
- * It is used within this file but not by other parts of OED.
+ * Middleware function to require authentication on protected routes.
+ * Verifies the request's token, ensures the user exists, and checks that
+ * the token has not been invalidated.
+ * This middleware is currently used within this file.
```
huss: Thank you for updating the comment. Unless you feel otherwise, I think this comment would be better as "This middleware was created to only be used within this file."

containers/web/Dockerfile
```diff
@@ -12,6 +12,16 @@ FROM node:18.17.1
 RUN mkdir -p "/usr/src/app"
 WORKDIR "/usr/src/app"
 
+# Create a dedicated non-root runtime user/group
```
huss: I don't think the changes in this file relate to this PR. I suspect they came from changes to a different PR that is sharing a branch. Unless there is a reason, these should go.

src/server/routes/verification.js
huss: 4/21: Not seeing any change/comment.

src/server/routes/verification.js
huss: 4/21: No change.

src/server/routes/verification.js
huss: 4/21: no change

src/server/routes/authenticator.js
huss: I see you added a comment but I'm not completely satisfied with the code. Is there a reason that the exact time for the token date cannot be gotten? I would think then a proper comparison (such as using moment) would not have an issue within one second. Can you please help me understand this.

src/server/routes/authenticator.js
huss: 4/21: no change

huss (changes requested):
@Oykunle Thank you for the update. I've added some comments and tried to put a date comment on all the ones that did not seem addressed. Previously I commented:

> I would like to request that you comment on every open comment to indicate it was done, why not or your thoughts.

I am again asking for this. I would like you to let me know once that is done. I do not intend to review this again nor do final testing until that is complete. I hope I have not missed something here.

Please let me know if something is not clear or you have thoughts.

Oykunle:
@huss

Good catch on that, you’re right, the Dockerfile changes were not related to this PR.
Those changes came from work on a different branch and were accidentally included here. I’ve now removed them from this PR.

src/server/routes/verification.js
Oykunle: @huss

I’ve now updated the verification route to return a generic authentication failure response instead of exposing specific error details.

Please let me know if this addresses your concern or if you’d prefer a different approach.

Oykunle:
@huss

I’ve updated this to compare the token issue time and token_invalid_before at millisecond precision instead of relying on a one-second comparison boundary.

That should avoid the edge case where issuance and invalidation happen within the same second while still preserving the intended invalidation behavior.

Oykunle:
@huss

I’ve now gone through the remaining open comments and either addressed them in code or replied with my reasoning where no additional code change was made.

The main updates in this round were:
- removed the unrelated Dockerfile change from this PR
- added the explicit logout route comment
- updated verification to use a generic authentication failure response
- moved the CSV test user into the session invalidation test file
- split admin and CSV tests into separate describe blocks
- updated the token invalidation comparison to use millisecond precision
- renamed the login route to use an explicit `/login` path and updated usages accordingly

Please let me know when you’d like me to make any further adjustments.

huss:
> I’ve now gone through the remaining open comments and either addressed them in code or replied with my reasoning where no additional code change was made.

@Oykunle This comment indicates that some items were not done and there is a reply to my comment. However, when I look at the open comments I don't see any responses from you. Am I missing something? Did this only apply to older items?

Also, for your future information, I find it easier if there are replies to comments since it keeps them together. (Unfortunately GitHub is creating new comments while I am doing a review so my replies are separated and/or show up twice which I regret.) If that is not feasible then doing a Quote reply of the necessary text helps put the response in context. I understand you are using a different method and appreciate your consideration.

src/server/routes/loginLogout.js
```diff
@@ -10,8 +10,8 @@ const secretToken = require('../config').secretToken;
 const validate = require('jsonschema').validate;
```
huss, line 10: Thank you for changing the logout route name. Previously I said about another previous comment:

> I previously commented that a good name might be needed for this file. I think the current one of login.js is a misnomer. Maybe loginLogout.js but I'm open to other ideas.

I tried to think of a better name. authentication could work but OED already has a file close to that name for a different purpose. Thus, I think it should be renamed to loginLogout.js (not copied but renamed in git) unless you have a better name or thoughts.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,193 @@
...
+	const LOGOUT_ENDPOINT = '/api/login/logout';
+	const VERIFY_ENDPOINT = '/api/verification';
+	const PROTECTED_ENDPOINT = '/api/users';
+
+	let token;
+	let csvUser;
+
+	/**
+	 * Logs in the provided user and returns the issued token.
+	 * @param {User & {password: string}} user
```
huss: I'm unclear on the param typing. I think both username and password are strings inside a JSON structure. This does not seem to match that.

src/server/routes/loginLogout.js
huss: Partial addressed but replaced with new comment.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,193 @@
...
+mocha.describe('Session Invalidation Security', () => {
+	const LOGIN_ENDPOINT = '/api/login/login';
+	const LOGOUT_ENDPOINT = '/api/login/logout';
+	const VERIFY_ENDPOINT = '/api/verification';
+	const PROTECTED_ENDPOINT = '/api/users';
+
+	let token;
+	let csvUser;
+
+	/**
```
huss: Small point but I would use /* not /** since this is not exported.

src/server/test/routes/sessionInvalidationTest.js
huss: 4/24 I don't see the part about separating the tests addressed.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,193 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const bcrypt = require('bcrypt');
```
huss: The name of the file is wrong and differs from other uses in OED. It should be bcryptjs. Once you address comments in this PR you really should run the code and test code to verify it works correctly.

src/server/test/common.js
```diff
@@ -96,7 +97,7 @@ module.exports = {
 	mocha,
 	expect,
 	app,
-	testUser,
+	testUser,,
```
huss: The syntax is wrong - only one comma. This makes the tests fail.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,193 @@
...
+	mocha.beforeEach(async () => {
+		const conn = common.testDB.getConnection();
+
+		csvUser = new User(
+			undefined,
+			'test-csv@example.invalid',
+			bcrypt.hashSync('csv-password-2', 10),
+			User.role.CSV
+		);
+		csvUser.password = 'csv-password-2';
```
huss: A User instance does not have a password. This works because there is no testing in the model when directly used. However, it is sent to the login route below and that should fail. I cannot test because the code is broken. This should not be on the user instance but separate. Also, please carefully test that the login route does reject params when extra ones are included. As I indicated in another comment, these routes need systematic testing with the new system that would catch this.

huss (changes requested):
I started a review but stopped it due to the fact that I could not easily get the code to work. After fixing the ones in comments to the PR I started to get this when running the new tests:

> (node:380) Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
> (Use `node --trace-warnings ...` to show where the warning was created)
> (node:380) Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
> (node:380) Warning: Accessing non-existent property 'getDB' of module exports inside circular dependency
> (node:380) Warning: Accessing non-existent property 'stopDB' of module exports inside circular dependency

Right now I don't have the time to figure out what is wrong. Circular dependencies have become an issue but I suspect some change you made caused this one since it wasn't there before.

Also, you changed a route but did not, from what I see, change anything on the client code where it seems to be used in (at least) two files. Once the actual file name is resolved per another comment, please update all uses and test.

Given the concerns found so far, I am stopping my review until the code is in a better state for review. Please let me know if anything is not clear and/or you have thoughts/questions.

Oykunle:
> I started a review but stopped it due to the fact that I could not easily get the code to work. After fixing the ones in comments to the PR I started to get this when running the new tests:
> 
> > (node:380) Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
> > (Use `node --trace-warnings ...` to show where the warning was created)
> > (node:380) Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
> > (node:380) Warning: Accessing non-existent property 'getDB' of module exports inside circular dependency
> > (node:380) Warning: Accessing non-existent property 'stopDB' of module exports inside circular dependency
> 
> Right now I don't have the time to figure out what is wrong. Circular dependencies have become an issue but I suspect some change you made caused this one since it wasn't there before.
> 
> Also, you changed a route but did not, from what I see, change anything on the client code where it seems to be used in (at least) two files. Once the actual file name is resolved per another comment, please update all uses and test.
> 
> Given the concerns found so far, I am stopping my review until the code is in a better state for review. Please let me know if anything is not clear and/or you have thoughts/questions.

@huss

Thanks for taking a look at this and for pointing that out.

You were right, there were a few issues in the test setup that were causing the failures and the circular dependency warnings. I tracked those down and fixed them:

* removed the leftover testUser2 usage in common.js
* fixed the syntax error (extra comma) that was breaking the test setup
* updated the tests to use bcryptjs to stay consistent with the project
* refactored the login helper to pass username/password directly instead of attaching a password to the User instance
* cleaned up the test structure (separated admin and CSV cases)
* aligned expected responses with the current generic authentication failure behavior

After these changes, I reran the tests in Docker:

NODE_ENV=test npx mocha --timeout 15000 src/server/test/routes/sessionInvalidationTest.js

They are now passing (8/8), and the blocking errors you hit are no longer occurring.
On the route change , you’re right, I haven’t updated the client-side usages yet. I’ll go through and update those once the final naming decision (e.g., loginLogout.js) is settled so everything stays consistent.

I’ll also go back and reply directly to the remaining inline comments so it’s easier to track what’s been addressed.

Please let me know when you’re ready to take another look or if anything still seems off on your side.

Oykunle:
> Also, you changed a route but did not, from what I see, change anything on the client code where it seems to be used in (at least) two files. Once the actual file name is resolved per another comment, please update all uses and test.
> 

@huss,

Good catch on the client-side usage , you’re right, the login route was still pointing to api/login.

I’ve updated it to api/login/login in authApi.ts so it matches the server changes.

I also double-checked logout and verification, and those were already consistent.

Oykunle:
Hi @huss,

I’ve made the latest updates and pushed them.

Updates include:
- resolved merge conflicts with upstream/development
- renamed `login.js` to `loginLogout.js` using `git mv`
- updated the client login route in `authApi.ts` to `api/login/login`
- removed duplicate JWT verification and standardized on `verifyActiveTokenAndGetUser`
- cleaned up duplicate assertions and unused imports
- reran the session invalidation tests successfully

Let me know if anything else needs adjustment.

Oykunle:
>Not seeing how this was addressed.
>
Thanks for pointing this out.

I’ve updated the verification route to use a generic authentication failure response instead of returning specific error messages (such as token invalidated or user not found).

This aligns it with the rest of the authentication flow and avoids leaking sensitive information.

Oykunle:
>4/21: No change.
>

Thanks for the follow-up.

This has now been updated to consistently use verifyActiveTokenAndGetUser with a generic failure response. The previous conditional error handling has been removed.

Oykunle:
>I see you added a comment but I'm not completely satisfied with the code. Is there a reason that the exact time for the token date cannot be gotten? I would think then a proper comparison (such as using moment) would not have an issue within one second. Can you please help me understand this.
>
Great question.

JWT stores the iat value in seconds, while the database timestamp (token_invalid_before) is stored with millisecond precision. Because of this difference in precision, a direct comparison at second-level granularity can lead to edge cases when issuance and invalidation occur within the same second.

To address this, I converted the JWT iat value to milliseconds and performed the comparison at millisecond precision. This avoids edge cases without requiring additional libraries and keeps the comparison consistent with the database format.

Oykunle:
> > I started a review but stopped it due to the fact that I could not easily get the code to work. After fixing the ones in comments to the PR I started to get this when running the new tests:
> > > (node:380) Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
> > > (Use `node --trace-warnings ...` to show where the warning was created)
> > > (node:380) Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
> > > (node:380) Warning: Accessing non-existent property 'getDB' of module exports inside circular dependency
> > > (node:380) Warning: Accessing non-existent property 'stopDB' of module exports inside circular dependency
> > 
> > 
> > Right now I don't have the time to figure out what is wrong. Circular dependencies have become an issue but I suspect some change you made caused this one since it wasn't there before.
> > Also, you changed a route but did not, from what I see, change anything on the client code where it seems to be used in (at least) two files. Once the actual file name is resolved per another comment, please update all uses and test.
> > Given the concerns found so far, I am stopping my review until the code is in a better state for review. Please let me know if anything is not clear and/or you have thoughts/questions.
> 
> @huss
> 
> Thanks for taking a look at this and for pointing that out.
> 
> You were right, there were a few issues in the test setup that were causing the failures and the circular dependency warnings. I tracked those down and fixed them:
> 
> * removed the leftover testUser2 usage in common.js
> * fixed the syntax error (extra comma) that was breaking the test setup
> * updated the tests to use bcryptjs to stay consistent with the project
> * refactored the login helper to pass username/password directly instead of attaching a password to the User instance
> * cleaned up the test structure (separated admin and CSV cases)
> * aligned expected responses with the current generic authentication failure behavior
> 
> After these changes, I reran the tests in Docker:
> 
> NODE_ENV=test npx mocha --timeout 15000 src/server/test/routes/sessionInvalidationTest.js
> 
> They are now passing (8/8), and the blocking errors you hit are no longer occurring. On the route change , you’re right, I haven’t updated the client-side usages yet. I’ll go through and update those once the final naming decision (e.g., loginLogout.js) is settled so everything stays consistent.
> 
> I’ll also go back and reply directly to the remaining inline comments so it’s easier to track what’s been addressed.
> 
> Please let me know when you’re ready to take another look or if anything still seems off on your side.


I traced this back to issues in the test setup and fixed the root causes, including a syntax error, inconsistent imports, and incorrect test user handling.

After these fixes, I reran the tests and they now execute successfully without the blocking errors.

Oykunle:
Hi @huss,

I reran the checks after the latest fix:

- Working tree is clean
- No merge conflict markers remain
- Key files pass node syntax checks
- Session invalidation tests pass: 8/8
- Related login/auth/users tests pass: 47/47

The latest fix also updates the session invalidation tests to use HTTP_CODES from util/httpCodes.

Please let me know when if you want me to change something.

src/server/app.js
```diff
@@ -17,7 +17,7 @@ const users = require('./routes/users');
 const readings = require('./routes/readings');
 const meters = require('./routes/meters');
 const preferences = require('./routes/preferences');
-const login = require('./routes/login');
+const login = require('./routes/loginLogout');
```
huss, line 20: While I think I see the appeal of leaving the name as login so all the current routes work, I think making the actual server route file differ from the route name is confusing and prone to mistakes. The only other case of doing this is for readings where the actual name is used in the app.use(). I think this was an attempt to decouple the two names but it was not done in any consistent way. Something that can be fixed. For now, I think it should become loginLogout with all uses updated including app.use below.

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,198 @@
...
+			expect(verifyRes.body).to.not.have.property('message');
+		});
+
+		mocha.it('should invalidate a token after logout', async () => {
+			const beforeVerify = await chai.request(app)
+				.post(VERIFY_ENDPOINT)
+				.send({ token });
+
+			expect(beforeVerify).to.have.status(HTTP_CODES.OK);
+			expect(beforeVerify.body).to.have.property('success', true);
```
huss, line 63: Above has a check for no message property. Should it be here (and I think there is another usage below)?

src/server/test/routes/sessionInvalidationTest.js
```diff
@@ -0,0 +1,198 @@
...
+			expect(firstLogoutRes).to.have.status(HTTP_CODES.OK);
+			expect(firstLogoutRes.body).to.have.property('success', true);
+
+			const secondLogoutRes = await chai.request(app)
+				.post(LOGOUT_ENDPOINT)
+				.send({ token });
+
+			expect(secondLogoutRes).to.have.status(HTTP_CODES.OK);
+			expect(secondLogoutRes.body).to.have.property('success', true);
+			expect(secondLogoutRes.body).to.have.property('message', 'Logout successful.');
```
huss, line 95: Here you check the message but not right above for the first attempt. Is there a reason?

src/server/routes/verification.js
huss: This is being postponed and will be addressed when all the routes with similar issues are addressed.

huss (changes requested):
@Oykunle First, I want to say I'm sorry this took so long. I communicated with you earlier about why and I appreciate your patience.

This is looking good. I merged in development to get rid of the DB startup issue from a recently merged PR as this sat for a while due to me. Testing after that did not see issues. I also went through all the previous comments and hid all the resolved ones so it would be easier to see what is left. I think all the old ones are done (thanks) and there are three new ones to consider. I don't think they are complex but let me know if you have questions/thoughts.

I know this is past the time you intended to work on OED so let me know if you want someone else to address the final comments.

I will want to do a careful read of the code once it looks done to be sure this security patch is all good.

huss:
@Oykunle Thank you for your work on this issue. I know it is after the time you originally intended to work on OED. Given OED wants to finish this up & it has been a few weeks since my last comment, I'm going to get someone else to complete this work unless you indicate by 6/15/26 that you intend to work on it and it is completed in the near future.

huss:
All outstanding comment on this PR are now done in the follow on PR #1672. This PR is being closed.

PR #1595, opened 2026-04-02
Add configurable trusted origins for clickjacking protection

Oykunle:
Description

This change extends clickjacking protection by allowing trusted framing origins to be configured.

Previously, enforcing strict framing protection (e.g., frame-ancestors 'self') could break valid use cases such as embedding charts via generated links. This update introduces a configuration-based approach that preserves security while allowing controlled embedding.

Changes:
	•	Added support for OED_TRUSTED_FRAME_ANCESTORS environment variable
	•	Parsed and exposed trusted origins via server configuration
	•	Updated CSP header to include 'self' plus configured trusted origins

Security impact:
	•	Maintains clickjacking protection by default
	•	Allows controlled embedding without breaking expected behavior
	•	Avoids hard-coded origins

Testing:
	•	Verified default: `frame-ancestors 'self'`
	•	Verified configured: frame-ancestors 'self' https://example.com

## Type of change

## Checklist

- [x] All contributors have signed the OED CLA and are listed

Limitations
	•	Trusted origins are environment-based (no admin UI yet)
	•	Minimal validation of origin format
	•	Changes require restart to take effect

### DIFF
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -155,6 +155,13 @@ router.get('*', (req, res) => {
 	fs.readFile(path.resolve(__dirname, '..', 'client', 'index.html'), (err, html) => {
 		const subdir = config.subdir || '/';
 		let htmlPlusData = html.toString().replace('SUBDIR', subdir);
+
+		// Clickjacking protection: restrict framing to same-origin by default,
+		// with optional admin-configurable trusted origins via environment variable.
+		// This preserves OED's chart embedding functionality while mitigating UI redress attacks.
+		const frameAncestors = ["'self'", ...config.trustedFrameAncestors].join(' ');
+		res.setHeader('Content-Security-Policy', `frame-ancestors ${frameAncestors};`);
+
 		res.send(htmlPlusData);
 	});
 });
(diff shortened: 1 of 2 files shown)


huss:
@Oykunle Thank you for looking into this and working on it. I'm sorry to say that I don't think OED can use this solution. OED has a "Admin Settings" page with a "Site Settings" section to allow admins to add customization. While OED does use environment variables for some setup, it was intended to be limited to software setup and not items that can change after OED is running. I think a site might want to add or remove more exceptions over time (such as if they switch the server that provides web pages). Exactly how this would be done to modify these values to the DB and allow for entry of one or more trusted origins would need to be worked out. If you want to pursue this then I suggest you communicate with the project about design ideas. If you are not going to do this then an issue should be opened to indicate this is needed. You are welcome to do this or I can do it. Also, as we have discussed, OED probably cannot merge in the proposed clickjacking fix until this or another mechanism is in place to allow for using the OED chart links. Thus, if you are not going to pursue this then the new issue can reference this PR and someone else can use it to move forward but OED would likely close this PR for now.

I welcome any thoughts, ideas or concerns you have.

Oykunle:
@huss
I have opened Issue #1601 to track the OED-compatible solution for trusted origins through Admin Settings / Site Settings.

Given your feedback, I understand this PR’s environment-variable approach is better treated as a prototype/reference for the CSP logic rather than the final implementation.

Please let me know whether you would prefer this PR to remain open as a reference or be closed now that Issue #1601 is in place.

Thanks again.

huss:
Per issue #1601 this is being closed.

PR #1604, opened 2026-04-13
Fix Undue Information Disclosure

Zachary-Squires:
# Description

Adds centralized backend error handling to prevent information disclosure. Changes:

Global error handler in app.js
Generic 500 responses in response.js

Fixes OED Security Issue 14

## Type of change

## Checklist

## Limitations

N/A

### DIFF
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -155,7 +155,7 @@ router.get('*', (req, res) => {
 	fs.readFile(path.resolve(__dirname, '..', 'client', 'index.html'), (err, html) => {
 		if (err) {
 			log.error('Failed to read index.html for client router; logging caught err object.', err);
-			return res.status(500).send('Internal Server Error. Site admins can check logs for details.');
+			return res.status(500).send('Internal Server Error. Details are in the OED logs that are available to your site admin(s).');
 		}
 
 		const subdir = config.subdir || '/';
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -155,7 +155,7 @@ router.get('*', (req, res) => {
 	fs.readFile(path.resolve(__dirname, '..', 'client', 'index.html'), (err, html) => {
 		if (err) {
 			log.error('Failed to read index.html for client router; logging caught err object.', err);
-			return res.status(500).send('Internal Server Error');
+			return res.status(500).send('Internal Server Error. Site admins can check logs for details.');
 		}
 
 		const subdir = config.subdir || '/';
@@ -173,6 +173,9 @@ app.use((req, res) => {
 // Global error handler, errors will still be logged internally but keep client response generic
 app.use((err, req, res, next) => {
 	// Malformed JSON needs to return bad request for tests to pass
+	// err instanceof SyntaxError: body-parser throws SyntaxError when JSON cannot be parsed
+	// err.status === 400: confirms this parse failure maps to HTTP 400 Bad Request
+	// 'body' in err: indicates the error came from request body parsing and not an unrelated SyntaxError
 	if (err instanceof SyntaxError && err.status === 400 && 'body' in err) {
 		return res.status(400).send('Bad Request');
 	}
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -153,6 +153,11 @@ const router = express.Router();
 // Accept all other endpoint requests which will be handled by the client router
 router.get('*', (req, res) => {
 	fs.readFile(path.resolve(__dirname, '..', 'client', 'index.html'), (err, html) => {
+		if (err) {
+			log.error('Failed to read index.html for client router.', err);
+			return res.status(500).send('Internal Server Error');
+		}
+
 		const subdir = config.subdir || '/';
 		let htmlPlusData = html.toString().replace('SUBDIR', subdir);
 		res.send(htmlPlusData);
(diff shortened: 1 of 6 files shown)


src/server/routes/groups.js
```diff
@@ -293,8 +293,9 @@ router.post('/create', adminAuthMiddleware('create groups'), async (req, res) =>
 			});
 			success(res);
 		} catch (err) {
+			// Map duplicate-name DB errors to a safe 400 response
```
huss, line 296: Should "Map" be "Group"?

src/server/routes/response.js
```diff
@@ -25,8 +25,10 @@
  *
  */
 function failure(res, code = 500, comment = '') {
+	// Return a generic message for 500-level failures.
+	const responseBody = code >= 500 ? 'Internal Server Error' : comment;
```
huss, line 29: I have some thoughts and want your input. I like that you are stopping error messages from going back to the user. I also noticed that a number of routes don't use this function (or success) but directly send back the info where you fixed some of those. Given this, here are a few thoughts:

1. I think the logging should happen here rather than in each location where the error occurs. Then the message will not go back to the user but will be recorded in the DB.
2. If this is done, should there be a parameter to disable logging with the default of false? That means logs are created unless the developer feels there is a good reason not to do that.
3. All routes should be checked so logging info is sent on for failures and not done in the route.
4. All routes should be changed to use the modified functions and stop sending back responses directly. This would centralize this process and make it safer/easier to change in the future.
5. For modified message with a generic server error, I was thinking to make it "Internal Server Error. Details are in the OED logs that are available to your site admin(s)." so users know to look there or ask someone to look if they cannot.

Let me say this may be beyond what you wanted to do so OED can open an issue to fix up all the routes if you agree with the idea but don't want to do it.

src/server/app.js
```diff
@@ -165,4 +170,18 @@ app.use((req, res) => {
 	res.status(404).send('<h1>404 Not Found</h1>');
 });
 
+// Global error handler, errors will still be logged internally but keep client response generic
```
huss, line 173: I think I get the idea but I'm not understanding how to get this to actually happen. I tried removing the try/catch in a route so the error rippled but that did not seem to do it. I also tried to throw an error but did not work. Could you help me understand the overall intention of this?

src/server/log.js
```diff
@@ -171,10 +171,15 @@ const defaultLogger = new Logger(logFile);
  */
 process.on('unhandledRejection', (reason, p) => {
 	p.catch(e => {
-		defaultLogger.error(`Unhandled Promise Rejection: ${reason}`, e);
+		defaultLogger.error('Unhandled Promise Rejection:', reason, e);
```
huss, line 174: I don't know what this used to put out so unsure if this is a new concern but I see this logging but it does not seem to show a reason or e (error). Also, I think the message should say it is the reason with reason and the error with e.

src/server/app.js
```diff
@@ -165,4 +170,18 @@ app.use((req, res) => {
...
+app.use((err, req, res, next) => {
+	// Malformed JSON needs to return bad request for tests to pass
+	if (err instanceof SyntaxError && err.status === 400 && 'body' in err) {
```
huss, line 176: Could you please add some more comments on how each if works and why it is checking each value so future people understand?

src/server/app.js
```diff
@@ -153,6 +153,11 @@ const router = express.Router();
 // Accept all other endpoint requests which will be handled by the client router
 router.get('*', (req, res) => {
 	fs.readFile(path.resolve(__dirname, '..', 'client', 'index.html'), (err, html) => {
+		if (err) {
+			log.error('Failed to read index.html for client router.', err);
```
huss, line 157: Do you know an easy way to make this happen for testing?

Also, all the messages (here and below) should clearly indicate it is the err from the item that is being added to the log as discussed in another file.

huss (changes requested):
@Zachary-Squires Thanks for addressing this issue. Testing was tricky for this work so I hope my comments are correct and make sense. (I note that testing this led me to realize the group create route is broken due to the new param testing. All the routes need careful review and fixing as needed but that is not work for this PR.) Given the nature of some of the comments, I want to see how they are addressed and your thoughts before I do testing on some parts. If anything is not clear them please let me know. I also happy to discuss.

src/server/app.js
Zachary-Squires: The problem is that when you remove the try/catch it creates a promise rejection and express never sees it, this is specifically to handle when a route calls next(err)

src/server/routes/groups.js
Zachary-Squires: Changing that.

src/server/routes/response.js
Zachary-Squires: I do agree those are good ideas, but I don't think I'll have time to implement them.

src/server/app.js
Zachary-Squires: I changed the messages, an easy way to test this is to create a temp route that calls next(new Error('test')) which will hit the global handler

huss: I'm sorry I did not label this well and it is not clear. I think the res string should indicate that the details are in the logs that a site admin can see. That way if someone wants to know the details they know where to go. This was done in another change you recently made. The actual msg logged is clear in the log so that isn't really needed.

src/server/app.js
huss: What I meant by this is that there are 3 parts of the logic in the if statement so could you describe what each is doing for the overall goal?

src/server/routes/response.js
huss: Okay. I think an issue should be opened. Is this something you want to do or should I?

huss (changes requested):
@Zachary-Squires Thank you for the update addressing my comments. I have resolved what I could and left a few new comments to existing ones to consider. I think this is close so just cleaning up these should do it. Please let me know if anything is not clear or you have thoughts.

src/server/app.js
Zachary-Squires: Will add an expanded comment explaining it.

src/server/app.js
Zachary-Squires: Changed it.

src/server/routes/response.js
Zachary-Squires: I think you should probably do it, you have a better understanding of it than I do for sure.

huss: Issue #1614 was opened to address this.

huss (approved):
@Zachary-Squires Thank you for addressing my comments. Review and testing found everything is now good. Congratulations on another accepted contribution to OED.

PR #1581, opened 2026-03-05
Switch Numbers to Constants

Nespina24:
# Description

Added constant imports throughout 64 src/server files in order to change hard-coded numeric values to fit the listed format in a newly added file src/server/util/httpCodes.js.

Fixes #1573

## Type of change

- [ ] Note merging this changes the database configuration.
- [ ] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations

No limitations.

PR #1605, opened 2026-04-13
Rate Testing

### LINKED ISSUE #1564: limit login rate

A malicious person could try to get into an account with lots of tries. OED tries to mitigate that by not telling which of the username and password are wrong. OED also has a rate limiter on all routes in app.js that should significantly slow down such an attack and probably make it unlikely to succeed.

While the current situation is very good, it is proposed to have an explicit, lower limit on login attempts. This will make attacks harder, esp. if specific values are being probed. Specifically, the route in src/server/routes/login.js will be limited. If the rate limit is hit in normal usage then valid logins will fail. OED does not require login for most actions; only for admin, export, csv and some systems sending OED data. Thus, there is unlikely to be a lot of login at any time. It is unclear how many should be allowed to be safe but I'll throw out 900 requests per hour (3600 seconds) or an average of 1 every 4 seconds. By making it per hour it seems less likely to hit the limit during normal usage in which there are a larger number of people wanting to login in a short amount of time but still allow lots of logins. Note the general limit is (3600 sec/hour) / (5 sec/rate interval)  * (200 requests/rate interval) =  144,000 requests/hour. Thus, the new limit is 144,000 / 900 = 160 times lower. It limits tries to a little over 20k/day which should give a site time to stop such attacks if they monitor the logs. I'm very open to other ideas.

OED could allow the site to set the limit on the admin site page. I propose we not do this unless issues are encountered. The reason is that one needs to know what are better values and it isn't clear it is needed. If OED ever does this it could check the value proposed to warn/exclude the value trying to be set.

### Additional context

None.

Zach-O-Bates:
# Description

This pull request resolves issues with login rate limiting and test reliability. The login rate limiter was updated to behave differently based on environment variables, allowing normal tests to run without interference while enforcing strict limits during dedicated rate limit testing. A separate ratetest environment was introduced to properly validate 429 responses without impacting other test suites.

Fixes #1564 

## Type of change

## Checklist

## Limitations

N/A


huss:
This is to note that PR #1593 may change the route location of the login and this could impact this PR.

OED really should check all route names to verify they are correct compared to usage here and the upcoming test code (this part is outside the scope of the PR).

src/server/app.js
```diff
@@ -116,6 +120,16 @@ const exportRawLimiter = rateLimit({
 // Apply the raw export limit
 app.use('/api/readings/line/raw/meters', exportRawLimiter);
 
+// Limit the number of login attempts to 1 per 4 seconds
```
huss: Issue #1564 did suggest a login limit that is equivalent to 1 per 4 seconds. However, it discussed making it 900 per hour (or 3,600,000 ms) (see issue for rationale). I do think it is important not to be this way but am open to other limits if you have thoughts.

src/server/tempRatetest/rateTest.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* This file tests the login route rate limiting functionality.
+   It ensures that repeated login attempts are blocked after
+   exceeding the configured rate limit. */
+
+/* Run in OED Docker web container terminal/shell:
```
huss: Please remove this comment as every test runs this way.

src/server/tempRatetest/rateTest.js
```diff
@@ -0,0 +1,37 @@
...
+            expect(first).to.have.status(401);
+    
+            const second = await chai.request(app)
+                .post('/api/login')
+                .send({
+                    username: 'invalidUser',
+                    password: 'invalidPassword'
+                });
+    
+            expect(second).to.have.status(429);
```
huss: src/server/util/readingsUtils.js now has HTTP_CODE for naming all the return codes OED uses. These should be used here, below and it the comment where making it in addition to 429 is fine if you prefer.

Note PR #1581 will be moving this import - FYI. It can be done when it merges if this PR is merged after that one.

src/server/tempRatetest/rateTest.js
```diff
@@ -0,0 +1,37 @@
...
+
+/* This file tests the login route rate limiting functionality.
+   It ensures that repeated login attempts are blocked after
+   exceeding the configured rate limit. */
+
+/* Run in OED Docker web container terminal/shell:
+npm run testsome src/server/test/routes/loginRateTest.js */
+
+const { chai, mocha, expect, app } = require('../test/common');
+
```
huss: Please remove second blank line. Also, indenting should be by tabs not spaces.

src/server/app.js
```diff
@@ -116,6 +120,16 @@ const exportRawLimiter = rateLimit({
...
+const loginLimiter = rateLimit({
+	windowMs: 4 * 1000, // 4 seconds
```
huss: Though OED is not consistent, it prefers comments above lines of code.

src/server/app.js
```diff
@@ -116,6 +120,16 @@ const exportRawLimiter = rateLimit({
...
+	limit: isRateTest ? 1 : 1 * testMultiplier, // 1 requests
```
huss: I think the comment should describe why it is set to one for rate testing. Also, the comment is only correct when that is true so maybe modify to explain the general limit for all other cases.

src/server/tempRatetest/rateTest.js
```diff
@@ -0,0 +1,37 @@
...
+	mocha.it('Should block repeated login attempts with 429', async () => {
+            const first = await chai.request(app)
+                .post('/api/login')
+                .send({
+                    username: 'invalidUser',
+                    password: 'invalidPassword'
+                });
+            expect(first).to.have.status(401);
+    
+            const second = await chai.request(app)
```
huss: This implicitly assumes the rate is 1 request per a unit of time and that the test runs fast enough to not cross that amount of time between the first and second attempt. That should be okay but a comment should make it explicit. Note if the time frame of the limit is changed as suggested in another comment (to per hour) then the odds of this happening goes to zero but the rationale should still be commented.

src/server/tempRatetest/rateTest.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* This file tests the login route rate limiting functionality.
```
huss: The name of this file does not seem to scope it as only for login (rateTest.js). I think the name should stay the same but the comment state it currently only does the login route. There should also be a TODO to check every rate limit in app.js.

I think an issue should be opened to say this needs to be done (unless you do it now which I'm assuming you are not). Would you like to do that or should I?

src/server/app.js
```diff
@@ -131,7 +145,7 @@ app.use('/api/users', users);
 app.use('/api/meters', meters);
 app.use('/api/readings', readings);
 app.use('/api/preferences', preferences);
-app.use('/api/login', login);
+app.use('api/login')
```
huss: Two thoughts. First, you removed the second argument of login but the const is still defined above. This may not be important given the second thought. Second, when I try to start OED I get an error starting with ``TypeError: app.use() requires a middleware function``. The first time the OED Docker web container did not properly run but it did the second time. However, this needs to be resolved. I'm not really knowledgeable about this area but it seems you cannot make this change. My first thoughts are either put it back or try removing it and making sure it does not cause any issues elsewhere in OED. Or maybe you have other ideas/thoughts. If it stays, it needs a semicolon at the end of the line.

src/server/tempRatetest/rateTest.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
huss: The location of this file seems off. package.json has ``src/server/rateTest/**/*.js`` which seems reasonable. The test will not run at this time with the package script. I'm putting off testing until this and other comments are resolved.

huss (changes requested):
@Zach-O-Bates Thanks for working on this issue. Overall it is okay but I've made some comments to consider. Please let me know if anything is not clear or you have questions/thoughts.

src/server/tempRatetest/rateTest.js
Zach-O-Bates: I can open it

src/server/app.js
```diff
@@ -116,6 +120,23 @@ const exportRawLimiter = rateLimit({
 // Apply the raw export limit
 app.use('/api/readings/line/raw/meters', exportRawLimiter);
 
+// Limit the number of login attempts
+const loginLimiter = rateLimit({
+	// Window of 1 hour 
+	windowMs: 60 * 60 * 1000, 
+	/* Rationale: The login route requires a more specifc and strict rate limit that must be tested seperately.
```
huss, line 127: Misspelled: specifc, seperately & enviroment

src/server/rateTest/rateTest.js
```diff
@@ -0,0 +1,38 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* This file currently only tests the login route rate limiting functionality.
+   It ensures that repeated login attempts are blocked after
+   exceeding the configured rate limit. */
+
+// TODO: Create additonal rate limits tests for other OED Routes
```
huss, line 9: spelling: additonal

src/server/tempRatetest/rateTest.js
huss: I'm not seeing the issue yet.

src/server/tempRatetest/rateTest.js
huss: Thanks for fixing the location of the file. I know I said the proposed location was okay but I've reconsidered. How would you feel about changing the directory name from rateTest/ to testRate/. That would mean it would be listed next to the test/ directory and that seems useful to me.

src/server/rateTest/rateTest.js
```diff
@@ -0,0 +1,38 @@
...
+	mocha.it('Should block repeated login attempts with 429', async () => {
+			const first = await chai.request(app)
+				.post('/api/login')
+				.send({
+					username: 'invalidUser',
+					password: 'invalidPassword'
+				});
+
+			//First request that makes it through to authentication
+			expect(first).to.have.status(HTTP_CODE.UNAUTHORIZED);
```
huss, line 25: This test is failing on my machine. It is giving a 404/NOT_FOUND instead of 401/UNAUTHORIZED. I have not looked into why. Do you see that when you do ``npm run rateTest``?

I'm going to put off testing all the new package script choices until this is resolved.

src/server/rateTest/rateTest.js
```diff
@@ -0,0 +1,38 @@
...
+	
+			const second = await chai.request(app)
```
huss, line 27: Redoing comment since code moved. I had:

> This implicitly assumes the rate is 1 request per a unit of time and that the test runs fast enough to not cross that amount of time between the first and second attempt. That should be okay but a comment should make it explicit. Note if the time frame of the limit is changed as suggested in another comment (to per hour) then the odds of this happening goes to zero but the rationale should still be commented.

I think discussing here that the limit should be 1 for an hour so this fails with a comment has value.

huss (changes requested):
@Zach-O-Bates Thank you for addressing my previous comments. I have made a couple of new ones and also commented on a couple of older ones. I think this is getting much closer. Please let me know if anything is not clear or you have thoughts/questions.

huss:
@Zach-O-Bates I was sweeping issues and wanted to see if you could give me any information on the status of this work. Thanks.

src/server/tempRatetest/rateTest.js
huss: I created issue #1680 for this.

huss:
I'm noting that several outstanding issues/comments are being marked resolved/hidden because they are completed in PR #1670 but they were never done in this PR. This is helping track needed changes.

huss:
PR #1670 continues this work so this PR is being closed since that PR has now addressed all the open comments in this PR.

PR #1609, opened 2026-04-20
Adding Support for Infisical Password Vault

Zachary-Squires:
# Description

This PR adds optional support for Infisical, an open source free secrets manager that can be self hosted. This PR allows the POSTGRES_PASSWORD and OED_DB_PASSWORD variables to be stored inside the Infisical password vault instead of the .env for increased security. ~~I've created a tutorial as to how to do this which is here: https://gist.github.com/Zachary-Squires/b7d673922d251ec4d0edc5aa752ec06f~~ There is an [infisical  design doc](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/infisicalIntegration/infisicalIntegration.md) with this information and some ideas on moving forward for anyone looking at this.

## Type of change

## Checklist

## Limitations

Currently Infisical can only host the passwords for the database, in future this could be expanded to also host user passwords.


huss:
@Zachary-Squires Thanks for working on this. I realize it is a work in progress but I had two thoughts from only looking at this a little:

1. Normally OED keeps all the related documents in the DesignDocs repository. Is there a reason not to move your documentation there via a PR to that repo?
2. ~~There are a number of files that are changed in ways that seem unrelated to this work. Some seem related to open PRs and some from recent work. I have not looked carefully at why. I did see you did a merge from a development branch but maybe it wasn't from the OED repo. Is there a way to clean this up?~~

Please let me know if anything is not clear or your have thoughts.

huss:
I'm noting this PR contains code from PR #1554 as it is necessary. That PR will should be merged first so all those changes go away before this PR merged.

src/server/util/loadInfisicalSecrets.js
```diff
@@ -0,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const childProcess = require('child_process');
+const path = require('path');
+
+function loadInfisicalSecretsSync() {
+    // Keep vault support opt-in so default setups continue to use local .env values.
```
huss: This is using space indenting and OED likes tabs. Same for other new .js file.

src/server/util/loadInfisicalSecrets.js
```diff
@@ -0,0 +1,57 @@
...
+    const vaultEnabled = /^(true|yes|1)$/i.test(process.env.PASSWORD_VAULT || 'no');
```
huss: This isn't a big deal but OED has, I think, up to now only had yes and no for these values. Does OED want to allow true and 1? Given this is part of the config settings, I am inclined toward not unless others feel otherwise. Also, if this is done then it should probably be consistent with other so production, etc. is also changed.

src/server/util/loadInfisicalSecrets.js
```diff
@@ -0,0 +1,57 @@
...
+        require.resolve('@infisical/sdk');
+    } catch (e) {
+        console.error('PASSWORD_VAULT is enabled but @infisical/sdk is not installed. Run: npm install @infisical/sdk');
+        process.exit(1);
+    }
+
+    const scriptPath = path.resolve(__dirname, 'loadInfisicalSecretsChild.js');
+    try {
+        // Fetch secrets in a child process so this call remains synchronous for startup.
+        const output = childProcess.execFileSync(
```
huss: Exec always makes me nervous for security reasons. I see that no input values are used but env values are passed along (I think). Is this necessary vs. directly calling the function. I'm assuming yes but wanted to know why and if there is any reasonable alternative.

src/server/util/loadInfisicalSecretsChild.js
```diff
@@ -0,0 +1,65 @@
...
+// Need to be set in .env for Infisical support to work
+const clientId = env.INFISICAL_CLIENT_ID;
+const clientSecret = env.INFISICAL_CLIENT_SECRET;
+const siteUrl = env.INFISICAL_SITE_URL || 'https://app.infisical.com';
+const projectId = env.INFISICAL_PROJECT_ID;
+const environment = env.INFISICAL_ENVIRONMENT || 'dev';
+const secretPath = env.INFISICAL_PATH || '/';
+
+if (!clientId || !clientSecret) {
+    console.error('PASSWORD_VAULT is enabled but INFISICAL_CLIENT_ID or INFISICAL_CLIENT_SECRET is not set.');
```
huss: Maybe message should be more specific to say not set in .env file? Applies to all.

docker-compose.yml
```diff
@@ -47,13 +53,15 @@ services:
             - OED_MAIL_FROM=mydomain@example.com # The email address that the email will come from
             - OED_MAIL_TO=someone@example.com # Set the destination address here for where to send emails
             - OED_MAIL_ORG=My Organization Name # Org name for mail that is included in the subject
+            - PASSWORD_VAULT=no # Set to yes to load DB passwords from Infisical
```
huss, line 56: When the design doc  is in place for OED, it might be nice to put a link to it here so people can easily find the directions for enabling the vault.

huss (changes requested):
Thanks to @Zachary-Squires for looking into this, adding the code and the documentation. Overall, this seems good. I have not yet tried it because it is a draft and I wanted to get my first thoughts back to you. ~~I made a few comments to consider. I also put in a comment on the docs on your GitHub account. Please let me know if anything is not clear or you have any thoughts/questions.~~

src/server/util/loadInfisicalSecrets.js
Zachary-Squires: Fixed

src/server/util/loadInfisicalSecrets.js
Zachary-Squires: Changed it to just yes.

src/server/util/loadInfisicalSecrets.js
Zachary-Squires: To my understanding execFileSync means that no shell ever gets invoked, so that there's no injection risk. With exec there would be, so I understand the concern. I used execFileSync here to make sure that Infisical can get the secrets before the config loads because execFileSync keeps it paused until the child process is done.

src/server/util/loadInfisicalSecretsChild.js
Zachary-Squires: I'll make it more specific.

src/server/util/loadInfisicalSecrets.js
huss: Thank you. I read up on this and child_process is safer for the reasons you gave. Since the code uses the process.env it is indirectly getting some user input but it would be tricky to hack that. Also, the later parsing of the output uses JSON.parse that tries to avoid any injection, etc. Thus, it does seem safe. The code could use jsonschema as done in testing but I don't think that is necessary, esp since the two results are only used as passwords.

huss (changes requested):
@Zachary-Squires Thank you for the update addressing most of the comments. The one step left is to put in a pull request to the OED dev docs and then add the link in the one file (per other comment). After that someone else can finalize the work for general usage in OED and test the system. Please let me know if you need anything.

huss:
This is to document that the current developer will not be continuing this work. It will be done by others per the design doc linked in the description.

Valex609:
I'm working on this

huss:
> I'm working on this

@Valex609 Great. Could you please put a comment in issue #1650 so I can assign that to you. OED assigns issues for continuing PR work by someone else. Once you do that I can assign it to you. Thanks.

Also, please let me know if OED can help in any way.

PR #1610, opened 2026-04-20
fix: improve DB retry logic and prevent install loop + OOM handling (…

### LINKED ISSUE #1608: Non-deterministic database initialization caused by overlapping retry logic and async lifecycle handling
## Description
During local Docker startup, the application produces repeated PromiseRejectionHandledWarning messages and exhibits non-deterministic database initialization behavior. The issue appears to stem from overlapping retry logic between the bash installation script and Node.js database initialization code, combined with improper async lifecycle termination.

This results in noisy logs, repeated DB creation attempts, and unclear startup state during docker compose up.

## Reproduction Steps
Clone the repository
Ensure no existing containers or volumes:
```
docker compose down -v
docker system prune -af --volumes
```
Rebuild and start the stack:
```
docker compose build --no-cache
docker compose up
```
Observe startup logs during web service initialization

## Observed Behavior

During startup, the following occurs:

1. Node warning spam
(node:23) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: XXXXX)
(node:23) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: XXXXX)
(node:23) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: XXXXX)
2. Schema initialization retry behavior
npm run createdb is invoked repeatedly from installOED.sh
Node-side DB initialization (createDB.js) also performs async schema setup
These two mechanisms overlap and are not coordinated

## Expected Behavior
Database initialization should occur exactly once per startup
No repeated retries when DB init is already in progress
No PromiseRejectionHandledWarning logs
Deterministic startup lifecycle:
DB ready → schema created → web server starts

## Root Cause Analysis (based on investigation)
1. Overlapping retry systems
Bash script retries:

installOED.sh → npm run createdb (loop with retries)
Node script performs async DB initialization internally

Result: duplicate execution attempts and race conditions

2. Improper async lifecycle termination in Node

Current pattern in createDB.js:

(async function createSchemaWrapper() {
  try {
    await createSchema(conn);
    await insertStandardUnits(conn);
    await insertStandardConversions(conn);
    await redoCik(conn);
    process.exitCode = 0;
  } catch (err) {
    process.exitCode = 1;
  }
}());

## Issues:

Uses IIFE without explicit process termination
Relies on process.exitCode instead of deterministic process.exit()
Allows event loop to continue during Docker retry cycles
3. Delayed Promise rejection handling
Errors are not always handled immediately

Leads to Node emitting:

PromiseRejectionHandledWarning
Likely caused by asynchronous retry overlap between bash + Node

## Impact
Noisy and confusing startup logs
Unclear DB initialization state during development
Potential race conditions in schema creation
Reduced confidence in Docker startup reliability
Harder debugging due to repeated retries masking root errors

## Suggested Fixes
1. Make DB initialization deterministic (Node side)
Replace IIFE with explicit main function:
async function main() {
  try {
    const conn = getConnection();
    await createSchema(conn);
    await insertStandardUnits(conn);
    await insertStandardConversions(conn);
    await redoCik(conn);

    process.exit(0);
  } catch (err) {
    console.error(err);
    process.exit(1);
  }
}

main().catch(err => {
  console.error("Fatal error:", err);
  process.exit(1);
});
2. Remove overlapping retry logic (choose one system)
Either:
Keep bash retry loop
OR move retry logic into Node
Avoid dual-layer retries
3. Fix bash piping (error propagation)

Replace:
```
npm run createdb |& tee /tmp/oed.error > /dev/null
```
With:
```
set -o pipefail
npm run createdb 2>&1 | tee /tmp/oed.error
createdb_code=${PIPESTATUS[0]}
```
4. Add global async safety net (optional but recommended)
process.on("unhandledRejection", (err) => {
  console.error("UNHANDLED REJECTION:", err);
  process.exit(1);
});

## Notes
Issue is reproducible in clean Docker environment with no existing volumes
Database itself is not corrupted; issue is strictly initialization orchestration + async handling
Removing volumes does not resolve behavior

johnny603:
# Description

This change addresses non-deterministic database initialization during Docker startup that results in repeated PromiseRejectionHandledWarning logs, overlapping retry behavior, and inconsistent startup state.

The root issue is caused by two independent retry/initialization systems running concurrently:

installOED.sh retrying npm run createdb
Node.js createDB.js performing async schema initialization without deterministic process termination

Together, these create race conditions, repeated schema execution attempts, and noisy startup logs.

This change improves startup reliability by clarifying DB initialization flow and reducing overlapping execution paths.

Fixes #1608

## Type of change
Checklist
- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/)
 ideas
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/)
 and each author is listed in the Description section.

## Limitations

This fix improves initialization determinism and reduces retry-related race conditions, but does not fully eliminate underlying circular dependency complexity in the model layer (models/database.js).

Further improvements may include:

Refactoring schema initialization into a single orchestration layer
Reducing circular dependencies between model modules
Migrating DB initialization to a fully centralized startup service

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,117 +1,96 @@
-# *
-# * This Source Code Form is subject to the terms of the Mozilla Public
-# * License, v. 2.0. If a copy of the MPL was not distributed with this
-# * file, You can obtain one at http://mozilla.org/MPL/2.0/.
-# *
-
-version: "3.8"
 services:
-    # Database service. It's PostgreSQL, see the
-    # Dockerfile in ./database
-    database:
-        build: ./containers/database/
-        environment:
-            # Custom PGDATA per recommendations from official Docker page
-            - PGDATA=/var/lib/postgresql/data/pgdata
-            - POSTGRES_PASSWORD=pleaseChange # default postgres password that should be changed for security.
-        volumes:
-            - ./postgres-data:/var/lib/postgresql/data/pgdata
-        healthcheck:
-            test: ["CMD-SHELL", "pg_isready -U postgres"]
-            interval: 10s
-            timeout: 10s
-            retries: 3
-        # ports:
-        #    - "5432:5432"
-        # Uncomment the above lines to enable access to the PostgreSQL server
-        # from the host machine.
-        # Web service runs Node
-    web:
-        # Configuration variables for the app.
-        environment:
-            - OED_PRODUCTION=no
-            - OED_SERVER_PORT=3000
-            - OED_DB_USER=oed
-            - OED_DB_DATABASE=oed
-            - OED_DB_TEST_DATABASE=oed_testing
-            - OED_DB_PASSWORD=opened
-            - OED_DB_HOST=database # Docker will set this hostname
-            - OED_DB_PORT=5432
-            - OED_TOKEN_SECRET=?
-            - OED_LOG_FILE=log.txt
-            - OED_MAIL_METHOD=none # Method of sending mail. Supports "secure-smtp", "none". Case insensitive.
-            - OED_MAIL_SMTP=smtp.example.com # Edit this
-            - OED_MAIL_SMTP_PORT=465 # Edit this
-            - OED_MAIL_IDENT=someone@example.com # The user email that is used for sending emails (SMTP)
-            - OED_MAIL_CREDENTIAL=credential # Set the email password for sending email here
-            - OED_MAIL_FROM=mydomain@example.com # The email address that the email will come from
-            - OED_MAIL_TO=someone@example.com # Set the destination address here for where to send emails
-            - OED_MAIL_ORG=My Organization Name # Org name for mail that is included in the subject
-            # Changing this value does not impact what OED displays.
-            # What it will change is the date/time stamp on logs, notes and change dates that place the current date/time.
-            # It can also impact the interpretation of readings sent to OED such as Unix timestamps.
-            - TZ=Etc/UTC # Set the timezone of the Docker container where OED runs the web services.
-            # If in a subdirectory, set it here
-            # - OED_SUBDIR=/subdir/
-            # Set the correct build environment.
-        build:
-            context: ./
-            dockerfile: ./containers/web/Dockerfile
-        # Link to the database so the app can persist data
-        links:
-            - database
-        # Load the source code into the container.
-        # Using a volume allows autorebuild to work.
-        volumes:
-            - ./:/usr/src/app
-        # Map the default port.
-        ports:
-            - "3000:3000" # Should be commented out if you uncomment 80:3000 below.
-            - "9229:9229" # Debug port, should be commented out for production
-            # For production you might want something like:
-            # - "80:3000"
-            # and comment out the debug port and 3000:3000 line above
-            # Don't bring this up without the DB
-        depends_on:
-            # - database
-            database:
-              # We need the database and it has to be ready for work (see healthcheck above).
-              condition: service_healthy
-        healthcheck:
-            test: ["CMD", "curl", "-f", "http://localhost:3000"]
-            interval: 10s
-            timeout: 5s
-            retries: 5
-        # Lets docker compose up work right
-        # If environment variable install_args is not set then it becomes blank without warning user.
-        command:
-            [
-                "bash",
-                "./src/scripts/installOED.sh",
-                "${install_args:-}"
-            ]
-        # Use this if you are using a docker-compose that is earlier than version 2 and comment out the one above.
-        # command: ["bash", "./src/scripts/installOED.sh"]
+  # -------------------------
+  # PostgreSQL Database
+  # -------------------------
+  database:
+    build: ./containers/database/
+    environment:
+      PGDATA: /var/lib/postgresql/data/pgdata
+      POSTGRES_PASSWORD: pleaseChange
+    volumes:
+      - ./postgres-data:/var/lib/postgresql/data/pgdata
 
-    # Cypress testing service
-    cypress:
-      image: cypress/included
-      profiles:
-        - ui-testing
-      environment:
-        - CYPRESS_BASE_URL=http://web:3000
-        - DISPLAY=:99
-      working_dir: /usr/src/app
-      depends_on:
-        web:
-          condition: service_healthy
-      volumes:
-        - ./:/usr/src/app
-      entrypoint: >
-        /bin/sh -c "
-        rm -f /tmp/.X99-lock &&
-        Xvfb :99 -screen 0 1024x768x16"
-      read_only: true
-      security_opt:
-        - no-new-privileges:true
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
+      retries: 10
 
+    restart: unless-stopped
+
+  # -------------------------
+  # Web App (Node)
+  # -------------------------
+  web:
+    build:
+      context: ./
+      dockerfile: ./containers/web/Dockerfile
+
+    environment:
+      OED_PRODUCTION: "no"
+      OED_SERVER_PORT: "3000"
+      OED_DB_USER: "oed"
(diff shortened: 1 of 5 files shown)


huss:
@johnny603 You have these ideas in the description:

> Refactoring schema initialization into a single orchestration layer
> Reducing circular dependencies between model modules
> Migrating DB initialization to a fully centralized startup service

I think it would be valuable for OED to look into these. Would you like to open issue(s) about this? Any additional information on your thoughts are also very welcome.

docker-compose.yml
```diff
@@ -1,117 +1,96 @@
...
+
+  # -------------------------
+  # Web App (Node)
+  # -------------------------
+  web:
+    build:
+      context: ./
+      dockerfile: ./containers/web/Dockerfile
+
+    environment:
```
huss, line 29: I'm wondering about moving the environment variable location in this file. Is this your preference, a good idea or something else. I'm thinking about this because the OED site installation directions talk about these so moving them does have some pain threshold.

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
 #!/bin/bash
```
huss, line 1: The copyright notice was removed. OED wants that in every file that is possible.

Also, you removed or modified a comments in this file. I'm wondering why since some may have changed the meaning or removed comments designed to describe what was intended.

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
...
-			# Empty string so just ignore. Esp. happens if no install_args on docker compose up.
-			shift
-			;;
-		*)
-			printf "Unknown argument \"%s\"\n" $1
-			printf "%s\n" $USAGE
-			exit 1
-	esac
+  case "$1" in
+    --production) production=yes ;;
```
huss, line 25: The indenting was changed from tabs to spaces. While OED is not always consistent, esp. in script files, the project does prefer tab indenting when that is possible. Is there are reason not to use tab indenting?

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
...
+    --nostart) dostart=no ;;
```
huss, line 26: While your new formatting is okay, OED prefers each line of code on a separate line and having items inside a loop/logic indented on separate lines (so it is clear the association).

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
...
+    *) echo "Unknown argument: $1"; echo "$USAGE"; exit 1 ;;
+  esac
+  shift
 done
 
-# Load .env if it exists
+# ----------------------------------------------------
+# Load env
+# ----------------------------------------------------
+[ -f ".env" ] && source .env
```
huss, line 37: While this is the same, OED prefers the explicit if/else to make the logic clear. This is esp. true as we work with a lot of student developers who may not be a familiar with scripting syntax.

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
...
 
-if [ -f ".env" ]; then
-	source .env
+# ----------------------------------------------------
+# NPM install
+# ----------------------------------------------------
+if [ "$keep_node_modules" = "yes" ]; then
```
huss, line 42: I'm wondering why the logic about changed package files was removed. It was designed only do the install when package files were changed since OED only wants an update then and not every time it is run to keep all the file standard across implementations. This seems to run every time. Thoughts?

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
...
-	if [ $? == 0 ]; then
-		printf "\n%s\n\n" "NPM install finished."
-	else
-		# npm reported an error. Sometimes it does so can skip steps.
-		# Using printf since it is more reliable.
-		printf "\n%s\n\n" "NPM reported an error so stopping"
-		exit 2
-	fi
-fi
+  echo "Running database creation..."
```
huss, line 57: The script was using printf throughout to be consistent. It also did extra lines to make it easier to see certain items-steps. Could you explain why this was changed?

src/scripts/installOED.sh
```diff
@@ -1,198 +1,120 @@
...
-		./src/scripts/devstart.sh
-	fi
-else
-	printf "%s\n" "Not starting OED due to --nostart"
+# ----------------------------------------------------
+# Build frontend
+# ----------------------------------------------------
+if [ "$production" = "yes" ] || [ "$OED_PRODUCTION" = "yes" ]; then
+  npm run webpack:build
+elif [ "$dostart" = "no" ]; then
```
huss, line 105: You systematically changed the == to = for string comparisons. While they are the aliases for each other and = is POSIX compliant (we are tied to bash so less of an issue), the == was used to be explicit it is a string comparison. The reuse of = in other languages has led to many bugs (such as in C++ where you assign instead of just compare) so it was felt that being explicit was better here. What do you think?

huss (commented):
@johnny603 Thank you for thinking about this problem and addressing it. I have not done a complete review but have looked at this some. I probably have other comments but wanted to start here to begin the conversation. I've made some comments on this file to consider. In the bigger picture, where I try to start OED using ``docker compose up``, it fails to start and goes into a loop where it keeps reporting this error:

```
Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency
```

I'm assuming you don't see this. Do you have thoughts on what is happening?

Please let me know if anything is not clear, any thoughts or questions.

huss:
@johnny603 I was sweeping issues and wanted to see if you could give me any information on the status of this work. Thanks.

johnny603:
Thanks for checking in! Here's a quick breakdown of where things stand:

**What the PR addresses:**
This PR tackles installation and setup improvements, specifically around how OED handles its initialization workflow — ensuring the process is more reliable and predictable across different environments.

**What I've been working through:**
- Investigating issues that surfaced with the `installOED.sh` workflow in containerized environments
- Identifying edge cases in the setup flow that need to be handled before this is ready for review
- Running additional testing and validation to make sure the fix is solid

**Root cause & proposed fix:**
The `Warning: Accessing non-existent property 'sqlFile' of module exports inside circular dependency` error is caused by a circular dependency in `models/database.js`. During Node.js module initialization, when two modules import each other, the one loaded second receives an incomplete export object — so `sqlFile` resolves to `undefined` at that point.

The fix I'm planning is to convert top-level `require()` calls into lazy (inline) requires inside the functions that actually use them:

```js
// Instead of this at the top of the file:
const { sqlFile } = require('../models/database');

// Move it inside the function that needs it:
function someInitFunction() {
    const { sqlFile } = require('../models/database');
    // sqlFile is fully defined by the time any function is called
}
```

This is the least invasive fix and directly unblocks the startup crash without requiring a broader refactor. A deeper architectural improvement (centralizing DB init into a dedicated service) would be a good follow-up issue as you suggested.

**Current status:**
I haven't abandoned the PR and plan to work on this throughout this month due to school. I'll push updates as I make progress and flag here if anything needs further discussion.

Thanks for your patience!

Meanwhile you can check PR 1627 as I am sure it is ready to be rereviewed
https://github.com/OpenEnergyDashboard/OED/pull/1627

huss:
> The fix I'm planning is to convert top-level `require()` calls into lazy (inline) requires inside the functions that actually use them:

Okay. I looked at this a little to see the impact. It seems okay but there can be issues (https://reactnative.dev/docs/optimizing-javascript-loading).

> A deeper architectural improvement (centralizing DB init into a dedicated service) would be a good follow-up issue as you suggested.

If you want to write up the idea as an issue that would be nice. Then you or someone else can move it forward at some point.

Thanks for the update and I will get to the other PR.

huss:
@johnny603 Just seeing if you need anything.

huss:
This has not changed in a while so I'm making it available for anyone to work on. Contact OED to get started and be assigned this work.

PR #1611, opened 2026-04-22
Password Policy Implementation

### LINKED ISSUE #1612: Passwords longer than 72 bytes are silently truncated
### Describe the bug

OED does not properly handle passwords longer than 72 bytes. Since bcrypt only processes the first 72 bytes of a password, any additional bytes are silently truncated. 

### Expected behavior

Passwords longer than 72 bytes should be rejected during user creation. The system should return a validation error instead of allowing the password to be truncated.

### Screenshots

N/A

### Machine info (please complete the following information)

N/A

### Additional context

N/A

Zach-O-Bates:
# Description

This pull request resolves issues with password security and inconsistent validation behavior. A uniform password policy is now enforced across user creation and authentication, including minimum length requirements (8 characters for standard users and 14 for administrators), prevention of weak or common passwords using a strength estimator, and disallowing passwords that contain the username. Additionally, this PR addresses bcrypt’s 72-byte limitation by rejecting passwords that exceed this length using a UTF-8 byte check, preventing silent truncation before hashing. Tests were updated to validate both the password policy and truncation protection, ensuring these controls are consistently enforced.

Fixes #1612 

## Type of change

## Checklist

## Limitations
N/A

### DIFF
--- a/src/server/routes/users.js
+++ b/src/server/routes/users.js
@@ -11,6 +11,7 @@ const validate = require('jsonschema').validate;
 const { getConnection } = require('../db');
 const jwt = require('jsonwebtoken');
 const secretToken = require('../config').secretToken;
+const { validatePasswordPolicy } = require('../util/validatePassword');
 const { STRING_GENERAL_MAX_LENGTH, PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, TOKEN_MAX_LENGTH, USERNAME_MIN_LENGTH, USERNAME_MAX_LENGTH, NUMERIC_ID_MAX_LENGTH } = require('../util/validationConstants');
 
 const router = express.Router();
@@ -120,6 +121,21 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
 	} else {
 		try {
 			const { username, password, role, note } = req.body;
+
+			/* Determine the password’s size in bytes (using UTF-8 encoding) since some characters
+			take more than one byte, then reject the request if it exceeds 72 bytes to prevent
+			bcrypt from silently truncating the password before hashing */
+			const byteLength = Buffer.byteLength(password, 'utf8');
+			if (byteLength > 72) {
+				return res.status(400).send({ message: 'Password must not exceed 72 bytes.' });
+			}
+			
+			// Password policy validation
+			const errorMessage = validatePasswordPolicy(password, username, role);
+			if (errorMessage) {
+				return res.status(400).send({ message: errorMessage });
+			}
+
 			const conn = getConnection();
 			// Check if user already exists
 			const currentUser = await User.getByUsername(username, conn);
--- a/src/server/test/util/passwordPolicy.js
+++ b/src/server/test/util/passwordPolicy.js
@@ -1,3 +1,9 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
 const { chai, mocha, app } = require('../common');
 const expect = chai.expect;
 
--- a/src/server/test/util/passwordPolicy.js
+++ b/src/server/test/util/passwordPolicy.js
@@ -1,0 +1,66 @@
+const { chai, mocha, app } = require('../common');
+const expect = chai.expect;
+
+const { validatePasswordPolicy } = require('../../util/validatePassword');
+
+mocha.describe('Password Policy Validation', () => {
+
+  mocha.it('Rejects short passwords', () => {
+    const result = validatePasswordPolicy('short', 'user1', 'user');
+    expect(result).to.be.a('string');
+  });
+
+  mocha.it('Rejects admin passwords under 14 chars', () => {
+    const result = validatePasswordPolicy('shortpassword', 'admin1', 'admin');
+    expect(result).to.include('14');
+  });
+
+  mocha.it('Rejects passwords containing username', () => {
+    const result = validatePasswordPolicy('user1password', 'user1', 'user');
+    expect(result).to.include('username');
+  });
+
+  mocha.it('Rejects weak passwords (zxcvbn)', () => {
+    const result = validatePasswordPolicy('aaaaaaaa', 'user1', 'user');
+    expect(result).to.include('weak');
+  });
+
+  mocha.it('Accepts strong passwords', () => {
+    const result = validatePasswordPolicy('StrongPassphrase123!', 'user1', 'user');
+    expect(result).to.equal(null);
+  });
+
+});
+
+const chaiHttp = require('chai-http');
+chai.use(chaiHttp);
+
+mocha.describe('User Creation Password Policy', () => {
+
+  let adminToken;
+
+mocha.before(async () => {
+  const res = await chai.request(app)
+    .post('/api/login')
+    .send({
+      username: 'test@example.invalid',
+      password: 'password'
+    });
+
+  adminToken = res.body.token; 
+});
+
+  mocha.it('Rejects weak password on create', async () => {
+    const res = await chai.request(app)
+      .post('/api/users/create')
+      .set('token', adminToken)
+      .send({
+        username: 'testuser',
+        password: 'password',
+        role: 'user',
+        note: 'test'
+      });
+
+    expect(res).to.have.status(400);
+  });
+});
--- a/src/server/util/validatePassword.js
+++ b/src/server/util/validatePassword.js
@@ -1,6 +1,12 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
 const zxcvbn = require('zxcvbn');
 
-//Temproarly list, until we decide a better approach like a specific file
+//Temporary list, until we decide a better approach like a specific file
 const COMMON_PASSWORDS = new Set([
   'password',
   '12345678',
--- a/src/server/util/validatePassword.js
+++ b/src/server/util/validatePassword.js
@@ -1,0 +1,33 @@
+const zxcvbn = require('zxcvbn');
+
+//Temproarly list, until we decide a better approach like a specific file
+const COMMON_PASSWORDS = new Set([
+  'password',
+  '12345678',
+]);
+
+function validatePasswordPolicy(password, username, role = []) {
+  const minLength = role === 'admin' ? 14 : 8;
+  if (password.length < minLength) {
+    return `Password must be at least ${minLength} characters long.`;
+  }
+
+  // Username inclusion
+  if (username && password.toLowerCase().includes(username.toLowerCase())) {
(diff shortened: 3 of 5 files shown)


src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
+const { chai, mocha, app } = require('../common');
```
huss, line 7: This and the other new file are missing the copyright notice.

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
+const zxcvbn = require('zxcvbn');
```
huss, line 7: I think it might be good to put a TODO linking to https://github.com/dropbox/zxcvbn#script-load-latency. This isn't huge (400k) but it isn't trivial either. OED hopes to make a future effort to reduce the bundle size and these ideas might help. It might also be good to note that at this time it isn't used on the client.

huss:
I know @Zach-O-Bates analyzed this package but I wanted to leave a few notes from my looking. Positive is it has > 1M downloads a week and it seems like a decent package for this. Having said that, it does not seem to be maintained even given that not many changes might be needed. The last change was in 2017 and issues/pull requests are not being processed, including ones with security issues. I also saw that they used English to build the model. That is okay but may mean it isn't as good in other languages and OED is internationalized for usage around the world. So, I think it is fine to use this package, esp. compared to doing nothing, but this and any thoughts by others may help with future reconsideration.

Update: I did not try to find a lot of updates but I did find zxcvbn-ts that appears to be currently active. It isn't used as much but something to consider along with other ones that are found.

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+
+//Temproarly list, until we decide a better approach like a specific file
```
huss: ~~spelling: Temproarly.~~ Also, I think this should be a TODO as an action is needed. It also might be nice to say what a good list would contain.

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+const COMMON_PASSWORDS = new Set([
+  'password',
```
huss, line 11: Space intending instead of tab. Same in other file.

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+  '12345678',
+]);
+
+function validatePasswordPolicy(password, username, role = []) {
```
huss, line 15: JSDoc would be good.

I see where this is used in src/server/routes/users.js for /create. I'm wondering about all the other places where passwords are entered: Same route in /edit, src/server/services/user/createUser.js & src/server/services/user/editUser.js. Also, this could be used on the front-end to give feedback without contacting the server but that would mean putting this into the client bundle and this isn't done very often so I'm okay with not doing that now (or maybe ever).

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+  const minLength = role === 'admin' ? 14 : 8;
+  if (password.length < minLength) {
+    return `Password must be at least ${minLength} characters long.`;
```
huss, line 18: I think it might be good to add "for role ${role}" at the end since people might be confused as it varies by role.

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  // Common passwords
+  if (COMMON_PASSWORDS.has(password.toLowerCase())) {
+    return 'Password is too common.';
+  }
+
+  // Strength check
+  const result = zxcvbn(password);
+  if (result.score < 3) {
```
huss, line 33: I'm thinking about several things related to score usage:

1. Should the required score relate to the type of user as does the length? If so, any thoughts on what it might be?
2. I'm thinking this should be added to the admin settings page and DB so sites could increase/decrease the score as desired. OED should warn if it is a low value and a security concern. This can be a new issue since I'm unsure you want it as part of this effort and the current setup is acceptable. What do you think?
3. https://github.com/dropbox/zxcvbn#usage indicates the package can give an estimate of the number of tries it takes to break the password, warnings and feedback for that password. Should these be used?

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+    return `Password must be at least ${minLength} characters long.`;
+  }
+
+  // Username inclusion
+  if (username && password.toLowerCase().includes(username.toLowerCase())) {
+    return 'Password must not contain your username.';
+  }
+
+  // Common passwords
+  if (COMMON_PASSWORDS.has(password.toLowerCase())) {
```
huss, line 27: zxcvbn has an optional argument of user_inputs that seems to not allow those as passwords. Maybe that should be checked to see if direct usage or inside the password is excluded. If so, would it make sense just to use it?

src/server/util/validatePassword.js
```diff
@@ -0,0 +1,33 @@
...
+    return 'Password is too common.';
+  }
+
+  // Strength check
+  const result = zxcvbn(password);
+  if (result.score < 3) {
+    return 'Password is too weak.';
```
huss, line 34: See above for ideas about making the feedback more informative.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+const expect = chai.expect;
+
+const { validatePasswordPolicy } = require('../../util/validatePassword');
+
+mocha.describe('Password Policy Validation', () => {
+
+  mocha.it('Rejects short passwords', () => {
+    const result = validatePasswordPolicy('short', 'user1', 'user');
+    expect(result).to.be.a('string');
```
huss, line 16: For this and the other tests: I think that one could get more info and test the result in a stronger way. See other file about providing more feedback for the user that might help here.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
+const { chai, mocha, app } = require('../common');
+const expect = chai.expect;
+
+const { validatePasswordPolicy } = require('../../util/validatePassword');
+
+mocha.describe('Password Policy Validation', () => {
+
+  mocha.it('Rejects short passwords', () => {
+    const result = validatePasswordPolicy('short', 'user1', 'user');
```
huss, line 15: "user" is not a valid user role. While it isn't "admin", and I think that is what is intended, it really should use valid roles. Note User.role has the allowed roles that would be good to use here and probably as a test in the actual function.

Also, should this test 7 characters since that is right on the border. Even better if the other file had exported constants that would define these values and then used here. Along those lines, the new, stronger parameter testing on routes has PASSWORD_MAX_LENGTH in src/server/routes/obvius.js. I think the value you have and this needs to be tied together. This probably means making the tested value be dependent on the role and some other changes. A sweep of all places using a password and length test should be made.

Note some tests below have similar considerations.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(result).to.be.a('string');
+  });
+
+  mocha.it('Rejects admin passwords under 14 chars', () => {
+    const result = validatePasswordPolicy('shortpassword', 'admin1', 'admin');
```
huss, line 20: It might be worth a comment that the password is 13 characters - just under the min needed. If this converts to using consts for length then it should generate the correct length of password to try based on that.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(result).to.include('14');
+  });
+
+  mocha.it('Rejects passwords containing username', () => {
+    const result = validatePasswordPolicy('user1password', 'user1', 'user');
```
huss, line 25: Would this be a stronger test if the user name was in the middle of the password and not at the start?

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  mocha.it('Accepts strong passwords', () => {
+    const result = validatePasswordPolicy('StrongPassphrase123!', 'user1', 'user');
+    expect(result).to.equal(null);
+  });
+
+});
+
+const chaiHttp = require('chai-http');
```
huss, line 41: OED prefers all require/include at the top of a file.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+chai.use(chaiHttp);
+
+mocha.describe('User Creation Password Policy', () => {
+
+  let adminToken;
+
+mocha.before(async () => {
```
huss, line 48: Indenting off.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+  const res = await chai.request(app)
+    .post('/api/login')
+    .send({
+      username: 'test@example.invalid',
```
huss, line 52: This really should use testUser to get the info as in other tests.

src/server/test/util/passwordPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+      password: 'password'
+    });
+
+  adminToken = res.body.token; 
+});
+
+  mocha.it('Rejects weak password on create', async () => {
```
huss, line 59: These will need to be expanded if other routes are used during password testing per another comment.

src/server/routes/users.js
```diff
@@ -120,6 +121,21 @@ router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) =
 	} else {
 		try {
 			const { username, password, role, note } = req.body;
+
+			/* Determine the password’s size in bytes (using UTF-8 encoding) since some characters
+			take more than one byte, then reject the request if it exceeds 72 bytes to prevent
+			bcrypt from silently truncating the password before hashing */
+			const byteLength = Buffer.byteLength(password, 'utf8');
+			if (byteLength > 72) {
+				return res.status(400).send({ message: 'Password must not exceed 72 bytes.' });
```
huss, line 130: All the numeric values are being replaced by constants. Thus, this should happen when PR 1581 is merged.

I also think the message could be more user friendly. Maybe "Password must not exceed 72 bytes which often means 72 characters."

huss (changes requested):
@Zach-O-Bates Thank you for looking into this. I've made a number of comments. I tried to do a good review but given the possible changes that may be made, other thoughts may come at a later time. I also only did minimal testing at this point. Please let me know if anything is not clear, you have thoughts/questions or I can help.

huss:
I also forgot to add this: How will current site passwords be handled that do not meet the new standards? Do they need to be reentered or do they just stay until updated? If forcing to reenter then how should that be done? Any thoughts?

huss:
@Zach-O-Bates I looked at your update from a few weeks ago and found it resolved a couple of minor comments which I marked as done. There still remains a number of unresolved comments. At this point I wanted to touch base to see if you intend to address them or if this work needs to be passed on to someone else. I appreciate any information.

huss:
This is to note that PR #1554 has a simple length test for quality of new database passwords in src/server/util/changePostgresPasswords.js. There is a TODO saying this new work should be added to that when it is ready.

huss:
@Zach-O-Bates I was sweeping issues and wanted to see if you could give me any information on the status of this work. Thanks.

PR #1620, opened 2026-05-01
1563 enhance route checks for moment items

### LINKED ISSUE #1563: enhance route checks for moment items

PR #1528 added checks for time and duration in a route. See src/server/routes/compareReadings.js with ``if (!isValidIsoDateTime(currStartRaw) || !isValidIsoDateTime(currEndRaw) || !isValidIsoDuration(shiftRaw)) {``. This seems valuable and it should be considered for all other routes. If any other tests seem valuable then they can also be proposed.

Roll in these new tests as desirable.

The routes will not have this checking.

### Additional context

~~@omarraf did the original test and may also continue with this work. Anyone else interested in working on this should first check if @omarraf is working on this.~~ Obsolete given comment below so this is available for anyone.

GoodKimchi:
## Description

Closes #1563

This PR enhances route-level validation for time and duration inputs across relevant endpoints.

- Added validation for `timeInterval` inputs
- Centralized validation logic in `util/timeValidation.js`
- Enforced strict ISO datetime and duration formats
- Added test coverage for route parameter validation and time validation utilities
- Ensured consistent validation across readings, unitReadings, logs, meters, and preferences routes

## Type of change

- [ ] This change requires a documentation update.

## Checklist

- [x] I have removed text in () from the issue request
- [x] I acknowledge that every person contributing to this work has signed the OED Contributing License Agreement

## Limitations

- CSV pipeline changes were removed from this PR to keep scope aligned with route validation
- Additional validation for non-route inputs can be handled in a separate issue


huss:
@GoodKimchi I'm sorry for the slow reply.  Unfortunately, per a recent Discord msg, I won't be able to do a review for a number of days. I'm sorry about this.

huss:
@GoodKimchi & @audreydng First, I'm very sorry this took so long for me to get to.

Thank you for this contribution. The PR description states the CLA has been signed by all contributors. However, OED records do not have one for any of you. I'm sorry I did not notice this sooner. Thus, I would like you to:

- Please can each of you sign the CLA via the link in the description or let me know that you think OED records are off.

Once this is done, this PR can be formally reviewed. Please let me know if you have any questions/thoughts.

GoodKimchi:
We have now both signed the CLA. Thank you!

src/server/util/timeValidation.js
```diff
@@ -0,0 +1,47 @@
...
+	return moment.parseZone(value, moment.ISO_8601, true).isValid();
+}
+
+/**
+ * Returns true if value is a valid ISO 8601 duration string (e.g. P1Y2M3DT4H5M6S).
+ * @param {string} value
+ * @returns {boolean}
+ */
+function isValidIsoDuration(value) {
+	return ISO_DURATION_REGEX.test(value);
```
huss: I know this is how the original test did this. However, given this is for general use, I decided to think about it more. moment does have a duration check as isDuration (https://momentjs.com/docs/#/durations/is-a-duration/). If you take the string, create a duration and then apply this function it will return false if it is invalid. From reading up (e.g., https://stackoverflow.com/questions/38692972/how-to-validate-a-moment-js-duration but this is older), it does not detect certain bad inputs (may return 0) or validate it is only ISO but not positive. I think the check code is at https://github.com/moment/moment/blob/develop/src/lib/duration/valid.js and used in the constructor at https://github.com/moment/moment/blob/develop/src/lib/duration/constructor.js. If we never send zero length durations then the info above may allow for a stronger test but I have not looked at it deeply. I think it is worth using the isDuration check as the first shot, maybe with extra checks for zero if we can. Then it can be combined with the regular expression (if needed) via &&. Some testing might be needed to see what is best. What do you think? Do you want to look into this?

src/server/util/timeValidation.js
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Returns true if value is a valid timeInterval string as used by OED's TimeInterval class.
+ * Accepted forms: 'all', 'ISO_ISO', 'ISO_' (right unbounded), '_ISO' (left unbounded).
+ * Each non-empty timestamp component must be a valid ISO 8601 datetime.
+ * @param {string} value
+ * @returns {boolean}
+ */
+function isValidTimeInterval(value) {
+	if (value === 'all') return true;
```
huss: OED wants braces ({ }) around the clauses in ``if`` statement.

src/server/util/timeValidation.js
```diff
@@ -0,0 +1,47 @@
...
+	const underscoreIndex = value.indexOf('_');
+	if (underscoreIndex === -1) return false;
```
huss: I know what this is doing but I think a comment(s) on these lines might help others in the future.

src/server/util/timeValidation.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Returns true if value is a valid timeInterval string as used by OED's TimeInterval class.
+ * Accepted forms: 'all', 'ISO_ISO', 'ISO_' (right unbounded), '_ISO' (left unbounded).
+ * Each non-empty timestamp component must be a valid ISO 8601 datetime.
+ * @param {string} value
+ * @returns {boolean}
+ */
+function isValidTimeInterval(value) {
```
huss: I'm not positive but I think the right/left bounded is only allowed on line readings requests. If that is correct, would it make sense to add a parameter (maybe with default value of false?) that indicates if those are allowed for testing? The idea is if they are not allowed then they would be rejected. I think the other routes need to be checked (maybe by use/testing) to see if this is okay to use. What do you think? If yes, can you do the testing or should I?

src/server/routes/readings.js
```diff
@@ -42,23 +43,12 @@ router.get('/line/count/meters/:meter_ids', optionalAuthMiddleware, async (req,
 	};
 	if (!validate(req.params, validParams).valid || !validate(req.query, validQueries).valid) {
 		res.sendStatus(HTTP_CODES.BAD_REQUEST);
+	} else if (!isValidTimeInterval(req.query.timeInterval)) {
```
huss: What you did was fine. However, given no values are stored and it returns the same code should the ``else if`` be removed with the logic added to the ``if`` or did you want more parallelism with other routes?

src/server/routes/readings.js
```diff
@@ -42,23 +43,12 @@ router.get('/line/count/meters/:meter_ids', optionalAuthMiddleware, async (req,
...
-			return;
-		}
-		if ((timeInterval.startTimestamp && !timeInterval.startTimestamp.isValid())
-			|| (timeInterval.endTimestamp && !timeInterval.endTimestamp.isValid())) {
-			log.warn(`Invalid moment parsed for readings count: ${req.query.timeInterval}`);
-			res.sendStatus(HTTP_CODES.BAD_REQUEST);
-			return;
-		}
+		const timeInterval = TimeInterval.fromString(req.query.timeInterval);
 		try {
```
huss, line 49: I realize the fromString above should not fail but given the try/catch was removed I'm thinking the main try should encompass that. It might as well go above the conn to be really safe.

src/server/routes/readings.js
```diff
@@ -107,24 +97,13 @@ router.get('/line/raw/meter/:meter_id', optionalAuthMiddleware, async (req, res)
 	};
 	if (!validate(req.params, validParams).valid || !validate(req.query, validQueries).valid) {
 		res.sendStatus(HTTP_CODES.BAD_REQUEST);
```
huss: See above for comments that apply here.

src/server/routes/preferences.js
```diff
@@ -109,8 +111,15 @@ router.post('/', adminAuthMiddleware('edit site preferences'), async (req, res)
 			}
 		}
 	};
+	const prefs = req.body.preferences || {};
```
huss: I believe you are trying to protect if the body has no preferences item. However, I think the validate() only passes the req.body and will fail if the required preferences is not present. Given this, is this necessary? I think stopping without resetting passed values is better if possible.

huss (commented):
@GoodKimchi & @audreydng Thank you for this contribution to OED. Review and testing found it generally works as desired. I merged in the current development and pushed that change since the missing changes made it impossible to easily start OED. I've made a few comments to consider. Also, I tried to check all the routes for any places that needed the new testing. Overall, you were doing them but noticed:

- getLogsByDateRangeAndType in src/server/routes/logs.js seems to accept a timeInterval. There is a comment about why it is not currently checked but hopefully it can now be checked.
- Two routes in src/server/routes/maps.js have modifiedDate and I wonder if that should be checked.

Also, I ran the full test suite (npm run test) and it had 4 failed tests:

```
1) Compare Readings Parameter Validation
       GET /api/compareReadings/meters/:meter_ids - Meter Compare Readings
         Query Parameter Validation
           should reject invalid curr_start format:

      AssertionError: expected 200 to equal 400
      + expected - actual

      -200
      +400
      
      at Context.<anonymous> (src/server/test/routes/compareReadingsParamsTest.js:111:28)

  2) Compare Readings Parameter Validation
       GET /api/compareReadings/meters/:meter_ids - Meter Compare Readings
         Query Parameter Validation
           should reject invalid curr_end format:

      AssertionError: expected 200 to equal 400
      + expected - actual

      -200
      +400
      
      at Context.<anonymous> (src/server/test/routes/compareReadingsParamsTest.js:127:28)

  3) timeValidation utility
       isValidIsoDateTime
         should reject date-only strings (no time or timezone):

      2023-01-01
      + expected - actual

      -true
      +false
      
      at Context.<anonymous> (src/server/test/util/timeValidationTests.js:38:41)
      at process.processImmediate (node:internal/timers:476:21)

  4) timeValidation utility
       isValidTimeInterval
         should reject strings with invalid ISO timestamps:

      2023-01-01_2023-12-31
      + expected - actual

      -true
      +false
      
      at Context.<anonymous> (src/server/test/util/timeValidationTests.js:119:42)
      at process.processImmediate (node:internal/timers:476:21)
```

Please let me know if anything is not clear or you have thoughts/questions.

audreydng:
Hi @huss, 

Thank you for the detailed feedback and for catching the failing tests.

I've received your comments. Since there are a few failing tests regarding the ISO date validation and some architectural suggestions (like using `moment.duration()` and checking other routes), I need to dive into the codebase and test everything locally first. 

I'll review all of your comments thoroughly and update the PR as soon as the fixes are ready!

huss:
@GoodKimchi & @audreydng It has been about a month since we commented so I wanted to touch base to see if you were making progress or needed anything.

src/server/util/timeValidation.js
audreydng: I looked into this and updated the duration validation. I kept the regex as well because moment doesn't check for ISO-style duration strings.

I also added a positive duration check, so zero-length durations such as `PT0S` are rejected.

audreydng:
Hi @huss, I went through the review comments and made a few updates.

For the shared time validation helper, I updated `isValidIsoDateTime()` so it rejects date only strings like `2023-01-01`. Moment's strict ISO parsing still treats those as valid, so I added a simple regex check before calling Moment.

For durations, I added the` moment.duration() / moment.isDuration()` validation you suggested. I kept the regex as well so we only accept ISO duration strings, and I also reject zero and negative durations.

For `isValidTimeInterval()`, I added `allowOneSided` = false as the default. My thinking was that the shared helper should stay strict by default, and only the routes that already support one-sided intervals should enable them explicitly. So logs uses the default behavior, while readings and unit readings pass true.

For compare readings, I kept support for the existing `YYYY-MM-DD HH:mm:ss` date-time format. It includes a time, but it is not strict ISO because it uses a space instead of `T`. The compare specific validator preserves that behavior while still rejecting date-only values like `YYYY-MM-DD`. So I added a compare specific validator that still rejects date-only strings but continues to support that format to avoid breaking the charts. 

I also cleaned up the readings validation logic, moved `TimeInterval.fromString()` inside the try block, removed the unnecessary preferences fallback, and added modifiedDate validation for `maps`. 

I ran the unit tests successfully, with 921 tests passing and 1 pending.

Please let me know if you have any other suggestions or if you would prefer any of these changes to be handled differently.

src/server/util/timeValidation.js
audreydng: I agree that the left/right bounded should not be allowed everywhere by default. I think that the shared helper should stay conservative by default such as logs , and only routes that is right/left bounded should explicitly pass `true`. I added an allowOneSided parameter to isValidTimeInterval() with a default value of false, so these intervals are rejected unless allowed. I also added tests for both cases.

src/server/routes/preferences.js
audreydng: Yeah, I agree that this is not needed. I removed it and simplified the update flow with early returns.

src/server/routes/readings.js
audreydng: Agreed, I simplified the conditional statement.

src/server/util/timeValidation.js
huss: I don't see new comments. If what I want is unclear then let me know and I could easily add them if that is okay with you.

huss (changes requested):
@GoodKimchi & @audreydng Thank you for the updated code. I appreciate you being careful in your changes. It addressed all but one minor comment where I can help (see comment) as needed. I think this is virtually ready to merge.

audreydng:
Hi @huss,

Thanks for reviewing our updates. I added comments in `isValidTimeInterval` covering the unbounded and the left/right bounded cases. Does this meet your expectations? If you wanted the comments in a different location or a different format, I'm happy to update them.

huss:
@audreydng First, thanks for the quick commenting and that looks great. I thought this was ready to merge when CodeQL found a security concern. I had to think about it for a minute but I think they are right. Because this is JS (not TS), the type of arguments is not strictly enforced. Thus, when you did indexOf() on the strings it is possible to pass an array instead of a string and trick the code. Since these come from routes, this can be an issue. It pointed (not sure if you can see the alert) to these lines: src/server/util/timeValidation.js:43, src/server/util/timeValidation.js:37 & src/server/routes/logs.js:88. I'm hoping those are the only ones. I think if the values were checked to be strings before using that would clear the concern. Having said that, the parameter checking probably also covers this. If so we could skip the check but it is cheap/fast so maybe easier to check string than dismissing the security warning. Do you agree? Are you comfortable doing this?

audreydng:
Hi @huss,

Yes, I agree with the concern.

I added explicit string checks in the time validation helpers at the beginning. That covers the path from logs.js into isValidTimeInterval, and also protects the other routes that reuse the same helpers. I also updated the tests for non-string values, including array inputs that could be produced by duplicate query parameters.

I ran full test and also checked CodeQL locally. Everything looks good. There are still a few CodeQL findings for logs.js, but they are related to potential log injection (writing untrusted data to log files). I think it is outside the scope of this issue.

Let me know if you'd like to make any additional changes or updates.

huss:
@GoodKimchi Thanks for the quick update and doing the extra work.

> I added explicit string checks in the time validation helpers at the beginning. That covers the path from logs.js into isValidTimeInterval, and also protects the other routes that reuse the same helpers. I also updated the tests for non-string values, including array inputs that could be produced by duplicate query parameters.

I really appreciate you adding the new testing to validate the security fix concern. Everything looks good to me.

> I ran full test and also checked CodeQL locally. Everything looks good. There are still a few CodeQL findings for logs.js, but they are related to potential log injection (writing untrusted data to log files). I think it is outside the scope of this issue.

I agree there are some other issues listed but none relate to this PR. OED will clean them up later and some are already underway. I'm less concerned since (I think without careful checking) all the problematic routes involves some form of login so it is much less likely that a malicious input would be sent.

huss (approved):
@GoodKimchi Thank you for this contribution and all of your efforts. This is now complete and ready to merge (per other comments). Congratulations on your first accepted contribution to OED.

Issue #962, opened 2023-07-06
testing readings


huss:
### Is your feature request related to a problem? Please describe.

The[ Design Document on testing](https://github.com/OpenEnergyDashboard/DesignDocs/blob/main/testing/testing.md) specifies readings tests that are desired. Once a test is complete and merged it is labeled as done. We would like the others completed.

### Describe the solution you'd like

Implement the tests. This is an issue that is generally done by a new developer to get used to working with OED. They normally do 1 or 2 tests for this process and then move on to something else so there are tests for others.

### Additional context

See information in the design document on doing a test, You must allocate the test as described before beginning work on a test.

aliceduongg:
Hi, I am working with this file (in Line Readings) with the description: "should have daily points for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as kg of CO2"

dzheng13341:
Hi, I will work on the file with the description: "should have daily points for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as metric ton of CO2 & chained"

aliceduongg:
Hi, I am working with this file (in Line Readings) with the description: "should have daily points for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as lbs of CO2 & chained & reversed"

mrathgeber:
I have completed test cases with the descriptions:

- partial days/hours for daily gives only full days (line reading)
- 13 day bars for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as kWh (bar reading)

Dominier:
Hello, I am working with bar reading test case B14 with the description: "1 day bars for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as lbs of CO2 & chained & reversed"

TamNganLy:
Hi, I am working on test B11: "1 day bars for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as BTU reverse conversion"

ngocvbh:
Hello, I am working on line reading test L20: "should have daily points for 15 minute reading intervals and quantity units with +-inf start/end time & kWh as lbs of CO2 & chained & reversed"

amancior23:
Hello my team is working on B13

alexis0705:
Hello my team is working on test B8

abccodes:
Hello, my team is working on test B12.

yadhiramorales:
Hello, my team is working on test B6.

[...]

PR #1626, opened 2026-05-20
Fix: LR14 expected value

Angel43v3r:
# Description

On test case LR14, u6 preferredDisplay was set to false. In the #962 , the expected value for u6 preferredDisplay is true. 
Updated preferredDisplay: false -> preferredDisplay: true

Contributors:
@Angel43v3r 
@OrgitoV 

Partly Addresses #962 

## Type of change

## Checklist

## Limitations

No issues known


src/server/test/web/readingsLineMeterRangeRaw.js
huss, line 188: I'm unsure why the line above that had ``// Add LR22 here`` is listed as deleted in this PR since I don't see that line in the current development branch. However, it should not be there so this is simply a comment about it.

huss (approved):
@Angel43v3r & @OrgitoV Thank you for this PR. Review and testing found it is fine. For the record, I'm noting that this currently does not change the test behavior since preferred display of units are not yet implemented but it is good to have this correct in case a future change would have caused an issue. Congratulations on another accepted contribution to OED.

PR #1628, opened 2026-05-22
Show additional pipeline values in logStatus output

### LINKED ISSUE #1618: show more pipeline values in logging

src/server/services/pipeline-in-progress/processData.js has a logStatus() function to standardize the information provided when logging is done. This is supposed to give a full dump of the status and values that are being used. More values have been added over time that are not passed to or displayed by this function. They should be shown for complete information.

These values (at the current time) are not passed or used by logStatus():
- honorDst
- relaxedParsing
- useMeterZone
- warnOnCumulativeReset
- useMeterFrequency
- useMeterFrequencyVariation

They should be added as parameters in the function definition and added to each call via the arguments. This list was easily created by looking at the parameters for processData() and seeing if there are new ones at the end. It is possible more will be added so checking again would be good. Finally, the string used for logging needs to be updated with these values within logStatus() and the values can go at the end in the order they appear in the processData() parameters.

None if we want to show the values

### Additional context

None

desusaiteja:
Resolves #1618

Adds the newer pipeline flags to `logStatus()` and updates each call site so they appear in the standardized logging output.

Also adds a focused server-side test for the expanded log message. CLA submitted as per the guidelines.

### Verification

- `node --check src/server/services/pipeline-in-progress/processData.js`
- `node --check src/server/test/db/processDataTests.js`
- `./node_modules/.bin/mocha --timeout 15000 "src/server/test/db/processDataTests.js"`


oed-pr-body.md
```diff
@@ -0,0 +1,11 @@
+Resolves #1618
```
huss: This seems to be a description of your work that is in the PR description. It should not be part of the OED repository and this file should be removed.

src/server/test/db/processDataTests.js
```diff
@@ -0,0 +1,59 @@
+/*
```
huss: I think this is a nice test of the work you did. My thoughts on it are:

- It does not test any other values in the log msgs so it isn't a general test of the code.
- OED has not tested logging across the code base so this is different than the rest of the code. That does not mean it is bad/wrong but if OED starts doing this then it should be continued across more of the code base. Given logging has not direct impact on the OED results, I have mixed feelings about doing this.
- A stronger test would be to make sure all the values passed to the function are actually shown so any new ones would not be missed in the future. That would be more complex but likely of greater value moving forward.

Given all of this, I am inclined to remove this test even though it is a nice usage of sinon. I welcome your thoughts.

Note if this is removed then the new export at the end of processDataTests.js can be removed.

src/server/services/pipeline-in-progress/processData.js
```diff
@@ -837,13 +842,17 @@ function row(index, isAscending, length) {
  * @returns {string} The message just logged.
  */
 function logStatus(meterName, rowNum, prevReading, currentReading, timeSort, readingRepetition, isCumulative,
-	cumulativeReset, resetStart, resetEnd, readingGap, readingLengthVariation, onlyEndTime) {
+	cumulativeReset, resetStart, resetEnd, readingGap, readingLengthVariation, onlyEndTime, honorDst,
+	relaxedParsing, useMeterZone, warnOnCumulativeReset, useMeterFrequency, useMeterFrequencyVariation) {
 	let message = 'For reading #' + rowNum + ' on meter ' + meterName + ' in pipeline: ' + 'previous reading has value ' + prevReading.reading + ' start time '
```
huss, line 847: I ran the pipeline tests (``npm run testsome src/server/test/web/csvPipelineTest.js``) and 57 failed. I forgot when I created the issue that many of these tests check the returned information that is modified by this work. As a result, the new values need to be added to each of these tests. This is non-trivial work that will require looking at each failure message to see it is okay and copying the new values into the test code. Is this something you are willing to do? If not, someone else can do it but it needs to be done for this to merge.

Note I think this will effectively test what you added in processDataTests.js and I commented on it in that file. Given that, I am even more inclined to remove that test.

I also ran the full tests (``npm run test``) and no other tests failed due to your changes.

huss (changes requested):
@desusaiteja Thank you for this contribution. Review and testing found it works as desired. I made a few comments to consider. One involves an oversight in the issue that will require more work. If you want to discuss then just let me know. If anything is not clear or you have questions/thoughts then just let me know.

src/server/services/pipeline-in-progress/processData.js
desusaiteja: Thanks for catching this. I updated `csvPipelineTest.js` so the expected pipeline log output includes the newly added status fields, including the `honorDst` cases. I also reran `./node_modules/.bin/mocha --timeout 15000 "src/server/test/web/csvPipelineTest.js"`, and it passes locally with 67 passing tests.

oed-pr-body.md
desusaiteja: Good catch. I removed `oed-pr-body.md` from the branch since that content belongs in the PR description, not the repository.

src/server/test/db/processDataTests.js
desusaiteja: Thanks, that makes sense. Since the existing CSV pipeline tests cover the affected output, I removed the focused logging test and the temporary `logStatus` export.

huss:
@desusaiteja I'm sorry it took me so long to reply. I had an unexpected GitHub access issue while traveling. I am hoping to review this PR in the near future.

src/server/test/web/csvPipelineTest.js
```diff
@@ -32,6 +32,17 @@ const CHAI_METERS_REQUEST_EMAIL = `chai.request(app).post('${UPLOAD_METERS_ROUTE
 // but all other keys are arrays of length number of uploads in test.
 // Note the use of double quotes for strings because some have single quotes within.
 
+function appendPipelineLogStatusValues(message, honorDst = false) {
```
huss: First, I want to complement you on how you addressed the issue of new values in the information returned from the route. It was elegant and it makes it easier to change in the future.

There are just a few minor thoughts on this:

- It would be nice to have JSDoc including what these two new functions are designed to do. It is clear to me but may not be for people in the future.
- I tried to figure out why honorDst has a default value of false. It seemed to me that it should always be true/false if properly used. Is that correct? If so, would it just print the bogus value if it was not (such as undefined) and then likely fail the test (which is likely to catch the bad usage)? If this is correct, you can either leave the default value with a comment explaining this or remove it. I know some people prefer default values, others don't in some cases and some have parameter checks to see it is set in a reasonable way. All are fine.

huss (changes requested):
@desusaiteja Thank you for the updated code. Review and testing found it works as desired. I made one small comment on the new functions to consider. Please let me know if anything is not clear or you have thoughts/questions.

src/server/test/web/csvPipelineTest.js
desusaiteja: Thank you for the feedback @huss. 

1. I agree, I’ll add JSDoc for both helper functions.
2. Good point. With the current default, an omitted second argument would silently become `false`, so it would not expose bad usage.I think removing the default is better here. I’ll remove the default and add JSDoc documenting that `honorDst` is the expected value for the pipeline log field, derived from whether the upload request includes `.field('honorDst', true)`. Let me know if that sounds okay to you?

huss: > 2\. I think removing the default is better here. I’ll remove the default and add JSDoc documenting that `honorDst` is the expected value for the pipeline log field, derived from whether the upload request includes `.field('honorDst', true)`. Let me know if that sounds okay to you?

Sound good to me.

huss (approved):
@desusaiteja Thank you for addressing my comment. Everything looks good. Congratulations on your first accepted contribution to OED.

desusaiteja:
> @desusaiteja Thank you for addressing my comment. Everything looks good. Congratulations on your first accepted contribution to OED.

@huss  Thank you and am glad my first PR got merged finally. Please let me know if you have any potential next issue or enhancement in mind, I would love to contribute more

huss:
> lease let me know if you have any potential next issue or enhancement in mind, I would love to contribute more

@desusaiteja OED welcomes people to continue to contribute. I prefer to understand what people are interested in and make sure tasks are well matched to people. Thus, I like to chat with the person. If you use Discord, you could go to the OED server (directions on the [OED contact page](https://openenergydashboard.org/contact/)) and message me (It is easy to find me there). That might be a better forum for this than a closed PR. Let me know if you want to find another way to do this.

PR #1642, opened 2026-06-13
Issue 1473 Translate Improvements

### LINKED ISSUE #1473: untranslated compare line strings
### Describe the bug

There are several strings in the newer compare line component that are not translated.

### To Reproduce

If you change language they will not be translated when using the compare line plot.

### Expected behavior

src/client/app/components/CompareLineChartComponent.tsx has the following untranslated strings:

- Shifted
- The original line has ... (only part of string shown here)
- The shifted line overlaps the original line ... (only part of string shown here)
- Days of week align (unless missing readings)
- The month and day of the month align for the original and shifted readings
-  until original reading at date (only part of string shown here)

These should be translated as other strings are done and the key/values placed in src/client/app/translations/data.ts. Note some are not in components but functions in a component file and may need different treatment. PR #1471 did similar work if it is needed.

Nespina24:
# Description

Translated 6 strings in src/client/app/components/CompareLineChartComponent.tsx, shown below:

- Shifted
- The original line has ... (only part of string shown here)
- The shifted line overlaps the original line ... (only part of string shown here)
- Days of week align (unless missing readings)
- The month and day of the month align for the original and shifted readings
- until original reading at date (only part of string shown here)

Added the necessary key/value translations in src/client/app/translations/data.ts in English, Spanish, and French.

Fixes #1473

## Type of change

## Checklist

## Limitations

The line "const translate = useTranslate();" is used on both line 30 and line 198 in different functions, which feels redundant and a better way might exist.


src/client/app/translations/data.ts
```diff
@@ -1074,6 +1084,16 @@ const LocaleTranslationData = {
 		"select.unit": "Select Unit\u{26A1}",
 		"shift.date.interval": "Shift Date Interval\u{26A1}",
 		"shifted.data.crosses.leap.year.to.non.leap.year": "Shifted data crosses a leap year so the graph might not align appropriately\u{26A1}",
+		"shifted": "Décalé\u{26A1}",
```
huss: Here and in es are similar. The ``\u{26A1}`` is put at the end of untranslated strings. However, shifted seems translated. Should it be there?


Also, are you a fluent translator to French and Spanish? If not, these should be left in English with the symbol for such a person to translate.

src/client/app/components/CompareLineChartComponent.tsx
```diff
@@ -194,19 +195,21 @@ export default function CompareLineChartComponent() {
  * @param shiftedReading shifted data to compare
  */
 function checkReceivedData(originalReading: any, shiftedReading: any) {
+	const translate = useTranslate();
```
huss: You have limitation in the PR description about doing this twice. While that in and of itself is not a big issue, I believe the need is actually a problem. The ``useTranslate()`` function can, as far as I know, only be used inside a component. While it can work outside a component this is not guaranteed and I think it violates a rule in React. If you look at src/client/app/components/groups/EditGroupModalComponent.tsx in function ``deepMetersToList()`` you will see it is inside the component (and does not have another use statement). I think this is the correct usage. This is a subtle rule that is often/easily overlooked and we learned about it the hard way. I think the change should be easy - move the code up inside the component function.

src/client/app/translations/data.ts
```diff
@@ -486,6 +486,16 @@ const LocaleTranslationData = {
...
+		"shifted": "Shifted",
+		"compare.line.original.shifted.count": "The original line has {original} readings but the shifted line has {shifted}, which means the points will not align horizontally.",
+		"compare.line.original.shifted.count.a": "The original line has ",
+		"compare.line.original.shifted.count.b": " readings but the shifted line has ",
+		"compare.line.original.shifted.count.c": ", which means the points will not align horizontally.",
+		"compare.line.shifted.overlaps.start": "The shifted line overlaps the original line starting at {start}",
+		"compare.line.shifted.overlaps.start.prefix": "The shifted line overlaps the original line starting at ",
+		"compare.line.days.align": "Days of week align (unless missing readings)",
+		"compare.line.month.day.align": "The month and day of the month align for the original and shifted readings",
+		"compare.line.month.day.align.until": " until original reading at date {date}",
```
huss: I think the code might be using ``compare.line.month.day.align.until.prefix`` so this one is not used.

src/client/app/translations/data.ts
```diff
@@ -486,6 +486,16 @@ const LocaleTranslationData = {
 		"select.unit": "Select Unit",
 		"shift.date.interval": "Shift Date Interval",
 		"shifted.data.crosses.leap.year.to.non.leap.year": "Shifted data crosses a leap year so the graph might not align appropriately",
+		"shifted": "Shifted",
+		"compare.line.original.shifted.count": "The original line has {original} readings but the shifted line has {shifted}, which means the points will not align horizontally.",
+		"compare.line.original.shifted.count.a": "The original line has ",
```
huss: I don't really like names that differ by a single letter. I think it tends to lead to mistakes in usage and they are usually not descriptive. I know OED is not consistent in this regard but could these be renamed?

src/client/app/translations/data.ts
```diff
@@ -486,6 +486,16 @@ const LocaleTranslationData = {
 		"select.unit": "Select Unit",
 		"shift.date.interval": "Shift Date Interval",
 		"shifted.data.crosses.leap.year.to.non.leap.year": "Shifted data crosses a leap year so the graph might not align appropriately",
+		"shifted": "Shifted",
+		"compare.line.original.shifted.count": "The original line has {original} readings but the shifted line has {shifted}, which means the points will not align horizontally.",
```
huss: I cannot find where this is used.

huss (changes requested):
@Nespina24 Thank you for this contribution. Overall, it works well and is good. I have made a few comments to consider. Please let me know if you have any thoughts/questions.

Nespina24:
@huss I was unaware about leaving untranslated strings with an emoji. I had them auto-translated by AI at first, but I personally cannot read the automated translations, so I reverted them to the same as the English strings and put the symbol at the end.

For the `useTranslate()` function, I interpreted your comment as nesting `checkReceivedData()` inside `CompareLineChartComponent()`, but if you meant anything different, please let me know.

src/client/app/translations/data.ts
```diff
@@ -485,7 +485,16 @@ const LocaleTranslationData = {
 		"select.shift.amount": "Select shift amount",
 		"select.unit": "Select Unit",
 		"shift.date.interval": "Shift Date Interval",
+		"shifted": "Shifted",
 		"shifted.data.crosses.leap.year.to.non.leap.year": "Shifted data crosses a leap year so the graph might not align appropriately",
+		"compare.line.original.shifted.count.first": "The original line has ",
+		"compare.line.original.shifted.count.second": " readings but the shifted line has ",
+		"compare.line.original.shifted.count.third": ", which means the points will not align horizontally.",
+		"compare.line.shifted.overlaps.start": "The shifted line overlaps the original line starting at {start}",
```
huss: I cannot find where this is used.

src/client/app/translations/data.ts
```diff
@@ -485,7 +485,16 @@ const LocaleTranslationData = {
 		"select.shift.amount": "Select shift amount",
 		"select.unit": "Select Unit",
 		"shift.date.interval": "Shift Date Interval",
+		"shifted": "Shifted",
 		"shifted.data.crosses.leap.year.to.non.leap.year": "Shifted data crosses a leap year so the graph might not align appropriately",
```
huss: I cannot find where this is used.

huss (changes requested):
@Nespina24 During a sweep of PRs today I realized this was ready for re-review but I had not done that. I'm sorry this sat for a while.

Thank you for addressing my previous comments. This seems to be working well. I made two minor comments to consider. I think this is very close to merging. Please let me know if anything is not clear or you have questions/thoughts.

src/client/app/translations/data.ts
Nespina24: As far as I know, this line of code has been here since before I was working on it. Regardless, I deleted it.

huss: Thanks. I missed that you did not make the change.

huss (approved):
@Nespina24 Thank you for the update. I am sorry this took me so long to review. Review and testing found it addressed all comments and works well. This is ready to merge. Congratulations on another successful contribution.

Issue #588, opened 2021-03-02
enhance log error messages


huss:
- There are a few error messages that do not give IP addresses such as:
“At 2020-06-17T04:01:27.161Z, an ERROR event occurred. ERROR ON REQUEST TO METER [NO IP ADDRESS AVAILABLE], 401 - "401 Unauthorized (lots of text removed)”
- This message be made clearer for what went wrong?:
“At 2020-06-12T11:01:31.076Z, an ERROR event occurred. ERROR ON REQUEST TO METER <IP address was here but removed>, Cannot read property 'trim' of undefined”

huss:
Just for the record, here are some more such log messages:

- At 2021-02-07T23:01:20.285Z, an ERROR event occurred. ERROR ON REQUEST TO METER [NO IP ADDRESS AVAILABLE], Number of columns is inconsistent on line 131 
- At 2021-02-07T23:03:24.292Z, an ERROR event occurred. ERROR ON REQUEST TO METER [NO IP ADDRESS AVAILABLE], Error: connect ETIMEDOUT <IP address was here but removed>
    - This one might be okay.

IssEdgarH:
I would like to work on this.

Bolshialex:
@IssEdgarH Are you still working on this? We would like to pick this up if not.

huss:
> [@IssEdgarH](https://github.com/IssEdgarH) Are you still working on this? We would like to pick this up if not.

@Bolshialex Since it has been a while since @IssEdgarH asked to work on this issue and there has not been a reply, I am assigning you this issue. Please let me know if I can help or if you don't want this issue assigned to you.

Jaronie:
I am claiming this issue with my CodeDay group! @JPbooya @sabinMas

huss:
@Jaronie Thank you and your team for offering to work on this. I have assigned you this issue. Please let me know if you need anything.

huss:
While PR #1643 indicated it only partly addressed this issue, I could not see any of the cases above as not covered. Given this, I'm closing this issue. If someone sees an issue with the error message then this can reopened with a comment or a new issue can be created.

PR #1643, opened 2026-06-14
enhance error log messages to include meter name, ID, and URL for iss…

JPbooya:
Issue #588 

# Description
Enhanced error log messages in updateMeter.js to provide more descriptive information when a meter request fails.

Partly addresses #588

## Type of change

## Checklist

## Limitations
No limitations at this time, but there could be more log messages in the codebase that could be refined in the future.

### Contributors
GitHub Handles: @JPbooya @sabinMas @Jaronie

### DIFF
--- a/src/server/services/updateMeters.js
+++ b/src/server/services/updateMeters.js
@@ -23,15 +23,17 @@ async function updateAllMeters(dataReader, metersToUpdate, conn) {
 		// Issues from the pipeline will be logged by called functions.
 		await Promise.all(
 			metersToUpdate
-				.map(meter => dataReader(meter, conn))
-				.map(p => p.catch(err => {
-					let ipAddress = '[NO IP ADDRESS AVAILABLE]';
-					if (err.options !== undefined && err.options.ipAddress !== undefined) {
+				.map(meter => dataReader(meter, conn)
+					.catch(err => {
+						let ipAddress = '[NO IP ADDRESS AVAILABLE]';
+						if (err.options !== undefined && err.options.ipAddress !== undefined) {
 						ipAddress = err.options.ipAddress;
 					}
-					log.error(`ERROR ON REQUEST TO METER ${ipAddress}, ${err.message}`, err);
+					log.error(`ERROR ON REQUEST TO METER ${ipAddress}, Meter name: ${meter.name}, Meter ID: ${meter.id}, URL: ${meter.url}, ${err.message} `,  err);
 					return null;
-				})));
+				})
+			)
+		);
 		log.info('Update finished');
 	} catch (err) {
 		log.error(`Error updating all meters: ${err}`, err);
--- a/src/server/services/updateMeters.js
+++ b/src/server/services/updateMeters.js
@@ -27,12 +27,12 @@ async function updateAllMeters(dataReader, metersToUpdate, conn) {
 					.catch(err => {
 						let ipAddress = '[NO IP ADDRESS AVAILABLE]';
 						if (err.options !== undefined && err.options.ipAddress !== undefined) {
-						ipAddress = err.options.ipAddress;
-					}
-					log.error(`ERROR ON REQUEST TO METER ${ipAddress}, Meter name: ${meter.name}, Meter ID: ${meter.id}, URL: ${meter.url}, ${err.message} `,  err);
-					return null;
-				})
-			)
+								ipAddress = err.options.ipAddress;
+						}
+						log.error(`ERROR ON REQUEST TO METER ${ipAddress}, Meter name: ${meter.name}, URL: ${meter.url}, ${err.message} `,  err);
+						return null;
+					})
+				)
 		);
 		log.info('Update finished');
 	} catch (err) {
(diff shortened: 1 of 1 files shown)


src/server/services/updateMeters.js
```diff
@@ -23,15 +23,17 @@ async function updateAllMeters(dataReader, metersToUpdate, conn) {
...
 			metersToUpdate
-				.map(meter => dataReader(meter, conn))
-				.map(p => p.catch(err => {
-					let ipAddress = '[NO IP ADDRESS AVAILABLE]';
-					if (err.options !== undefined && err.options.ipAddress !== undefined) {
+				.map(meter => dataReader(meter, conn)
+					.catch(err => {
+						let ipAddress = '[NO IP ADDRESS AVAILABLE]';
+						if (err.options !== undefined && err.options.ipAddress !== undefined) {
 						ipAddress = err.options.ipAddress;
```
huss: I think the intending is a little off here. If you are using VSC then format document should fix.

src/server/services/updateMeters.js
```diff
@@ -23,15 +23,17 @@ async function updateAllMeters(dataReader, metersToUpdate, conn) {
...
 					}
-					log.error(`ERROR ON REQUEST TO METER ${ipAddress}, ${err.message}`, err);
+					log.error(`ERROR ON REQUEST TO METER ${ipAddress}, Meter name: ${meter.name}, Meter ID: ${meter.id}, URL: ${meter.url}, ${err.message} `,  err);
```
huss: OED is not logging the meter.id and it is never returning it to the user now. I think that can be removed.

src/server/services/updateMeters.js
```diff
@@ -23,15 +23,17 @@ async function updateAllMeters(dataReader, metersToUpdate, conn) {
 		// Issues from the pipeline will be logged by called functions.
 		await Promise.all(
 			metersToUpdate
-				.map(meter => dataReader(meter, conn))
-				.map(p => p.catch(err => {
-					let ipAddress = '[NO IP ADDRESS AVAILABLE]';
-					if (err.options !== undefined && err.options.ipAddress !== undefined) {
+				.map(meter => dataReader(meter, conn)
+					.catch(err => {
```
huss, line 27: There is a subtle difference in the removal of the ``.map``. This type of code seems to always confuse me so I wanted your input. Was it done to fix how it worked or to get direct access to meter? I suspect testing this is a little tricky so I wanted to know if you had been able to do that, esp. when multiple meters were being done at the same time to verify the correct error was logged for each.

huss (changes requested):
@JPbooya, @sabinMas & @Jaronie Thank you for looking into this issue. I made a few comments based on reading the code. I have not yet tested and will wait until you have a chance to give your thoughts, esp. on one of the comments. Please let me know any thoughts/questions.

src/server/services/updateMeters.js
JPbooya: For the original two.map() version, the second .map() was iterating over the array of Promises returned by the dataReader from the first .map(). By merging them into one .map(), .catch() connects directly onto one Promise inside the same callback where meter is still defined. Thus able to use the variable. For the testing, I have only verified through the existing tests, I have not written a dedicated test case that triggers the failure. But since each call to the arrow function creates its own closure over its own meter value, each . catch() callback should still refer to the correct meter value it was created with.

huss: So, this turned out to. be trickier than I thought. First, I had to set up a way to test the code. For the records, I created two meters of type MAMAC that were enabled. I made one have a completely bogus URL and the other went to localhost:3000. This caused different errors (first fails on URL & second fails due to a bad response). I then ran ``npm run updateMamacMeters`` to cause the changes to run. This showed that the value being looked at in ``err`` generally did not exist. I tried to find another place in that object but which one is set seems to depend on the type of error. Thus, I think it is risky to use the values from ``err`` to get the URL. Given you added the meter info and it should match the request/response, this should provide the needed information. It would only be an issue if the URL constructed by Axios differed in a way that was not clear but that seems unlikely. So, I changed the code to do this and pushed the change. I welcome anyone's thoughts on what I have done. I'm sorry to directly make the change but it would not be easy for someone without a deep understanding and the change you already made enabled the ability to remove this URL info as it had been.

I also want to say I tested your logic on the .map and you are correct. It works exactly as you said and desired. I believed you but wanted to note I could confirm empirically during my testing.

huss (commented):
@JPbooya, @sabinMas & @Jaronie Thank you for your changes. Overall, they we what was desired. I have left a new reply comment on a change I made and pushed to this PR. I would appreciate you looking at it and saying what you think. If you have any thoughts/concerns/questions then please let me know. If you think my change is okay then this is ready to merge.

huss:
> @JPbooya, @sabinMas & @Jaronie Thank you for your changes. Overall, they we what was desired. I have left a new reply comment on a change I made and pushed to this PR. I would appreciate you looking at it and saying what you think. If you have any thoughts/concerns/questions then please let me know. If you think my change is okay then this is ready to merge.

It has been more than a week since my comment. Since there was no reply and my change was straightforward, I'm considering it okay. Given this, this PR is now ready to merge.

huss (approved):
@JPbooya, @sabinMas & @Jaronie Thank you for this contribution and your efforts. Per other comments, this is now ready to merge. Congratulations on an accepted contribution to OED.

PR #1566, opened 2026-01-16
Issue869 - Implemented Failure/Success Notifications to Edit/Create pages

aduques:
# Description

These changes address Issue #869. These changes aim to reimplement the missing success and failure notifications across all of the available Edit/Create pages (excluding Maps and Admin Settings). The additions of these notifications do not intend to change the functionality of the edit/create pages, but instead help future developers understand if a request is successful or not. The messages on each of the notifications come from the predetermined messages listed on [data.ts](http://data.ts/).

In this pull request:
- showSuccessNotifications() and showFailureNotifications() are added to all edit/create pages and are all made to display a unique message based on the page.
- Each success notification includes the predetermined message from [data.ts](http://data.ts/) + specific information of the object (meter, conversion, user, etc.) to help the developer identify what was specifically updated.
- Each failure notification includes the predetermined message from [data.ts](http://data.ts/) + the specific error
- Each file contains TODO DEBUG code that is used to force the error notification. This is only included for testing and is intended to be removed before merging.
- The [data.ts](http://data.ts/) was slightly modified to include missing predetermined messages and to fix a small typo on one of the messages.
- Success and Failure Notifications were added for the delete feature on the edit pages and includes some specific information of the involved object.

Motivation: After working on Issue #1191 and Issue #1517, I have noticed that there was inconsistency with the failure/success notifications across the edit/create pages. Some pages had it implemented with vague messages, while some did not have the notifications implemented at all. After discussing with the project maintainer, it was brought to my attention that there was an existing issue that had overlapping problems. As someone who is familiar with the edit/create pages, I wanted to take the opportunity to work on this issue.

Partly Addresses #869

## Type of change

- [ ] Note merging this changes the database configuration.
- [x] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations

As discussed in the comments of the Issue page, I was not sure whether or not the implemented notifications have the desired information. In addition, no progress was done on the PreferencesComponent.tsx (Admin Settings page) at the time of initially creating the pull request. I plan on addressing it after getting feedback on the other parts of the implementation.

To reiterate, this pull requests is intended to show the current progress and to improve once feedback is given.

[...]

PR #1663, opened 2026-07-06
Issue1651 - failure on group creation and check others

### LINKED ISSUE #1651: failure on group creation and check others
### Describe the bug

If you try to create a group without a default graphic unit then it fails in the route. Recent changes added ``id: { type: 'integer', minimum: 1 },`` which excludes no unit with value of -99.

### To Reproduce

Create any group and add a meter or group as the child but do not set the default graphic unit. It will fail with a message in the web browser console and the terminal where OED was started. The terminal will indicate why it failed in an error message.

### Expected behavior

Groups are allowed with no unit so the parameter testing needs to be fixed. If possible, only all -99 and not other negative values.

### Additional context

This page and others should be tested to see if any others have the same issue with id. These should be fixed in the same way. It may be possible to find ones by looking for which ones allow special, negative values in the model or component page. If any other problematic checks are found then they should be documented and, if possible, fixed. All changes should be carefully checked.

aduques:
# Description

These changes aim to address the existing bug that occurs when a user tries to create a group without a default graphic unit but fails despite having valid default values. Before these changes, the schema was rejecting the 'id' value -99 (the value representing "no unit," which is one of the default values). In addition to the bug, through discussions with the maintainer, we have identified other places in the codebase that can potentially have the same issues. In order to keep the changes consistent across the codebase, the bug fix for groups has been implemented in related files.

Motivation: This issue was brought to my attention by @huss while working on the pull request (https://github.com/OpenEnergyDashboard/OED/pull/1566). @huss Suggested a suggestion that would also be used to fix the bugs described in this issue.

Fixes #1651

Through discussions with the maintainer, we have found that this issue can be broken down into 5 parts. The discussions in the related issue (https://github.com/OpenEnergyDashboard/OED/issues/1651#issuecomment-4834127393) breaks down the initial plan and implementation which has received approval from the maintainer. The purpose of this pull request is to verify if those proposed changes work without any issues.

**1. Fixing defaultGraphicUnit (-99 issue)**

Ported the anyOf fix (which was initially suggested by @huss) to meters.js's validateMeterParams(), so defaultGraphicUnit accepts positive integers or exactly -99, instead of any integer with no bound.

**2. Decoupling schema validation**

validateMeterParams() now takes an isEdit boolean, id is only added to the schema's required properties when isEdit is true, so /edit still requires id while /addMeter now rejects it, using a single shared schema rather than two duplicated ones.

**3. Remove ID from Group create**

id is removed from group.js /create schema properties entirely (as it was discussed DB assigns it automatically), and maxProperties: 9 is added as an additional defensive bound.

**4. Fixing the deletion bounds**

conversions.js /delete route had minimum 0 on sourceId, destinationId, meterIds, and groupIds. The minimum was increased to 1 as desired.

**5. Fixing payload compatibility**

CreateGroupModalComponent.tsx no longer includes id in its submitted payload.

## Type of change

## Checklist

## Limitations

I will go into details about any notes and feedback I had while testing the implementations for each section in a later comment. For the most part, I would appreciate a confirmation if the way that I initially tested the changes or if there was anything that I have overlooked.

I will go into more details below, but I noticed the following with the changes in meters.js:

In meters.js, the defaultGraphicUnit fix could not be shown to fix an reproducible bug as there are existing variables implemented that prevent the create request from being called in the first place. The changes are still applied so that it is consistent with groups.js, but this might good to note.


aduques:
## Fixing defaultGraphicUnit (-99 issue)

Make changes in meters.js and groups.js and test the changes through CreateGroupModalComponent.tsx.

**Changes made:**

defaultGraphicUnit: {'anyOf': [{ type: 'integer', minimum: 1 }, { type: 'integer', 'enum': [-99] }]}, //defaultGraphicUnit: { type: 'integer', minimum: 1 },

This comment (https://github.com/OpenEnergyDashboard/OED/issues/1651#issuecomment-4828715677) shows the resulting notification before and after making the changes to defaultGraphicUnit. This specific comment shows the result when the change has only been made to groups.js. In the screenshot, it shows a group being created with the “Default Graphic Unit” not being changed. The “no unit” is equivalent to “-99,” which is the default value we want the bounds to accept. With the changes, the backend will accept the payload because all values (including the default values) are now accepted.

Based on my current understanding, the problematic default value was just defaultGraphicUnit as after the changes were made, a group can be successfully created with the bare minimum required properties (name and child meters/child group).

**Regarding the changes for meters.js** 

I am under the assumption that these changes can be made through the CreateMeterModalComponent.tsx as I figured this is the flow for the create and edit operations:

meters.js (/addMeter) → metersApi (defines the POST request for addMeter) → CreateMeterModalComponent.tsx (imports metersApi as submitAddMeter)

So when I create a meter, and leave the defaultGraphicUnit as a default value (do not change it), I will have a successful creation theoretically. 
However, with the existence of 
const defaultGaphicUnitIsSelected = meterDetails.defaultGraphicUnit !== -999; 

It is created as a safeguard for this issue, which did not exist for groups.js which is probably why the defaultGraphicUnit error was raised because of the lack of guardrails. The hard coded defaultGraphicUnitIsSelected prevents the error message from being raised.

**TODO:** In order to test the changes for meters.js, I would need to disable defaultGraphicUnitIsSelected to possibly see the defaultGraphicUnit error.


                         {/* On click calls the function handleSaveChanges in this component */}
                         {/* TODO DEBUG: disabling this line disables the defaultGraphicUnitIsSelected which prevents the
                         defaultGraphicUnit error from being raised unlike in CreateGroupModalComponent.tsx */}
                         {/*<Button color='primary' onClick={handleSubmit} disabled={!meterIsValid || !canSave}>*/}
                         <Button color='primary' onClick={handleSubmit} disabled={false}>


Screenshot 2026-07-05 at 10 27 05 PM

After making this change in meters.js

                 //defaultGraphicUnit: { type: 'integer' },
                 defaultGraphicUnit: {'anyOf': [{ type: 'integer', minimum: 1 }, { type: 'integer', 'enum': [-99] }]},


Screenshot 2026-07-05 at 10 27 42 PM


After enabling defaultGraphicUnitIsSelected again

                   {/* On click calls the function handleSaveChanges in this component */}
                   {/* TODO DEBUG: disabling this line disables the defaultGraphicUnitIsSelected which prevents the
                   defaultGraphicUnit error from being raised unlike in CreateGroupModalComponent.tsx */}
                   <Button color='primary' onClick={handleSubmit} disabled={!meterIsValid || !canSave}>
                   {/*<Button color='primary' onClick={handleSubmit} disabled={false}>*/}

Screenshot 2026-07-05 at 10 28 13 PM


It is not possible, due to the meter is not valid if nothing has been selected, we cannot even attempt to create the meter. Choosing “no unit” would result in a value that’s not “-999.” With that in mind, I am not sure how to assess what the change does for meters.js defaultGraphicUnitIsSelected prevents the defaultGraphicUnit error from being raised as the user is prevented from even attempting the create request.

Screenshot 2026-07-05 at 10 28 38 PM

## Decoupling Schema Validation

As discussed before, my attempted solution of requiring a single schema that utilizes a boolean to control the addition of the id parameter (as the db will handle it automatically), I added a boolean parameter to validateMeterParams().

      +// This checks params for both edit and create. The id property is only validated on edit since the DB assigns it on create.
      +function validateMeterParams(params, isEdit = false) {

The implementation involves removing id from the required properties (excluding it here affect an id being added because it is handled automatically by the db, it only matters in the case of edit). 


      -       const validParams = {
      -               type: 'object',
      -               additionalProperties: false,
      -               // We can get rid of some of these if we defaulted more values in the meter model.
      -               required: ['name', 'url', 'enabled', 'displayable', 'meterType', 'timeZone', 'note', 'area'],
      -               properties: {
      -                       id: { type: 'integer', minimum: 1 },
      
      
      +       const validParams = {
      +               type: 'object',
      +               additionalProperties: false,
      +               // We can get rid of some of these if we defaulted more values in the meter model.
      +               required,
      +               properties
      +       };


If the isEdit is true (when validateMeterParams() is called in an edit request), the id property will be added to the required properties (the same singular schema, so no two schemas are created with one excluding id).

      +       if (isEdit) {
      +               properties.id = { type: 'integer', minimum: 1 };
              }
      +
      +       const required = ['name', 'url', 'enabled', 'displayable', 'meterType', 'timeZone', 'note', 'area'];
      +       if (isEdit) {
      +               required.push('id');
      +       }
      +
      +       const validParams = {
      +               type: 'object',
      +               additionalProperties: false,
      +               // We can get rid of some of these if we defaulted more values in the meter model.
      +               required,
      +               properties
      +       };


In order to use the updated validateMeterParams(), I have updated the request parameters in /edit and /create accordingly.

       router.post('/edit', adminAuthMiddleware('edit meters'), async (req, res) => {
      -       const response = validateMeterParams(req.body)
      +       // isEdit=true: id is required here since the client must tell us which meter to update.
      +       const response = validateMeterParams(req.body, true)
              if (!response.valid) {
                      log.warn(`Got request to edit a meter with invalid meter data, errors: ${response.errors}`);
                      failure(res, HTTP_CODES.BAD_REQUEST, 'validation failed with ' + response.errors.toString());
      @@ -327,7 +338,8 @@ router.post('/edit', adminAuthMiddleware('edit meters'), async (req, res) => {
        * Route for POST add meter.
        */
       router.post('/addMeter', adminAuthMiddleware('add meter'), async (req, res) => {
      -       const response = validateMeterParams(req.body)
      +       // isEdit=false: id must not be present, since it's assigned by the DB on insert.
      +       const response = validateMeterParams(req.body, false)
              if (!response.valid) {
                      log.warn(`Got request to create a meter with invalid meter data, errors: ${response.errors}`);
                      failure(res, HTTP_CODES.BAD_REQUEST, 'validation failed with ' + response.errors.toString());


**Testing the changes:**

With the new additions, the server will reject a payload that contains id. To test this, we temporarily remove the omit(meterDetails, ‘id’) from the submitState which removes id from the payload. This test will demonstrate that the server will now reject the payload and demonstrate that the boolean isEdit works as intended.

                     // TODO DEBUG: commenting out omit to make sure that the create request payload gets sent with an id and confirm the server rejects it.
                     // With the new changes to validateMeterParams(), an id will be rejected since isEdit flag is not raised for /create requests.
                     //...omit(meterDetails, 'id'),
                     ...meterDetails,

Screenshot 2026-07-05 at 10 30 36 PM


As you can see, with the isEdit, and changes to the /edit and /create requests. The id parameter will only be added if it is an edit request. Otherwise, id is left out as a required parameter.

Screenshot 2026-07-05 at 10 31 19 PM


## Remove ID in Group

As discussed, the create routes should not be passing ID into the payload for create requests. Therefore, I have made the changes to remove id from the properties. In addition, I have included maxProperties as an additional guardrail to prevent id from being considered in the payload.


      @@ -228,8 +228,8 @@ router.post('/create', adminAuthMiddleware('create groups'), async (req, res) =>
                      type: 'object',
                      additionalProperties: false,
                      required: ['name', 'childGroups', 'childMeters'],
      +               maxProperties: 9,
                      properties: {
      -                       id: { type: 'integer', minimum: 1 },


The screenshot shows that with the changes, newly created groups will not pass an ‘id’ property.

Screenshot 2026-07-05 at 10 32 01 PM


## Fixing the Deletion Bounds

As suggested, “src/server/routes/conversions.js in delete has the 3 id minimum as 0 but I think they should be 1.”

**NOTE:** I have made the changes, but I will appreciate help in testing if the following changes work as intended.

      @@ -171,21 +171,21 @@ router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, re
                      properties: {
                              sourceId: {
                                      type: 'integer',
      -                               minimum: 0
      +                               minimum: 1
                              },
                              destinationId: {
                                      type: 'integer',
      -                               minimum: 0
      +                               minimum: 1
                              },
                              meterIds: {
                                      type: 'array',
      -                               items: { type: 'integer', minimum: 0 },
      +                               items: { type: 'integer', minimum: 1 },
                                      uniqueItems: true,
                                      maxItems: 1000
                              },
                              groupIds: {
                                      type: 'array',
      -                               items: { type: 'integer', minimum: 0 },
      +                               items: { type: 'integer', minimum: 1 },
                                      uniqueItems: true,
                                      maxItems: 1000
                              }
      (END)


## Fixing Payload Compatibility

      @@ -239,9 +239,14 @@ export default function CreateGroupModalComponent() {
                      if (inputOk) {
                              // The input passed validation.
                              // GPS may have been updated so create updated state to submit.
      -                       const submitState = { ...state, gps: gps };
      +                       //const submitState = { ...state, gps: gps };
      +                       const { ...stateWithoutId } = state;
      +                       const submitState = { ...stateWithoutId, gps: gps };
       
      -                       createGroup(submitState)
      +                       // groupsApi.ts's createGroup mutation already strips id/deepMeters/deepGroups internally
      +                       // before building the request; this cast is needed here only because stateWithoutId no
      +                       // longer structurally matches GroupData once id is removed client-side.
      +                       createGroup(submitState as unknown as GroupData)
                                      .unwrap()
                                      .then(() => {
                                              showSuccessNotification(
      (END)

I have tested the changes by simply creating a new group object. However, I am expecting that this is not sufficient enough for testing the changes, so I appreciate any insight on this.

Screenshot 2026-07-05 at 10 33 32 PM

src/client/app/components/groups/CreateGroupModalComponent.tsx
```diff
@@ -239,9 +239,14 @@ export default function CreateGroupModalComponent() {
 		if (inputOk) {
 			// The input passed validation.
 			// GPS may have been updated so create updated state to submit.
-			const submitState = { ...state, gps: gps };
+			//const submitState = { ...state, gps: gps };
+			const { ...stateWithoutId } = state;
+			const submitState = { ...stateWithoutId, gps: gps };
 
-			createGroup(submitState)
+			// groupsApi.ts's createGroup mutation already strips id/deepMeters/deepGroups internally
```
huss, line 245: Given my comment above, the part of this comment about the added casting of state is no longer needed. It should go back to how it was.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -179,7 +179,10 @@ export default function CreateMeterModalComponent(props: CreateMeterModalProps):
 		if (inputOk) {
 			// The input passed validation.
 			const submitState = {
+				// TODO DEBUG: commenting out omit to make sure that the create request payload gets sent with an id and confirm the server rejects it.
```
huss: All three commented out lines should be removed (this, next and then one past the uncommented out line) as testing is now done. You noted but I'm carefully listing all. For the record, I did this type of testing by going into src/server/routes/meters.js and inserting one of these lines to arbitrarily change the req so it appears the client sent it that way. This covers this test and the next one. I find this easier than trying to hack the client (you asked about my testing).

```
	req.body = {...req.body, id: 1001};
	req.body = {...req.body, defaultGraphicUnit: -999 };
```

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -785,6 +788,9 @@ export default function CreateMeterModalComponent(props: CreateMeterModalProps):
 						<FormattedMessage id="discard.changes" />
 					</Button>
 					{/* On click calls the function handleSaveChanges in this component */}
+					{/* TODO DEBUG: disabling this line disables the defaultGraphicUnitIsSelected which prevents the
```
huss: As above, please remove the commented out test code. Once done I think there will be no changes in this file. You noted but I'm carefully listing all.

src/client/app/components/meters/CreateMeterModalComponent.tsx
```diff
@@ -179,7 +179,10 @@ export default function CreateMeterModalComponent(props: CreateMeterModalProps):
 		if (inputOk) {
 			// The input passed validation.
 			const submitState = {
+				// TODO DEBUG: commenting out omit to make sure that the create request payload gets sent with an id and confirm the server rejects it.
+				// With the new changes to validateMeterParams(), an id will be rejected since isEdit flag is not raised for /create requests.
 				...omit(meterDetails, 'id'),
```
huss, line 182: The code in group does the omit in the API and that seems cleaner. I searched the code for ``...omit`` and found it in src/client/app/redux/selectors/chartQuerySelectors.ts, src/client/app/components/unit/CreateUnitModalComponent.tsx & src/client/app/components/conversion/CreateConversionModalComponent.tsx. The one in the chartQuerySelectors is probably okay but the others could be pushed to the API. If you agree, an issue could be opened to do this. Would you like to open? Would you be interested in having it assigned to you. Wanted to give you first dibs.

src/server/routes/meters.js
```diff
@@ -237,7 +230,8 @@ function validateMeterParams(params) {
 				]
 			},
 			unitId: { type: 'integer' },
-			defaultGraphicUnit: { type: 'integer' },
+			//defaultGraphicUnit: { type: 'integer' },
```
huss: Commented out line should go.

src/server/routes/meters.js
```diff
@@ -256,14 +250,31 @@ function validateMeterParams(params) {
...
+
+	const required = ['name', 'url', 'enabled', 'displayable', 'meterType', 'timeZone', 'note', 'area'];
+	if (isEdit) {
+		required.push('id');
+	}
+
+	const validParams = {
+		type: 'object',
+		additionalProperties: false,
+		// We can get rid of some of these if we defaulted more values in the meter model.
```
huss: This comment would probably be better placed right above the ``const required`` line above since it really refers to that.

src/server/routes/meters.js
```diff
@@ -158,16 +158,9 @@ router.get('/:meter_id', optionalAuthMiddleware, async (req, res) => {
...
-function validateMeterParams(params) {
-	const validParams = {
-		type: 'object',
-		additionalProperties: false,
-		// We can get rid of some of these if we defaulted more values in the meter model.
-		required: ['name', 'url', 'enabled', 'displayable', 'meterType', 'timeZone', 'note', 'area'],
-		properties: {
-			id: { type: 'integer', minimum: 1 },
+// This checks params for both edit and create. The id property is only validated on edit since the DB assigns it on create.
+function validateMeterParams(params, isEdit = false) {
```
huss: First, I really like your solution. I had not thought about the fact that this is a function so you could do logic. It is cleaner than using multiple schema. I think it would be good to consider a new issue to roll this out across routes. I quickly looked at groups and it has copies of most of edit and create that this would avoid. I think this is a larger issue since all the routes need to be looked at to determine which ones would benefit and then do it. Would you like to do the issue (assuming you agree) or should I? Again, you can ask to be assigned it if you want. Second, I wonder if the default value on isEdit should be true. The current code does not use the default but if a developer left it off then the stronger test would be run and that seems safer. If it was not what they wanted then they would fix it in the call.

src/server/routes/meters.js
```diff
@@ -256,14 +250,31 @@ function validateMeterParams(params) {
...
+		required,
+		properties
+	};
 	const paramsValidationResult = validate(params, validParams);
 	return { valid: paramsValidationResult.valid, errors: paramsValidationResult.errors };
 }
 
 router.post('/edit', adminAuthMiddleware('edit meters'), async (req, res) => {
-	const response = validateMeterParams(req.body)
+	// isEdit=true: id is required here since the client must tell us which meter to update.
```
huss, line 276: Just noting for the record that I tested this by adding this line (you asked about my testing):

```
	delete req.body.id;
```

src/server/routes/conversions.js
```diff
@@ -171,21 +171,21 @@ router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, re
...
 			},
 			destinationId: {
 				type: 'integer',
-				minimum: 0
+				minimum: 1
 			},
 			meterIds: {
 				type: 'array',
-				items: { type: 'integer', minimum: 0 },
+				items: { type: 'integer', minimum: 1 },
```
huss, line 182: You asked how to test this. source/destination are like other comments. For meterIds I did (group similar):

```
req.body.meterIds.push(0);
```

It is good I did this since it showed that the error message did not properly show on the client. I had to edit line 423 of src/client/app/components/conversion/EditConversionModalComponent.tsx to be:

```
error.data
```

and then it mirrors save error handling. That made it work. I did a search and found these files also handle the error differently but am unsure if it is wrong: src/client/app/components/admin/users/CreateUserModalComponent.tsx, src/client/app/components/admin/users/EditUserModalComponent.tsx, src/client/app/components/groups/EditGroupModalComponent.tsx. There is a chance it is done somewhere else but I tried to catch them all by searching for variants (error.data.message, err.data., data.message). Maybe this should be a new issue (unless you want to do it right now but it isn't directly related to the PR). Same questions as other new issues: you create? assigned to you?

huss (changes requested):
@aduques Thank you for this contribution. Review and testing found it works as desired. You asked about testing in your comments and I added my method to my comments. I also added comments to remove your testing code to be complete but you mentioned those too. There are few other comments to consider but I think they are probably straightforward. A few comments involve creating new issues. If you agree then they can be done (see comments for options). Please let me know if anything is unclear or you have thoughts/questions.

src/client/app/components/groups/CreateGroupModalComponent.tsx
aduques: I would appreciate clarification on this comment. I have interpreted it as only the createGroup() call should be changed. The second part of the comment about adding a casting of state has been removed.

This is the current implementation now:

			      const { ...stateWithoutId } = state;
			      const submitState = { ...stateWithoutId, gps: gps };
      
			      // groupsApi.ts's createGroup mutation already strips id/deepMeters/deepGroups internally
			      // before building the request
			      createGroup(submitState)

src/client/app/components/meters/CreateMeterModalComponent.tsx
aduques: I have uncommented the TODO DEBUG lines used for testing.

				    // TODO DEBUG: commenting out omit to make sure that the create request payload gets sent with an id and confirm the server rejects it.
				    // With the new changes to validateMeterParams(), an id will be rejected since isEdit flag is not raised for /create requests.
				    //...meterDetails,

					{/* TODO DEBUG: disabling this line disables the defaultGraphicUnitIsSelected which prevents the
					defaultGraphicUnit error from being raised unlike in CreateGroupModalComponent.tsx */}
					{/*<Button color='primary' onClick={handleSubmit} disabled={!meterIsValid || !canSave}>*/}

src/client/app/components/meters/CreateMeterModalComponent.tsx
aduques: I have made the changes here: https://github.com/OpenEnergyDashboard/OED/pull/1663#discussion_r3540969263

src/client/app/components/meters/CreateMeterModalComponent.tsx
aduques: Please correct my understanding if needed. I understand that in this case, the omit() solution works, but it is more preferable to have all related files follow the approach done in group.

In /src/client/app/redux/api/groupsApi.ts, as mentioned, the omit() is done inside of the API, rather than in the handleSubmit().

		    createGroup: builder.mutation<void, GroupData>({
			    query: groupData => ({
				    url: 'api/groups/create',
				    method: 'POST',
				    // omit properties not accepted by the server schema
				    body: omit(groupData, ['id', 'deepMeters', 'deepGroups'])
			    }),
			    invalidatesTags: ['GroupData', 'GroupChildrenData']
		    }),

For the sake of this issue, no changes need to be made, but the implementation will eventually need to be changed and match with the group implementation.

I would like to work on this issue, and I can open the issue after this issue is closed. Before I do create this issue, I would appreciate any other information regarding this issue that I have not yet mentioned that would be helpful to include in the issue description.

src/server/routes/meters.js
aduques: I have done this.

src/server/routes/meters.js
aduques: I have made the changes.

	      // We can get rid of some of these if we defaulted more values in the meter model.
	      const required = ['name', 'url', 'enabled', 'displayable', 'meterType', 'timeZone', 'note', 'area'];

src/server/routes/meters.js
aduques: I would like to do the issue, and would not mind being the one to create the issue. Similarly to the other proposed issue, I will appreciate any additional information that will be helpful to include in the issue details if possible. I am assuming it is enough to note that it is not necessarily the case that all routes need this change, but primarily only the routes that have parameters that don't need to be passed in both create/edit requests (i.e. parameters that have similar cases with the id property).

I have noted the change about using a default value. I had it initially set to false for the default value, but I made the switch to true as suggested. I may have overlooked this initially as the function calls for validateMeterParams() were written to have their proper boolean values for create and edit cases, so I didn't need the default value in my use cases.

      // This checks params for both edit and create. The id property is only validated on edit since the DB assigns it on create.
      function validateMeterParams(params, isEdit = true) {

src/server/routes/meters.js
aduques: No changes have been made here.

src/server/routes/conversions.js
aduques: I would not mind working on this issue. It is a part of the codebase/ features that I am not too familiar with and would be beneficial if I got the opportunity to work on it so that I can work on future issues that involve different error messages.

However, I am not too sure how to describe this issue, so I will appreciate if you were able to create the issue with the proper issue context. In addition, I don't mind being immediately assigned to this issue, but since I would be assigned to potentially 3 more issues, it can be left open for anyone to take until I finished my unfinished issues.

aduques:
@huss I have addressed the recent comments, but there was one feedback that I would appreciate clarification on if needed (the suggestion related to the casting of state). I also gave my thoughts about opening the issue. I plan to open up those issues later this week, but there is one issue that I would appreciate getting help opening as mentioned in one of my comments.

src/server/routes/conversions.js
aduques: I have created the issue: https://github.com/OpenEnergyDashboard/OED/issues/1666

As I have noted before that I was not sure if I understood the request completely, I would appreciate any additional notes that I may have left out in the issue description.

src/client/app/components/meters/CreateMeterModalComponent.tsx
aduques: I have created the issue: https://github.com/OpenEnergyDashboard/OED/issues/1664

src/server/routes/meters.js
aduques: I have created the issue: https://github.com/OpenEnergyDashboard/OED/issues/1665

src/server/routes/conversions.js
huss: Thanks for creating the issue. It looked good to me and I assigned it to you.

src/client/app/components/meters/CreateMeterModalComponent.tsx
huss: Your understanding is correct. I've assigned you the issue you created.

src/client/app/components/groups/CreateGroupModalComponent.tsx
huss: You made a change and it was correct. The call does not need a cast and the comment should reflect that (removed).

huss (approved):
@aduques Thank you for the updated code that correctly addressed all my comments. This PR looks good and is ready to merge. Thank you for another accepted contribution to OED.

PR #1671, opened 2026-07-14
Issue 1646 login timingFix login timing for missing usernames

### LINKED ISSUE #1646: Continue work started in PR #1599 to fix login timing differences
### Description

PR #1599 began work to fix login timing differences. The original developer(s) will not be continuing this work. As a result, OED would like the work continued by others to address any comments and complete the task as needed.

If any comments, ideas or needed work is unclear or you have questions, then please reach out to the OED project.

### Additional context

New developers will not be able to push commits to the PR branch. The PR branch should be cloned to your machine, create a new branch and pushed to your own GitHub fork. Once the work is complete and on your GitHub fork, a new PR can be created from your branch on GitHub. The description should reference the original PR as the starting point and acknowledge the previous work's authors. Note OED requires that the original commit chain be present in the new PR and properly document who did what work. OED will close the original PR after any new one is merged. If you make progress but cannot address all needed work then OED would still like to have a draft PR created with the work you did complete. If you have questions about this process then please contact the OED project.

Adam-Alhakam:
# Description

Fixes #1646

This PR continues the work started in PR #1599 to reduce login timing differences between a valid existing username with a wrong password and a fake non-existing username.

The problem was in `src/server/routes/login.js`. Before this change, the login route only ran `bcrypt.compare()` when the username existed in the database. If the username did not exist, the route skipped bcrypt and returned `401 Unauthorized` faster. Since bcrypt comparison is intentionally slow, this created a timing difference that could help someone guess which usernames exist in OED.

This PR changes the login flow so both paths use a bcrypt comparison:

- If the user exists, the code compares the submitted password with the real stored password hash.
- If the user does not exist, the code compares the submitted password with a fixed dummy bcrypt hash.
- Login only succeeds when the user exists and the password matches.
- A fake username still returns `401 Unauthorized`.
- A valid username with the wrong password still returns `401 Unauthorized`.
- A valid username with the correct password still returns `200 OK`.

This also avoids the issue in PR #1599 where the code attempted to access `user.passwordHash` when `user` was `null`. In this PR, the password hash is selected safely before calling `bcrypt.compare()`.

The result is that the fake-user path now follows a similar bcrypt timing path to the valid-user wrong-password path, reducing the username enumeration risk.

This PR acknowledges the original work and idea from PR #1599 by @Andrew-Bonner. The final implementation in this PR corrects the null-user issue and safely applies the dummy-hash approach.

Authors/contributors:

- @Adam-Alhakam - Alnagdi Mohsen 
- @JTRBG123 - Joseph Ramer

## Type of change

## Checklist

## Testing

I tested the login route by comparing the behavior and timing for both a valid existing username and a fake non-existing username.

### 1. Valid existing username with correct password

I first confirmed that a real user can still log in successfully:

```bash
curl -i -X POST http://localhost:3000/api/login \
  -H "Content-Type: application/json" \
  -d '{"username":"test","password":"password"}'

### CODE REVIEW

**Comment by huss** [comment]
@Adam-Alhakam Thank you for this contribution. The PR description states the CLA has been signed by all contributors. However, OED records do not have one for @JTRBG123. Thus, I would like them to sign the CLA via the link in the description or let me know that you think OED records are off.

Once this is done, this PR can be formally reviewed. Please let me know if you have any questions/thoughts.

**Comment by JTRBG123** [comment]
Sorry for the delay in signing the CLA, @huss. I've just signed it. Thanks for your patience!

File: src/server/routes/login.js [hunk 1]
```diff
@@ -46,13 +47,15 @@ router.post('/', credentialsRequestValidationMiddleware, async (req, res) => {
...
 			const user = await User.getByUsername(req.body.username, conn);
-			let isValid;
-			if (user === null) {
-				// User did not exist so return false.
-				isValid = false;
-			} else {
-				isValid = await bcrypt.compare(req.body.password, user.passwordHash);
-			}
+
+// User did not exist so return false.
```
huss: I think this comment should go (and the empty comment next) since it isn't what the code is doing.

src/server/routes/login.js
```diff
@@ -46,13 +47,15 @@ router.post('/', credentialsRequestValidationMiddleware, async (req, res) => {
...
+//
```
huss: The indenting is off. If using VSC then format file will fix it.

src/server/routes/login.js
```diff
@@ -15,6 +15,7 @@ const { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, USERNAME_MIN_LENGTH, USERNAME_
 const { HTTP_CODES } = require('../util/httpCodes');
 
 const router = express.Router();
+const DUMMY_PASSWORD_HASH = '$2a$10$7EqJtq98hPqEX7fNZaFWoOHIoQStbSNRaCbkWa3vgKwK3/q5YLhKa';
```
huss: First, is there a reason to not put this inside the function to scope it to that? Second, where did this hash come from? I just want to be sure there is no chance it is likely to be a real password hash. Please don't post if you hashed a known string so it is not publicly revealed. Either say it is okay and/or message me privately.

huss (changes requested):
@Adam-Alhakam & @JTRBG123 I've reviewed/tested the code and found it functions correctly. I've made a few minor comments to consider. Please let me know if anything is not clear or you have thoughts/questions.

Adam-Alhakam:
Thanks @huss. I updated this in the latest commit.

Changes made:
- Moved the dummy hash inside the login route so it is scoped to this function.
- Removed the old `// User did not exist so return false.` comment and the empty comment line because they no longer matched what the code does.
- Formatted `src/server/routes/login.js`.

For the dummy hash, I generated it from a random value using:

```js
crypto.randomBytes(32).toString('hex')

File: src/server/routes/login.js [hunk 4]
```diff
@@ -15,6 +15,7 @@ const { PASSWORD_MAX_LENGTH, PASSWORD_MIN_LENGTH, USERNAME_MIN_LENGTH, USERNAME_
 const { HTTP_CODES } = require('../util/httpCodes');
 
 const router = express.Router();
+const DUMMY_PASSWORD_HASH = '$2a$10$N6cWKczGlZaT2ReVzJ48pu8t87bpatdCnpI50fXQ7SnHO23LL7Nfe';
```
huss: I see you moved the value into the function below and declared it as a separate const. Thus, I'm unsure what this one is for. Should it be removed?

huss (changes requested):
@Adam-Alhakam Thank you for the quick updates. They addressed my previous comments. I made one new one that I think will be easy to address. I'll do one final test pass on the final version but I think this is all okay.

Adam-Alhakam:
Thanks @huss. Yes, that top-level constant was left over from the earlier version. I removed it so the dummy hash is only declared inside the login route where it is used.

huss (approved):
@Adam-Alhakam & @JTRBG123 Thank you for the latest update. Review and testing found everything works as desired. Congratulations on an accepted contribution to OED.

PR #1673, opened 2026-07-16
Issue1664 - Moved omit() logic for unneeded properties from client layer to API layer

### LINKED ISSUE #1664: Move stripping properties with omit() from client-side React components to the Redux API definitions

There is an inconsistency in how API request payloads are formatted across the frontend. In some routes, payload formatting is handled directly within the UI components. For example, removing id from the schema during a create request.

As mentioned by the maintainer, ...omit(...) is used inside of

src/client/app/redux/selectors/chartQuerySelectors.ts
src/client/app/components/unit/CreateUnitModalComponent.tsx
src/client/app/components/conversion/CreateConversionModalComponent.tsx

before passing the data to the submit handler. This mixes API payload formatting logic with the UI. This is not an implementation issue, but more of an issue to clean up the current implementation.

The maintainer also notes: "The one in the chartQuerySelectors is probably okay but the others could be pushed to the API."

The logic for stripping out unnecessary properties should be centralized within the Redux API definitions rather than handled inside of the React components on the client side. The omit() calls should be removed from the React components and moved to their respective API slice files.

This change should mirror the pattern already established in:
/src/client/app/redux/api/groupsApi.ts

            	    createGroup: builder.mutation<void, GroupData>({
            		    query: groupData => ({
            			    url: 'api/groups/create',
            			    method: 'POST',
            			    // omit properties not accepted by the server schema
            			    body: omit(groupData, ['id', 'deepMeters', 'deepGroups'])
            		    }),
            		    invalidatesTags: ['GroupData', 'GroupChildrenData']
            	    }),  

No other alternatives. It is possible that this change does not need to happen in all routes (since not all routes would need to exclude a property like id), despite the goal of wanting this implementation be consistent across the codebase.

There could be other methods of removing the omit() from the client to somewhere else in the server, but the identified approach appears to be the cleanest.

### Additional context

This issue was brought to attention by this issue/pull request: https://github.com/OpenEnergyDashboard/OED/pull/1663

The suggested changes were from this comment: https://github.com/OpenEnergyDashboard/OED/pull/1663#discussion_r3537216809

aduques:
# Description

As discussed in a previous pull request, it was desirable to move the omit() logic from the client files, such as CreateGroupModalComponent.tsx, to their corresponding API layers (groupsApi.ts, in this case). This change was requested to be made consistent with the other client files that use omit() to remove unneeded properties (primarily id property). This change does not fix any existing bug, but instead, it is a change meant to clean up the implementations in the client files.

Fixes #1664

Minimal changes were made. In this issue, only the omit() calls were moved from the client files to the corresponding API files.

CreateGroupModalComponent.tsx and groupsApi.ts originally had this desired implementation, so the goal was to have the other files be consistent.

The changes have been made to the following files:
- CreateUnitModalComponent.tsx and unitsApi.ts
- CreateMeterModalComponent.tsx and metersApi.ts

The following files were considered to follow the proposed implementation, but required different approaches:
- CreateConversionModalComponent.tsx

Instead of moving the omit() to its corresponding API layer, helper functions were added which separates the omit() calls and computes bidirectional based on the current source/destination selections.

The following files include comment changes as discussed in the issue discussion thread:
- CreateUserModalComponent.tsx
- CreateUnitModalComponent.tsx

## Type of change

## Checklist

## Limitations

The following changes have been briefly tested by manually checking if the Create routes for each of the client files work successfully. I did not initially check to see if the other requests for each client (such as edit and delete) as the changes I've made were strictly in the functions and API calls that directly relate to. However, after this pull request is created, I plan to skim through and check the basic functions of each page (create/edit/delete). I assume that there will be no issues with these changes, but if there is, I will mention it as additional comments if they come up.

Changes were discussed around chartQuerySelectors.ts, but it was found that the use of omit() here was used to exclude local properties. This is different than the other clients that have properties that have been fetched from the API layer. Therefore, the maintainer and I have discussed that omit() can remain as is on this file.

The following changes have been tested using npm run check and npm run test.


aduques:
After going through the pages again, I have noticed the following errors:

image



Conversions (create/edit/delete) works.
Groups (create/edit/delete) works.
Meters (edit) does not work. --> error message: Failed to edit meter with message: "7/15newmeter" (identifier: 7/15newmeter, type: egauge) invalid date/time format

Units (create/edit/delete) works.

I did not catch this error with EditMeterModalComponent.tsx initially as I did not work with the file directly. I will look into why this error happens.


UPDATE: I want to make the assumption that an error could have occurred because the test cases have been written to expect omit() (or by having omit() moved to API layer, it causes the omit() to happen before/after where the checks expect it to happen). I am not too familiar with how the test cases are handled, but I've only traced the error messages back to csvPipelineTest.js and found the testCases object and the numUploads object that contained some parts of the error message structure that I noticed.

This was part of the error message that made me assume that the omit() change could be problematic. (specifically about 'Error parsing Reading')

      -<h1>FAILURE</h1>CSVPipelineError: User Error: Meter with identifier 'pipe29' not found.
            +<h1>FAILURE</h1><h2>It looks like the insert of the readings had issues with some or all of the readings where the processing of the readings returned these warning(s)/error(s):</h2><br>For meter pipe29: Error parsing Reading #1. Reading value gives 24 with error message:<br>The first ever reading must be dropped when dealing with cumulative data.<br>For reading #1 on meter pipe29 in pipeline: previous reading has value 0 start time 1970-01-01T00:00:00Z end time 1970-01-01T00:00:00Z and current reading has value 24 start time 2021-06-01T00:00:00Z end time 2021-06-02T00:00:00Z with timeSort increasing; duplications 1; cumulative true; cumulativeReset false; cumulativeResetStart 00:00:00; cumulativeResetEnd 23:59:59.999999; lengthGap 0; lengthVariation 0; onlyEndTime false; honorDst false; relaxedParsing false; useMeterZone false; warnOnCumulativeReset false; useMeterFrequency false; useMeterFrequencyVariation 0<br>For meter pipe29: Error parsing Reading #4 with cumulative data. The reading value provided of 240.x is not considered a number so all reading are rejected.<br>
      
    
I plan to investigate this further tomorrow, and test the edit behavior after reverting the omit() changes to see if it is responsible for it.

huss:
I wondered about why this would happen so I looked. The development branch shows the same issue. I looked at edit meter to start and found that the new regular expression check on the date fails. I think it has to do with how the date is formatted vs. what the check is doing. This is my fault for missing it during review and it has nothing to do with your PR. I'm going to track this down/fix and then I assume the PR will be fine. I will update when I have further details.

huss:
PR #1674 is now in to fix this issue. The problem was that the start/endTimestamp for meters does not have to have a timezone. I updated the checks to reflect this. @aduques Since you found this error would you be interested in reviewing my proposed change? It is okay if not and I'll find someone else. Once this clears and is merged into your code then this PR should stop failing for this reason (at least I hope so :-). Sorry for this issue.

src/client/app/redux/api/unitsApi.ts
```diff
@@ -7,12 +7,14 @@ import { RootState } from 'store';
 import { UnitData, UnitDataById } from '../../types/redux/units';
 import { baseApi } from './baseApi';
 import { conversionsApi } from './conversionsApi';
+import { omit } from 'lodash';
 export const unitsAdapter = createEntityAdapter<UnitData>({
 	sortComparer: (unitA, unitB) => unitA.identifier?.localeCompare(unitB.identifier, undefined, { sensitivity: 'accent' })
 });
 export const unitsInitialState = unitsAdapter.getInitialState();
 export type UnitDataState = EntityState<UnitData, number>;
```
huss, line 16: Could you please remove the added, second blank line.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
```diff
@@ -161,11 +161,18 @@ export default function CreateConversionModalComponent() {
 	};
 	/* End Warning Modal */
 
+	// The helper function will handle the omit() which separates it from the addConversionMutation.
+	// The addConversionMutation will call this helper function, so that omit() is not directly called.
```
huss: Not a big deal but I don't think addConversionMutation calls the new function. I think it is called to fix up the argument that is used for that function.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -109,6 +109,10 @@ export default function CreateUserModal() {
 	// End Modal show/close
 
 	const handleSubmit = async () => {
+		// id is not used when creating a newUser. omit() could be used to exclude id for userDetails to be consistent with the implementations
```
huss: I think I confused you during our conversion. userDetails contains 8 key/value pairs of which 4 are used in the route for createUser. id is not one of them. The others are used internally to this component to track status (such as passwordMatch). I think this comment should be updated to reflect this (if you agree).

aduques:
@huss I'm working on reviewing PR #1674. I plan to pull and apply the changes made to my current branch. Then, I will see if I can use the requests on each client file.

huss (changes requested):
@aduques Thank you for another contribution. Review and testing found it works as desired. I made three minor comments to consider. This PR will wait for the fix in PR #1674 but it should be very close to done.

src/client/app/redux/api/unitsApi.ts
aduques: I have noted this change.

src/client/app/components/conversion/CreateConversionModalComponent.tsx
aduques: I have updated the comments.

	      // This helper function will fix up the argument that will be used in addConversionMutation().
	      // The helper function will handle the omit() which separates it from the addConversionMutation().
	      // This helper function will also computes bidirectional based on the current source/destination selections.
	      // This helper function is introduced to allow the CreateConversion to be similar to the Create requests on other client files.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
aduques: I agree with this change, and have made revisions.

		      // id is not used when creating a newUser, and is one of several userDetails fields that are only used internally to this component
		      // (e.g. to track status like passwordMatch) rather than being sent to the createUser route. omit() could be used to exclude these unused
		      // fields from userDetails to be consistent with the implementations on other client files, but it is not necessary at the moment. In addition,
		      // introducing omit() in the client files for User could cause unexpected issues. We are acknowledging that the codebase currently has a different
		      // implementation for handleSubmit() compared to other client files.

aduques:
@huss I have finished reviewing the comment suggestions and pushed the changes.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
huss: Okay. I pushed a minor wording change. I hope this is okay with you.

aduques: I have seen the change, and yes it is okay. Thank you for applying the corrections.

huss (approved):
@aduques Thank you for the update on minor comments. Everything looks good now. Congratulations on another accepted OED contribution. OED appreciates your continuing contributions.

PR #1603, opened 2026-04-13
Improved Error Messages in PreferencesComponent.tsx

BrianRaymond800:
# Description

This is a simple change to address a concern from Issue Nineteen. Most of Issue Nineteen's problems have already been fixed by other contributors, but this pull request addresses an issue with error messages that were not fully descriptive.

Fixes Issue 19

## Type of change

- [ ] Note merging this changes the database configuration.
- [ ] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations

A few of the areas mentioned in Issue 19 could have further error message improvements in order to be more clear for users, but this would best be done in a larger assessment of all of the routes and their associated error messages.

PR #1672, opened 2026-07-16
Session invalidation

Better-Code-Saul-E:
# Description

Continue the work started in #1593 by @Oykunle  to implement server-side session invalidation for JWT authentication. This PR addresses the three outstanding review coments from #1593 that were left unresolved when the original author became unavailable to continue the work. See #1652.

Authors: @Oykunle  (original implementation), @Better-Code-Saul-E (review comment fixes)

1. Added a missing 'expect(beforeVerify.body).to.not.have.property('message')' assertion in the "token verify before logout" test
2. Added the missing 'message' property assertion on the first logout call in the "repeated logout" test
3. Renamed the 'login' variable in 'src/server/app.js' to 'loginLogout', and updated its usage in the same file to match the already renamed "loginLogout.js" route file

Fixes #1652 

## Type of change
- [x] Note merging this changes the database configuration.
- [ ] This change requires a documentation update

## Checklist
- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations
- As with the original implementation in #1593, this invalidates all tokens globally per user, not per-device or per-session.
- Requires a database migration for existing deployments

Issue #1616, opened 2026-04-27
Improved Error Messages throughout preferences page.


BrianRaymond800:
### Is your feature request related to a problem? Please describe.

Many of the error messages on the Preferences page are lackluster and should be improved to better communicate the issue to the user.

### Describe the solution you'd like

Error messages should clearly communicate to the user what the issue is without using jargon that can only be understood by a developer. Numbers should be accurate. Phrasing should make sense.

### Describe alternatives you've considered

This is a low priority issues, as the page is usable as is. However, this change could improve quality of life.

### Additional context

This Issue is related to the drafted PR #1603 .

GoodKimchi:
I’d like to work on this issue. I’ll start looking into the preferences page error handling and submit a PR once I have an update. This would be my second contribution to this project.

huss:
@GoodKimchi OED would be happy to have you work on this and I've assigned it to you. In the past I reviewed the linked PR so let me know if anything is not clear. Also, if you want to discuss what messages are unclear or possible solutions then just let me know.

huss:
@GoodKimchi There is a team that is starting issue #1645 to fix up PR #1603. This relates to this issue. Thus, I wanted to see if you are actively working on this and what the status might be to minimize overlap of work. Thanks.

GoodKimchi:
@huss me and @aduques are working on both of these #1645 / #1603 . I was planning on starting #1616 tomorrow but if the work overlaps me and @aduques can just work on #1645 / #1603 instead rather than #1616 let us know please.

huss:
> [@huss](https://github.com/huss) me and [@aduques](https://github.com/aduques) are working on both of these [#1645](https://github.com/OpenEnergyDashboard/OED/issues/1645) / [#1603](https://github.com/OpenEnergyDashboard/OED/pull/1603) . I was planning on starting [#1616](https://github.com/OpenEnergyDashboard/OED/issues/1616) tomorrow but if the work overlaps me and [@aduques](https://github.com/aduques) can just work on [#1645](https://github.com/OpenEnergyDashboard/OED/issues/1645) / [#1603](https://github.com/OpenEnergyDashboard/OED/pull/1603) instead rather than [#1616](https://github.com/OpenEnergyDashboard/OED/issues/1616) let us know please.

[...]

Issue #1592, opened 2026-03-30
proper validation of passed ids


huss:
### Describe the bug

The following routes send values after the route name using the colon (:) syntax (found by searching for /:):

- src/server/routes/compareReadings.js
- src/server/routes/groups.js
- src/server/routes/maps.js
- src/server/routes/meters.js
- src/server/routes/readings.js
- src/server/routes/unitReadings.js
- src/server/routes/users.js

When the values are validated, it uses a string match that does not exclude all bad values. There are two possible (maybe more) fixes:

- Enhance the validation on these strings which can include comma separated values.
- Send the values as parameters inside the request as done for all other values. This would allow for direct validation as an array of integer values and (I think) min of 1.

Personally, I think the second option might be better but here are some considerations:

- Was there a good reason it was done this way or is it a historical artifact from early on in the project?
- Does this impact any current scripts or endpoints that are directly used outside the OED web app? This needs to be determined and a plan if it is the case.
- Testing should be done to verify the route is fairly easy and the validation is too. This could be done for a sample route with comma separated values.
- All the uses in OED need to be updated to reflect this change. This should be easier now that OED is mostly using the Redux Toolkit. Note maps have a long-standing PR (PR #1314) that would need to be rolled in if it is not yet merged.
- All the new tests need to check/validate this.  PR #1583 has a TODO in src/server/test/routes/groupsParamsTest.js that indicates the group tests also return 200 for bad values (it should not). This is one place but all tests related to the routes updated would need updating. This includes route tests but also general tests that send these values.

### Expected behavior

OED should properly validate all values sent. OED is now mostly doing that and this would be a remaining item.

### Additional context

None

huss:
This is to note that some tests that look at these values tried to use and empty string (''). This did not test what was desired and actually tried to hit a different route. Once this change is made then checking blank could be checked again.

PR #1677, opened 2026-07-16
Issue1645 followup Fix PreferencesComponent file size validation behavior

### LINKED ISSUE #1645: Continue work started in PR #1603 for preferences component error messages
### Description

PR #1603 began work to to fix up preferences component error messages. The original developer(s) will not be continuing this work. As a result, OED would like the work continued by others to address any comments and complete the task as needed. Note the referenced issue is not part of OED issues.

If any comments, ideas or needed work is unclear or you have questions, then please reach out to the OED project.

### Additional context

New developers will not be able to push commits to the PR branch. The PR branch should be cloned to your machine, create a new branch and pushed to your own GitHub fork. Once the work is complete and on your GitHub fork, a new PR can be created from your branch on GitHub. The description should reference the original PR as the starting point and acknowledge the previous work's authors. Note OED requires that the original commit chain be present in the new PR and properly document who did what work. OED will close the original PR after any new one is merged. If you make progress but cannot address all needed work then OED would still like to have a draft PR created with the work you did complete. If you have questions about this process then please contact the OED project.

GoodKimchi:
# Description

This pull request continues the work from PR #1603 and addresses issue #1645.

The change restores the intended client-side validation behavior for the default file size settings in `PreferencesComponent.tsx`. The maximum validation bound is restored to `Infinity`, which prevents invalid file size values from being submitted while keeping the validation consistent with similar fields on the meter and unit pages.

This pull request also removes a duplicated line in `src/server/routes/preferences.js` that was introduced during a merge conflict and fixes the related indentation.

Contributors:
- Oscar Bedolla (@GoodKimchi)
- Andrei Solomon Duque (@aduques)

Fixes #1645

## Type of change

## Checklist

## Limitations

`npm run check` passes.

`npm run test` one failure caused by a PostgreSQL deadlock in the test:

`Retrieves 15 min reading when asked for 1 hour interval so looking at hourly readings`

The reported error was `deadlock detected`. This is a known test issue currently affecting the development branch and is unrelated to the changes in this pull request. @aduques did not have this issue


src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -380,6 +421,8 @@ export default function PreferencesComponent() {
 								showErrorNotification(translate('failed.to.submit.changes') + err.data);
 							});
 					}}
+					/* TODO DEBUG: removed the invalidFuncs check so that we can test the new error messages that display specific reasons */
```
huss: The code needs to go back to how it was and not for debugging.

src/client/app/translations/data.ts
```diff
@@ -215,7 +215,8 @@ const LocaleTranslationData = {
 		"failed.to.delete.map": "Failed to delete map",
 		"failed.to.edit.map": "Failed to edit map",
 		"failed.to.link.graph": "Failed to link graph",
-		"failed.to.submit.changes": "Failed to submit changes ",
+		"failed.to.submit.changes": "Failed to submit changes: ",
```
huss: Changing this was a little risky. It is also passed to UnsavedWarningComponent. However, that does not actually seem to use the the passed value so it is okay. I'm noting this and have an item to look into why it is passed.

huss (changes requested):
@GoodKimchi & @aduques Thank you for this contribution. I've made two comments on specific code to consider. I did not do a full review because I am unsure about the changes. It appears you now check if values are invalid at save and pop up a toast message if they are. What OED does on other pages is, when possible/reasonable, detect bad input when it happens, put a red error box around the item, put a red error message below and disable the save button. This means the user immediately knows the issue and cannot save the bad values. This should also be correctly carried to the unsaved warnings so it also will not save. There are limited circumstances where OED does wait until save to check values but that should be where immediate checks cause issues or are too complicated/interrelated to do at field entry. I'm not sure those apply here so I'm seeking your thoughts on whether the popup on save makes sense or if it should be the other way. I'm also noting the discard button does not work but I don't know yet if that is from your changes. Once these are settled/addressed, I should be able to do a fuller review. Please let me know if anything is not clear or you have thoughts/questions.

aduques:
@huss I wanted to clarify our approach with the changes. Our changes has updated the error notifications that appear when the user tries to save with invalid changes, but we've also preserved the original red box and red text errors that appear as soon as the user tries to input invalid data. This will also disable the save button for the user.

The reason why the TODO DEBUG for disabled={!hasChanges} was included was because it was necessary to show the newly updated error notifications when a user attempts to save with invalid changes (to demonstrate the changes in this PR strictly). As mentioned before, this should never be shown to a user in a normal use case due to disabled={!hasChanges || Object.values(invalidFuncs).some(check => check())} preventing the user from encountering it in the first place. The TODO DEBUG will ultimately be removed as it doesn't satisfy the requirement of showing the red box and red text error and disabling the save functionality.

image


As intended, the red box and red text error displays when it detects an invalid input + disables the save button. As per the original comments from the original PR, the bounds have also been reverted back to Infinity to catch extremely large numbers that would bypass invalidFuncs.

Screenshot 2026-07-17 at 10 28 24 AM

I would appreciate confirmation if this approach satisfies the intended behavior, or if we should make further changes. In addition, I have noted the changes of the invalidFuncs not being carried over to the UnsavedWarning component and have that button disabled.

I don't recall making changes that affected the discard feature, but I will go through the file and make sure that the button properly fetches the current admin state as expected.

src/client/app/components/admin/PreferencesComponent.tsx
aduques: This TODO DEBUG was included to demonstrate the changes to the save fail error notification. It will ultimately be removed, and the user will not encounter the updated error notification due to the checks in place preventing the save functionality. I will remove the TODO DEBUG as requested, but wanted to point out that this was how I was able to test the updated error notification that displays which invalidFunc fails during the condition of a invalid input bypassing the existing checks in place.

src/client/app/translations/data.ts
aduques: I acknowledge this potential problem and will revert the changes. Instead of updating the existing message to contain a ":" and possibly affecting other instances of it that don't need the change, I will make a new one containing the ":" so that it can be used in this context. With this approach, the new message will be updating only the error notification for a save fail in PreferencesComponent.tsx and not anywhere else that's not intended.

aduques: I have made the following changes in data.ts:

		      "failed.to.submit.changes": "Failed to submit changes ",
		      "failed.to.submit.changes.saved": "Failed to submit changes: ",

The TODO DEBUG was used to display the error message during the case of invalid inputs somehow bypassing the invalidFuncs checks. With the changes, the error message is retained, but the new addition will not affect the implementation of failed.to.submit.changes in the unsaved warning component.

image

When attempting to leave the page with unsaved changes, the unsaved warning component pops up. When attempting to save with the invalid inputs, this error message pops up.

image

However, as you have pointed out, the error message that points out is the error message that is declared in UnsavedWarningComponent.tsx and not the UnsavedWarningComponent object in PreferencesComponent.tsx

In UnsavedWarningComponent.tsx

      export function UnsavedWarningComponent(props: UnsavedWarningProps) {
	      const { hasUnsavedChanges, submitChanges, changes } = props;
	      const blocker = useBlocker(hasUnsavedChanges);
	      const handleSubmit = async () => {
		      submitChanges(changes)
			      .unwrap()
			      .then(() => {
				      showSuccessNotification(translate('unsaved.success'));
				      if (blocker.state === 'blocked') {
					      blocker.proceed();
				      }
			      })
			      .catch(() => {
				      showErrorNotification(translate('unsaved.failure'));
				      if (blocker.state === 'blocked') {
					      blocker.proceed();
				      }
			      });
	      };

In PreferencesComponent.tsx

			      <UnsavedWarningComponent
				      hasUnsavedChanges={hasChanges}
				      changes={localAdminPref}
				      submitChanges={submitPreferences}
				      successMessage='updated.preferences'
				      failureMessage='failed.to.submit.changes'
			      />

If desired, a possible issue can be created in the future to update the error messages from UnsavedWarningComponent.tsx to be more specific. If I recall, SimpleUnsavedWarningComponent.tsx was created because UnsavedWarningComponent.tsx could not be applied to other client files cleanly. In the client files that use SimpleUnsavedWarningComponent.tsx, the showErrorNotification() and showSuccessNotification() have been updated to have specific error messages. This change was not done for UnsavedWarningComponent.tsx and only contains generic error/success messages.

huss:
I have not looked at the new code yet. Having said that, I would appreciate you explaining the advantage of the new error  check and message on save. You state the user should never see it in a normal use case. So, does it only happen if the code is not working properly? If a bad value was passed to the server then hopefully it would be rejected and the user would see that without the new check. Thanks.

aduques:
https://github.com/OpenEnergyDashboard/OED/issues/1645#issuecomment-4952542518

By asking the question about the error message, "Failed to submit changes Bad Request," we were under the assumption that this error message was to be made more specific in addition to the end user never seeing it as the checks are in place to prevent it. The condition of preventing a user from attempting the save was always intended to be left untouched. The changes to the error check were included so that the invalid input of a large value in file size limit could be captured by invalidFuncs() (large value such as 9999999999999). I can see why this might have caused confusion as this addition to the error checks was only meant to check the specific case that was mentioned in the issue description. This error check addition should have been added with a TODO DEBUG as it was added to help display the new message.

Specifically these were supposed to be included as the TODO DEBUG and not meant to be added as a new condition for the error check:

      || Number(localAdminPref.defaultWarningFileSize) > Number.MAX_SAFE_INTEGER;
      || Number(localAdminPref.defaultFileSizeLimit) > Number.MAX_SAFE_INTEGER;

As for the new updated message it self, if I have made the wrong assumption about your answer to the question about the error message, "failed to submit changes Bad Request," and updated the error message unnecessarily, then I can quickly revert the additions if desired.

The benefit of this updated message was that it would be updated to follow a similar pattern as other save error messages on other pages and it would contain a specific message that lets a developer know which of the functions/checks in invalidFuncs() was raised. Before the changes, the error message would just say "Failed to submit changes Bad Request" when a save failed. However, with the changes, it would say "Failed to submit changes: (Default File Size Limit has invalid values)" when it detects that fileSizeLimit() returns true in invalidFuncs(). This lets the developer know which check produces the error given. Though, these messages won't be accessed due to the invalidFuncs() disabling the save button.

The core changes of this issue would just be strictly the comment suggestions of the original pull request. The added TODO DEBUGs were related to additions that I could have mistakenly thought were needed about the error message on save.

huss:
Sorry this sat so long.

> By asking the question about the error message, "Failed to submit changes Bad Request," we were under the assumption that this error message was to be made more specific in addition to the end user never seeing it as the checks are in place to prevent it..........

I agree with this goal. What I was wondering is if the user should (virtually) always see the updated messages without changing the save because the component is doing the same validation with a good message.

> As for the new updated message it self, if I have made the wrong assumption about your answer to the question about the error message, "failed to submit changes Bad Request," and updated the error message unnecessarily, then I can quickly revert the additions if desired.
> 
> The benefit of this updated message was that it would be updated to follow a similar pattern as other save error messages on other pages and it would contain a specific message that lets a developer know which of the functions/checks in invalidFuncs() was raised. Before the changes, the error message would just say "Failed to submit changes Bad Request" when a save failed. However, with the changes, it would say "Failed to submit changes: (Default File Size Limit has invalid values)" when it detects that fileSizeLimit() returns true in invalidFuncs(). This lets the developer know which check produces the error given. Though, these messages won't be accessed due to the invalidFuncs() disabling the save button.

If the only change is to show the user the error message when a save happens then that is fine. I thought it might be redoing the checks and that I feel more ambivalent about. It is more code to maintain with little chance the check will fail (assuming I have the correct idea of the situation). And, the unlikely missed case from not having it in save will return a server error that you now show. Does this make sense? If I have the wrong idea then just let me know.

> The core changes of this issue would just be strictly the comment suggestions of the original pull request. The added TODO DEBUGs were related to additions that I could have mistakenly thought were needed about the error message on save.

I'm okay with them if they are for debugging. Maybe you can remove them now so only the code you want permanently in OED is in the PR. That way I can see what will be merged. I can either test another way or look at an earlier commit to temporarily put back a debug statement. How does that sound?

As always, let me know anything unclear/thought/questions.

huss:
The recent merge of PR #1672 is now causing a merge conflict. Please let me know if you need any help with this. The route moved from login to loginLogout and with new subroutes.

aduques:
> I agree with this goal. What I was wondering is if the user should (virtually) always see the updated messages without changing the save because the component is doing the same validation with a good message.

I'm not quite sure if I understand this statement, but I am interpreting it if a user types an invalid input, then the updated message (from showErrorNotification) would always be displayed like the red box/red-line text error. Or if the updated message be used to replace the existing message of the red-line text error. To clarify, the validation for save will remain the same. The only difference is the new updated error message for save (showErrorNotification). If it is desired to have the updated error notification be displayed on the page when the server detects invalid changes, I would be initially against the idea. The reason for this is because I don't recall showErrorNotification or showSuccessNotification being used when the component reloads (like with React.useEffect() or with the red-text/ red box errors). These notifications are primarily used for when a request has been made. If this is desired (given that I didn't misinterpret your statement), I would suggest that other pages have a similar change where the notifications appear when the server detects invalid changes so that there is consistency.

To close my thought, the idea of having the updated messages (error notification for save) appear for the user in the same way as the red box/red text errors is a good idea on the Admin Settings page. However, I would assume it might be problematic if this behavior only existed on Admin Settings page and not consistent with other pages. I would appreciate insight on this as I believe I may have misinterpreted the initial statement.

If desired, I would propose creating an issue that applies this behavior on other pages to let users know why the save button is disabled for them. If not already, the error messages would display to the users which checks are being raised to disable the save button. The error messages would keep in mind to the specific checks in each page, but this would required to be looked into more to see if it would make sense to add.

As for the other comments, you have answered all of my points and understood my thought process with the code additions. Any changes to the checks were strictly for testing and I will restore it back in the next commits. As for the updated error messages, I will assume that the changes are welcome and will leave it in for you to test along with the other changes.

huss:
I'm feeling as if I am having trouble expressing my thoughts so they are clear. If you want to talk about this verbally then just let me know. Having said that, let me add this. I was not trying to propose that the error popup happens on page load/reload. What OED has on other pages (and I think here too) is that red boxes with error messages are placed on the page when issues are detected and the save is disabled. If associating with a given input box does not make sense then (I think) the message goes at the bottom of the page just before the save button. The idea was that this should make it clear why the save is disabled. (For now I'm not discussing a modification will happen on save to clean up the input that still allows the save.) If this is correct, then I was wondering why recheck on save on the preferences page since they should never get there if there are issues - at least in normal cases. If that is not the case then I would like to discuss which ones and why. Now, every page should display if the server rejects the request and why since the page checks either miss something or cannot readily check for issues. That is the final fail safe. That is okay and acceptable. Does this make more sense? I hope I'm understanding what you are saying.

Note PR #1672 changed the login/logout route to loginLogout/ and added subroutes. It was recently merged and is likely causing the merge conflict. Please let me know if you want any assistance in resolving it.

aduques:
> If associating with a given input box does not make sense then (I think) the message goes at the bottom of the page just before the save button.

I understand that having the showErrorNotification was not what you meant. I'm interpreting it that we should apply the red box / red text error messages that are used throughout OED to highlight textboxes with invalid inputs to the save button when it is disabled. The red box/red text errors will display the updated messages that specify which flags are raised in the error checks to let the user know why they can't save.


> If this is correct, then I was wondering why recheck on save on the preferences page since they should never get there if there are issues - at least in normal cases. If that is not the case then I would like to discuss which ones and why.

I want to clarify that when I previously said, 

"The TODO DEBUG was used to display the error message during the case of invalid inputs somehow bypassing the invalidFuncs checks. With the changes, the error message is retained, but the new addition will not affect the implementation of failed.to.submit.changes in the unsaved warning component." 

I was not referring to a known bug that allows the user to bypassing the invalidFuncs check. I talked about this to describe the TODO DEBUG case where the invalidFuncs validation checks were removed to test out the updated error messages that displayed specific functions from invalidFuncs. I have not experienced any bugs with this guardrail, and as far as I am aware, invalidFuncs' only issue was the max bounds for fileSizeLimit not being Infinity (which is addressed and fixed in this PR).

I'm assuming the "recheck on save" refers to the addition of the getInvalidFieldNames() which would call invalidFuncs() again:

      onClick={() => {
						      const invalidFieldNames = getInvalidFieldNames();
						      if (invalidFieldNames) {
							      showErrorNotification(
								      translate('failed.to.submit.changes.saved') + '(' + invalidFieldNames + translate('failed.to.submit.changes.fields')
							      );
							      return;
						      }
      
						      submitPreferences(localAdminPref)
							      .unwrap()
							      .then(() => {
								      showSuccessNotification(translate('updated.preferences'));
							      })
							      .catch(err => {
								      showErrorNotification(translate('failed.to.submit.changes.saved') + err.data);
							      });
					      }}

The reason why this helper function calls invalidFuncs() again is to strictly confirm which of the functions in invalidFuncs is raised to determine how to update the error message. I understand that the first call of invalidFuncs() would prevent the recheck from even happening as the save is disabled (the original call that is a condition for disabling save). This recheck would only work with the TODO DEBUG case where the first check is removed and allowing the save button to be available.

With that in mind, I did think of alternatives, but I was hesitant to do so. This alternative solution might work with wanting to implement the red-box/red text error to the disabled save button. This alternative solution would include modifying the invalidFuncs() to update the error message (essentially what getInvalidFieldNames() does). This would prevent a recheck on invalidFuncs() as the initial check will handle disabling the save button if needed and updating the error message. I was hesitant to do this as I was not sure how changing invalidFuncs() might affect the behaviors elsewhere (if there is any beyond PreferencesComponents.tsx, but I assume not).

I think removing the "recheck" is definitely the approach since it would not be accessible in a normal use case when the first check disables the button. Something like the alternative solution I proposed above might be worth trying if you agree.

> Note PR #1672 changed the login/logout route to loginLogout/ and added subroutes. It was recently merged and is likely causing the merge conflict. Please let me know if you want any assistance in resolving it.

We are currently working on merging the new changes and we will have a follow up message if we encounter any merge conflicts.

aduques:
We have pulled and merged the recent changes for PR #1672. We didn't experience any merge conflicts, but noticed an issue with the login/logout. The login and logout still work, but there is an error message after the logout request in console. We are assuming this is not related to a merge conflict as the option to resolve a merge conflict did not pop up and the application started up without any issues.

Screenshot 2026-07-22 at 1 34 02 PM

Screenshot 2026-07-22 at 1 34 25 PM

Initially, I ran into an issue that was preventing me from logging in with either 'test' or 'test@example.com'. However, this was quickly resolved by refreshing the page and trying again (I'm assuming the changes were not live when I initially tried). I wanted to point this out as it seems other students ran into this issue based on the discord channel for codeday-sose.

huss:
Thanks for the info on login/out. I received several reports yesterday. I posted to the OED Discord server on the developer channel about it so see that if you didn't already. It probably requires a DB change to work completely correctly.

huss:
Thanks for the comment about the recheck. Given the red boxes & messages catch all the known issues and stops saves, I don't think OED needs to run a second set of checks on save to clarify the error message. It seems much more likely that the error message is from something else and not the checks being done in the component. Thus, the second check will not likely add any useful information. Note OED has decided not to provide the actual error from the server but say something generic as a recommenced security change. The detailed information is in the logs. So, if I understood your message and you agree with this, I think the secondary check on save can go along with all the debug code. Then I can do what is hopefully a final sweep. Please let me know any thoughts.

aduques:
I understand that it is a design choice for OED to not include these actual error messages to the users and instead shows a generic error message. Therefore, the specific error messages will be scrapped from this PR.

However, I wanted to show you what the specific error messages would look like as a red text/ red box error on the save button. This code is removed from the PR and will not be committed, but I thought it would be interesting to consider adding these messages to OED where buttons would be disabled and it could give info to users as to why.
I've kept the specific messages as placeholder, but it could easily be replaced with generic messages to follow the design choice. I'd imagine this being something potentially helpful to users, but this idea would need to be refined more to give users information that is helpful and not already provided elsewhere. In addition, if this were to be considered in the future, it would require a better design implementation visually as I found what I've tried out to not be as polished visually.

The changes involve assigning the result of invalidFuncs to a variable, invalidFuncsResult, which allows it to be reused and only have one check as desired. We are able to get specific error messages by creating a dictionary/map which has the field name for each function in invalidFuncs. If the function returns true, the function name is used as a key to look up the field name in its bucket. The action of checking invalidFuncs and updating the error message is in the same pass and would not require a second check.

Screenshot 2026-07-23 at 1 00 18 PM


To summarize the code that will be commited in this PR, it will strictly be the requested changes of the original PR we were tasked with. The secondary check, specific error messages, and the debug code will all be removed for the final sweep.

I apologize for the  time spent looking at the proposed ideas with the updated error messages for it to not be included in this PR, but I hope the basis of it can inspire future additions to OED.

huss:
Please down worry at all about the discussion. I find your ideas interesting and they make me think. I'm open to UI changes but I want to be sure I understand the change. It can also be propose to a wider audience to see if they want it (I don't override, in general, the consensus of the group.) Currently OED tries to show each issue on the input page with a red box/message and not allowing save. Your proposal seems a similar look but done during submit. Is it correct that a user should not see your new messages if it is properly covered by the current style that stops saving? If it is only other ones and you need to know in advance to create your map then could these be added to the current checks that disable saving? If it involves an unexpected error then I'm unsure how OED could help the user with a message without showing the actual error (which we don't want to do for other potential issues). My hope is that over time the save button will not allow many errors to come back where the user sees the generic error pop up.

I still feel I'm missing something and I welcome your thoughts.

aduques:
> Your proposal seems a similar look but done during submit. Is it correct that a user should not see your new messages if it is properly covered by the current style that stops saving?

In the screenshot of the error message below the save button, it appears with the current behavior of the save button. So, the error message below the submit does not appear during the submit request, but instead it gets updated when the invalidFuncs() is called during the check. In other words, the current validation behavior that determines if the saved button is disabled is kept exactly the same. The only addition is that the error message would be displayed at the bottom to let the users know why it's still disabled.

As before, I am unaware of ways that users will be able to fail the invalidFuncs() checks (or any other unexpected errors that will raise these error messages), so I cannot provide a scenario at the moment. While I acknowledge that the error message itself can provide useful information, the current guardrails prevent the user as much as possible from even seeing these error messages (as intended).

If interested, here was the code snippets that I left out from the PR that were related to that error message:

      const fieldTranslationIds: { [key: string]: string } = {
	      readingFreq: 'default.meter.reading.frequency',
	      minDate: 'default.meter.minimum.date',
	      maxDate: 'default.meter.maximum.date',
	      readingGap: 'default.meter.reading.gap',
	      meterErrors: 'default.meter.maximum.errors',
	      warningFileSize: 'default.warning.file.size',
	      fileSizeLimit: 'default.file.size.limit'
      };
      
      // Computed once by iterating invalidFuncs a single time - reused for the Submit button's disabled
      // state, the FormFeedback visibility, and the detailed list of which fields are invalid.
      const invalidFuncsResult = Object.entries(invalidFuncs)
	      .filter(([, check]) => check())
	      .map(([key]) => translate(fieldTranslationIds[key]))
	      .join(', ');
      
      
      <div className='d-flex justify-content-end mt-3'>
	      <Button
		      type='button'
		      onClick={discardChanges}
		      disabled={!hasChanges}
		      style={{ marginRight: '20px' }}
		      color='secondary'
	      >
		      {translate('discard.changes')}
	      </Button>
	      <Button
		      type='submit'
		      onClick={() => {
			      submitPreferences(localAdminPref)
				      .unwrap()
				      .then(() => {
					      showSuccessNotification(translate('updated.preferences'));
				      })
				      .catch(err => {
					      showErrorNotification(translate('failed.to.submit.changes') + err.data);
				      });
		      }}
		      disabled={!hasChanges || !!invalidFuncsResult}
		      color='primary'
	      >
		      {translate('submit')}
	      </Button>
      </div>
      <FormFeedback style={{ display: invalidFuncsResult ? 'block' : 'none' }}>
	      {translate('failed.to.submit.changes.saved') + '(' + invalidFuncsResult + translate('failed.to.submit.changes.fields')}
      </FormFeedback>

huss:
Thank you for explaining this several times. I think I am finally starting to understand. I think the idea is fine but I'm not yet convinced it is worth doing across the pages. Here is my current thinking:

- It is code that needs to be maintained.
- The pages we are discussing are all admin-only pages. This means they are not the typical OED user and they are much more likely to look at the help pages and/or have experience with the UI. If these were used by many users then that would be different. For example, OED works to remove choices from graphics pages that are not allow and shows a message in large font on the center of the graphic or as popups that are specific.
- Many admin pages would not need scrolling to see the red box and text so it isn't that hard to see if you know to look for it (whenever the save is disabled).
- This is a common web UI technique. I acknowledge some fancy pages jump to the bad value(s) on save but that would be more work for what (I perceive) as small payback on admin pages.

As I said before, you are making me think about alternatives. Given what you have said and done, this already come to mind. A non-specific message could be placed near the save button to emphasize that the admin needs to find all the red boxes and fix them. This might help because OED does use red boxes for items that will be modified on save from what is shown and this clearly shows the difference. The unclickable save button does that too but if people think this would help then I'm open/okay with this. This would be consistent across pages so maintaining it would be much easier. I'm unsure if it would be easy to do but I would think it isn't too bad.

Let me know what you think and you can still advocate for your original solution.

aduques:
I agree that my proposed code would primarily be used by admin users who will interact with the create/edit features, whereas actual non-admin users would never see these changes. I wanted to suggest these changes, but I can understand that at the moment it has niche applications.

I support the idea of having a generic message be displayed throughout the pages that explain why the buttons are disabled, but I understand that the effectiveness varies on each page. I believe that it would be important to make sure that the generic message doesn't just repeat what the red text/red box error messages display to be effective.

I think that my original solution is not ready for implementation and it has clear room to be improved on. This idea should definitely be reconsidered if OED has gotten a lot of reports from users stating they aren't clear as to why they cannot access the save buttons. However, the existing red text messages should already clear any confusion, so it is likely not the case. For right now, I think it would make sense to put this idea on hold until it becomes apparent that users need changes to clear any future confusions that they might have. Or, it would also make sense to revisit this idea if there is a future request to improve the UI to make it more new user friendly in any way.

As I continue to work on OED, I will keep a look out for any features that might be potentially confusing for end users and discuss these with you when they arise.

huss:
> As I continue to work on OED, I will keep a look out for any features that might be potentially confusing for end users and discuss these with you when they arise.

I appreciate your thoughts so far and any you have in the future. OED welcomes them.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -345,7 +358,13 @@ export default function PreferencesComponent() {
 					invalid={invalidFuncs.fileSizeLimit()}
```
huss, line 358: I played with this and it isn't showing the max warning as I had hoped. I debugged for a little while but not completely. It seems to relate to the fact that it is using a general function (makeLocalChanges) to handle changes and not one that thinks it is a number. This may unwind back to how it is stored in OED (here and/or server). I looked at src/client/app/components/meters/CreateMeterModalComponent.tsx and it does it differently but seems to check the max value.

Another issue is that src/server/routes/preferences.js has a max of 1000000000. I'm unsure where that came from. src/server/routes/meters.js on maxVal does not seem to have any limit and assumes it could not come through as a number if it was invalid (I think). I'm unclear on why OED has an upper limit on preferences and if the one in meter does what is desired. Again, it may relate to treating as a string instead of a number.

I could not add above so am putting here. The date check here is close to but a little different than meter. Unsure if it matters but wonder about standardizing them.

The old PR has a comment that also included unit for some of these. That may also help/need coordination. 

I'm unsure the right overall solution for this and welcome thoughts.

huss (changes requested):
@GoodKimchi & @aduques Thank you for the updated code and your willingness to reconsider items done. I have now reviewed and tested. I made one comment to consider. I'm open to ideas on what to do and when/who might address it. Please let me know any thoughts.

src/client/app/components/admin/PreferencesComponent.tsx
GoodKimchi: Hi @huss, @aduques and I dug into this a bit more and reproduced the issue. When entering a very large value, such as 9999999999999, for Default File Size Limit, the frontend accepts it without inline validation, but the backend rejects it with a Bad Request because preferences.js currently validates both file-size fields with a maximum of 1000000000.

I noticed you mentioned you were unsure where that limit came from, so I was thinking about a few possible directions:

1. Keep the current maximum (1000000000) and align the frontend validation/messages with the backend.
2. Keep a finite maximum, but define it as a shared/documented constant so the frontend and backend can't drift apart in the future.
3. Remove the application-level maximum and allow any non-negative value if the intent is for deployments to determine their own upload limits.

Since these settings appear to be used mainly for CSV or spreadsheet imports, I would not expect most deployments to approach a 1 GB upload. That made me wonder whether the current limit is an intentional design decision or primarily a safeguard against unrealistic values. A shared constant seems maintainable and would be easy to update later if larger files need to be supported. I’m happy to implement whichever direction makes the most sense.

While tracing this, I also noticed what appears to be the same type of frontend/backend validation mismatch for Default Meter Reading Gap: the frontend effectively allows values above the backend maximum. I think that should probably be handled in a separate issue or PR to keep this one focused, but I wanted to mention it because it appears to be the same class of bug.
Screenshot 2026-07-28 235125

huss: @GoodKimchi  &  @aduques Thank you for thinking this through. My thoughts are below.

> dug into this a bit more and reproduced the issue. When entering a very large value, such as 9999999999999, for Default File Size Limit, the frontend accepts it without inline validation, but the backend rejects it with a Bad Request because preferences.js currently validates both file-size fields with a maximum of 1000000000.
> 
> I noticed you mentioned you were unsure where that limit came from, so I was thinking about a few possible directions:
> 
>     1. Keep the current maximum (1000000000) and align the frontend validation/messages with the backend.
> 
>     2. Keep a finite maximum, but define it as a shared/documented constant so the frontend and backend can't drift apart in the future.
> 
>     3. Remove the application-level maximum and allow any non-negative value if the intent is for deployments to determine their own upload limits.
> 
> 
> Since these settings appear to be used mainly for CSV or spreadsheet imports, I would not expect most deployments to approach a 1 GB upload. That made me wonder whether the current limit is an intentional design decision or primarily a safeguard against unrealistic values. A shared constant seems maintainable and would be easy to update later if larger files need to be supported. I’m happy to implement whichever direction makes the most sense.

Yes. OED should not allow files that are too large. 1 GB seems more than enough. Aligning the front/back-end is a good idea and centralizing so they can never differ is even better. You can proceed with this unless you have thoughts/questions.

> While tracing this, I also noticed what appears to be the same type of frontend/backend validation mismatch for Default Meter Reading Gap: the frontend effectively allows values above the backend maximum. I think that should probably be handled in a separate issue or PR to keep this one focused, but I wanted to mention it because it appears to be the same class of bug. Screenshot 2026-07-28 235125

Nice catch. I agree they should be consistent. Reading gap is in seconds so 86400 is 60 days. That was undoubtedly deliberate as a more than max value to ever use. I think it would stop really silly values while still allowing it to be so big that one can turn off the warnings about gaps. So, that value should consistently be used. Do you want to create the issue or should I? If you want to work on it then put in a comment on the new issue or let me know.

I welcome any thoughts.

GoodKimchi: Thanks @huss . That all makes sense. @aduques and I will go ahead with the shared constant approach for the file-size validation so the frontend and backend stay in sync.
I'd also be happy to work on the Reading Gap validation afterward. I'll create a separate issue to keep this PR focused, then submit a follow-up PR for that consistency fix as well.

GoodKimchi: Thanks for the guidance, @huss . We implemented the shared constant approach and pushed the changes.

The frontend and backend now both use a shared MAX_FILE_SIZE_LIMIT constant from src/common, so the validation is consistent and the frontend prevents values that the backend would reject.

We kept the Reading Gap changes out of this PR as discussed to keep it focused. @aduques has reviewed and tested the changes, so I believe it's ready for your review. I'll open a separate issue and follow-up PR for the Reading Gap consistency changes.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -324,12 +327,28 @@ export default function PreferencesComponent() {
...
+					max={Math.min(
+						Number(localAdminPref.defaultFileSizeLimit),
+						MAX_FILE_SIZE_LIMIT
+					)}
 					maxLength={50}
 					invalid={invalidFuncs.warningFileSize()}
 				/>
 				<FormFeedback>
-					<FormattedMessage id="error.bounds" values={{ min: 0, max: Number(localAdminPref.defaultFileSizeLimit) }} />
+					{Number(localAdminPref.defaultWarningFileSize) < 0 ? (
```
huss: First, this new message is off since it indicated zero is not allowed. If that is why you added it then it can go. Second, I don't think it is needed. It isn't done on other pages. I tried removing the conditional and true part and I get all the messages desired. Thus, if not for the first item I still think this should go. Do you agree?

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -324,12 +327,28 @@ export default function PreferencesComponent() {
 					value={localAdminPref.defaultWarningFileSize}
 					onChange={e => makeLocalChanges('defaultWarningFileSize', Number(e.target.value))}
 					min='0'
-					max={Number(localAdminPref.defaultFileSizeLimit)}
+					max={Math.min(
```
huss, line 330: I'm uncertain when the min is needed. It is not in the form feedback below. Since the defaultFileSizeLimit cannot exceed MAX_FILE_SIZE_LIMIT I thought that the check on defaultFileSizeLimit would be sufficient. What do you think?

huss (approved):
@GoodKimchi & @aduques Thank you for the updated code. Overall, it is good. I have made two comments to consider. I want to note that I think all comments before this one are now covered by newer comments or resolved. Please let me know if anything is not clear or you have thoughts/questions.

huss (changes requested):
Accidentally approved so now marking as Requested changes.

huss:
> We kept the Reading Gap changes out of this PR as discussed to keep it focused.

I'm trying to get this to where only new comments apply so I want to check if you plan to open an issue and/or work on this? If not, I can do the issue.

huss:
> What OED does on other pages is, when possible/reasonable, detect bad input when it happens, put a red error box around the item, put a red error message below and disable the save button. This means the user immediately knows the issue and cannot save the bad values. This should also be correctly carried to the unsaved warnings so it also will not save.

Sorry I missed this before but am now sweeping all previous comments. When I set the warning size to too big a value, it shows a red box and disables the save button. However, if I click on the OED icon (without discarding so I try to leave the page) then the unsaved warning popup still has the save enabled. I was able to save an invalid value to the DB.

I looked into why. First, the route cannot easily check this because it depend on another routed value. However, I think some routes do checks outside the param validation to allow this type of check. The DB does not stop this (maybe it should). Second, src/client/app/components/SimpleUnsavedWarningComponent.tsx has a prop that disables the save button but src/client/app/components/UnsavedWarningComponent.tsx does not. That would need to be fixed to get the save button to be disabled.

Do you have thoughts on this? If you agree, I think it would be fine to open an issue (I can do it) to resolve these as they are not a result of your changes.

aduques:
> Sorry I missed this before but am now sweeping all previous comments. When I set the warning size to too big a value, it shows a red box and disables the save button. However, if I click on the OED icon (without discarding so I try to leave the page) then the unsaved warning popup still has the save enabled. I was able to save an invalid value to the DB.
> 
> I looked into why. First, the route cannot easily check this because it depend on another routed value. However, I think some routes do checks outside the param validation to allow this type of check. The DB does not stop this (maybe it should). Second, src/client/app/components/SimpleUnsavedWarningComponent.tsx has a prop that disables the save button but src/client/app/components/UnsavedWarningComponent.tsx does not. That would need to be fixed to get the save button to be disabled.
> 
> Do you have thoughts on this? If you agree, I think it would be fine to open an issue (I can do it) to resolve these as they are not a result of your changes.

Thanks for the review. We agree with both points. We'll remove the extra conditional FormFeedback logic and simplify the max prop since the defaultFileSizeLimit validation already guarantees it cannot exceed MAX_FILE_SIZE_LIMIT.
We also agree that the UnsavedWarningComponent behavior appears to be a separate issue from this PR. We'd be happy to work on that as a follow-up issue, and we can create the issue soon.

I want to give an insight on the differences between UnsavedWarningComponent.tsx and SimpleUnsavedWarningComponent.tsx as someone who has created the SimpleUnsavedWarningComponent.tsx. SimpleUnsavedWarningComponent.tsx was created as UnsavedWarningComponent.tsx has been hardcoded to work strictly with PreferencesComponents.tsx (Admin Settings page) and could not cleanly because the handleSubmit logic was coded directly into the component. The reason is that the handleSubmit logic not be universally used with the different schemas on different routes. The handleSubmit could only work with the requests and endpoints related to PreferencesComponents.tsx.

The creation of SimpleUnsavedWarningComponent.tsx separates the handleSubmit logic so that it is compatible with the existing submit logic on the other client files. I'm assuming that issues after mine related to updating and debugging files that used SimpleUnsavedWarningComponent.tsx have overlooked UnsavedWarningComponent.tsx usage in PreferencesComponent.tsx. In other words, I should have made sure that UnsavedWarningComponent.tsx had the same functionality and behavior as SimpleUnsavedWarningComponent.tsx in that initial issue.

src/client/app/components/admin/PreferencesComponent.tsx
aduques: I agree that min in this case is redundant. I have removed the min property and have confirmed that you cannot type a value below 0 regardless. I believe that this min property was initially included to capture a case where the user might attempt to input a minimum value. I have not investigated what exactly prevents no negative values from being inputed, but I can confirm that the min property is not responsible and can be removed.

src/client/app/components/admin/PreferencesComponent.tsx
aduques: I agree that since a value less than 0 cannot be typed (negative value), then there is no case that the "error.greater" message is displayed. Therefore, that condition is never reached. When the condition is removed, then the desired messages are still displayed. I have updated the FormFeedback line to the following:

      <FormFeedback>
					      <FormattedMessage
						      id="error.bounds"
						      values={{
							      min: 0,
							      max: Number(localAdminPref.defaultFileSizeLimit)
						      }}
					      />
      </FormFeedback>

huss:
I saw the update. I'm sorry that it is going to take a little while for me to get to this. Let me know if the delay becomes an issue for you.

aduques:
Hello @huss, as @GoodKimchi and I have spent the last few weeks working on OED, we thought we could propose features that could be considered.

## Summary:

After investigating the open and closed issues, we became aware of the longstanding goal of migrating the project to RTK, but the delays in addressing maps have prevented this goal from making any progress. We are aware that maps have been initially considered in the migration, but the longstanding pull request and updates to other areas of the codebase has made it harder to complete the migration. As of currently, Maps lacks the consistent message logging for success/error requests, behavior of handling unsaved changes, UI displays, etc. that the other pages have been receiving work to update and evolve.

Therefore, we are making the proposal of wanting to pick up the work done for map migration to RTK, and contribute to the efforts of getting the implementation done in Maps to be consistent with the other files. Once consistent implementation is achieved, it would allow the maintainer and other contributors to better suggest new features that can universally be applied to all files. An updated Maps implementation would allow more progress for uniform implementation.

## Context/Background:

After reviewing both the open and closed issues list, we have found that OED has been prioritizing the following goals:

- migration to RTK
- uniform implementation on client-server schema
- uniform implementation on UI-server messaging (success/failure notifications)

With these goals in mind, we’ve continued investigating the codebase and found that the maps component has not been completely migrated to RTK as the other components. We have come to this assumption as the maps files were the only ones to make use of 

- import { connect } from 'react-redux';
- extends React.Component

which is not from the RTK framework. In addition, a key takeaway that we found that maps was not completely migrated to RTK was that it was the only client page to not make use of the redux store and slices to maintain state management.

An issue that I’ve previously worked on, https://github.com/OpenEnergyDashboard/OED/pull/1504, has maps excluded from the changes as well.

Knowing that maps was being left out of the consistent implementations on other pages, we wanted to investigate the current efforts towards map migration to RTK. We have found the following:

- https://github.com/OpenEnergyDashboard/OED/pull/1314: Bulk of the efforts to implement the card modals onto the Maps page. This pull request aims to resolve multiple issues, but it ultimately became too messy and required merge conflict efforts. These changes would need to merged in if not already.
- https://github.com/OpenEnergyDashboard/OED/issues/1592: Is semi-related as a followup to 1314. It has yet to be worked on as it is waiting for the changes from 1314 to roll in so that the changes can be applied to maps. It was mentioned that these changes would be relatively easy to implement once the pages have migrated to RTK. This is a supporting reason as to why maps migration is a current longstanding bottleneck to the migration to RTK goal.
- https://github.com/OpenEnergyDashboard/OED/issues/1145: It was noted that the current schema used for Maps was not compatible with RTK. Given this, it is evident that the efforts to migrate Maps to RTK has been broken down into multiple issues.

## Motivation:

Because PR #1314 contains the bulk of the effort for these migrations, its current merge conflicts appear to be directly halting the progress of these dependent issues (and the other issues created for map migration). We want to be mindful not to overlap with or duplicate any existing open features for the map migration, so we wanted to check the status of this PR.

We noticed your comment mentioning that you had recently worked on resolving the conflicts with the development branch, but that there is still a list of remaining items needed to finish up the PR.

## Scope & Execution:

If no one else is currently working on the remaining work, we would like to offer our help. We can take on the remaining items from the list to help finally get the PR’s changes merged in.

We anticipate that once PR #1314 is merged, the Maps page will basically have an “early” version of the RTK migration which the other pages have long been updated and built upon (such as adding SimpleUnsavedWarningComponent.tsx, updated message logging, updated success/error messages, etc.). Therefore, once these changes can be merged, then followup issues can be created to help start work towards updating the Maps page to be consistent with other pages.

## Testing & Verification:
As we are aware that testing has been completed for the existing features of OED, we assume that testing the map migration to RTK can be tested using the existing test suite (npm run test and npm run check). Migration to RTK won’t change the features of maps, but only change its dependencies. Once maps is migrated to RTK and further efforts can be made to add missing features that other pages have (such as handling unsaved warnings), then manual testing will be done in the same way that unsaved warnings were handled elsewhere.

Please let us know if you are open to sharing that list of items with us so we can help clear this blockade!

huss:
First, I'm still very behind in getting to PRs that are ready for review and I'm sorry for that. I am trying to find the time to remove the PR backlog. Then my plan is to organize OED work so the project is stable again after so many changes (laong with some other items). This is going to take a while and it included the incorporation of the map work. Since you have asked about maps, I'll make a few comments at this time:

- The issue with PR #1314 is not really the conflicts. Those exist since I have put off the final work on the PR. I have a list of issues it resolves, issues it does not seem to have fixed and items that should be done. If you want to move forward with this then we should probably chat to discuss the status and for me to provide you with more information since it is not yet written up.
- I was planning to propose significant changes to the UI to update maps as a future item (meaning not likely for the next release(s) that are under development but that could be modified). This includes fixing the calibration so it looks like other pages for setting the values. I wanted to also take the time to determine the needed work for OED to change the use of a 300x500 pixel overlay on maps. This forces that aspect ratio on all maps no matter what is actually uploaded. I don't think it is fundamental to how maps are used and relates to how the points clicked for calibration are detected in Plotly. Thus, I suspect it can be changed but this will need to ripple through the map codebase.

I think you were proposing a more focused update to maps. I'm willing to discuss that and anything else you might want to do with maps. The details on which branch and the exact work would need to be worked out. Again, discussing in a (verbal) conversation might be the most useful.

Please let me know any thoughts and I'm happy to pursue this further.

aduques:
@huss I am interested to hearing about the possible work planned for the Maps feature and see if there is anything that can be done to contribute. As Maps have been a feature that I have not been familiar with both implementing-wise and use-wise, I would appreciate being directed to potential work that can help set the groundwork for the big UI changes later down in development.

I am interested in having a conversation discussing the feature and plans for development. While I don't have any immediate suggestions for the Maps UI, I would still be interested in following along with the discussions around it so that I could propose ideas in the future.

If some of the immediate work that could be done is related to #1314, those can also be discussed in the meeting as I am still open to help with that issue if desired. A meeting time can be discussed through Discord.

huss:
> A meeting time can be discussed through Discord.

FYI. I think I messaged you early today on Discord about a meeting time. Let me know if you don't get it.

src/client/app/components/admin/PreferencesComponent.tsx
huss: My mistake. I came to realize that the min stops the user from using the UI arrows on the field to go below that value. It is a small point but I put it back via a small commit.

huss:
> We also agree that the UnsavedWarningComponent behavior appears to be a separate issue from this PR. We'd be happy to work on that as a follow-up issue, and we can create the issue soon.

My scan did not see an issue for this. Do you think it should be created? If so, will you do it or should I?

huss:
I wanted your thoughts on if this PR also finishes up issue #1616? If so, the description can be modified to state this so it will close when this is merged. If not, it might be nice to determine what is left on that issue so it is clear.

huss (approved):
@GoodKimchi & @aduques I'm sorry this took so long to review (see Discord msg on OED developer channel).

Review and testing found this works as desired. I pushed a small commit to fix a very small issue due to a mistake in a comment I made (the only comment unresolved so you could easily see it). You can review if you wish and let me know if there are any thoughts.

Also, I made two comments a little while ago. These relate to how this relates to an open issue and potential follow-on work. I welcome your thoughts. I wanted to get those covered before I merge this PR so they are not lost.

I'm approving this PR. I'll wait to merge so you have a chance to give any thoughts on the above. Congratulations on another accepted contribution to OED.

aduques:
> > We also agree that the UnsavedWarningComponent behavior appears to be a separate issue from this PR. We'd be happy to work on that as a follow-up issue, and we can create the issue soon.
> 
> My scan did not see an issue for this. Do you think it should be created? If so, will you do it or should I?

@huss I can create an issue for this PR. I anticipate that it would require small work since it would just involve updating UnsavedWarningComponent to behave the same on the Admin Settings Page as the normal save feature.

aduques:
> I wanted your thoughts on if this PR also finishes up issue #1616? If so, the description can be modified to state this so it will close when this is merged. If not, it might be nice to determine what is left on that issue so it is clear.

I believe that issue has not been completely addressed. It was created to address the drafted PR #1603 that this issue also addresses. However, our issue fixes the inconsistency between the maximum/minimum bounds between the frontend and backend of PreferencesComponent.tsx. Issue #1616 seems to be wanting to make the vague error messages more detailed while being easy to understand as an end-user.

I believe that our previous discussion (https://github.com/OpenEnergyDashboard/OED/pull/1677#issuecomment-5067218784) actually attempted to address Issue #1616 more. To summarize what we did discuss: I talked about plans about adding more detailed error messages that would display when one of the checks for invalidFuncs would fail. We ended up drafting this idea as it would implement more code to maintain and does not have a practical use case at the moment. I believe that issue #1616 wanted to improve the messages that were displayed because the original messages from the drafted PR #1603 mentioned error messages that the previous contributor thought it'd be good to update alongside the PR comment suggestions.

To summarize, I think issue #1616 contains additional proposed changes that we basically discussed and ultimately drafted. This would be done alongside finishing up the drafted PR #1603 (which we did already). It can be remain open but should be updated saying that the mentioned drafted PR has already been addressed.

huss:
This has sat for a while so it seems fine with everyone. To make it cleaner for PR #1705 I am merging it now.

PR #1678, opened 2026-07-17
Enforce file size limits and user role checks on server

### LINKED ISSUE #1675: enforce CSV download limits on server
### Describe the bug

The client checks the expected file size and user role to check, if it is larger, that the site allows this to happen. This happens during raw (meter level) graphic data download. The server route is not enforcing this limit.

### Expected behavior

The server should enforce the limit as done in the client to avoid mistakes and bad requests.

### Additional context

This issue was reported to OED by @Santoshkumarpuppala via private channels and a CVE is planned.

Nespina24:
# Description

The export graph meter data route in the client (`src/client/app/redux/thunks/exportThunk.ts`) enforces a file size limit that prevents downloads of meter data if the estimated file size is greater than the set limit and the user’s role is not EXPORT or ADMIN. This change adds the same file size and user role checks to the server route (`src/server/routes/readings.js`). Added unit tests to `src/server/test/routes/readingsParamsTest.js`.

Fixes #1675 

Authors: @Nespina24 and @stockingstocker 

@huss added: This also fixed an issue with the parameter checking of the meter_id in this route and used the standard return functions.

## Type of change

## Checklist

## Limitations

No limitations.


src/client/app/components/ExportComponent.tsx
```diff
@@ -40,4 +40,4 @@ export default function ExportComponent() {
 			}
 		</>
 	);
-}
+}
```
huss: Is it possible the removed blank line at the end of the file could be added back as that is the only change in this file?

src/client/app/redux/thunks/exportThunk.ts
```diff
@@ -136,14 +136,14 @@ export const exportRawReadings = createAppThunk(
...
 		// Estimated file size in MB. Note that changing the language effects the size about +/- 8%.
 		// This is just a decent estimate for larger files.
+		// This estimate is also present in src/server/routes/readings.js and must be kept consistent between files
 		const fileSize = (count * 0.082 / 1000);
 		// Decides if the readings should be exported, true if should.
 		let shouldDownload = false;
 		if (fileSize <= adminState.defaultWarningFileSize) {
 			// File sizes that anyone can download without prompting so fine
 			shouldDownload = true;
 		} else if (fileSize > adminState.defaultFileSizeLimit) {
```
huss, line 146: Could you please add back in the next blank line and the one below that was removed.

src/server/routes/readings.js
```diff
@@ -96,24 +96,45 @@ router.get('/line/raw/meter/:meter_id', optionalAuthMiddleware, async (req, res)
 		}
 	};
 	if (!validate(req.params, validParams).valid || !validate(req.query, validQueries).valid || !isValidTimeInterval(req.query.timeInterval, true)) {
-		res.sendStatus(HTTP_CODES.BAD_REQUEST);
+		failure(res, HTTP_CODES.BAD_REQUEST);
+	} else if (req.params.meter_id == '2147483648' || req.params.meter_id == '0') {
```
huss, line 102: I appreciate you adding this test. Having staid that, I'm not sure checking the one large value and zero is too valuable. There are many values that would fail (any id that does not exist). Each of these will still generate a DB error that will cause an error for the request. Finally, these would go away when OED sends them as part of the req.query so it is an actual integer sent. Given all of this, I would remove remove this new check.

src/server/routes/readings.js
```diff
@@ -96,24 +96,45 @@ router.get('/line/raw/meter/:meter_id', optionalAuthMiddleware, async (req, res)
...
+		failure(res, HTTP_CODES.BAD_REQUEST);
 	} else {
-		let meterID;
-		let timeInterval;
+		const conn = getConnection();
+		// Get the routed meter id and time for the desired readings.
+		const meterID = req.params.meter_id;
+		const timeInterval = TimeInterval.fromString(req.query.timeInterval);
 		try {
```
huss, line 107: This is a very small point but I suggest moving the try above the getConnection() (as it was before) in the unlikely chance that any of the early lines cause an error so it is caught in the usual way.

src/server/routes/readings.js
```diff
@@ -96,24 +96,45 @@ router.get('/line/raw/meter/:meter_id', optionalAuthMiddleware, async (req, res)
...
-			const conn = getConnection();
-			// Get the routed meter id and time for the desired readings.
-			meterID = req.params.meter_id;
-			timeInterval = TimeInterval.fromString(req.query.timeInterval);
+			//check if user is allowed to export
+			let shouldDownload = false;
+			//estimate file size
+			//this estimate is also present in src/client/app/redux/thunks/exportThunk.ts and must be kept consistent between files
+			const count = await Reading.getCountByMeterIDAndDateRange(meterID, timeInterval.startTimestamp, timeInterval.endTimestamp, conn);
```
huss, line 117: There is a subtle point here that I think should be documented via a comment above this line. In the server testing, it sends all the meter ids and the totals up the size to enforce the limit. Here, it is checking a single meter. This means this can be a looser check if there are multiple meters. I think that is okay and I have not thought of a good alternative for direct API requests to this route but it would be good to note that they are slightly different.

src/server/routes/readings.js
```diff
@@ -96,24 +96,45 @@ router.get('/line/raw/meter/:meter_id', optionalAuthMiddleware, async (req, res)
...
+		const timeInterval = TimeInterval.fromString(req.query.timeInterval);
 		try {
-			const conn = getConnection();
-			// Get the routed meter id and time for the desired readings.
-			meterID = req.params.meter_id;
-			timeInterval = TimeInterval.fromString(req.query.timeInterval);
+			//check if user is allowed to export
+			let shouldDownload = false;
+			//estimate file size
+			//this estimate is also present in src/client/app/redux/thunks/exportThunk.ts and must be kept consistent between files
```
huss: Some comments, esp. the one about the estimate of size are on the client but not here. I think that is okay but I might add to this comment that people should look there for the full explanation of the estimate used.

src/server/routes/readings.js
```diff
@@ -96,24 +96,45 @@ router.get('/line/raw/meter/:meter_id', optionalAuthMiddleware, async (req, res)
...
+				const token = req.headers.token || req.body.token || req.query.token;
+				if (await isTokenAuthorized(token, User.role.EXPORT)) {
+					shouldDownload = true;
+				}
+			}
+			if (shouldDownload == false) {
+				failure(res, HTTP_CODES.FORBIDDEN);
+				return;
+			}
 			// Get the raw readings for this meter over time range desired.
```
huss: I understand that an else is not necessary. Having said that, I think that logic where you don't have another return before the end (by studies) is less prone to mistakes/errors/. Thus, I would remove the return, add an else and put the next two lines inside that else clause unless you feel/find otherwise.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,185 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+			expect(res).to.have.status(HTTP_CODES.OK);
+			expect(res.body).to.have.property('token');
+			return res.body.token;
+		}
+		//used to make users for roles CSV, EXPORT, OBVIUS
+		async function createUserWithRole(role, conn) {
+			const user = new User(
+				undefined,
+				`raw-export-${role}-${Date.now()}@example.invalid`,
+				bcrypt.hashSync('password', 10),
```
huss: It is a little safer to make the password depend on the user to avoid accidental use of another user. (Also see about logging out below that also helps.)

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,185 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+				role
+			);
+			user.password = 'password';
```
huss: Am I reading this correctly that you set the password via the new user above and then change it here before usage? If so, is there a reason?

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,185 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+			meterID = testData.meterID;
+			timeInterval = testData.timeInterval;
+			await Preferences.update({ defaultFileSizeLimit: limit }, conn);
+		}
+		mocha.describe('Estimated File Size is within File Size Limit', () => {
+			let csvToken;
+			let exportToken;
+			let obviusToken;
+			let adminToken;
+			mocha.beforeEach(async () => {
```
huss, line 327: I might suggest you look at the places in the test code that loop over roles to test access. For example, see src/server/test/web/maps.js but others might be closer to what you want. I think using a variable with needed values and a loop would cut down on the very similar code.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,185 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+				await setUpRawExportTest(100000000); // Set a high file size limit to ensure the estimated file size is within the limit;
```
huss: First, OED wants comments above the line of code and not at the end of the line. Second, is there a reason not to set the limit close to the estimated file size (the formula would tell you the value) so it just passes?

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,185 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+			await Reading.insertAll([
+				new Reading(meter.id, 10, start.clone(), start.clone().add(1, 'hour')),
+				new Reading(meter.id, 20, start.clone().add(1, 'hour'), start.clone().add(2, 'hours'))
+			], conn);
+			return {
+				meterID: meter.id,
+				timeInterval: '2020-01-01T00:00:00Z_2020-01-01T02:00:00Z'
+			};
+		}
+		async function getTokenForUser(user) {
```
huss, line 240: Overall, you have ok comments. It is always nicer to use JSDoc for a function though not critical here since it is not used anywhere else. However, maybe it will be give the generality.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,185 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+			await Preferences.update({ defaultFileSizeLimit: limit }, conn);
+		}
+		mocha.describe('Estimated File Size is within File Size Limit', () => {
+			let csvToken;
+			let exportToken;
+			let obviusToken;
+			let adminToken;
+			mocha.beforeEach(async () => {
+				await setUpRawExportTest(100000000); // Set a high file size limit to ensure the estimated file size is within the limit;
+				csvToken = await getTokenForRole(User.role.CSV, conn);
```
huss: It is safer to login as each user, to the test for that user and then logout. This might be easy if you were looping within the test.

huss (changes requested):
@Nespina24 & @stockingstocker Testing found this works as desired. I have made a few comments to consider. In addition, I am noting these items to see if you have thoughts or want to address (if not then an issue will be created):

- If the readings route you worked on returns an error then the client hangs up for a while waiting and does not give any message. I did not check but it may have existed before. I saw it since I modified your file size check to force it to fail.
- src/client/app/components/admin/PreferencesComponent.tsx has an issue that is stopping the admin from saving the default file size limits for export. It is because it sends a string and not a number when updated. The changes to route param checks now detects this. On line 42 I added the Number() as shown here:
```js
setLocalAdminPref({ ...localAdminPref, [key]: Number(value) });
```
This seems to have fixed it. There are also other Number casts that can probably go away given this change but I did not check that.

Please let me know if anything is not clear or you have thought/questions.

src/server/routes/readings.js
stockingstocker: > I appreciate you adding this test. Having staid that, I'm not sure checking the one large value and zero is too valuable. There are many values that would fail (any id that does not exist). Each of these will still generate a DB error that will cause an error for the request. Finally, these would go away when OED sends them as part of the req.query so it is an actual integer sent. Given all of this, I would remove remove this new check.

This check was added because we noticed a failing test in Readings Route Parameter Validation, specifically the test for invalid meter_id patterns which include '0' and INT32_MAX + 1. We didn't feel comfortable modifying a test that we didn't write so we added this check as a bandaid fix. Would it be preferable to remove '0' and INT32_MAX + 1 from invalidValues[] instead?

huss: > This check was added because we noticed a failing test in Readings Route Parameter Validation, specifically the test for invalid meter_id patterns which include '0' and INT32_MAX + 1. We didn't feel comfortable modifying a test that we didn't write so we added this check as a bandaid fix. Would it be preferable to remove '0' and INT32_MAX + 1 from invalidValues[] instead?

Sigh....Given where the code is at, the path of least resistance is to leave your check in the route. Please put a TODO there that once the meter_id is an integer instead of a string then this should be removed. That way it is in the file where the change will happen so the next developer should see it when making this change. Does that make sense to you?

src/server/test/routes/readingsParamsTest.js
stockingstocker: That is correct. Since createUserWithRole directly inserts a new User into the database, we needed to pass the hashed password for the stored DB value. We are also using the login API afterward, which expects a plain-text password. So after inserting the hashed password, we update the local User object to hold the plain-text password for the login request.

stockingstocker:
>     * If the readings route you worked on returns an error then the client hangs up for a while waiting and does not give any message. I did not check but it may have existed before. I saw it since I modified your file size check to force it to fail.

I don't think I've been able to reproduce this. Could you elaborate on how you modified the file size check to force it to fail? 

I did think that maybe its a scenario where the client accepts the request but the server rejects which causes lineReadings = await response.unwrap() to throw an error so I added a try-catch with corresponding error notification around it, does that fix the issue? I didn't manage to find a good error message in `src/client/app/translations/data.ts` so I just wrote a unique one for now.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,169 @@ mocha.describe('Readings Route Parameter Validation', () => {
 		});
 	});
 
+
+	mocha.describe('Raw Export File Size and User Role Authorization', () => {
+		let conn;
+		let meterID;
```
huss: It took me a little while to figure out how timeInterval was set. Given this, I think it would be nice to put comments here about these two quasi-global variables (conn, too, if you want). For timeInterval, it is set in a returned structure from createRawExportTestData with the same name but not the variable and then that structure value is used in setUpRawExportTest to set the actual variable. I hope I have this right. meterID is more straightforward but worth noting.

src/server/test/routes/readingsParamsTest.js
huss: I'm not seeing this done. Could it be as simple as a variable that sets the desired password to ${role}password and then used in the bcrypt and to set user.password?

src/server/test/routes/readingsParamsTest.js
```diff
@@ -184,6 +193,169 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+
+		/**
+		 * Logs in as given user and returns the generated auth token.
+		 * The User object is expected to contain a plain-text password to use for login.
+		 * @param user The user to log in as. 
+		 * @returns The auth token generated by the login route.
+		 */
+		async function getTokenForUser(user) {
+			const res = await chai.request(app)
+				.post('/api/login')
```
huss: PR #1672 is changing the name of this route to loginLogout and adding the subroute /login for this. I'm not sure which will be merged first but that will need to be fixed in one of the PRs.

src/server/test/routes/readingsParamsTest.js
huss: This was marked resolved but I'm unclear on how it was resolved. Could you please help me?

huss:
I'm just noting I pushed a commit to clarify two comments since that was easy. Let me know if you have any thoughts.

huss:
>     src/client/app/components/admin/PreferencesComponent.tsx has an issue that is stopping the admin from saving the default file size limits for export. It is because it sends a string and not a number when updated. The changes to route param checks now detects this. On line 42 I added the Number() as shown here:

> setLocalAdminPref({ ...localAdminPref, [key]: Number(value) });

> This seems to have fixed it. There are also other Number casts that can probably go away given this change but I did not check that.

I wanted to know if you want to do this as part of this PR or if it should be an issue (by you or me) for someone else. Okay either way.

huss:
> don't think I've been able to reproduce this. Could you elaborate on how you modified the file size check to force it to fail?

I went into the raw reading route (currently line 118) and changed it to ``const fileSize = 1000000;`` so it rejected the file size on check. This causes an error if you ask for a raw/meter readings export from the line graphic page on the UI.

> I did think that maybe its a scenario where the client accepts the request but the server rejects which causes lineReadings = await response.unwrap() to throw an error so I added a try-catch with corresponding error notification around it, does that fix the issue? I didn't manage to find a good error message in `src/client/app/translations/data.ts` so I just wrote a unique one for now.

Yes, it now works by putting up an error msg rather than taking a while and showing nothing.

huss (changes requested):
@Nespina24 and @stockingstocker Thank you for the updates. Overall, they worked well. I have made a little over a half dozen comments to consider where about half are new and half are about previous items.  I think all resolved items are done. To try to help me keep items needing my attention as future changes are made, I'm asking you not resolve any comments. You are welcome (encouraged) to put a thumbs up if you think you addressed it or put in a reply. Thanks. Please let me know if anything is not clear or you have thoughts/questions.

src/server/test/routes/readingsParamsTest.js
stockingstocker: Your interpretation is correct. I tried changing the variable names a little and left comments at the top to clarify the intent. Please let me know if that helps.

src/server/test/routes/readingsParamsTest.js
Nespina24: Sorry for resolving this prematurely. I added in the logout to `expectRawExportStatus()` since the loop is calling this function every time. That said, I'm not completely sure if the format is correct since I didn't see any other instance of logging out for testing in the codebase.

stockingstocker:
>     * src/client/app/components/admin/PreferencesComponent.tsx has an issue that is stopping the admin from saving the default file size limits for export. It is because it sends a string and not a number when updated. The changes to route param checks now detects this. On line 42 I added the Number() as shown here:
> 
> 
> ```js
> setLocalAdminPref({ ...localAdminPref, [key]: Number(value) });
> ```
> 
> This seems to have fixed it. There are also other Number casts that can probably go away given this change but I did not check that.


With the Number cast on line 42, I was unable to change the default minimum and maximum reading date checks on the admin settings page. I believe this is because `makeLocalChanges() `is also used for non-numerical fields so casting value to Number breaks those fields. I opted to add Number casts for the fields `defaultWarningFileSize` and `defaultFileSizeLimit` instead and I was able to successfully save those preferences.

huss:
> I opted to add Number casts for the fields `defaultWarningFileSize` and `defaultFileSizeLimit` instead and I was able to successfully save those preferences.

Thanks for figuring that out. I looked at the code and tried it and it works well for me too.

src/server/test/routes/readingsParamsTest.js
huss: Yes, OED is trying to get better and better at its testing. In some cases the logout is not needed since the token cannot be used across tests if it is not stored somehow. But this is a start on this aspect of the changes. What you did seems fine.

src/server/test/routes/readingsParamsTest.js
huss: Your PR wound up right behind PR #1672 that merged very recently. Given this, I merged in development to add that PRs work and updated your code. It worked fine on my machine and the GitHub tests will do the final confirmation.

huss:
Noting that all comments before now are resolved and the code seems to work fine. Awaiting GitHub CI for final validation.

huss (approved):
@Nespina24 and @stockingstocker Thank you for the updates. Review and testing found everything is good. Congratulations on an accepted contribution to OED.

PR #1685, opened 2026-07-29
Fix disabled group delete button after parent group removal (Issue 1653)

### LINKED ISSUE #1653: dashboard failure if delete a group after it failed the first time
### Describe the bug

If you try to delete a group that is contained in another group then OED will not allow it. If you delete the other group and then retry the original delete the button does not work. If you reload the OED web page then it work to delete the original group.

### To Reproduce

1. Create a group without any child groups but with a child meter.
2. Create another group with the first group as a child group.
3. Delete the first group. There will be a failure popup stating it is used in the second group.
4. Go to the second group and delete it.
5. Go back to the first group. The delete button is present but you cannot click it. 
6. Refresh the web page and redo the previous step and it works fine.

### Expected behavior

This should work without the need to refresh.

### Additional context

The reason for this bug is unknown. However, it may be that the Redux state is not properly reset when the failure occurs. The edit page should be doing this when the cancel button is pressed so that may help show what is needed if this is the issue.

kujoon226:
# Description

This PR fixes the issue where a group cannot be deleted after its parent group has been deleted unles the page is refreshed.
This fix uses forceRefetch to request the latest group data from the server instead of using the data already stored in Redux. After the parent group is deleted, the new response reflects that the original group is no longer being used as a child group. Redux then updates the page with the latest data, which re-enables the delete button and allows the group to be deleted without refreshing the browser.

Fixes #1653 

Contributors:

- Kujoon Kwon (@kujoon226  )
- Neil Cabanilla (@neilcabanilla )

## Type of change

## Checklist

## Limitations

N/A

### DIFF
--- a/src/client/app/components/groups/EditGroupModalComponent.tsx
+++ b/src/client/app/components/groups/EditGroupModalComponent.tsx
@@ -992,7 +992,12 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 	 */
 	async function validateDelete() {
 		// Get all parent groups of this group.
-		const { data: parentGroupIDs = [] } = await store.dispatch(groupsApi.endpoints.getParentIDs.initiate(groupState.id, { subscribe: false }));
+		const { data: parentGroupIDs = [] } = await store.dispatch(
+			groupsApi.endpoints.getParentIDs.initiate(groupState.id, {
+				subscribe: false,
+				forceRefetch: true
+			})
+		);
 
 		// If there are parents then you cannot delete this group. Notify admin.
 		if (parentGroupIDs.length !== 0) {
--- a/src/client/app/components/groups/EditGroupModalComponent.tsx
+++ b/src/client/app/components/groups/EditGroupModalComponent.tsx
@@ -991,7 +991,14 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 	 * If not, then continue delete process.
 	 */
 	async function validateDelete() {
-		// Get all parent groups of this group.
+		/**
+		 * Gets all parent groups of this group since it is easy and this is not done often.
+		 * In principle, the Redux state for groups has this information. However, due to the
+		 * recursive nature of groups, it is a harder to get it from that state so the
+		 * code simply uses the existing server/DB functions to do this.
+		 * Redux does not always think the state has changed so it may use stale state, esp. after a delete.
+		 * Thus, the fetch is forced to avoid this.
+		 */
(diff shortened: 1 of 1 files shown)


src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -992,7 +992,12 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
 	 */
 	async function validateDelete() {
 		// Get all parent groups of this group.
```
huss: I thought about why this happens given your solution. I know you did not put in this comment (looks like I did) so I can do this change or you can. I think clarifying this comment would be good. I propose to change it to:

```js
// Get all parent groups of this group since it is easy and this is not done often.
// In principle, the Redux state for groups has this information. However, due to the
// recursive nature of groups, it is a harder to get it from that state so the
// code simply uses the existing server/DB functions to do this.
```

src/client/app/components/groups/EditGroupModalComponent.tsx
```diff
@@ -992,7 +992,12 @@ export default function EditGroupModalComponent(props: EditGroupModalComponentPr
...
-		const { data: parentGroupIDs = [] } = await store.dispatch(groupsApi.endpoints.getParentIDs.initiate(groupState.id, { subscribe: false }));
+		const { data: parentGroupIDs = [] } = await store.dispatch(
+			groupsApi.endpoints.getParentIDs.initiate(groupState.id, {
+				subscribe: false,
+				forceRefetch: true
```
huss, line 1005: I think a comment on why this line is here would be nice. Maybe something like;

```js
// Redux does not always think the state has changed so it may use stale state, esp. after a delete.
// Thus, the fetch is forced to avoid this.
```

huss (approved):
@kujoon226 & @neilcabanilla Thank you for another contribution. Review and testing found it works fine. I did make two small comments about potential comments for your consideration. I think this is almost ready to merge.

neilcabanilla:
Hello @huss,

We've implemented the comment for the code that we made. Please let us know if there are any other things we need to do or add.

Thank you so much!

huss:
I hope to review this in about a week. Sorry for the delay.

huss (approved):
@kujoon226 & @neilcabanilla First, I'm sorry this took so long to review (see OED Discord server developer msg).

Review and testing found that this addressed all the comment and it work fine. Congratulations on another accepted contribution to OED.

PR #1686, opened 2026-07-29
Replace duplicated raw export size calculations with a shared estimator

### LINKED ISSUE #1684: Share raw export file size estimator between client and server

While working on issue #1675, we added file size estimation to the raw export server route using the same formula and values already used by the client in `exportThunk.ts`. This means the client and server now contain separate copies of the same estimation logic. If the formula or its constants are changed in the future, one copy could be updated without the other, causing the client and server to calculate different estimated file sizes.

Move the raw export file size estimation logic into a shared module. The shared module would contain the estimator function and its related constants. Both the client and server would import and use this function instead of calculating the estimate independently. This would  ensure that future changes are automatically applied to both the client and server. The comments explaining how the estimate was derived could also be moved into the shared module.

Leave the current implementation unchanged and require any future changes to the file size estimation formula or constants to be made separately in both the client and server.

### Additional context

A unit test should be added to verify that the shared estimator returns the expected result for known reading counts.

stockingstocker:
# Description

This PR moves the raw export file size estimation logic into a shared module used by both the client and server. The client uses the estimate to warn users and determine whether an export should proceed, while the server uses the same estimate to enforce the configured file size limit.

Previously, `exportThunk.ts` and `readings.js` each contained their own copy of the file size estimation formula:

`count * 0.082 / 1000`

Keeping separate copies creates a risk that the client and server could use different values or formulas if one is updated without the other.

Now, the client and server import `estimateRawExportSizeMB()` instead of calculating the estimate independently. Unit tests have also been added under`src/server/test/util` to verify the estimator returns the expected values for:

- zero readings
- a known reading count with an expected estimated size
- a non-round reading count to verify the calculation does not depend on values that simplify cleanly

Fixes #1684 

## Type of change

## Checklist

## Limitations

No limitations.


src/server/test/util/rawExportFileSizeTests.js
```diff
@@ -0,0 +1,30 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const chai = require('chai');
```
huss: While this works, many files get the chai, mocha & expect from common file to keep them consistent across files. I think that would be better. Having said that, OED probably should open an issue to do that across all tests to be consistent, assuming it causes no problem. Would you like to do that or should I? This should also get rid of the blank line between items as OED prefers.

src/server/test/util/rawExportFileSizeTests.js
```diff
@@ -0,0 +1,30 @@
...
+
+const expect = chai.expect;
+const mocha = require('mocha');
+
+const { estimateRawExportSizeMB } = require('../../../common/RawExportFileSize');
+
+mocha.describe('Raw Export File Size Estimator', () => {
+    mocha.it('returns zero for zero readings', () => {
+        expect(estimateRawExportSizeMB(0)).to.be.closeTo(0, 0.001);
```
huss: Is there a specific reason you use 0.001 tolerance on these tests? In readings (see src/server/util/readingsUtils.js) it has a DELTA that is stricter. Stricter is better when it works. I would have thought this would be basic quasi-random numerical variation in the final digit given what is being tested. So, I tried 1e-15 and that worked fine. Maybe that should be used and set to a const in the file..

src/common/RawExportFileSize.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const ESTIMATED_KB_PER_RAW_READING = 0.082;
+const KB_PER_MB = 1000;
+/**
```
huss, line 9: Could you please add a blank line before the JSDco.

src/common/RawExportFileSize.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+const ESTIMATED_KB_PER_RAW_READING = 0.082;
```
huss: I have two thoughts. First, given the estimate is in MB, would it be easier to make this const be in MB and avoid the KB_PER_MB conversion? I might make it 8.2e-5 so it is easier to understand. Second, I'm wondering why this is exported. I could not find uses outside this file except in the d.ts file but then that usage does not seem to be used anywhere else. I'm uncertain this const would ever be used anywhere else in OED and encapsulating it to the function may make sense. Thus, I was wondering about removing the export, placing the const inside the function and removing it from the t.ds file. It worked fine for me.

huss (changes requested):
@stockingstocker Thank you for another contribution. Review and testing found it works well. I have made a few comments to consider. Please let me know if anything is not clear or you have thoughts/questions.

src/common/RawExportFileSize.js
stockingstocker: For the first point, I was keeping the structure of the formula the same as it originally was in the client but you're right that changing the estimate to be in MB makes a lot more sense now that its in a function. For the second point I was being safe just in case the values may be needed elsewhere but I realize now that encapsulating it for just the function is more reasonable.

src/server/test/util/rawExportFileSizeTests.js
stockingstocker: I didn't notice the delta value in readingsUtils.js. I now see that I gave the tests too much wiggle room. I will make the change

src/server/test/util/rawExportFileSizeTests.js
stockingstocker: Thanks for catching that. I was working from habit and forgot about OED’s common test file.. For the issue, I would be happy to open it after I take a closer look at the existing test files.

huss:
I saw the update. I'm sorry that it is going to take a little while for me to get to this. Let me know if the delay becomes an issue for you.

huss (approved):
@stockingstocker I'm sorry this review took so long (please see Discord msg on the OED server in the developer channel). Review and testing found it works as desired so the changes addressed all comments. I did push a small change so the admin preferences page indicates that file size limits are in MB. This is not from your work but I thought of it due to these changes. It was very straightforward so I did it via a commit. Congratulations on another accepted contribution to OED.

Issue #1676, opened 2026-07-16
Upgrade Express from v4 to v5


CarlyAThomas:
### Is your feature request related to a problem? Please describe.

OED is currently on Express ~4.19.2. While working on #1614 (standardizing server error responses), it became clear that Express 4 has a real gap: it doesn't forward rejected promises from async route handlers to the global error handler on its own, so an unhandled error in an async route can leave a request hanging indefinitely instead of returning a clean error response. Express 5 fixes this natively. Separately, OED needs to move to v5 for the next release along with other package updates, so this issue is to track that migration specifically, since it's a larger, breaking change that shouldn't be bundled into #1614.

### Describe the solution you'd like

Upgrade the express dependency to v5, following the [official migration guide](https://expressjs.com/en/guide/migrating-5/). From an initial pass over this codebase, a few concrete breaking changes will need addressing:

- app.js**'s SPA catch-all route** (router.get('*', ...)) — Express 5's updated path-to-regexp no longer accepts a bare * wildcard; this needs to be rewritten to the new named-wildcard syntax (e.g. /*splat) or the server won't start.
- req.query **becomes read-only** — src/server/middleware/paramsProcessing.js's lowercaseAllParamNames middleware directly reassigns req.query = newQuery, which will throw under Express 5. This middleware is only wired into obvius.js, so the blast radius is scoped to that route, but it needs a rework (e.g. mutating the existing query object's keys instead of reassigning it).
- req.param() **removal** — turns out not to be an issue: obvius.js calls req.param() extensively, but OED already defines its own req.param mixin in paramsProcessing.js that shadows whatever Express provides, so this isn't dependent on Express's own (now-removed) method.
- **Unverified dependency compatibility** — multer, express-rate-limit, body-parser, and morgan versions currently pinned in package.json haven't been checked against Express 5 yet.
- Remove the express-async-errors dependency — added in #1614 as a temporary Express 4 patch (forwards rejected async route handler promises to the global error handler, which Express 5 does natively). Once this upgrade lands, delete the require('express-async-errors') line in app.js (already marked with a TODO referencing this issue) and run npm uninstall express-async-errors — nothing else in the codebase depends on it directly.

### Describe alternatives you've considered

[...]

Issue #1666, opened 2026-07-10
Have save error messages show up on the client for ID values on all routes


aduques:
### Is your feature request related to a problem? Please describe.

It was discovered recently that the save error message was not properly being shown on the client for src/client/app/components/conversion/EditConversionModalComponent.tsx.

The maintainer mentioned that line 423 of that file had to be changed to 

      error.data

in order to properly mirror the save error handling.

The maintainer found that error handling was different in multiple files and wanted to confirm if the implementations on those files are handling and displaying the error properly:

src/client/app/components/admin/users/CreateUserModalComponent.tsx
src/client/app/components/admin/users/EditUserModalComponent.tsx
src/client/app/components/groups/EditGroupModalComponent.tsx

The maintainer has provided the following note: 
"There is a chance it is done somewhere else but I tried to catch them all by searching for variants (error.data.message, err.data., data.message)"

### Describe the solution you'd like

All routes should consistently display the save error messaging and have a consistent implementation. All routes need to be verified that the save error is properly being displayed.

### Describe alternatives you've considered

No other alternatives. It is possible that the error is not being handled properly on all routes, so this issue requests that all the routes be checked to have the correct implementation.

### Additional context

The issue was originally brought up in this issue: https://github.com/OpenEnergyDashboard/OED/pull/1663

The comment that directly references the issue and a possible fix:
https://github.com/OpenEnergyDashboard/OED/pull/1663#discussion_r3537604440

aduques:
I plan to work on this issue after completing my other open issues.

aduques:
## EditConversionModalComponent.tsx

To reproduce the bug, I added the following in conversions.js:

      router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, res) => {
         // TODO DEBUG: to force the showErrorNotification to pass in EditConversionsModalComponent.tsx
         req.body.meterIds.push(0);

<img width="560" height="348" alt="Image" src="https://github.com/user-attachments/assets/f1833b80-9fb6-4126-af9a-b0360cd97c65" />


This displays the problematic error message. The maintainer prefers when the error message for error notifications uses error.data instead of error.data.message.


                     showErrorNotification(
                         translate('conversion.delete.failure') +
                         ' (' + translate('conversion.source') + ' "' + unitDataById[payload.sourceId]?.identifier + '"' +
                         ', ' + translate('conversion.destination') + ' "' + unitDataById[payload.destinationId]?.identifier + '") ' +
[...]

PR #1591, opened 2026-03-30
Update DOMPurify Sanitizer

Zach-O-Bates:
# Description
Merges changes from PR #1488 and  PR #1544. This PR fixes the issue where the update to DOMPurify and jsdom was not reflected in the package-lock.json file. It also addresses the modification issues mentioned in the previous PR. Now that DOMPurify is properly implemented the cross-site scripting vulnerability when users upload meter data should be remediated.

Developed and Implemented by:
Zachary Squires - https://github.com/Zachary-Squires
Zachary Bates - https://github.com/Zach-O-Bates
Krista Clanton - https://github.com/Quillen09

## Type of change
- [ ] Note merging this changes the database configuration.
- [ ] This change requires a documentation update

## Checklist
- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations
N/A

PR #1528, opened 2025-08-19
Implement parameter validation testing across all server routes

omarraf:
# Description

This PR implements comprehensive parameter validation testing across all server routes as requested in #1497. The testing framework systematically validates that each route properly handles invalid parameters, preventing malformed requests from reaching application logic.

### 1.) Reusable Testing Framework

  Created src/server/test/util/validationHelpers.js with helper functions for:
  - String validation (length constraints, enum values)
  - Integer/boolean validation (type checking, bounds)
  - Token security testing (DoS prevention)
  - Parameter injection prevention (extra field rejection)
  - Malicious input testing (SQL injection, XSS, path traversal)

  ### 2.) Centralized Validation Constants

  Added src/server/util/validationConstants.js to eliminate hardcoded validation limits and ensure
  consistency across routes

  ### 3.) Comprehensive Test Coverage

  Added 20+ parameter validation test files covering:
  - Core Resources: units, meters, groups, maps
  - Data & Readings: readings, compare, baseline, CSV import/export
  - Conversions: unit conversions and bulk operations
  - Auth: login, 2FA, user management, verification
  - System: preferences, logs, version, CIKS, Obvius integration

## Related
This work builds upon  #1512

Fixes #1497 

## Type of change

- [ ] Note merging this changes the database configuration.
- [x] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

PR #1688, opened 2026-07-31
Standardize server responses to use success/failure functions

### LINKED ISSUE #1614: standardize server responses to use success/failure functions

In src/server/routes/response.js, OED has standard success/failure functions to route back the desired response. PR #1604 enhanced these functions and used this in some places. During review, it was noticed that how it now works and where it is used could be enhanced to better encapsulate the ideas and make sure all places use these as they have better security/practices.

- The logging on failure should happen here rather than in each location where the error occurs. Then the message will not go back to the user but will be recorded in the DB. This means passing the sanitized message to this function and the error to log to the DB.
- Add a parameter to disable logging with the default of false. That means logs are created unless the developer feels there is a good reason not to do that.
- All routes should be checked so logging info is sent on for failures and not done in the route or any other uses.
- All routes should be changed to use the modified functions and stop sending back responses directly. This would centralize this process and make it safer/easier to change in the future. Care should be taken not to send back internal error information but only have it logged per OED's new security desires not to expose internal details of the application to users except the logs available to admins.

In addition, src/server/services/csvPipeline/failure.js & src/server/services/csvPipeline/success.js have separate versions. There is also another specialized one in src/server/routes/obvius.js that should be checked out. These send back different strings but it should be investigated if it could be merged with the version above. If not, they should be reviewed to make them as consistent as possible and comments in all locations indicating where the others are. 

Leave it as is with its limitations and reduced security.

### Additional context

If anyone has other ideas then they are welcome. If anyone working on this has questions about some uses or what to do then please reach out.

CarlyAThomas:
# Description

This standardizes how every server route responds, moving all 21 route files (63 endpoints) onto `response.js`'s `success()`/`failure()` functions instead of the mix of hand-rolled `res.*` calls that had grown up across the codebase. The issue laid out four specific things to address, plus a follow-up question about the CSV pipeline and `obvius.js`, so I've gone through each of those in order below.

**1. "The logging on failure should happen here rather than in each location where the error occurs... This means passing the sanitized message to this function and the error to log to the DB."**

`failure()` now takes the raw error as its own parameter, separate from an optional `safeMessage` meant for the client. The raw error is only ever logged internally; it is never sent to the client under any circumstance, so a route can no longer accidentally leak a DB error or a stack trace just by passing the wrong thing into a response call. Routes no longer call `log.warn()`/`log.error()`/etc. themselves — that all happens inside `failure()` now.

**2. "Add a parameter to disable logging with the default of false. That means logs are created unless the developer feels there is a good reason not to do that."**

I went slightly beyond the literal ask here, with @huss's approval on the issue. Instead of a plain on/off boolean, `failure()` takes a `LogLevel` severity (`DEBUG`/`INFO`/`WARN`/`ERROR`/`SILENT`), so a route can keep its original log priority instead of every failure collapsing to one level. This turned out to matter more than I expected once I started migrating files: `log.js`'s `emailLevel` only e-mails admins for `ERROR`-priority events, so if every failure had collapsed to a single severity, routine `WARN`-level validation failures would have started paging admins for things that were never meant to reach them. Logging is still on by default (`LogLevel.ERROR`); a route has to explicitly opt out with `LogLevel.SILENT` if it has good reason not to log something, which satisfies the "logs are created unless the developer feels there is a good reason not to" part of the ask.

**3. "All routes should be checked so logging info is sent on for failures and not done in the route or any other uses."**

I went through all 63 endpoints individually, by hand, rather than trusting a mechanical find-and-replace pass. That review is what turned up most of the real bugs described below — hang bugs where a route logged but never actually responded, catch blocks that responded but bypassed the shared functions entirely, and about 40 call sites where an earlier, more mechanical migration pass had silently dropped the descriptive part of the original log message. All of that is fixed and restored now, endpoint by endpoint.

**4. "All routes should be changed to use the modified functions and stop sending back responses directly... Care should be taken not to send back internal error information but only have it logged."**

Every one of the 21 route files (63 endpoints) has been migrated onto `success()`/`failure()`, with two deliberate, discussed exceptions explained under Limitations below. The client-facing sanitization is structural, not something each route has to remember to do correctly: `failure()` always substitutes a generic message for any `500`-level response regardless of what's passed in, so internal detail can only ever reach the admin-visible logs, never the client.

**On the CSV pipeline and `obvius.js`:** both are addressed under Limitations below, since the outcome for each ended up being "leave it alone, for a specific reason" rather than a straightforward migration.

Fixes #1614

## Real bugs found and fixed along the way

Going through every endpoint by hand (not just grepping for patterns) turned up a handful of real, pre-existing bugs:

- **Multiple "hang bugs"**: catch blocks that only logged and never sent any response at all, leaving the client's request pending forever. Found in `groups.js` (5 endpoints: `GET /`, `/idname`, `/deep/groups`, `/children/:group_id`, `/allChildren/`), `conversions.js`'s `GET /`, `meters.js`'s `GET /`, `users.js`'s `GET /`, `maps.js`'s `GET /`, `preferences.js`'s `GET /`, and `ciks.js`'s `GET /` — all the same shape of bug. `maps.js`'s `/edit` validation branch had a different flavor of the same problem: `res.status(400)` called with no `.send()`/`.json()`/`.end()`, which sets the status code but never completes the response.
- **`baseline.js` had a completely broken logger**: `const log = require('../log')` imports the log *module* (an object), but the file called it directly as `log(message, 'error')` — a `TypeError` on every single failure, meaning neither of this file's two endpoints ever actually logged anything, they just crashed silently. Fixed as a side effect of routing through `failure()`.
- **Several "centralized in name only" bugs**: catch blocks that *did* log and respond, but via raw `res.sendStatus()`/`res.send()` calls that bypassed the shared response functions entirely, meaning any internal error detail passed to them went straight to the client unsanitized. Examples: `groups.js`'s `POST /delete`, `meters.js`'s `GET /:meter_id`, `users.js`'s `GET /:user_id`, and `maps.js`'s `GET /:map_id` all logged the error correctly but then responded with a raw `res.sendStatus(500)` instead of going through `failure()`.
- **A double-response bug in `conversions.js`'s `/edit`**: `success(res)` was called unconditionally after the try/catch, so it fired a second time even after `failure()` had already responded inside a caught error.
- **Two un-awaited `Promise.all(...).then(...)` chains in `groups.js`** (`GET /` and `GET /deep/groups`): neither awaited nor given a `.catch()`, so an error inside either callback fired after the enclosing `try/catch` had already exited — an unhandled rejection with nowhere to land, independent of the Express-4 async-forwarding gap below.
- **Restored `~40 call sites`' worth of lost log context**: the initial mechanical migration pass replaced `log.error("descriptive text: " + err, err)`-style calls with bare `failure(res, code, err)`, which loses everything except `err.message`. Every catch block across all 21 files was reviewed individually and had its original descriptive context restored via `new Error(message, { cause: err })`, which also required a small `log.js` fix so the console/file output continues to show the *original* error's stack trace (via `.cause`), not just the wrapper's — this only affects console/file output, never what gets persisted to the DB (which only ever stores `message`).
- **Express 4 doesn't forward rejected async route handler promises to the global error handler on its own** (Express 5 does this natively) — added `express-async-errors` as a standalone, one-line fix. This is temporary; tracked for removal in #1676 (the Express 5 migration, split out of this issue's scope at @huss's request) via a `TODO` comment already in `app.js` pointing at that issue.

## A few smaller things fixed in passing

While I was already touching these exact lines:
- `units.js`'s `/addUnit` validation log said "edit units" instead of "add units" (copy-paste from the edit endpoint).
- `maps.js`'s `/delete` catch block said "Error while deleting **group**" instead of "map" (copy-paste from `groups.js`).
- `conversions.js`'s `/simulate-delete` client-facing validation message said "delete conversions" instead of "simulate deletion of conversions" (copy-paste from `/delete`, right above it).
- `meters.js` had a TODO left by the original authors: `/edit` and `/addMeter`'s success paths called `res.json(...)` directly instead of `success()`, with a comment saying they weren't sure `success()` could return values. I checked — `res.send()` JSON-encodes objects identically to `res.json()`, so it works fine — and converted both. Once I'd confirmed that pattern was safe, I applied it consistently anywhere else a route's success path was bypassing `success()` for the same reason (`preferences.js`, `logs.js`, `baseline.js`, `ciks.js`, `compareReadings.js`, `unitReadings.js`, `conversionArray.js`).

## Test coverage added

`responseParamsTest.js` already existed and covered `success()`/`failure()`'s response-sending mechanics (status codes, comment handling, edge cases), but every one of those tests passed `error = null`, so `failure()`'s actual logging behavior — the core of what this issue asked for — was never exercised. Added 5 tests, using a `sinon` stub on `log.log()` so nothing actually hits the real DB/e-mail during the test:

- **Logs at the given severity** — passing `LogLevel.WARN` actually results in `log.log()` being called with `WARN`, the error's message, and the error object itself.
- **Defaults to `LogLevel.ERROR` when no severity is passed** — this is the "logs are created unless the developer opts out" behavior from the issue.
- **`LogLevel.SILENT` suppresses logging entirely** — even with a real error present, `log.log()` is never called.
- **No error, no log** — if `error` is `null`/falsy, nothing gets logged regardless of severity.
- **500-level responses never leak the raw error to the client** — constructs an `Error` with a fake secret in its message, sends it through `failure()` at a 500 code, and asserts the response body is only ever the generic message, never containing the secret — while confirming the error was still logged internally. This is the core safety guarantee of the whole redesign, and previously nothing verified it.

I also considered adding tests characterizing the "phantom success" delete bug noted above (`units`/`conversions`/`users`), but held off — whether deleting a nonexistent id should return `200` or something else is an API-design decision, not something to lock in via a test as a side effect of this PR. Flagging it here rather than guessing.

## Catching this branch up with `development`

This took long enough that `development` moved a lot underneath it — 106 commits, 48 files, including the new session-invalidation/auth work (`login.js` became `loginLogout.js`) and changes overlapping 14 of the 18 route files this PR touches. I rebased onto current `development` and went through every conflict by hand rather than taking either side wholesale, so both upstream's new validation logic/features and this issue's response-standardization work survive together. Full test suite passes after the rebase — 942 passing, 1 pending (down from 4 pending before the rebase, since upstream's own date-validation fix let 3 previously-skipped `compareReadingsParamsTest.js` tests get re-enabled).

## Type of change

- [ ] This change requires a documentation update (left unchecked since none of the existing developer/admin docs are part of this repo to update directly — see the note under Limitations)

## Checklist

Author: @CarlyAThomas

## Limitations

**Two files stay outside the `response.js` consolidation, both discussed and agreed with @huss on the issue:**

`csvPipeline/success.js` and `csvPipeline/failure.js` are left as-is. They were originally in scope, but PR #1591 (an in-flight, unmerged XSS fix using DOMPurify) is independently modifying the same two files for a different reason, and consolidating them now would conflict with that work and undermine its fix. Fusing them into one shared function with a mode parameter is still a real possibility, just as a separate follow-up once #1591 lands.

`obvius.js` keeps its own separate `success`/`failure` implementation — different signature, a hardware-specific `406` status, and a plaintext body format that shouldn't change without risking ~7-8 year old field hardware we can't test against. The only change there is renaming the local functions to `successObvius`/`failureObvius`, so the distinction from `response.js`'s functions is visible at every call site instead of only in a comment.

**A few pre-existing bugs and gaps turned up while reviewing every endpoint.** These are input-validation/data-shaping issues, not response-sanitization ones, so I left them alone as out of scope for this issue, but wanted them on record:

- `groups.js`'s `GET /children/:group_id` has no input validation on `:group_id`, unlike its siblings, so malformed input reaches the database directly instead of being rejected cleanly.
- `readings.js`'s `GET /line/count/meters/:meter_ids` has no numeric pattern on `meter_ids` (its sibling endpoint and `compareReadings.js`'s equivalent both do), so non-numeric input reaches the DB as `NaN` and turns into an `ERROR`-level, admin-emailed log entry instead of a clean `400`.
- The same "phantom success" pattern shows up three times: `POST /api/units/delete`, `POST /api/conversions/delete`, and `POST /api/users/delete` all run a `DELETE ... WHERE` with no `RETURNING` clause through `conn.none()`, which only throws if rows are *returned*, not based on how many were affected — so deleting something that doesn't exist still comes back as a `200` success. Confirmed live for all three.
- `GET /api/users/:user_id` sends the raw DB row back to the client, including the bcrypt `passwordHash` field. Nothing this PR changes, but probably worth a look — there's no real reason an admin caller needs the hash itself.
- Some `400`s leak raw `jsonschema` validator text (like `instance.group_id does not match pattern "^\d+$"`) to the client — that's the library's own internal formatting, not something OED wrote on purpose, though it's not a server-internals leak either (nothing about DB structure or file paths). Left the wording exactly as it was, same as everywhere else in this PR.

**Two of the route test files have a pre-existing bug that leaves them testing nothing.** Both `baselineParamsTest.js` and `conversionArrayParamsTest.js` (from PR #1528) hit the wrong URL — `/api/baseline` and `/api/conversionArray/refresh` instead of the actual mounted paths, `/api/baselines` and `/api/conversion-array`. Every request in both files silently lands on the SPA's `index.html` catch-all instead of the real route, so ~30 tests per file are commented out with a note saying the route "isn't properly mounted" — it is, the URLs just have a typo. I verified both endpoints thoroughly by hand with `curl` instead. I haven't checked whether other files from #1528 have the same issue; the ones I actually ran this session all passed with real, meaningful counts.

**Two flaky test failures showed up while I was running the full suite, and I chased both down before assuming they were mine.** `csvParamsTest.js`'s "should handle multiple file upload attempts" fails intermittently with an `EPIPE` (multer doesn't drain the request stream after a file-count error — nothing in `csv.js`/`csvPipeline` is touched by this branch at all). `logsRouteTests.js`'s "should handle date range filtering" inserts a log, waits 1s, records a cutoff timestamp, waits another 1s, inserts a second log, then queries by date range — but it assumes each `POST /api/logs/info` has finished writing to the DB by the time it responds `200`. It hasn't: `log.js`'s `Logger.log()` fires its DB insert in an un-awaited async IIFE, so the HTTP response can return before the row is actually committed. That's invisible in an isolated run (the insert finishes in milliseconds either way), but under the full suite's DB connection-pool contention from hundreds of concurrent tests, that gap widens enough to occasionally break the test's own timing assumptions. Neither is something this branch touches or introduces. I reproduced both failing on the *unmodified* base commit on a repeat run, and both pass cleanly every time in isolation on either branch — they're pre-existing flakiness under load, not something this branch causes.

**Heads up on overlap with #1666.** @aduques is working on #1666 (making client-facing error messages show the specific validation details instead of generic text), and their investigation found the same `users.js`/`groups.js` raw-response issues this PR fixes. The mechanism change (routing through `success()`/`failure()`, consistent response shape) is what I did here; I deliberately left every endpoint's original message *wording* untouched, so their work on message *content* — like replacing `users.js`'s generic "Invalid params" with the specific validation errors — is completely separate and still fully needed. @huss already weighed in on the issue that it's fine for both to proceed in parallel.

**No documentation currently covers any of this, and I think it should exist.** I checked the developer docs (they live in a separate repo, `OpenEnergyDashboard.github.io`, not this one) for anything on error handling, logging, log levels, or admin e-mail notifications. The closest existing page just says where log output ends up (`log.txt`, `nohup.out`) — nothing about how `success()`/`failure()` work, what `LogLevel` controls, or when an admin actually gets e-mailed. Since this affects how every future route is written, what the client sees on failure, and when admins get paged, I think it's worth a dedicated page rather than something I fold into this PR. I'd like to raise this with @huss directly — either as a documentation proposal in this PR thread, or as its own follow-up question — rather than guess at what should go where.

### DIFF
--- a/src/server/log.js
+++ b/src/server/log.js
@@ -60,6 +60,13 @@ class Logger {
 		if (error !== null) {
 			if (error.stack) {
 				messageToLog += `Stacktrace: \n${error.stack}\n`;
+				// If this error wraps another one (e.g. new Error(msg, { cause: originalErr })),
+				// also print the original's stack so console/file output isn't limited to just
+				// where the wrapper was created — this only affects console/file output, not the
+				// message stored in the DB, which is unaffected either way.
+				if (error.cause && error.cause.stack) {
+					messageToLog += `Caused by: \n${error.cause.stack}\n`;
+				}
 			} else {
 				// It's possible someone passed in an error that isn't actually an Error object
 				// because javascript lets you throw anything. In that case, the error won't have
--- a/src/server/routes/compareReadings.js
+++ b/src/server/routes/compareReadings.js
@@ -12,6 +12,7 @@ const Reading = require('../models/Reading');
 const { STRING_GENERAL_MAX_LENGTH, NUMERIC_ID_MAX_LENGTH } = require('../util/validationConstants');
 const { HTTP_CODES } = require('../util/httpCodes');
 const { isValidIsoDuration } = require('../util/timeValidation');
+const { success, failure } = require('./response');
 
 const DATE_TIME_WITH_TIME_REGEX = /^\d{4}-\d{2}-\d{2}(?:T| )\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:?\d{2})?$/;
 
@@ -118,7 +119,7 @@ function createRouter() {
 
 	router.get('/meters/:meter_ids', async (req, res) => {
 		if (!(validateMeterCompareReadingsParams(req.params) && validateQueryParams(req.query))) {
-			res.sendStatus(HTTP_CODES.BAD_REQUEST);
+			failure(res, HTTP_CODES.BAD_REQUEST);
 			return;
 		}
 		const meterIDs = req.params.meter_ids.split(',').map(id => parseInt(id));
@@ -128,20 +129,24 @@ function createRouter() {
 		const shiftRaw = req.query.shift;
 
 		if (!isValidCompareDateTime(currStartRaw) || !isValidCompareDateTime(currEndRaw) || !isValidIsoDuration(shiftRaw)) {
-			res.sendStatus(HTTP_CODES.BAD_REQUEST);
+			failure(res, HTTP_CODES.BAD_REQUEST);
 			return;
 		}
 
-		// The string sent should set the timezone to UTC so honor that as OED uses UTC.
-		const currStart = moment.parseZone(currStartRaw, moment.ISO_8601, true);
-		const currEnd = moment.parseZone(currEndRaw, moment.ISO_8601, true);
-		const shift = moment.duration(shiftRaw);
-		res.json(await meterCompareReadings(meterIDs, graphicUnitID, currStart, currEnd, shift));
+		try {
+			// The string sent should set the timezone to UTC so honor that as OED uses UTC.
+			const currStart = moment.parseZone(currStartRaw, moment.ISO_8601, true);
+			const currEnd = moment.parseZone(currEndRaw, moment.ISO_8601, true);
+			const shift = moment.duration(shiftRaw);
+			success(res, await meterCompareReadings(meterIDs, graphicUnitID, currStart, currEnd, shift));
+		} catch (err) {
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET meter compare readings: ${err.message}`, { cause: err }));
+		}
 	});
 
 	router.get('/groups/:group_ids', async (req, res) => {
 		if (!(validateGroupCompareReadingsParams(req.params) && validateQueryParams(req.query))) {
-			res.sendStatus(HTTP_CODES.BAD_REQUEST);
+			failure(res, HTTP_CODES.BAD_REQUEST);
 			return;
 		}
 		const groupIDs = req.params.group_ids.split(',').map(id => parseInt(id));
@@ -151,15 +156,19 @@ function createRouter() {
 		const shiftRaw = req.query.shift;
 
 		if (!isValidCompareDateTime(currStartRaw) || !isValidCompareDateTime(currEndRaw) || !isValidIsoDuration(shiftRaw)) {
-			res.sendStatus(HTTP_CODES.BAD_REQUEST);
+			failure(res, HTTP_CODES.BAD_REQUEST);
 			return;
 		}
 
-		// The string sent should set the timezone to UTC so honor that as OED uses UTC.
-		const currStart = moment.parseZone(currStartRaw, moment.ISO_8601, true);
-		const currEnd = moment.parseZone(currEndRaw, moment.ISO_8601, true);
-		const shift = moment.duration(shiftRaw);
-		res.json(await groupCompareReadings(groupIDs, graphicUnitID, currStart, currEnd, shift));
+		try {
+			// The string sent should set the timezone to UTC so honor that as OED uses UTC.
+			const currStart = moment.parseZone(currStartRaw, moment.ISO_8601, true);
+			const currEnd = moment.parseZone(currEndRaw, moment.ISO_8601, true);
+			const shift = moment.duration(shiftRaw);
+			success(res, await groupCompareReadings(groupIDs, graphicUnitID, currStart, currEnd, shift));
+		} catch (err) {
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET group compare readings: ${err.message}`, { cause: err }));
+		}
 	});
 
 	return router;
--- a/src/server/routes/groups.js
+++ b/src/server/routes/groups.js
@@ -10,9 +10,8 @@ const Unit = require('../models/Unit');
 const { getConnection } = require('../db');
 const Group = require('../models/Group');
 const { adminAuthMiddleware, optionalAuthMiddleware } = require('./authenticator');
-const { log } = require('../log');
 const Point = require('../models/Point');
-const { failure, success } = require('./response');
+const { failure, success, LogLevel } = require('./response');
 const { HTTP_CODES } = require('../util/httpCodes');
 const { STRING_GENERAL_MAX_LENGTH, STRING_SHORT_MAX_LENGTH: SHORT_STRING_MAX_LENGTH, NUMERIC_ID_MAX_LENGTH } = require('../util/validationConstants');
 
@@ -138,15 +137,14 @@ router.get('/', optionalAuthMiddleware, async (req, res) => {
 	const conn = getConnection();
 	try {
 		const rows = await Group.getAll(conn);
-		const promises = await rows.map(async (row) => {
+		const promises = rows.map(async (row) => {
 			const deepMeters = await Group.getDeepMetersByGroupID(row.id, conn);
 			return { ...row, deepMeters: deepMeters};
 		});
-		Promise.all(promises).then(function (values) {
-			res.json(values.map(formatGroupForResponse));
-		})
+		const values = await Promise.all(promises);
+		res.json(values.map(formatGroupForResponse));
 	} catch (err) {
-		log.error(`Error while preforming GET all groups query: ${err}`, err);
+		failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET all groups query: ${err.message}`, { cause: err }));
 	}
 });
 
@@ -159,7 +157,7 @@ router.get('/idname', optionalAuthMiddleware, async (req, res) => {
 		const rows = await Group.getAll(conn);
 		res.json(rows.map(formatToOnlyNameID));
 	} catch (err) {
-		log.error(`Error while performing GET all groups query: ${err}`, err);
+		failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET all groups id and name query: ${err.message}`, { cause: err }));
 	}
 });
 
@@ -171,15 +169,14 @@ router.get('/deep/groups', adminAuthMiddleware('view deep groups'), async (req, 
 	const conn = getConnection();
 	try{
 		const rows = await Group.getAll(conn);
-		const promises = await rows.map(async (row) => {
+		const promises = rows.map(async (row) => {
 			const deepGroups = await Group.getDeepGroupsByGroupID(row.id, conn);
 			return { ...row, deepGroups: deepGroups, deepMeters: [] };
 		});
-		Promise.all(promises).then(function (values) {
-			res.json(values.map(formatGroupForResponse));
-		});
+		const values = await Promise.all(promises);
+		res.json(values.map(formatGroupForResponse));
 	} catch (err) {
-		log.error(`Error while performing GET deep groups for all groups query: ${err}`, err);
+		failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET deep groups for all groups query: ${err.message}`, { cause: err }));
 	}
 });
 
@@ -200,7 +197,7 @@ router.get('/children/:group_id', optionalAuthMiddleware, async (req, res) => {
 		]);
 		res.json({ meters, groups, deepMeters });
 	} catch (err) {
-		log.error(`Error while preforming GET on all immediate children (meters and groups) of specific group: ${err}`, err);
+		failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET on all immediate children (meters and groups) of specific group: ${err.message}`, { cause: err }));
 	}
 });
 
@@ -217,7 +214,7 @@ router.get('/allChildren/', optionalAuthMiddleware, async (req, res) => {
 		const allChildren = await Group.getImmediateChildren(conn);
 		res.json(allChildren);
 	} catch (err) {
-		log.error(`Error while preforming GET on all immediate children (meters and groups) of all groups: ${err}`, err);
+		failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET on all immediate children (meters and groups) of all groups: ${err.message}`, { cause: err }));
 	}
 });
 
@@ -236,16 +233,15 @@ router.get('/deep/groups/:group_id', optionalAuthMiddleware, async (req, res) =>
 	};
 	const validatorResult = validate(req.params, validParams);
 	if (!validatorResult.valid) {
-		log.error(`Got request group deep group children with invalid data, errors: ${validatorResult.errors}`);
-		failure(res, HTTP_CODES.BAD_REQUEST, "Got request group deep group children with invalid data. Error(s): " + validatorResult.errors.toString());
+		const message = "Got request group deep group children with invalid data. Error(s): " + validatorResult.errors.toString();
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message);
 	} else {
 		const conn = getConnection();
 		try {
 			const deepGroups = await Group.getDeepGroupsByGroupID(req.params.group_id, conn);
 			res.json({ deepGroups });
 		} catch (err) {
-			log.error(`Error while preforming GET on all deep child groups of specific group: ${err}`, err);
-			res.sendStatus(HTTP_CODES.INTERNAL_SERVER_ERROR);
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET on all deep child groups of specific group: ${err.message}`, { cause: err }));
 		}
 	}
 });
@@ -265,16 +261,15 @@ router.get('/deep/meters/:group_id', optionalAuthMiddleware, async (req, res) =>
 	};
 	const validatorResult = validate(req.params, validParams);
 	if (!validatorResult.valid) {
-		log.error(`Got request group deep meter children with invalid data, errors: ${validatorResult.errors}`);
-		failure(res, HTTP_CODES.BAD_REQUEST, "Got request group deep meter children with invalid data. Error(s): " + validatorResult.errors.toString());
+		const message = "Got request group deep meter children with invalid data. Error(s): " + validatorResult.errors.toString();
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message);
 	} else {
 		const conn = getConnection();
 		try {
 			const deepMeters = await Group.getDeepMetersByGroupID(req.params.group_id, conn);
 			res.json({ deepMeters });
 		} catch (err) {
-			log.error(`Error while preforming GET on all deep child meters of specific group: ${err}`, err);
-			res.sendStatus(HTTP_CODES.INTERNAL_SERVER_ERROR);
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET on all deep child meters of specific group: ${err.message}`, { cause: err }));
 		}
 	}
 });
@@ -294,16 +289,15 @@ router.get('/parents/:group_id', optionalAuthMiddleware, async (req, res) => {
 	};
 	const validatorResult = validate(req.params, validParams);
 	if (!validatorResult.valid) {
-		log.error(`Got request group parents with invalid data, errors: ${validatorResult.errors}`);
-		failure(res, HTTP_CODES.BAD_REQUEST, "Got request group parents with invalid data. Error(s): " + validatorResult.errors.toString());
+		const message = "Got request group parents with invalid data. Error(s): " + validatorResult.errors.toString();
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message);
 	} else {
 		const conn = getConnection();
 		try {
 			const parentGroups = await Group.getParentsByGroupID(req.params.group_id, conn);
 			res.json(parentGroups);
 		} catch (err) {
-			log.error(`Error while preforming GET on all parents of specific group: ${err}`, err);
-			res.sendStatus(HTTP_CODES.INTERNAL_SERVER_ERROR);
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing GET on all parents of specific group: ${err.message}`, { cause: err }));
 		}
 	}
 });
@@ -313,8 +307,8 @@ router.post('/create', adminAuthMiddleware('create groups'), async (req, res) =>
 	const validatorResult = validateGroupsParams(req.body, false);
 
 	if (!validatorResult.valid) {
-		log.error(`Got request to create group with invalid data, errors: ${validatorResult.errors}`);
-		failure(res, HTTP_CODES.BAD_REQUEST, "Got request to create group with invalid data. Error(s): " + validatorResult.errors.toString());
+		const message = "Got request to create group with invalid data. Error(s): " + validatorResult.errors.toString();
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message);
 	} else {
 		const conn = getConnection();
 		try {
@@ -340,10 +334,10 @@ router.post('/create', adminAuthMiddleware('create groups'), async (req, res) =>
 		} catch (err) {
 			// Group duplicate-name DB errors to a safe 400 response
 			if (err.toString() === 'error: duplicate key value violates unique constraint "groups_name_key"') {
-				failure(res, HTTP_CODES.BAD_REQUEST, 'Group name already exists');
+				failure(res, HTTP_CODES.BAD_REQUEST, err, 'Group name already exists', LogLevel.SILENT);
 			} else {
-				log.error(`Error while inserting new group ${err}`, err);
-				failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, err.toString() + ' with detail ' + err['detail']);
+				const detail = err['detail'] ? ` with detail ${err['detail']}` : '';
+				failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while inserting new group: ${err.message}${detail}`, { cause: err }));
 			}
 		}
 	}
@@ -354,8 +348,8 @@ router.put('/edit', adminAuthMiddleware('edit groups'), async (req, res) => {
 	const validatorResult = validateGroupsParams(req.body, true);
 
 	if (!validatorResult.valid) {
-		log.error(`Got request to edit group with invalid data, errors: ${validatorResult.errors}`);
-		failure(res, HTTP_CODES.BAD_REQUEST, "Got request to edit group with invalid data. Error(s): " + validatorResult.errors.toString());
+		const message = "Got request to edit group with invalid data. Error(s): " + validatorResult.errors.toString();
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message);
 	} else {
 		try {
 			const conn = getConnection();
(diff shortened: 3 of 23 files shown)


huss:
@CarlyAThomas Thank you for this substantial contribution, esp. being your first one. I have reviewed your well documented changes and comments provided with this PR. I am very appreciative of the extra work you did outside the direct issue you were working on. At this point I wanted to give some early thoughts:

- It will take a little while to get through the many changes since they impact many files/areas. Please know I am trying to get the time to go through them all with the care they deserve. If I am delaying you then please let me know.
- You mention a number of items that are not done within the code changes of this PR but indicating they probably deserve consideration. In at least one case you talk about future work. These seem very useful and I will be commenting on them later. For now, I wanted to get a sense of what you might want for when I respond. I always prefer to give the person who located a concern the option to create the issue/documentation on the situation. Do you think you might want to do that for some/all of the items you discussed? If not, OED can create them as desired. This will help me know, if possible, what would be best to say/provide in my thoughts on each of these. Obviously, OED is not asking you to commit to definitely do any issues but it would remain an option if you want to after any discussion is done on them.

Again, thanks for your efforts and any information/thoughts/questions. I hope this is clear for you.

CarlyAThomas:
Thanks! And no worries on timing—I know this is a large review, so I'd much rather you take the time needed than rush through it.

As for the additional items, I'd be happy to create the issues for the ones that make sense. My main goal is simply to make sure they're valid observations and that the scope is appropriate before they're tracked. Once we've discussed them, I'm happy to write them up if you think they'd be worthwhile additions to the issue tracker.

huss:
> Once we've discussed them, I'm happy to write them up if you think they'd be worthwhile additions to the issue tracker.

Thank you and this sounds great. It will be a little while but I'll get to this.

src/server/routes/response.js
```diff
@@ -18,19 +19,35 @@ function success(res, comment = '') {
...
+ *  will not e-mail admins by default, while LogLevel.ERROR will.
  *
  */
-function failure(res, code = HTTP_CODES.INTERNAL_SERVER_ERROR, comment = '') {
-	// Return a generic message for 500-level failures.
-	const responseBody = code >= HTTP_CODES.INTERNAL_SERVER_ERROR ? 'Internal Server Error. Details are in the OED logs that are available to your site admin(s).' : comment;
+function failure(res, code = HTTP_CODES.INTERNAL_SERVER_ERROR, error = null, safeMessage = '', severity = LogLevel.ERROR) {
+	if (severity !== LogLevel.SILENT && error) {
+		const logMessage = error instanceof Error ? error.message : String(error);
+		log.log(severity, logMessage, error instanceof Error ? error : undefined);
```
huss, line 41: Two thoughts.
1. If there is no error then it is set to undefined. In log it has a default of null. While similar, is there a reason for this difference?
2. The third argument has logic here but for the second argument it is assigned to a const right above. I think being consistent is good. I would vote to create the const for both as that makes debugging a little easier.

src/server/routes/response.js
```diff
@@ -18,19 +19,35 @@ function success(res, comment = '') {
...
+		? 'Internal Server Error. Details are in the OED logs that are available to your site admin(s).'
+		: safeMessage;
 	res.status(code)
-	.send(responseBody);
-
+		.send(responseBody);
 }
 
-module.exports = { success, failure };
+// Re-export LogLevel so routes only need to import from here, not from log.js directly,
```
huss, line 51: I wanted to get your thoughts on re-exporting an item from a require. I don't think OED does it anywhere else and it makes it another step to get to the actual values. I'm not really knowledgeable but did find source(s) via a web search that indicate it can be an issue (esp. if using require and not import). I did a global search and I think the exported value is used in 7 files that would need a separate require if this is removed. Given this, do you think this is best and, if so, should OED do this more consistently across the code base?

src/server/routes/compareReadings.js
```diff
@@ -128,20 +129,24 @@ function createRouter() {
...
+			failure(res, HTTP_CODES.BAD_REQUEST);
 			return;
 		}
 
-		// The string sent should set the timezone to UTC so honor that as OED uses UTC.
-		const currStart = moment.parseZone(currStartRaw, moment.ISO_8601, true);
-		const currEnd = moment.parseZone(currEndRaw, moment.ISO_8601, true);
-		const shift = moment.duration(shiftRaw);
-		res.json(await meterCompareReadings(meterIDs, graphicUnitID, currStart, currEnd, shift));
+		try {
```
huss, line 136: Thanks for wrapping routes in try/catch to avoid issues. What I'm wondering is if this should wrap the entire route code. It only returns a generic message without any specific text details. Thus, if an error occurred in another part of the function it would still seem to apply and might be a little safer. What do you think? This would apply below and maybe to other routes (could be a separate issue to fix them all).

src/server/routes/compareReadings.js
```diff
@@ -118,7 +119,7 @@ function createRouter() {
 
 	router.get('/meters/:meter_ids', async (req, res) => {
 		if (!(validateMeterCompareReadingsParams(req.params) && validateQueryParams(req.query))) {
-			res.sendStatus(HTTP_CODES.BAD_REQUEST);
+			failure(res, HTTP_CODES.BAD_REQUEST);
 			return;
```
huss, line 123: Sometimes there is a return after success/failure and sometimes not. Which way do you think OED should do this?

src/server/routes/groups.js
```diff
@@ -340,10 +334,10 @@ router.post('/create', adminAuthMiddleware('create groups'), async (req, res) =>
 		} catch (err) {
 			// Group duplicate-name DB errors to a safe 400 response
 			if (err.toString() === 'error: duplicate key value violates unique constraint "groups_name_key"') {
-				failure(res, HTTP_CODES.BAD_REQUEST, 'Group name already exists');
+				failure(res, HTTP_CODES.BAD_REQUEST, err, 'Group name already exists', LogLevel.SILENT);
```
huss, line 337: I can see why this was set to silent. However, for security reasons, I think it might be best to make this an info (or warn) log level.

src/server/routes/groups.js
```diff
@@ -340,10 +334,10 @@ router.post('/create', adminAuthMiddleware('create groups'), async (req, res) =>
...
 			} else {
-				log.error(`Error while inserting new group ${err}`, err);
-				failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, err.toString() + ' with detail ' + err['detail']);
+				const detail = err['detail'] ? ` with detail ${err['detail']}` : '';
```
huss, line 339: Can you help me understand why this has err['detail'] but other ones use a different way? If needed, a comment for future developers would be nice.

src/server/routes/groups.js
```diff
@@ -393,13 +387,12 @@ router.put('/edit', adminAuthMiddleware('edit groups'), async (req, res) => {
 
 				return t.batch(flatten([adoptGroupsQueries, disownGroupsQueries, adoptMetersQueries, disownMetersQueries]));
 			});
-			res.sendStatus(HTTP_CODES.OK);
+			success(res);
 		} catch (err) {
 			if (err.message && err.message === 'Cyclic group detected') {
-				res.status(HTTP_CODES.BAD_REQUEST).send({ message: err.message });
+				failure(res, HTTP_CODES.BAD_REQUEST, err, err.message, LogLevel.SILENT);
```
huss, line 393: See other silent comment and this should be a warn.

src/server/routes/loginLogout.js
```diff
@@ -55,16 +55,15 @@ router.post('/login', credentialsRequestValidationMiddleware, async (req, res) =
...
 				const token = jwt.sign({ data: user.id }, secretToken, { expiresIn: 86400 });
-				res.json({ token: token, username: user.username, role: user.role });
+				success(res, { token: token, username: user.username, role: user.role });
 			} else {
 				throw new Error('Unauthorized password');
 			}
 		} catch (err) {
 			if (err.message === 'Unauthorized password' || err.message === 'No data returned from the query.') {
-				res.status(HTTP_CODES.UNAUTHORIZED).send({ text: 'Not authorized' });
+				failure(res, HTTP_CODES.UNAUTHORIZED, err, { text: 'Not authorized' }, LogLevel.SILENT);
```
huss, line 64: I think this should be a warn level. We want admins to know if there is a sequence of bad login attempts.

src/server/routes/loginLogout.js
```diff
@@ -106,15 +105,14 @@ router.post('/logout', async (req, res) => {
...
 		await User.invalidateTokensBeforeNow(user.id, conn);
-		res.status(HTTP_CODES.OK).json({ success: true, message: 'Logout successful.' });
+		success(res, { success: true, message: 'Logout successful.' });
 	} catch (error) {
 		if (error.code === 'TOKEN_INVALIDATED') {
-			res.status(HTTP_CODES.OK).json({ success: true, message: 'Logout successful.' });
+			success(res, { success: true, message: 'Logout successful.' });
 		} else if (error.message === 'No data returned from the query.') {
-			res.status(HTTP_CODES.UNAUTHORIZED).json({ success: false, message: 'Logout failed.' });
+			failure(res, HTTP_CODES.UNAUTHORIZED, null, { success: false, message: 'Logout failed.' });
```
huss, line 113: I think if this specific error is going to be caught separately then it should have a specific error for this case instead of a null error. Maybe "Logout failed due to no returned data while invalidating user tokens".

Update: Maybe this is for a different effort and not this PR?

src/server/routes/loginLogout.js
```diff
@@ -106,15 +105,14 @@ router.post('/logout', async (req, res) => {
...
 		} else {
-			log.error('Logout failed while invalidating user tokens.', error);
-			res.status(HTTP_CODES.INTERNAL_SERVER_ERROR).json({ success: false, message: 'Logout failed.' });
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Logout failed while invalidating user tokens: ${error.message}`, { cause: error }));
```
huss, line 115: Maybe I'm missing something but this seems to change what is returned when an error occurs. The old version was JSON with ``success: false``. Is that correct? If so, was there a reason for the change?

src/server/routes/maps.js
```diff
@@ -227,13 +226,12 @@ router.post('/create', adminAuthMiddleware('create maps'), async (req, res) => {
 				);
 				await newMap.insert(t);
 			});
-			res.sendStatus(HTTP_CODES.OK);
+			success(res);
 		} catch (err) {
 			if (err.toString() === 'error: duplicate key value violates unique constraint "maps_name_key"') {
-				res.status(HTTP_CODES.BAD_REQUEST).json({ error: `Map "${req.body.name}" is already in use.` });
+				failure(res, HTTP_CODES.BAD_REQUEST, err, { error: `Map "${req.body.name}" is already in use.` }, LogLevel.SILENT);
```
huss, line 232: Similar to another comment, don't have this be a silent level.

src/server/routes/meters.js
```diff
@@ -279,16 +279,16 @@ router.post('/edit', adminAuthMiddleware('edit meters'), async (req, res) => {
...
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message, LogLevel.WARN);
 	} else if (
 		(req.body.startTimestamp && !isValidIsoDateTime(req.body.startTimestamp, false)) ||
 		(req.body.endTimestamp && !isValidIsoDateTime(req.body.endTimestamp, false)) ||
 		(req.body.previousEnd && !isValidIsoDateTime(req.body.previousEnd)) ||
 		(req.body.minDate && !isValidIsoDateTime(req.body.minDate)) ||
 		(req.body.maxDate && !isValidIsoDateTime(req.body.maxDate))
 	) {
-		failure(res, HTTP_CODES.BAD_REQUEST, 'invalid date/time format');
+		failure(res, HTTP_CODES.BAD_REQUEST, null, 'invalid date/time format');
```
huss, line 291: Per other comments, this should log some message.

src/server/routes/meters.js
```diff
@@ -363,7 +361,7 @@ router.post('/addMeter', adminAuthMiddleware('add meter'), async (req, res) => {
...
```
huss, line 364: See above about logging.

src/server/routes/meters.js
```diff
@@ -140,7 +139,8 @@ router.get('/:meter_id', optionalAuthMiddleware, async (req, res) => {
 		}
 	};
 	if (!validate(req.params, validParams).valid) {
-		res.sendStatus(HTTP_CODES.BAD_REQUEST);
+		// No message given so as not to reveal whether the meter exists, matching the branch below.
+		failure(res, HTTP_CODES.BAD_REQUEST);
```
huss, line 143: The parameter validation does not check if the meter exists or is displayable. It is strictly looking to see if the format is okay. Thus, it should return the issue with the validation as in other routes. This was not addressed in the PR #1689 but it should be and maybe that is the right place to do this. Also applies below. Noting for now.

src/server/routes/users.js
```diff
@@ -101,11 +101,11 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
 		maxLength: TOKEN_MAX_LENGTH
 	};
 	if (!validate(token, validParams).valid) {
-		res.status(HTTP_CODES.FORBIDDEN).json({ message: 'No token provided or JSON was invalid.' });
+		failure(res, HTTP_CODES.FORBIDDEN, null, { message: 'No token provided or JSON was invalid.' });
```
huss, line 104: This and other failures in file should log it (not null).

src/server/routes/users.js
```diff
@@ -116,12 +116,12 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
 							role: userProfile.role
 						});
 				} catch (error) {
-					res.status(HTTP_CODES.UNAUTHORIZED).json({ message: 'User does not exist in database.' });
+					failure(res, HTTP_CODES.UNAUTHORIZED, error, { message: 'User does not exist in database.' }, LogLevel.SILENT);
```
huss, line 119: This does not have to be for this PR but I'm thinking this is providing information that should not be returned (as did the previous version). It is letting someone know the user does not exist. Now, I think it must be a valid token to get here but I might be more generic to be careful.

src/server/routes/users.js
```diff
@@ -116,12 +116,12 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
...
 				}
 			}
 		});
 	} else {
-		res.status(HTTP_CODES.FORBIDDEN).send({ message: 'No token provided.' });
+		failure(res, HTTP_CODES.FORBIDDEN, null, { message: 'No token provided.' });
```
huss, line 124: Again, maybe for future change. I think the message would be better if it said "Invalid or no token provided.".

src/server/routes/users.js
```diff
@@ -157,25 +156,23 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
 				const user = new User(undefined, username, hashedPassword, role, note);
 				await user.insert(conn);
-				res.sendStatus(HTTP_CODES.OK);
+				success(res);
 			}
 		} catch (error) {
-			// Log the error internally and return a generic response
-			log.error(`Error while performing POST request to create user: ${error}`, error);
-			res.status(HTTP_CODES.INTERNAL_SERVER_ERROR).send({ message: 'Internal Server Error' });
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing POST request to create user: ${error.message}`, { cause: error }));
```
huss, line 175: A note for the future. This message has POST when others do not. At some point OED needs to sweep all messages to be consistent.

huss (changes requested):
@CarlyAThomas First, per a recent Discord msg on the OED developer channel, I want to say I'm sorry this took so long. Second, I want to thank you for your through and careful work on this issue, esp. given it is your first contribution to OED. You not only addressed the original issue but carefully analyzed, found and fixed/documented other items.

Overall, review and testing found it works well. I have made some comments on specific parts of the code. Some are items that probably should be addressed as part of this PR (let me know if the delay in this review means you cannot address them) but some are likely for work outside this PR. I would appreciate you giving any thoughts on the items you want as you have valuable insights. I would appreciate you not resolving any comments as I use them to track items I need to consider/re-review. You are welcome to reply. If you think it is addressed then a simple thumbs up will let me know. I will resolve as I clear the item.

I also have some comments that are not only for a single place in the code so I'm making them here. I welcome your thoughts. It may well be that they are future items but settling what should be done would be nice.

- I noticed these files have res.send/status calls directly. Should they use the new functions? I was unsure why some were not included in the changes (maybe because the receiving client cannot deal with or ignores?).
  - src/server/app.js
  - src/server/routes/authenticator.js
- A number of files have res.json to return values. OED should think though if some/all should be switched to success. Unsure why some were not included.
- Some places use LogLevel.WARN for parameter issues in routes and some don't. I think they should probably all be warn but OED needs to make a decision. Given how often this occurs, it makes me wonder about a parameter failure wrapper that calls failure to make sure they are all consistent.

I notice the following during my manual testing using the OED web pages. These are all for future consideration but your thoughts are welcome. This involved deliberately causing errors that would not normally happen:

- Some routes that generate failure responses are silently absorbed but they might cause issues. For example, if the '/' route in src/server/routes/conversions.js fails it shows up as an error in the web console (client-side) as well as the terminal where OED was started (server-side). The web page appears as normal on a refresh. This might not break the user experience but I suspect some will. OED needs to consider detecting these errors and properly deciding what to do.
- Generating an error in routes is correctly displayed as an error message in the client. It also shows the expected route failure in the web terminal. However, it is also showing an uncaught exception in the web console. Examples are add and edit in src/server/routes/conversions.js. I made this happen by putting ``req.body.wrong = -1;`` at the start of the route before the param check.
- I'm noting I tried various routes but did not systematically do them all. OED needs to carefully test all before the release for these changes and other reasons so it seems okay to put off. I have no reason to think they will fail and this will not hold up this PR.

src/server/log.js
huss, line 84: This was thought through while reviewing PR #1689. I've changed my mind and decided the stack trace should go into the DB as a permanent record. Otherwise, someone cannot provide to OED if they don't grab the console and that is less likely in production or if one does not know the error occurred. Thus, I propose the code becomes:

```suggestion
			const logMsg = new LogMsg(level.name, messageToLog, logTime);
```

huss, line 78:
```suggestion
			if (error.stack) {
				// The calling error stack trace if it exists.
				messageToLog += `Stacktrace: \n${error.stack}\n`;
			}
			if (error.cause && error.cause.stack) {
				// If this error wraps another one (e.g. new Error(msg, { cause: originalErr })),
				// also print the original's stack so console/file output isn't limited to just
				// where the wrapper was created.
				// OED failure should set error.cause and error.cause.stack to the same value but
				// use stack to be consistent. If it ever turns out stack is not enough then the
				// cause can be show for other types of causes.
				messageToLog += `Caused by: \n${error.cause.stack}\n`;
			}
		} else {
			// It's possible someone passed in an error that isn't actually an Error object
			// because javascript lets you throw anything. In that case, the error won't have
			// a stack.

			// So we just try to convert whatever we got to a string and include it. It's better
			// than nothing.
			messageToLog += `An error was included, but it was not an Error object:\n${error}`;
		}
```

The change listed does not seem to have all the lines I indicated but hopefully the location is clear as it is replacement code. This was thought through while reviewing PR #1689. I'm making this suggestion for the following reasons:

- In case there is no original error stack the code still tries to add the cause one if possible.
- Some commenting was modified/extended since I had to think this through and wanted to document.

src/server/routes/response.js
huss, line 16: This was thought through while reviewing PR #1689. I was thinking that this leaves some msgs shown to the user via popups to have nothing at the end and it isn't clear why. Thus, I proposed (wording open to thoughts):

```suggestion
function success(res, comment = 'No known reason for issue.') {
```

PR #1689, opened 2026-08-01
Issue1666 - Include specific parameters to success/error messages on all routes

aduques:
# Description

This PR addresses inconsistent and vague error messages returned to the client across groups.js, units.js, users.js, meters.js, and conversions.js. The root causes fell into a few categories:

- error.data vs error.data.message mismatch

Some client components read error.data while others read error.data.message, depending on whether the corresponding server route sent a plain string (via response.js success()/failure()) or a raw JSON object (via res.json()/res.send()).

users.js did not previously import response.js at all, so its client components (CreateUserModalComponent.tsx, EditUserModalComponent.tsx) required .message to extract the string. This PR converts users.js to use response.js consistently, so error.data can be used uniformly. This change for consistency is intended to set up a future issue where a standard function will be created to handle the messages and validations of these routes.

- Generic "Invalid params" / hardcoded messages

users.js /create, /edit, and /delete routes previously discarded the actual validator output (validatorResult.errors) and always returned a generic "Invalid params" message. This has been fixed to display the specific validation error, matching the patterns already used in groups.js and conversions.js

- response.js failure() discarding custom messages at 500+

failure() always overrides the provided comment with a generic "Internal Server Error" message whenever the status code is 500 or above. Several routes (groups.js, meters.js) were passing expected errors (duplicate names, invalid check-constraint values) through failure() at 500, silently hiding the actual message. These routes now catch known error cases (duplicate key violations, check constraint violations) and re-send them at 400 so the specific message reaches the client, while unexpected errors continue to fall through to the generic 500 response as intended.

- Missing response bodies causing undefined in the client

groups.js /delete catch block previously called res.sendStatus() with no body, causing the client to display "undefined." This has been fixed to use failure() with a specific message.

All affected routes (conversions.js, groups.js, units.js, users.js, meters.js) and their corresponding client components have been manually tested for CREATE, EDIT, and DELETE success and failure paths to confirm specific, correct messages are now displayed.

Partially Addresses #1666

## Type of change

## Checklist

## Limitations

- meters.js /edit and /addMeter routes could not be converted to use response.js's success() on the success path, since the client depends on the DB-assigned data (e.g. id) being returned in the response body, and success() does not currently support returning structured data. These routes continue to use res.json() on success, per the pre-existing TODO comment in the file.
- A broader, standardized validation and response wrapper function (to centralize validate(), logging, and message formatting across all routes) was discussed with the maintainer but has been left to be addressed in a new issue, since it depends on this issue and #1614 to be completed first.


aduques:
I acknowledge that the there are still commented out TODO DEBUG lines of codes in the committed code. I will remove these in a future commit once the changes have been reviewed and the pull request is close to being wrapped up.

## users.js

**CREATE FAIL:** Failed to create the user: (username: himeko) Got request to insert user with invalid user data. Error(s): instance is not allowed to have the additional property "username",instance is not allowed to have the additional property "role",instance is not allowed to have the additional property "password",instance is not allowed to have the additional property "note",instance requires property "user"

**EDIT FAIL:** Failed to edit user: (username: himeko) Got request to edit users with invalid user data, errors: instance.user.username does not meet minimum length of 3

**DELETE FAIL:** Failed to delete the user: (username: himeko) Got request to delete users with invalid user data. Error(s): instance is not allowed to have the additional property "id"

**CREATE SUCCESS:** Successfully created the user: (username: welt) (role: admin)

**EDIT SUCCESS:** Successfully edited user: (username: welt) (role: admin)

**DELETE SUCCESS:** Successfully deleted user: (username: welt)

## conversions.js

**CREATE FAIL:** Failed to create a conversion. (Source: "7/26edit", Destination: "730unit") Got request to insert conversion with invalid conversion data. Error(s): instance.sourceId must be greater than or equal to 1,instance.destinationId must be greater than or equal to 1

**EDIT FAIL:** Failed to edit conversion. (Source: "730unit", Destination: "7/26unit") Got request to edit conversions with invalid conversion data, errors: instance.sourceId must be greater than or equal to 1,instance.destinationId must be greater than or equal to 1

**DELETE FAIL:** Failed to delete conversion. (Source: "730unit", Destination: "7/26edit") Got request to delete conversions with invalid conversion data. Error(s): instance.meterIds[0] must be greater than or equal to 1

**CREATE SUCCESS:** Successfully created a conversion. (Source: "7/26edit", Destination: "730unit")

**EDIT SUCCESS:** Successfully edited conversion. (Source: "7/26edit", Destination: "730unit")

**DELETE SUCCESS:** Successfully deleted conversion. (Source: "730unit", Destination: "7/26edit")

## groups.js

**CREATE FAIL:** Failed to create a group with message: (name: "730group") Got request to create group with invalid data. Error(s): instance.name does not meet minimum length of 1

**EDIT FAIL:** Failed to edit group with message: (Name: "730group") Got request to edit group with invalid data. Error(s): instance.name does not meet minimum length of 1

**DELETE FAIL:** Failed to delete group.Got request to delete group with invalid data. Error(s): instance.id must be greater than or equal to 1

**CREATE SUCCESS:** Successfully created a group. (name: "730group")

**EDIT SUCCESS:** Successfully edited group. (Name: "730group")

**DELETE SUCCESS:** Successfully deleted group. (Name: "730group")

## meters.js

**CREATE FAIL:**
Failed to create a meter with message: "731test" (identifier: 731test, type: egauge) invalid date/time format

Failed to create a meter with message: "730meter" (identifier: 730meter, type: egauge) Meter name "730meter" already exists

**EDIT FAIL:**
Failed to edit meter with message: "727meter" (identifier: 727meter, type: egauge) Invalid meter data: error: new row for relation "meters" violates check constraint "meters_identifier_check"

Failed to edit meter with message: "730meter" (identifier: 730meter, type: egauge) Meter name "730meter" already exists

**CREATE SUCCESS:** Successfully created a meter."731success" (identifier: 731success, type: egauge)

**EDIT SUCCESS:** Successfully edited meter."731success" (identifier: 731success, type: egauge)

**NOTE:** success() cannot be used currently in meters.js. There is an existing TODO note that states that /addMeter and /edit are expected to return a json object. Therefore, simply replacing it with success() does not work. For now, I have skipped over applying success() until I look into it more. If this is something that I can do within this pull request, I can apply the change, but if it requires more files to be changed, it may be better to open this specifically as a new issue to keep it separate from this one. Thoughts on this is appreciated.

**NOTE:** Every database error would fall through to the generic 500 branch in response.js's failure(). That function is designed to discard any custom message and always send back "Internal Server Error..." whenever the status code is 500 or higher, so no matter what specific comment the route passed in, the client would only ever see the generic text. By catching these particular error types here and re-sending them at 400 instead, failure() is allowed to pass the actual message through rather than suppressing it, so the user sees what's actually wrong instead of the message shown in failure().

## units.js

**CREATE FAIL:** 
Failed to create a unit. "731unit" (identifier: 731unit, type: unit) Got request to add units with invalid unit data, errors: instance.secInRate must be greater than or equal to 0

Failed to create a unit. "730unit" (identifier: 730unit, type: unit) Unit name "730unit" already exists

**EDIT FAIL:** 
Failed to edit unit."730unit" (identifier: 730unit, type: unit) Got request to edit units with invalid unit data, errors: instance.secInRate must be greater than or equal to 0

Failed to edit unit."730unit" (identifier: 730unit, type: unit) Unit name "730unit" already exists

**DELETE FAIL:** Failed to deleted unit with error: Got request to delete a unit with invalid data, error(s): instance.id must be greater than or equal to 1

**CREATE SUCCESS:** Successfully created a unit. "731unit" (identifier: 731unit, type: unit)

**EDIT SUCCESS:** Successfully edited unit. "731unit" (identifier: 731unit, type: unit)

**DELETE SUCCESS:** Successfully deleted unit: 731unit

src/server/routes/conversions.js
```diff
@@ -131,7 +131,8 @@ router.post('/addConversion', adminAuthMiddleware('add conversions'), async (req
 				);
 				await newConversion.insert(t);
 			});
-			res.sendStatus(HTTP_CODES.OK);
+			//res.sendStatus(HTTP_CODES.OK);
```
huss: I think this change overlaps a change in PR #1688 that came in at about the same time. My current thinking is their change should stay as all routes were aligned with this new methodology. Thought?

src/server/routes/groups.js
```diff
@@ -393,19 +393,24 @@ router.put('/edit', adminAuthMiddleware('edit groups'), async (req, res) => {
 
 				return t.batch(flatten([adoptGroupsQueries, disownGroupsQueries, adoptMetersQueries, disownMetersQueries]));
 			});
-			res.sendStatus(HTTP_CODES.OK);
+			//res.sendStatus(HTTP_CODES.OK);
```
huss: See other comment. The changes here also overlap the other PR. Thoughts?

src/server/routes/meters.js
```diff
@@ -340,7 +340,14 @@ router.post('/edit', adminAuthMiddleware('edit meters'), async (req, res) => {
 			res.json(formatMeterForResponse(meter, true));
 		} catch (err) {
 			log.error(`Error while editing a meter with detail "${err['detail']}"`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, err.toString() + ' with detail ' + err['detail']);
+			if (err.toString().includes('duplicate key value violates unique constraint')) {
```
huss, line 340: See other comment. The changes here also overlap the other PR but I think bring in other features. Do you think they could be merged? Thoughts?

src/server/routes/units.js
```diff
@@ -158,7 +158,15 @@ router.post('/edit', adminAuthMiddleware('edit units'), async (req, res) => {
 			success(res, 'Successfully edited unit');
 		} catch (err) {
 			log.error(`Failed to update unit: ${err}`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, 'Unable to update unit');
+			//failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, 'Unable to update unit');
```
huss: See other comment. The changes here also overlap the other PR but with new items. Thoughts?

src/server/routes/units.js
```diff
@@ -197,7 +205,15 @@ router.post('/addUnit', adminAuthMiddleware('add units'), async (req, res) => {
 			success(res, 'Unit created successfully');
 		} catch (err) {
 			log.error(`Error while inserting new unit: ${err}`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, `Error while inserting new unit: ${err}`);
+			//failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, `Error while inserting new unit: ${err}`);
```
huss: See comment above in this file.

src/server/routes/users.js
```diff
@@ -156,26 +157,31 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
 // Route for creating a new user.
 router.post('/create', adminAuthMiddleware('create a user.'), async (req, res) => {
 	// isEdit=false: id must not be present, since it's assigned by the DB on insert, and password is required.
-	if (!validateUsersParams(req.body, false).valid) {
-		res.status(HTTP_CODES.BAD_REQUEST).json({ message: 'Invalid params' });
+	const validatorResult = validateUsersParams(req.body, false);
```
huss, line 158: See other comment. The changes here also overlap the other PR. Thoughts?

src/server/routes/users.js
```diff
@@ -156,26 +157,31 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
+		failure(res, HTTP_CODES.BAD_REQUEST, `Got request to insert user with invalid user data. Error(s): ${validatorResult.errors}`);
 	} else {
 		try {
 			const { username, password, role, note } = req.body;
 			const conn = getConnection();
 			// Check if user already exists
 			const currentUser = await User.getByUsername(username, conn);
 			if (currentUser !== null) {
-				res.status(HTTP_CODES.BAD_REQUEST).send({ message: `user ${username} already exists so cannot create` });
+				//res.status(HTTP_CODES.BAD_REQUEST).send({ message: `user ${username} already exists so cannot create` });
```
huss: See other comment. The changes here also overlap the other PR. Thoughts?

Also, commented out line may not be for debugging.

src/server/routes/users.js
```diff
@@ -156,26 +157,31 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
+				failure(res, HTTP_CODES.BAD_REQUEST, `user ${username} already exists so cannot create`);
 			} else {
 				const hashedPassword = await bcrypt.hash(password, 10);
 				const user = new User(undefined, username, hashedPassword, role, note);
 				await user.insert(conn);
-				res.sendStatus(HTTP_CODES.OK);
+				//res.sendStatus(HTTP_CODES.OK);
```
huss: Commented out line. See other comment. The changes here also overlap the other PR but seem the same.

src/server/routes/users.js
```diff
@@ -156,26 +157,31 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
+				success(res);
 			}
 		} catch (error) {
 			// Log the error internally and return a generic response
 			log.error(`Error while performing POST request to create user: ${error}`, error);
-			res.status(HTTP_CODES.INTERNAL_SERVER_ERROR).send({ message: 'Internal Server Error' });
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, `Error while performing POST request to create user: ${error}`);
```
huss: See other comment. The changes here also overlap the other PR.  To simplify comments, this applies to all the remaining changes in this file. Thoughts?

huss (commented):
@aduques Thank you for another contribution. I need some more time to look over all the proposed changes. I did compare this to another recent PR #1688 and put in some comments about that. I hope I correctly did all the ones that I should have. I wanted you to be able to see them now to think it over. Please let me know any thoughts.

aduques:
@huss I am writing to update that I've seen the recent comments. I am anticipating that I will be spending some time comparing my changes with the changes made in #1688 as there are overlaps in the exact messages used in either the failure() or success() calls. From what I am seeing, the changes include detailed messages that may be generally similar to the ones that I have included in this PR. The reason for this is I believe we both took the existing error/success message and had it called in failure()/success(). The differences in our implementation appear to be that the implementations in #1688 had stored the messages into a message variable to clean up the actual failure()/success() calls. This alone can easily be merged with my changes without any issues. I will most likely merge the messaging style done in #1688 as it is very similar to what I included and it would be best for the entire codebase that there is a uniform messaging style for success() and failure() which #1688 has achieved.

There are some logic (such as the if-else statements in meters.js catch), that needs to be carefully looked at before merging. I believe that I will have no issue with merging the failure()/success() additions from #1688 to mine, but I need to test each message output again when I do so. I need to make sure that the merged changes from #1688 does not revert the fixes for undefined and generic error messages that this issue uniquely solves.

Therefore, I would appreciate guidance on how I should proceed with merging and testing the changes in #1688. Should I go through the changed files in #1688, copy paste the changes onto my branch, and test the changes. Or is the procedure more extensive, where I need to copy the working branch of #1688 onto mine and resolve the merge conflicts in that way. I would appreciate insight on the proper practice as this would be a great learning opportunity for me in the case I need to work with other active issues/pull requests.

huss:
Thank you for your thoughts and looking at the other PR. First, your overall plan/idea seems good to me. I also liked the standardization of dealing with messages in the other PR along with taking care of doing the logging so it always happens. Second, I don't have a precise answer on how you can move forward because I'm not an expert and I think different people do it differently. I don't care how as long as it meets the objectives.

Having said that, the copy/paste method can lead to merge issues and/or wipe the git line history for those lines when all the work is combined, esp. if great care is not taken. If you want to accept all the changes in the other PR then you can fetch/merge that branch/file into your branch where you always use their version for those lines. If done correctly, I think, it will leave the git line history as the other person having done it along with your other changes in the file as yours.  Let me add that if the entire file will be replaced with the PR one then you can revert your changes and leave it for when that PR is merged. Now, if you have overlapping changes where parts of each need to be merged then it gets trickier. Then, as far as I know, you need to manually decide what part of each is retained in the final version. This might cause merge issues for the other person (or you) depending on details but it might all be fine (others probably know better than I).

I'm not sure this is complete/perfect so I'm happy to have your thoughts/questions. Know I'm also here to help as you desire.

aduques:
@huss I have merged all of the changes made in #1688 and have resolved merge conflicts. As suspected, most of my changes have been replaced with the ones in #1688 to ensure that there is a consistent implementation in messaging across the codebase. I have tried to retain as much of my unique changes such as if-else conditions in the catch statements, sending specific param messages, and replacing generic "Invalid Params" messages.

While the primary overlap was the contents of the messages in the success() and failure() calls, there is a conflict in implementations that I would like to discuss. The implementations do not cause any errors, but it conflicts with the plan of wanting to create a standard function to handle message logs and validation across the routes. The reason for this is that there has been a change to have the success() and failure() take in strings with a message wrapper (and new Error() for error messages), rather than just a raw string parameter. The reason that this concerns me is because I have made the changes for failure to take in raw strings to handle the "error.data vs error.data.message" inconsistency problem.

For example, the messages in users.js don't work well with the changes. When using "error.data," the message will output [object Object] because it is not specifically getting the .message property of the data. Therefore, an immediate fix is to have error.data.message, but this would be one case where it is different from other routes. I plan to follow up with this message with updated outputs of each error message to visualize the current state of these messages.

If we were to proceed with the standard function, I would like an opinion on the design choice going forward. I would definitely like to prioritize #1688 as extensive work has been done to get a consistent implementation across the codebase.

My current TODO:
- Using the current merged code, test the error messages on localhost
- Take note of which routes are using .message and show [object Object]
- Adapt any leftover messages that are raw strings to match the design of #1688 (use message wrapper or new Error())

src/server/routes/conversions.js
aduques: Yes, I definitely agree. The changes made in #1688 should take priority as extensive work has been done to have consistent implementations across all the routes. When merging the work from #1688, their change has overwritten this one.

src/server/routes/groups.js
aduques: Noting that the merge has replaced this with the consistent implementation from #1688.

src/server/routes/meters.js
aduques: This has been updated to consider the changes from the other PR.

      } catch (err) {
			      if (err.toString().includes('duplicate key value violates unique constraint')) {
				      failure(res, HTTP_CODES.BAD_REQUEST, err, `Meter name "${req.body.name}" already exists`, LogLevel.SILENT);
			      } else if (err.toString().includes('violates check constraint')) {
				      failure(res, HTTP_CODES.BAD_REQUEST, err, `Invalid meter data: ${err.toString()}`, LogLevel.SILENT);
			      } else {
				      const detail = err['detail'] ? ` with detail "${err['detail']}"` : '';
				      failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while editing a meter${detail}: ${err.message}`, { cause: err }));
			      }
		      }

src/server/routes/units.js
aduques: This has been updated to consider the changes from the other PR.

      } catch (err) {
			      if (err.toString().includes('duplicate key value violates unique constraint')) {
				      failure(res, HTTP_CODES.BAD_REQUEST, err, `Unit name "${req.body.name}" already exists`, LogLevel.SILENT);
			      } else if (err.toString().includes('violates check constraint')) {
				      failure(res, HTTP_CODES.BAD_REQUEST, err, `Invalid unit data: ${err.toString()}`, LogLevel.SILENT);
			      } else {
				      failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Failed to update unit: ${err.message}`, { cause: err }));
			      }
		      }

src/server/routes/units.js
aduques: This has been updated to consider the changes from the other PR.

		      } catch (err) {
			      if (err.toString().includes('duplicate key value violates unique constraint')) {
				      failure(res, HTTP_CODES.BAD_REQUEST, err, `Unit name "${req.body.name}" already exists`, LogLevel.SILENT);
			      } else if (err.toString().includes('violates check constraint')) {
				      failure(res, HTTP_CODES.BAD_REQUEST, err, `Invalid unit data: ${err.toString()}`, LogLevel.SILENT);
			      } else {
				      failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while inserting new unit: ${err.message}`, { cause: err }));
			      }
		      }

src/server/routes/users.js
aduques: I have made the following changes to match with the other PR.

      const validatorResult = validateUsersParams(req.body, false);
      
	      if (!validatorResult.valid) {
		      const message = `Got request to insert user with invalid user data. Error(s): ${validatorResult.errors}`;
		      failure(res, HTTP_CODES.BAD_REQUEST, message, { message });

src/server/routes/users.js
aduques: I have made the following changes to match with the other PR.

			      if (currentUser !== null) {
				      failure(res, HTTP_CODES.BAD_REQUEST, null, { message: `user ${username} already exists so cannot create` });

The commented out line was handled.

src/server/routes/users.js
aduques: I have addressed the commented out line.

src/server/routes/users.js
aduques: I have merged the changes from the other PR and similar lines were also merged and replaced.

      } catch (error) {
			      failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing POST request to create user: ${error.message}`, { cause: error }));
		      }

The difference was primarily that the error message was passed as a new Error() object rather than a raw string. I gave my thoughts on this change below, but overall, I did not find any issues with the changes made. I need to review the outputs of these messages again to ensure that the merged lines are working. The changes to the routes and parameter changes for failure()/success() affect the planning for the standard function.

huss:
@aduques It will be about a week before I can review this. Thank you for the update.

src/server/routes/logs.js
```diff
@@ -54,40 +55,36 @@ router.post('/info', adminAuthMiddleware('create info log'), async (req, res) =>
 	const validationResult = validate(req.body, validLog);
 	if (validationResult.valid) {
 		log.info(req.body.message);
-		res.sendStatus(HTTP_CODES.OK);
+		success(res);
 	} else {
-		log.error('invalid input from client logger');
-		res.sendStatus(HTTP_CODES.BAD_REQUEST);
+		failure(res, HTTP_CODES.BAD_REQUEST, 'invalid input from client logger');
```
huss, line 60: I realize this is what the message was but maybe it should mirror the other parameter failure ones. It would also be good to add the route (/info here). This applies to similar ones below.

src/server/routes/meters.js
```diff
@@ -334,13 +334,17 @@ router.post('/edit', adminAuthMiddleware('edit meters'), async (req, res) => {
...
-			// TODO This is not using the success function since it needs to return values.
-			// At some point we probably should fuse the success and returning values.
 			// Need to format since some properties have different names than come from DB.
-			res.json(formatMeterForResponse(meter, true));
+			success(res, formatMeterForResponse(meter, true));
 		} catch (err) {
-			log.error(`Error while editing a meter with detail "${err['detail']}"`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, err.toString() + ' with detail ' + err['detail']);
+			if (err.toString().includes('duplicate key value violates unique constraint')) {
+				failure(res, HTTP_CODES.BAD_REQUEST, err, `Meter name "${req.body.name}" already exists`, LogLevel.SILENT);
```
huss, line 341: Not your change but I think making the message be: `` `Failure likely because meter name "${req.body.name}" already exists` ``. I'm suggesting this since it could also fail for another reason but only if bug or bad request.

Also, I think this pattern came from the other PR but per a [comment there](https://github.com/OpenEnergyDashboard/OED/pull/1688#discussion_r3918034273) I think it should not be silent so all issues are logged. This applies to other ones in this file.

src/server/routes/meters.js
```diff
@@ -405,13 +409,17 @@ router.post('/addMeter', adminAuthMiddleware('add meter'), async (req, res) => {
...
-			// TODO This is not using the success function since it needs to return values.
-			// At some point we probably should fuse the success and returning values.
 			// Need to format since some properties have different names than come from DB.
-			res.json(formatMeterForResponse(newMeter, true));
+			success(res, formatMeterForResponse(newMeter, true));
 		} catch (err) {
-			log.error(`Error while inserting new meter with detail "${err['detail']}"`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, err.toString() + ' with detail ' + err['detail']);
+			if (err.toString().includes('duplicate key value violates unique constraint')) {
+				failure(res, HTTP_CODES.BAD_REQUEST, err, `Meter name "${req.body.name}" already exists`, LogLevel.SILENT);
```
huss, line 416: See proposed string change above.

src/server/routes/meters.js
```diff
@@ -405,13 +409,17 @@ router.post('/addMeter', adminAuthMiddleware('add meter'), async (req, res) => {
...
+			} else if (err.toString().includes('violates check constraint')) {
+				failure(res, HTTP_CODES.BAD_REQUEST, err, `Invalid meter data: ${err.toString()}`, LogLevel.SILENT);
+			} else {
+				const detail = err['detail'] ? ` with detail "${err['detail']}"` : '';
```
huss, line 420: Curious why this is part of the message but above and in most cases it is the full message.

src/server/routes/meters.js
```diff
@@ -363,7 +367,7 @@ router.post('/addMeter', adminAuthMiddleware('add meter'), async (req, res) => {
 		(req.body.minDate && !isValidIsoDateTime(req.body.minDate)) ||
 		(req.body.maxDate && !isValidIsoDateTime(req.body.maxDate))
 	) {
-		failure(res, HTTP_CODES.BAD_REQUEST, 'invalid date/time format');
+		failure(res, HTTP_CODES.BAD_REQUEST, null, 'invalid date/time format');
```
huss, line 370: Per a [comment](https://github.com/OpenEnergyDashboard/OED/pull/1688#discussion_r3919142682) on another PR, I think this should not be null and log it.

src/server/routes/units.js
```diff
@@ -157,8 +155,13 @@ router.post('/edit', adminAuthMiddleware('edit units'), async (req, res) => {
 			await unit.update(conn);
 			success(res, 'Successfully edited unit');
 		} catch (err) {
-			log.error(`Failed to update unit: ${err}`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, 'Unable to update unit');
+			if (err.toString().includes('duplicate key value violates unique constraint')) {
```
huss, line 158: See comment in meters.js about not doing silent & changing duplicate name msg.

src/server/routes/units.js
```diff
@@ -196,8 +199,13 @@ router.post('/addUnit', adminAuthMiddleware('add units'), async (req, res) => {
 			});
 			success(res, 'Unit created successfully');
 		} catch (err) {
-			log.error(`Error while inserting new unit: ${err}`, err);
-			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, `Error while inserting new unit: ${err}`);
+			if (err.toString().includes('duplicate key value violates unique constraint')) {
```
huss, line 202: See comment above on silent & name msg.

src/server/routes/users.js
```diff
@@ -101,11 +101,11 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
 		maxLength: TOKEN_MAX_LENGTH
 	};
 	if (!validate(token, validParams).valid) {
-		res.status(HTTP_CODES.FORBIDDEN).json({ message: 'No token provided or JSON was invalid.' });
+		failure(res, HTTP_CODES.FORBIDDEN, null, { message: 'No token provided or JSON was invalid.' });
```
huss, line 104: Per other comments, use a logging msg not null. Applies to other similar uses in file.

src/server/routes/users.js
```diff
@@ -101,11 +101,11 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
...
 	} else if (token) {
 		jwt.verify(token, secretToken, async (err, decoded) => {
 			if (err) {
-				res.status(HTTP_CODES.UNAUTHORIZED).json({ message: 'Failed to authenticate token.' });
+				failure(res, HTTP_CODES.UNAUTHORIZED, err, { message: 'Failed to authenticate token.' }, LogLevel.SILENT);
```
huss, line 108: Per other comment, don't make it silent. Applies to other similar uses in file.

src/server/routes/users.js
```diff
@@ -116,12 +116,12 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
 							role: userProfile.role
 						});
 				} catch (error) {
-					res.status(HTTP_CODES.UNAUTHORIZED).json({ message: 'User does not exist in database.' });
+					failure(res, HTTP_CODES.UNAUTHORIZED, error, { message: 'User does not exist in database.' }, LogLevel.SILENT);
```
huss, line 119: Not your change but maybe make the message: ``User unknown.``

src/server/routes/users.js
```diff
@@ -197,10 +200,8 @@ router.post('/edit', adminAuthMiddleware('edit a user'), async (req, res) => {
 				const numberOfAdmins = await User.getNumberOfAdmins(conn);
 				if (numberOfAdmins < 2) {
 					const errorMessage = 'There must be at least one admin remaining to avoid lockout!';
-					log.error(errorMessage);
-					return res.status(HTTP_CODES.BAD_REQUEST).json({
-						message: errorMessage,
-					});
+					failure(res, HTTP_CODES.BAD_REQUEST, errorMessage, { message: errorMessage });
+					return;
```
huss, line 204: Per [comment](https://github.com/OpenEnergyDashboard/OED/pull/1688#discussion_r3917900254) on other PR, need to decide if return or not. You are welcome to give thoughts.

src/server/routes/users.js
```diff
@@ -140,42 +140,42 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
 				const user = new User(undefined, username, hashedPassword, role, note);
 				await user.insert(conn);
-				res.sendStatus(HTTP_CODES.OK);
+				success(res);
 			}
 		} catch (error) {
-			// Log the error internally and return a generic response
-			log.error(`Error while performing POST request to create user: ${error}`, error);
-			res.status(HTTP_CODES.INTERNAL_SERVER_ERROR).send({ message: 'Internal Server Error' });
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing POST request to create user: ${error.message}`, { cause: error }));
```
huss, line 178: When I put ``throw new Error('test');`` right before the ``user.insert`` above I saw an undefined error on the admin web page. When I look at the web console for the response I see the expect error message. This may relate to your comment about the new failure() causing issues.

src/server/routes/users.js
```diff
@@ -245,23 +244,24 @@ router.post('/delete', adminAuthMiddleware('delete a user'), async (req, res) =>
...
+		failure(res, HTTP_CODES.BAD_REQUEST, message, { message });
 	} else {
 		try {
 			const conn = getConnection();
 			const { username } = req.body;
 			const id = req.decoded.data;
 			const user = await User.getByID(id, conn);
 			if (user.username === username) {// Admins cannot delete themselves
-				res.sendStatus(HTTP_CODES.BAD_REQUEST);
+				failure(res, HTTP_CODES.BAD_REQUEST);
```
huss, line 258: When I try to delete the current admin I see an undefined in the web page and nothing web console response. It should have a better log msg & one back to the admin since it is only a bad request.

src/server/routes/users.js
```diff
@@ -245,23 +244,24 @@ router.post('/delete', adminAuthMiddleware('delete a user'), async (req, res) =>
...
 			} else {
 				await User.deleteUser(username, conn);
-				res.sendStatus(HTTP_CODES.OK);
+				success(res);
 			}
 		} catch (error) {
-			log.error('Error while performing delete user request', error);
-			res.sendStatus(HTTP_CODES.INTERNAL_SERVER_ERROR);
+			failure(res, HTTP_CODES.INTERNAL_SERVER_ERROR, new Error(`Error while performing delete user request: ${error.message}`, { cause: error }));
```
huss, line 264: When I put throw new Error('test'); right before the deleteUser above I saw an undefined error on the admin web page. When I look at the web console for the response I see the expect error message. May relate to your comment on issues with failure.

src/server/routes/users.js
```diff
@@ -197,10 +200,8 @@ router.post('/edit', adminAuthMiddleware('edit a user'), async (req, res) => {
 				const numberOfAdmins = await User.getNumberOfAdmins(conn);
 				if (numberOfAdmins < 2) {
```
huss, line 201: During testing I found a bug here that already existed before this work. This line needs to be ``if (numberOfAdmins.count < 2) {``. Could you fix it?

src/server/routes/users.js
```diff
@@ -197,10 +200,8 @@ router.post('/edit', adminAuthMiddleware('edit a user'), async (req, res) => {
...
 					const errorMessage = 'There must be at least one admin remaining to avoid lockout!';
-					log.error(errorMessage);
-					return res.status(HTTP_CODES.BAD_REQUEST).json({
-						message: errorMessage,
-					});
+					failure(res, HTTP_CODES.BAD_REQUEST, errorMessage, { message: errorMessage });
```
huss, line 203: Since this is a bad request with a known issue, the message should go back to the user. I think I saw it did not and may relate to your comment on failure issues.

huss:
> If we were to proceed with the standard function, I would like an opinion on the design choice going forward.

I think we should discuss this and figure out the best, standard way to do this.

huss:
From the PR description:
> meters.js /edit and /addMeter routes could not be converted to use response.js's success() on the success path, since the client depends on the DB-assigned data (e.g. id) being returned in the response body, and success() does not currently support returning structured data. These routes continue to use res.json() on success, per the pre-existing TODO comment in the file.

From a comment:
> **NOTE:** success() cannot be used currently in meters.js. There is an existing TODO note that states that /addMeter and /edit are expected to return a json object. Therefore, simply replacing it with success() does not work. For now, I have skipped over applying success() until I look into it more. If this is something that I can do within this pull request, I can apply the change, but if it requires more files to be changed, it may be better to open this specifically as a new issue to keep it separate from this one. Thoughts on this is appreciated.

I think these two are related. I tried to figure them out but I see a lot of success/failure usage in src/server/routes/meters.js. I do see two res.json in the get routes for meter info but not others. Could you help me understand this better?

huss (changes requested):
@aduques Thank you for another contribution to OED. First, per a message on the OED Discord server developer channel, I'm sorry this review took so long.

For the record, the following files only seem to have changes from PR #1688: package.json, package-lock.json, log.js, app.js, responseParamsTest.js, verification.js, unitReadings.js, response.js, readings.js, preferences.js, obvius.js, maps.js, logs.js, loginLogout.js, groups.js, conversions.js, compareReadings.js, ciks.js, baseline.js. Please let me know if I overlooked something.

Overall, this works well. I have made some comments on specific parts of the code to consider. Here are ones that did not fit well with specific code comments:

- I searched for "failure(res, HTTP_CODES.BAD_REQUEST" and found that a number of places send back the failure without any message. I have not gone through them (I did comment on a few but then stopped) but I think they should, in general, since it is not a server error. Some probably relate to the param checks that will be fixed more generally but some do not. I found that some of the others were causing undesirable web page messages. I welcome your thoughts on this.
- I did some page testing but stopped due to the overall issues found. I thought it best to wait until the best course of action was decided for the existing comments before doing more. You also have TODO items in a comment that overlap what I was starting to do so I'll wait for your analysis.

src/server/routes/logs.js
aduques: Updated the message to mirror the other parameter failure ones and included route information.

      const message = 'Invalid input from client logger for POST /info';
      failure(res, HTTP_CODES.BAD_REQUEST, message, message, LogLevel.WARN);

The same has been added for the /warn and /error

src/server/routes/meters.js
aduques: The message has been updated to note assume duplication and the error is no longer silent when logged.

      failure(res, HTTP_CODES.BAD_REQUEST, err, `Failure likely because meter name "${req.body.name}" already exists`);

src/server/routes/meters.js
aduques: The same behavior is used in /addMeter

      failure(res, HTTP_CODES.BAD_REQUEST, err, `Failure likely because meter name "${req.body.name}" already exists`);

src/server/routes/meters.js
aduques: The err.toString() and the LogLevel.SILENT originally came from the merged changes. I didn’t want to make any changes to the incoming log messages at the time as it would conflict with the other PR if changes to the log messages have already been suggested. I updated this section to match the existing logging behavior where the detailed error is logged while the client receives the ‘Invalid meter data’ message.

      failure(res, HTTP_CODES.BAD_REQUEST, err, 'Invalid meter data');

src/server/routes/meters.js
aduques: The failure no longer passes a null and gets logged.

      // For meter creation
      failure(res, HTTP_CODES.BAD_REQUEST, new Error('Invalid date/time format for meter creation request'), 'Invalid date/time format');
      
      // For meter editing
      failure(res, HTTP_CODES.BAD_REQUEST, new Error('Invalid date/time format for meter edit request'), 'Invalid date/time format');

src/server/routes/units.js
aduques: Silent logging was removed and the message no longer indicates duplicate as the main/only issue raised.

      failure(res, HTTP_CODES.BAD_REQUEST, err, `Failure likely because unit name "${req.body.name}" already exists`);

src/server/routes/units.js
aduques: Silent logging was removed and the message no longer indicates duplicate as the main/only issue raised.

      failure(res, HTTP_CODES.BAD_REQUEST, err, `Failure likely because unit name "${req.body.name}" already exists`);

src/server/routes/users.js
aduques: The logging messages no longer use null

      // For invalid token input
      failure(res, HTTP_CODES.FORBIDDEN, new Error('No token provided or JSON was invalid.'), 'No token provided or JSON was invalid.');
      
      // For no-token
      failure(res, HTTP_CODES.FORBIDDEN, new Error('No token provided.'), 'No token provided.');
      
      // Create/Edit branches
      failure(res, HTTP_CODES.BAD_REQUEST, message, message);
      failure(res, HTTP_CODES.BAD_REQUEST, message, message, LogLevel.WARN);

src/server/routes/users.js
aduques: The message no longer uses LogLevel.SILENT

      failure(res, HTTP_CODES.UNAUTHORIZED, err, 'Failed to authenticate token.');

The message gets logged and the client receives the message.

src/server/routes/users.js
aduques: The message has been updated to use ‘User unknown’ instead.

      failure(res, HTTP_CODES.UNAUTHORIZED, error, 'User unknown.');

src/server/routes/users.js
aduques: I think that return should be used to prevent the route from continuing with the update after a response has already been sent. The only reason I see that raising the warning log is sufficient is if it is important to not have the route interrupted at any point. In other words, if the route already handles an interruption (or can be updated to handle this), then return should be okay. In addition, in the old implementation, whenever an error message was thrown and logged, it was returned, which also halted the route. I see as not returning changing the behavior and could require additional debugging that will take more time to identify possible problems.

src/server/routes/users.js
aduques: If I recall correctly, with the current setup of failure() it is intended to pass in a String or object with message property, so when it receives error.data, it is not the String it is expecting.

For now, I propose a temporary solution where both String and the error.data can be accepted without seeing an Undefined error.

      (typeof error.data === 'string' ? error.data : error.data?.message) ?? 'Unknown error'
      
      showErrorNotification(
	      translate('users.failed.to.create.user') +
	      translate('users.successfully.edit.user.username') + userDetails.username + ') ' +
	      ((typeof error.data === 'string' ? error.data : error.data?.message) ?? 'Unknown error')
      );

This should prevent the undefined error, but I would appreciate having a second look at these changes.

src/server/routes/users.js
aduques: I have updated the log message.

      if (user.username === username) {
	      const message = 'Administrators cannot delete themselves.';
	      failure(res, HTTP_CODES.BAD_REQUEST, new Error(message), message);
	      return;
      }

src/server/routes/users.js
aduques: https://github.com/OpenEnergyDashboard/OED/pull/1689#discussion_r4023471914 I have made a temporary solution as I have done for this change. It accepts either a String response or object with a message property, but a singular Object type should be discussed later for consistency.

src/server/routes/users.js
aduques: I have made the requested changes.

      if (numberOfAdmins.count < 2) {

src/server/routes/users.js
aduques: The message was updated to be a string and is returned after failure()

      const errorMessage = 'There must be at least one admin remaining to avoid lockout!';
      failure(res, HTTP_CODES.BAD_REQUEST, errorMessage, errorMessage);
      return;

aduques:
@huss I have finished addressing the recent feedback. I anticipate that the changes made here could affect how the changes will be done in the other open PR as it involves updating the failure() message. As I have mentioned in the related comments, I introduced a temporary solution that accepts both Strings and objects with the message property to workaround the undefined error. I would like help testing this change to see if it displays the message as expected.

As for discussing the additional method that I proposed in this PR, I still need time to review the details of it, but was able to remember parts of the idea as I worked through the feedback. I still believe that it should only be implemented after this issue and the other open issue is closed. However, this implementation plan can still be discussed once I review more.

Screenshot 2026-09-16 at 12 40 17 AM

Screenshot 2026-09-16 at 12 40 50 AM

I tested the following by adding 

        throw new Error('test create-user failure');
        throw new Error('test delete-user failure');

right before their respective requests.

huss:
@aduques I have this queued to be looked at again. I don't have any thoughts yet on what you said but will reconsider after I look the changes over. FYI, there is a new merge conflict that probably resulted from a recently merged PR.

src/server/routes/users.js
huss: This worked correctly for me. I do wonder about pushing the choice for the string to the client. Why not have the server, where the type of error is known and created, do this fix so the current failure accepts it correctly? If so, one could create a utility function on the server that does this and is used where needed. Now, if the standard failure function could do this logic and centralized there then that would be great. However, I'm concerned it will sometimes leak undesirable information to the user but maybe that would not happen.

src/client/app/components/admin/users/CreateUserModalComponent.tsx
```diff
@@ -130,7 +130,7 @@ export default function CreateUserModal() {
 				showErrorNotification(
 					translate('users.failed.to.create.user') +
 					translate('users.successfully.edit.user.username') + userDetails.username + ') ' +
-					error.data.message);
+					(typeof error.data === 'string' ? error.data : error.data?.message) ?? 'Unknown error');
```
huss, line 133: I looked at this more. I finally realized that I think it can be simpler. If every route is now using failure, then it always returns a string (``responseBody`` that defaults to an empty string if all else does not work). I think it is good that it is consistent and removes the issue of it not existing and not a string. I tried the suggestiion and it seems to work for me. Also applies to edit. I think all routes should be doing this and the handful I checked do but maybe all should be checked.

```suggestion
error.data
```

src/server/routes/users.js
```diff
@@ -140,42 +140,43 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message);
 	} else {
 		try {
 			const { username, password, role, note } = req.body;
 			const conn = getConnection();
 			// Check if user already exists
 			const currentUser = await User.getByUsername(username, conn);
 			if (currentUser !== null) {
-				res.status(HTTP_CODES.BAD_REQUEST).send({ message: `user ${username} already exists so cannot create` });
+				failure(res, HTTP_CODES.BAD_REQUEST, new Error(`User ${username} already exists so cannot create.`), `User ${username} already exists so cannot create.`);
```
huss, line 170: We chatted about this and the suggested change is to:

```suggestion
				const msg = `The user '${username}' seems to already exist so cannot use for a new user.`;
				console.log('msg: ', msg);
				failure(res, HTTP_CODES.BAD_REQUEST, msg, msg);

```

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -209,7 +209,7 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
 				showErrorNotification(
 					translate('users.failed.to.edit.user') +
 					translate('users.successfully.edit.user.username') + userDetails.username + ') ' +
-					error.data.message
+					((typeof error.data === 'string' ? error.data : error.data?.message) ?? 'Unknown error')
```
huss, line 212: See create about this.

src/client/app/components/admin/users/EditUserModalComponent.tsx
```diff
@@ -228,7 +228,7 @@ export default function EditUserModalComponent(props: EditUserModalComponentProp
 				showErrorNotification(
 					translate('users.failed.to.delete.user') +
 					translate('users.successfully.edit.user.username') + props.user.username + ') ' +
-					error.data.message
+					((typeof error.data === 'string' ? error.data : error.data?.message) ?? 'Unknown error')
```
huss, line 231: Same as other in file.

src/server/routes/users.js
```diff
@@ -101,11 +101,11 @@ router.get('/token', optionalAuthMiddleware, async (req, res) => {
 		maxLength: TOKEN_MAX_LENGTH
 	};
 	if (!validate(token, validParams).valid) {
-		res.status(HTTP_CODES.FORBIDDEN).json({ message: 'No token provided or JSON was invalid.' });
+		failure(res, HTTP_CODES.FORBIDDEN, new Error('No token provided or JSON was invalid.'), 'No token provided or JSON was invalid.');
```
huss, line 104: For this one and the others in this file, I think it is better and more maintainable if the string is set to a const above and reused for both arguments.

src/server/routes/users.js
```diff
@@ -140,42 +140,43 @@ router.get('/:user_id', adminAuthMiddleware('get one user'), async (req, res) =>
...
 	} else {
 		try {
 			const { username, password, role, note } = req.body;
 			const conn = getConnection();
 			// Check if user already exists
 			const currentUser = await User.getByUsername(username, conn);
 			if (currentUser !== null) {
-				res.status(HTTP_CODES.BAD_REQUEST).send({ message: `user ${username} already exists so cannot create` });
+				failure(res, HTTP_CODES.BAD_REQUEST, new Error(`User ${username} already exists so cannot create.`), `User ${username} already exists so cannot create.`);
+				return;
```
huss, line 171: When we chatted we discussed that the code will systematically do a return after success/failure so just noting this is in line with that but not yet checking them all.

src/server/routes/users.js
huss: This is fine but I'm not sure it is temporary given our conversation. Let me know if I'm missing something.

src/server/routes/logs.js
```diff
@@ -54,40 +55,39 @@ router.post('/info', adminAuthMiddleware('create info log'), async (req, res) =>
 	const validationResult = validate(req.body, validLog);
 	if (validationResult.valid) {
 		log.info(req.body.message);
-		res.sendStatus(HTTP_CODES.OK);
+		success(res);
 	} else {
-		log.error('invalid input from client logger');
-		res.sendStatus(HTTP_CODES.BAD_REQUEST);
+		const message = 'Invalid input from client logger for POST /info';
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message, LogLevel.WARN);
```
huss, line 61: I would think it might be better to use LogLevel.ERROR for this one and all related ones. It really should never happen and is likely a bug and a lost log msg.

src/server/routes/logs.js
```diff
@@ -54,40 +55,39 @@ router.post('/info', adminAuthMiddleware('create info log'), async (req, res) =>
...
 	const validationResult = validate(req.body, validLog);
 	if (validationResult.valid) {
 		log.error(req.body.message);
-		res.sendStatus(HTTP_CODES.OK);
+		success(res);
 	} else {
-		log.error('invalid input from client logger');
-		res.sendStatus(HTTP_CODES.BAD_REQUEST);
+		const message = 'Invalid input from client logger for POST /error';
+		failure(res, HTTP_CODES.BAD_REQUEST, message, message, LogLevel.WARN);
```
huss: I would think it might be better to use LogLevel.ERROR for this one.

src/server/routes/meters.js
huss: I'm not sure what I put in the other PR but I'm uncertain about a SILENT log level since it suppresses it. What about WARN for these two?

src/server/routes/users.js
huss: I'm resolving since the ``{ message }`` actually does not have the braces.

src/server/routes/users.js
huss: This is fine but should have safe message too.

src/server/routes/users.js
huss: Agreed per another comment.

huss (changes requested):
@aduques Thank you for the updated code to address my previous comments. I have tried to resolve all comments that are now done. I also made some new ones to consider. There is also now a merge conflict. Finally, I still have not done a careful review of each route via testing given someone methods may change. I'll probably do a full re-review once it is stable. Please let me know if anything is not clear or you have thoughts/questions.

PR #1691, opened 2026-08-04
Add user logout routes to test files

### LINKED ISSUE #1683: Add user logouts to authentication tests

The goal of this issue is to make authentication tests consistently pair authentication setup and cleanup by logging out users after the test is finished. This can be done by using the `/api/loginLogout/logout` endpoint. When a test logs in, it should also clean up by logging out, so the authentication lifecycle follows a consistent pattern (login → perform the test → logout). Currently, many authentication tests perform the login and testing steps without a final logout.

For reference, files that have currently implemented a user logout are `src/server/test/routes/sessionInvalidationTest.js` and `src/server/test/routes/readingsParamsTest.js`.

Determine if each user test logs in only once or every testcase.
    • If a user logs in one time (e.g. inside a before hook), then have the user only log out one time after all the tests are finished. (e.g. src/server/test/routes/logsRouteTests.js)
    • If a user logs in for every test (e.g. inside a beforeEach hook or loop), then have the user log out after every test using the same token that was used for that test request. (e.g. src/server/test/web/login.js)
Do not generate a new token during cleanup; always reuse the token obtained during the test setup.

Listed below are all of the files I've found that appear to match these patterns.

Files that log in as multiple users in a loop without logging out: 
```
src/server/test/web/usersTest.js
src/server/test/web/groups.js
src/server/test/web/maps.js
src/server/test/web/meters.js
src/server/test/web/preferencesTest.js
```

Files that log in as admin testUser without logging out:
```
src/server/test/routes/logsRouteTests.js
src/server/test/routes/unitsRouteTests.js
src/server/test/web/login.js

(The files below overlap with the first criteria)

src/server/test/web/usersTest.js
src/server/test/web/groups.js
src/server/test/web/maps.js
src/server/test/web/meters.js
src/server/test/web/preferencesTest.js
```

Not all of the files listed above would benefit much from logging out the user (for example, tests that only verify login behavior), but it could still be implemented if the goal is to keep testing consistent. However, they could reasonably be left out.

Files with multiple logins or where a user is provided to multiple tests without logging out should have a logout implemented.

### Additional context

For tests with multiple users, instead of logging out all the users at the very end, it might be better to do one user at a time, log in, do the test & log out for each one.

Nespina24:
# Description

This PR cleans up tests suites by adding in a logout route to test users who login. When a test logs in, it should also clean up by logging out, so the authentication lifecycle follows a consistent pattern (login → perform the test → logout). This PR addresses the issue in the following files:

```
src/server/test/routes/logsRouteTests.js
src/server/test/routes/unitsRouteTests.js
src/server/test/web/login.js
src/server/test/web/usersTest.js
src/server/test/web/groups.js
src/server/test/web/maps.js
src/server/test/web/meters.js
src/server/test/web/preferencesTest.js
```

Fixes #1683

## Type of change

## Checklist

## Limitations

No limitations.


Nespina24:
I drafted this PR for now since I am not currently finished implementing all the logouts, but will undraft it once I am finished.

I just had a few clarifying questions about some of the tests.

src/server/test/web/meters.js
```diff
@@ -162,11 +162,20 @@ mocha.describe('meters API', () => {
 				// Since this .before is in the middle of tests, it should not have issues as
 				// documented in usersTest.js.
 				mocha.before(async () => {
+					// login
 					let res = await chai.request(app).post('/api/loginLogout/login')
 						.send({ username: testUser.username, password: testUser.password });
```
Nespina24: If this is supposed to test both ADMIN and CSV roles, then it looks like it doesn't access the CSV role. The test looks like it simply runs ADMIN both times instead of depending on the role being looped. If this behavior is unintended, then I can make the appropriate changes.

src/server/test/web/maps.js
```diff
@@ -51,6 +51,9 @@ function expectMapsToBeEquivalent(maps, length) {
 mocha.describe('maps API', () => {
 	mocha.beforeEach(async () => {
 		// TODO Why is there an empty body here?
+
+		// It seems to work fine if I comment it out.
+		// Deleting the beforeEach() hook should be possible.
 	});
```
Nespina24: I could delete this, but I wanted to know if there were any plans to use this hook that I'm unaware of.

src/server/test/web/meters.js
huss: What I see seems very similar to the next test for the other roles:

```js
mocha.describe('Admin role & CSV role:', () => {
		for (const role in User.role) {
			if (User.role[role] !== User.role.OBVIUS && User.role[role] !== User.role.EXPORT) {
```

I see what you mean and you seem correct. I think it should login with the desired role and use that token. Thanks for offering to make that change.

Unlike the next test which should test everyone else, this one should only test these roles. Thus, I think it would be much better to check if the role is admin or csv rather than not the others. Then it will still work if more roles are added. If you agree, could you also do that?

src/server/test/web/maps.js
huss: If you agree then I say remove it. It has been there a long time and it would not be hard to add one back if needed.

src/server/test/web/meters.js
Nespina24: I adjusted the test to login with the same role, but I noticed that it would only pass if I changed the before hook to a beforeEach hook in `src/server/test/web/meters.js` on line 162. Otherwise, I kept getting the following errors for both admin and csv:

```
 21 passing (35s)
  2 failing

  1) meters API
       Admin role & CSV role:
         should return all meters for ADMIN:
     AssertionError: expected { id: 1, name: null, url: null, …(31) } to have property 'name' of 'Meter 1', but got null
      at expectMetersToBeEquivalent (src/server/test/web/meters.js:54:26)
      at Context.<anonymous> (src/server/test/web/meters.js:213:6)

  2) meters API
       Admin role & CSV role:
         should return all meters for CSV:
     AssertionError: expected { id: 1, name: null, url: null, …(31) } to have property 'name' of 'Meter 1', but got null
      at expectMetersToBeEquivalent (src/server/test/web/meters.js:54:26)
      at Context.<anonymous> (src/server/test/web/meters.js:213:6)

```

I tested and found that the same error message is given for export and obvius roles for the same test, regardless of whether before or beforeEach is used (This might be the expected outcome, but shouldn't be for admin or csv). I also found that there's no difference between using an after or afterEach hook on line 176, but I kept it as afterEach to be consistent. I'm unsure of why I would strictly need a beforeEach for this test, but I plan on leaving it like this unless you have any input.

Nespina24:
~~At the time I'm writing this, the build checks are still running well past what it normally does. It gave the error "Error: write EPIPE" which I never saw while running on my local machine. Can you force stop it?~~

~~Edit: I just decided to close and reopen for now. Hopefully it won't give the same error.~~

Nespina24:
@huss I realized that the last 2 checks on my pull request never ran at all. Do you know how to fix this issue? My guess would be that it has to do with manually stopping a previous check.

src/server/test/routes/unitsRouteTests.js
```diff
@@ -5,15 +5,23 @@
...
 mocha.describe('Units Route', () => {
 	let token;
-
 	mocha.before(async () => {
+		// login
 		const res = await chai.request(app).post('/api/loginLogout/login')
 			.send({ username: testUser.username, password: testUser.password });
 		token = res.body.token;
 	});
+	mocha.after(async () => {
```
Nespina24, line 18: ~For this test, I wanted to apply the potential change I noted in the issue description:~

> ~For tests with multiple users, instead of logging out all the users at the very end, it might be better to do one user at a time, log in, do the test & log out for each one.~

~Even though it doesn't have multiple users, I thought this would be a good place to login/logout for every test since it runs for a long time. However, when I tried changing to beforeEach/afterEach, the checks had continually failed.~

~I'm ok with leaving it like this, but I think this should be addressed if the root cause of the failures can be found.~

Edit: this was the wrong file

huss:
I'll run the tests once I review the PR. If checks and tests pass on your machine then they should on GitHub. Hopefully I can get to this later in the week.

src/server/test/routes/logsRouteTests.js
```diff
@@ -24,8 +23,12 @@ mocha.describe('Log Routes', () => {
 		currentLogToDb = log.logToDb;
 		log.logToDb = true;
 	});
-
 	mocha.after(async () => {
```
Nespina24, line 26: For this test, I wanted to apply the potential change I noted in the issue description:

> For tests with multiple users, instead of logging out all the users at the very end, it might be better to do one user at a time, log in, do the test & log out for each one.


Even though it doesn't have multiple users, I thought this would be a good place to login/logout for every test since it runs for a long time. However, when I tried changing to beforeEach/afterEach, the checks had continually failed.

I'm ok with leaving it like this, but I think this should be addressed if the root cause of the failures can be found.

huss:
@Nespina24 I appreciate you keeping this up-to-date. Per my message on Discord, it will take me some time to get to review this.

src/server/test/routes/logsRouteTests.js
Nespina24: I'm actually considering removing the changes to this file entirely. After implementing this, I've noticed that even though it passed the most recent test, the node checks have a chance of failing different tests for a reason I'm unaware of. If it gets merged, I'm afraid that other people will start seeing similar issues.

huss: Curious. Do you want to wait until I review the PR to see if I have any ideas? One test should not, in a perfect world, impact another test. It makes me wonder if something else is wrong that really should be tracked down. If it cannot be tracked down during this work then I suspect a new issue should be opened to look at this further. Thoughts?

Nespina24: I'm okay with opening a new issue if it doesn't get resolved after a code review. If that's the case, then I think I should comment out the changes with a TODO describing the problem.

src/server/test/web/groups.js
```diff
@@ -99,10 +99,19 @@ mocha.describe('groups API', () => {
 		// Since this .before is in the middle of tests, it should not have issues as
 		// documented in usersTest.js.
 		mocha.before(async () => {
+			// login
 			let res = await chai.request(app).post('/api/loginLogout/login')
 				.send({ username: testUser.username, password: testUser.password });
 			token = res.body.token;
 		});
+		mocha.after(async () => {
```
huss, line 107: I tried to analyze these tests more. Some within this mocha.describe use this token as admin and some use a role specific one. However, I think the role specific one still has access to the pre-created admin token. I don't think the tests are making a mistake but it does partly circumvent your work to always log out. What I think might be a good idea is to separate the two types of tests so the ones that do role logins are not part of a mocha block that logs in as an admin with a before. I believe you did this idea in other file(s). What do you think?

src/server/test/web/login.js
```diff
@@ -16,6 +16,12 @@ mocha.describe('login API', () => {
...
 		expect(res).to.be.json;
 		expect(res.body).to.have.property('token');
+		let token = res.body.token;
+
+		if (token) {
+			await chai.request(app).post('/api/loginLogout/logout')
+				.set('token', token);
+		}
 	});
 	mocha.it('returns 401 for a wrong password', async () => {
```
huss, line 26: I think the setup is okay but I wanted to note that if this test ever failed then the provided token would still be around. Given it is a local variable I don't see how that is an issue so I say leave it as is.

src/server/test/web/login.js
```diff
@@ -16,6 +16,12 @@ mocha.describe('login API', () => {
 		expect(res).to.have.status(HTTP_CODES.OK);
 		expect(res).to.be.json;
 		expect(res.body).to.have.property('token');
+		let token = res.body.token;
+
+		if (token) {
```
huss, line 21: I'm noting that this token is scoped to this test so it should be impossible to use in another test. However, it seems good practice to logout to be safe. For this and the other tests changed I am saying that this change should stay.

src/server/test/web/maps.js
```diff
@@ -80,10 +77,19 @@ mocha.describe('maps API', () => {
 		// Since this .before is in the middle of tests, it should not have issues as
 		// documented in usersTest.js.
 		mocha.before(async () => {
```
huss, line 79: I'm simply noting that unlike another comment, this seems safe since the test using the admin role is separated from the ones using general roles. However, it might change based on another comment.

src/server/test/web/preferencesTest.js
```diff
@@ -13,24 +13,39 @@ const { HTTP_CODES } = require('../../util/httpCodes');
...
+				defaultChartToRender: 'line',
+				defaultBarStacking: true,
+				defaultLanguage: 'en',
+				defaultWarningFileSize: 5,
+				defaultFileSizeLimit: 25,
+				defaultAreaNormalization: true,
+				defaultAreaUnit: 'meters',
+				defaultMeterReadingFrequency: '1:13:17'
+			};
+			mocha.describe('Admin role: ', () => {
```
huss, line 27: I'm simply noting that the changes here seem to properly separate the admin from non-admin tests so there is no issue.

src/server/test/web/meters.js
huss: I was thinking about something similar while reviewing the code in other files. It seems that tests generally do a before/after if a single test and beforeEach/afterEach if multiple tests. For multiple tests it is generally needed since the different tests use different users. For a single test it should not matter. Now if someone adds more tests to the currently single test with before/after then it might pick up a token that is not intended. Thus, what do you think about consistently using beforeEach/afterEach as long as it does not cause issues? This would invalidate the token so if someone tries it would fail the token validation on the route. I'm only proposing for the tests you touched with login/out.

To be clear, this is an example of a test with the Each usage but only has a single test so, as you state, it should not matter.

huss:
> I'll run the tests once I review the PR. If checks and tests pass on your machine then they should on GitHub. Hopefully I can get to this later in the week.

Just noting that all tests run fine on my machine and are passing on GitHub.

src/server/test/routes/logsRouteTests.js
huss: I changed the tests in this file to use beforeEach/afterEach. I ran the single file of tests and the complete suite of tests multiple times but have yet to see it fail. Thus, I wanted to ask if you are still seeing this. If so, how often does it happen and is there anything special in your setup?

260921 update: This may be related to another comment on before/afterEach. I'm not adding any info but leaving it as part of that.

huss (changes requested):
@Nespina24 First, I want to apologize for how long this review took (see OED Discord server msg in developer channel).

Thank you for another contribution. Review and testing found it works well. I appreciate you fixing some other, minor issues while doing this work. I made a few comments to consider but nothing that seems substantial to me. Please let me know if anything is not clear or you have questions/thoughts.

src/server/test/routes/logsRouteTests.js
Nespina24: Thanks for responding. I also don't see it on my local machine, but I was referring to the GitHub testing suite which seems to either successfully pass or give an error on different files, such as `src/server/test/web/csvPipelineTest.js` or `src/server/test/routes/logsRouteTests.js` even though it ran mostly the same code.

Recently it seems to be working on GitHub, but it could potentially be an issue if other people pull in these changes and it returns a phantom failure.

I'll change the code to beforeEach/afterEach for now and see if the problem persists.

src/server/test/web/meters.js
Nespina24: > Thus, what do you think about consistently using beforeEach/afterEach as long as it does not cause issues?

I pushed a change to all the relevant before/after hooks to consistently use beforeEach/afterEach hooks, including those that only have a single test. One file that I don't think it's necessary for is `src/server/test/web/usersTest.js` since all the tests use the same admin token, but I included the change since it seemed to have no problems. I don't mind reverting the change though.

src/server/test/web/groups.js
```diff
@@ -95,149 +95,199 @@ mocha.describe('groups API', () => {
...
-						res = await chai.request(app).post('/api/groups/create').set('token', currentToken);
-						expect(res).to.have.status(expectedResponseStatus);
-					});
+					// Response status code should be 403 if improper role.
+					if (User.role[role] !== User.role.ADMIN) {
+						let token;
+						mocha.beforeEach(async () => {
+							// insert test user
+							const conn = testDB.getConnection();
+							const password = 'password';
```
huss, line 157: You didn't do this and it matters less now than before since you logout each time. However, I am pushing for each login to use a password that is specific to the user. For example, ``'password${role}'`` and this is similar to the login name/email.

Also, I'm realizing that a lot of tests do login/out. I was thinking about a utility function to do this and then having all tests use it. What do you think? It would be an issue of the future.

src/server/test/web/groups.js
```diff
@@ -95,149 +95,199 @@ mocha.describe('groups API', () => {
...
+							const hashedPassword = await bcrypt.hash(password, 10);
+							const unauthorizedUser = new User(undefined, `${role}@example.com`, hashedPassword, User.role[role]);
+							await unauthorizedUser.insert(conn);
+							unauthorizedUser.password = password;
 
+							// login
+							let res = await chai.request(app).post('/api/loginLogout/login')
```
huss, line 164: I think this can be const.

src/server/test/web/groups.js
```diff
@@ -95,149 +95,199 @@ mocha.describe('groups API', () => {
...
+			});
+			mocha.describe('Non-Admin roles', () => {
+				for (const role of Object.keys(User.role)) {
+					// Response status code should be 403 if improper role.
+					if (User.role[role] !== User.role.ADMIN) {
+						let token;
+						mocha.beforeEach(async () => {
+							// insert test user
+							const conn = testDB.getConnection();
+							const password = 'password';
```
huss, line 265: See above on password.

src/server/test/web/meters.js
```diff
@@ -157,16 +157,31 @@ mocha.describe('meters API', () => {
...
-				// Since this .before is in the middle of tests, it should not have issues as
-				// documented in usersTest.js.
-				mocha.before(async () => {
-					let res = await chai.request(app).post('/api/loginLogout/login')
-						.send({ username: testUser.username, password: testUser.password });
-					token = res.body.token;
-				});
-				mocha.it('returns all meters', async () => {
+				mocha.beforeEach(async () => {
+                    // insert test user
```
huss, line 163: space indenting should be tabs.

src/server/test/web/meters.js
huss: The change seems to work fine. However, I was unsure why it made such a difference so I tried to analyze it more. Unfortunately, I was unable to resolve it but have figured out there is something subtle (at least to me) going on here. Not being very knowledgeable here may mean I'm missing something. With debug statements I determined that the before/afterEach runs for both user roles that pass the if but it does it before both set of tests. Thus, the before is running for all tests for all user roles. You also see the loop starts and finishes all roles before it does any actual tests. It uses the correct token for each test but gets tokens for all roles that pass the if. I did some research and that is how mocha works. I thought not using a for loop but forEach by looping over the role Object types would fix it but it did not (at least so far). I have my code that I used for testing and some web references if someone is willing to look at this. This pattern is used in several places in OED testing so knowing exactly what is happening would be nice even though the tests pass.

src/server/test/routes/logsRouteTests.js
huss: I made another comment about a potential issues I could not resolve. If this is not figured out soon then I propose to create an issue that documents the code you add (probably via a commit id) and discussed what is happening. Then it can be addressed later since there is a good chance it is not a direct result of this work. How does that sound?

huss (changes requested):
@Nespina24 Thank you for updating the code. I have mostly resolved the previous comments and the code is running well. There are three remaining (may be related) that may become future issues. If you have thoughts on them I would love to hear them. I also made a few new comments but I think they should be straightforward. I think this is almost ready to merge. The larger issue(s) will probably either be new issue(s) or resolved here in a reasonable amount of time. Please let me know if anything is not clear or if you have thoughts/ideas.

PR #1561, opened 2025-12-11
Enhance suffix unit conversion deletion with dependency checks

hmhngx:
# Description

Enhances suffix unit conversion deletion to properly cascade delete related units/conversions when no dependencies exist, with improved UI warnings and consistency with unit edit/delete routes. Builds on the deletion branch of PR #1470.

**Key improvements:**
- Fixed async race condition in `removeAdditionalConversionsAndUnits()` using Promise.all
- Added dependency validation before deletion (checks meters/groups usage)
- Enhanced bidirectional conversion handling
- Added row-level locking for concurrent operation safety
- Improved UI warnings showing affected meters/groups
- Added DB layer methods for cascading deletions
- Enhanced simulation to account for suffix unit cascades
- Added depth limiting to prevent infinite loops in nested suffix chains
- Added orphan detection and performance optimizations

Fixes #1448 

## Type of change

- [ ] Note merging this changes the database configuration.
- [ ] This change requires a documentation update

## Checklist

- [x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [x] I have removed text in ( ) from the issue request
- [x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations

None. All edge cases from PR #1470 have been addressed including bidirectional conversions, nested suffix chains, concurrent deletions, and orphan prevention.

PR #1693, opened 2026-08-05
Improve suffix unit conversion deletion with cascade delete and UI warnings

BunnyBea83:
# Description

This PR solves OED's opporations for deleting conversions involving suffix units. Upon deletion of a suffix unit related conversion, OED created units and related conversions will be cascade deleted. UI warnings are inplace to notify the user of units and conversions that will be deleted, and groups/meters that will be affected by the deletion. This PR builds off of PR #[1561](https://github.com/OpenEnergyDashboard/OED/pull/1561) and PR #[1470](https://github.com/OpenEnergyDashboard/OED/pull/1470).

**Key improvements:**
- Resolved merge conflicts and implemented work from PR #1561 
- Suffix unit validations now account for Suffix Inputs and Suffix Analyzed during the simulateDeletion and Deletion phase of conversion deletion
- Deletion of a Suffix Unit conversion will cascade delete OED created units in relation to the parent conversion
- Cascade deletion within DB includes removing a units association to groups/meters, deleting conversions referencing the unit, and deleting units
- UI warnings include affected meters/groups, units, and conversions related to the deletion of a suffix unit conversion
- removeAdditionalConversionsAndUnits no longer 'hides' removed units. Removed units are completely deleted from the OED DB
   - Resolves discussion in issue#1448, when a unit is renamed the former unit is completely removed instead of 'hidden'
 - Included passing unit tests to confirm function of cascade deletion within suffixUnitDeletionTests.js

Partly Addresses #[1448](https://github.com/OpenEnergyDashboard/OED/issues/1448)

## Type of change

## Checklist

## Limitations

Multi-hop-generated suffix units are not yet covered by this cascade delete since it likely needs a
design decision before implementation.

This PR resolves cascade deletion for the direct/single-hop case (deleting a
conversion from a suffix root, or from an auto-created child, correctly
cleans up what it should). It does **not** yet handle units auto-created via
**multi-hop paths** (e.g. suffix unit → new auto-created unit→ further conversion). Since OED always attaches such units
directly to the root rather than the intermediate unit, deleting the
intermediate conversion currently leaves these orphaned. Possible implementations could include either a re-walk of valid paths after delete, or tracking explicit provenance on auto-created units, both of
which are a larger design change than this PR's scope. Flagging for
follow-up discussion with maintainers.

Before Deletion of multi-hop path 'metric Ton of TestUnit -> gallon'

image

After Deletion of multi-hop path

image

Deletion of TestUnit -> kg removes all OED created units

image


huss:
@BunnyBea83 Thank you for this PR. It will take me about a week to get to review this. The description seems to be fine from my quick review.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const Meter = require('../../models/Meter');
```
huss, line 5: Meter, Group & Conversion requires are showing as unused.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Checks if a unit has dependencies that prevent safe deletion.
+ * @param {number} unitId The unit ID to check
+ * @param {*} conn Database connection (can be a transaction)
+ * @returns {Promise<Object>} Object with hasDependencies boolean and details
+ */
+async function checkUnitDependencies(unitId, conn) {
+	const [meters, groups, conversions] = await Promise.all([
+		// Check if any meters use this unit (as unitId or default_graphic_unit)
+		conn.any(`
```
huss, line 20: OED typically uses the model class that calls sqlFile to execute queries. The goal is to separate the SQL from the JS code. I'm wondering if there is a reason not to do that here.

src/server/routes/conversions.js
```diff
@@ -6,11 +6,14 @@ const express = require('express');
...
 const { getConnection } = require('../db');
 const Conversion = require('../models/Conversion');
+const Unit = require('../models/Unit');
+const { removeAdditionalConversionsAndUnits } = require('../services/graph/handleSuffixUnits');
 const { success, failure } = require('./response');
 const { HTTP_CODES } = require('../util/httpCodes');
 const validate = require('jsonschema').validate;
 
 const { simulateDeleteConversion } = require('../services/conversionSimulation');
+const { checkUnitDependencies, getUnitDependencyDetails } = require('../services/graph/checkUnitDependencies');
```
huss, line 16: I don't see where the first item is used.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
...
+const Group = require('../../models/Group');
+const Conversion = require('../../models/Conversion');
+const Unit = require('../../models/Unit');
+const { log } = require('../../log');
+
+/**
+ * Checks if a unit has dependencies that prevent safe deletion.
+ * @param {number} unitId The unit ID to check
+ * @param {*} conn Database connection (can be a transaction)
+ * @returns {Promise<Object>} Object with hasDependencies boolean and details
```
huss, line 15: It might be nice to specify what details are returned.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -6,6 +6,9 @@ const Unit = require('../../models/Unit');
 const { pathConversion } = require('./pathConversion');
 const Conversion = require('../../models/Conversion');
 const { getAllPaths } = require('./createConversionGraph');
+const { log } = require('../../log');
+const { canSafelyHideSuffixUnit } = require('./checkUnitDependencies');
```
huss, line 10: I don't see where this is used.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Checks if a suffix-type unit (created by OED) can be safely hidden.
+ * These units should only be hidden if they're not used by meters/groups.
+ * Conversions are expected and will be cleaned up separately.
+ * @param {number} unitId The suffix unit ID to check
+ * @param {*} conn Database connection (can be a transaction)
+ * @returns {Promise<boolean>} True if safe to hide
+ */
+async function canSafelyHideSuffixUnit(unitId, conn) {
```
huss, line 57: I cannot find any usage of this function and am unsure what is intended.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
...
+			await conn.none('UPDATE meters SET default_graphic_unit = NULL WHERE id = $1', [meter.id]);
+		}
+	}
+
+	// Clear any group using this as its default graphic unit.
+	for (const group of deps.groups) {
+		await conn.none('UPDATE groups SET default_graphic_unit = NULL WHERE id = $1', [group.id]);
+	}
+
+	// Delete any other conversions still referencing this unit.
```
huss, line 117: I'm unclear if it is always safe to delete these. I think you may have a limitation in the PR about orphaning units that might apply here but it isn't trivial to tell. If this is correct then, at a minimum, at TODO should be in code as a comment.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
...
+	for (const conv of deps.conversions) {
+		log.info(`Deleting conversion ${conv.source_id}->${conv.destination_id} to allow deletion of unit ${unitId}.`);
+		await conn.none('DELETE FROM conversions WHERE source_id = $1 AND destination_id = $2', [conv.source_id, conv.destination_id]);
+	}
+
+	await conn.none('DELETE FROM cik WHERE source_id = $1 OR destination_id = $1', [unitId]);
```
huss, line 123: I have mixed feelings about removing single items from cik rather than doing a full recalculation. I worry about issues being created and the step away from the standard functions to do this. For example, would this be correct if you orphan a unit? I think a full refresh of cik should be done, esp. since this isn't a common operation and that does not take too long.

Note for future look: Should check if client properly waits for all this to complete.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -157,25 +160,73 @@ async function handleSuffixUnits(graph, conn) {
...
-	conversions.forEach(async (conversion) => {
-		const destinationUnit = await Unit.getById(conversion.destinationId, conn);
-		// The units that OED adds are suffix unit.
-		if (destinationUnit.typeOfUnit === Unit.unitType.SUFFIX) {
-			// Delete the conversion.
-			await Conversion.delete(conversion.sourceId, conversion.destinationId, conn);
-			// Hide the destination unit.
-			destinationUnit.displayable = Unit.displayableType.NONE;
-			await destinationUnit.update(conn);
+const MAX_SUFFIX_CLEANUP_DEPTH = 10;
```
huss, line 169: First, nothing should come between the JSDoc and the function declaration. Second, I think this might be best as a const inside the function if it is only used there. Also note that src/client/app/components/groups/EditGroupModalComponent.tsx has a depth limit (hardcoded) of 50. Should these be consistent; I'm not sure they are directly related? If so, then a global const in an appropriate place might be good.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -157,25 +160,73 @@ async function handleSuffixUnits(graph, conn) {
...
+
+				// Recursively clean up this unit's related conversions/units
+				// This handles nested suffix chains (A -> B -> C)
+				await removeAdditionalConversionsAndUnits(otherUnit, conn, depth + 1);
+
+				// Delete the auto-created unit (dependency checks + cik cleanup handled inside)
+				await deleteUnitSafely(otherUnitId, conn);
+
+			}
+		} catch (err) {
```
huss, line 223: It is good you caught this error. Do you have any idea if OED will be left in a valid state if an exception occurs, esp. for your new removeAdditionalConversionsAndUnits() function?

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -157,25 +160,73 @@ async function handleSuffixUnits(graph, conn) {
...
+	if (depth > MAX_SUFFIX_CLEANUP_DEPTH) {
+		log.error(`Max depth (${MAX_SUFFIX_CLEANUP_DEPTH}) reached cleaning up suffix unit ${suffixUnit.id}. Possible circular dependency.`);
+		throw new Error(`Suffix unit cleanup depth limit exceeded for unit ${suffixUnit.id}`);
+	}
+	// Get all conversions involving this suffix unit (as source, destination, or bidirectional)
+	const allConversions = await Conversion.getAll(conn);
+	const relatedConversions = allConversions.filter((conversion) =>
+	conversion.sourceId === suffixUnit.id ||
+	(conversion.bidirectional && conversion.destinationId === suffixUnit.id)
+);
```
huss, line 181: The formatting is off. Maybe do a document format if in VSC to fix any that exist.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -157,25 +160,73 @@ async function handleSuffixUnits(graph, conn) {
...
+	// Process all related conversions in parallel with proper async handling
+	await Promise.all(relatedConversions.map(async (conversion) => {
+		try {
+			// Determine which unit is the suffix unit and which is the destination
+			const isSource = conversion.sourceId === suffixUnit.id;
+			const otherUnitId = isSource ? conversion.destinationId : conversion.sourceId;
+			
+			const otherUnit = await Unit.getById(otherUnitId, conn);
+			
+			// Validate that the other unit exists
```
huss, line 192: Could this comment expand on why this happen? Also, I'm wondering if you didn't do these in parallel and you recalculated relatedConversions each time could this be avoided? If so, would that be desirable or would this found condition be missed and not reported (and desired).

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -157,25 +160,73 @@ async function handleSuffixUnits(graph, conn) {
...
+			if (!otherUnit) {
+				log.warn(`Unit ${otherUnitId} not found when cleaning up suffix unit ${suffixUnit.id}. Conversion ${conversion.sourceId}->${conversion.destinationId} may be orphaned.`);
+				return;
+			}
+			
+			// The units that OED adds are suffix units (typeOfUnit === SUFFIX)
+			if (otherUnit.typeOfUnit === Unit.unitType.SUFFIX) {
+				// Always delete the conversion
+				await Conversion.delete(conversion.sourceId, conversion.destinationId, conn);
```
huss, line 201: I wonder if orphaned units could be determined here by first seeing if the conversion.destinationId unit has other connections - not sure. Below you try to clean them up by deleting.

src/server/services/graph/handleSuffixUnits.js
```diff
@@ -157,25 +160,73 @@ async function handleSuffixUnits(graph, conn) {
 /**
```
huss: This is an overall comment. The user should probably be informed of all units and conversions that will be removed by this process in advance and asked if that is okay (usual cancel/confirm as in other places). This would allow the admin to consider and issues such as orphaned units would be noted before any action is taken. This does take extra code/work but it is done elsewhere. Note that to be safe/faster I think another use of this idea sends back the list(s) of items and then resends that list list to the server if the admin agrees so it is guaranteed to only do those items even if an unlikely change happens in between.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -354,7 +474,7 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 				cancel = true;
 			}
 		}
-
+		// Route to appropriate modal based on any above set cancle condition
```
huss, line 477: cancle misspelled. Also, maybe say processing instead of model since OED uses model for something else internally.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -95,11 +95,18 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 		return source?.typeOfUnit === UnitType.meter;
 	};
 
+	// Determine whether the selected unit is a Suffix unit
+	// Including Suffix input with unit type Unit and Suffix is filled in
+	const isSuffixRelated = (unit? : UnitData): boolean =>
+		!!unit && (
+			unit.typeOfUnit === UnitType.suffix || !!unit.suffix?.trim()
```
huss, line 102: I appreciate why you want to trim the .suffix. However, the unit modals do not do that and check for an empty string (''). I think the methods should be the same. I'm open to modifying the check but it should be consistent. In the bigger picture, it seems that a utility function to do this that is shared would make sense. This made me think what would happen if unit was undefined and that would need cleaning up - not terrible if done here but probably never happens.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -95,11 +95,18 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 		return source?.typeOfUnit === UnitType.meter;
 	};
 
+	// Determine whether the selected unit is a Suffix unit
+	// Including Suffix input with unit type Unit and Suffix is filled in
+	const isSuffixRelated = (unit? : UnitData): boolean =>
```
huss, line 100: OED prefers braces even if there is one statement. It is also nice to have the return be explicit and that may be needed after the braces are added.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 					</div>
 				);
 			}
+			// Suffix unit to Suffix Unit deletion
+		} else if (
```
huss, line 239: Given the total length of this else if I think it would be better on 1 line rather than 2.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			const suffixUnit = isSuffixRelated(source) ? source : dest;
+			// Find all conversions involving this suffix unit (as source or destination)
+			const suffixUnitConversions = conversionDetails.filter(c =>
+				c.sourceId === suffixUnit.id ||
+				(c.bidirectional && c.destinationId === suffixUnit.id)
+			);
+
+			// Find OED-created suffix-type units that would be cascade deleted
+			const suffixTypeUnitsToDelete = suffixUnitConversions
+				.map(c => {
```
huss, line 250: I think I understand how this works (checks all conversions that are suffix related, finds the side/unit that is not the suffix and then only takes the ones that are suffix related) but a clear comment would help me/others.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+					const otherId = c.sourceId === suffixUnit.id ? c.destinationId : c.sourceId;
+					return unitDataById[otherId];
+				})
+				.filter(isSuffixRelated);
+
+			// Get unique conversions that would be deleted
+			const conversionsToDelete = suffixUnitConversions.filter(c => {
```
huss, line 257: As above, a clear comment would help. I was also wondering if this result could be used to filter to get the above result as they seem very similar - not certain.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+				const otherId = c.sourceId === suffixUnit.id ? c.destinationId : c.sourceId;
+				const otherUnit = unitDataById[otherId];
+				return isSuffixRelated(otherUnit);
+			});
+
+			// Check for meters/groups using affected suffix units
+			const affectedSuffixUnitIds = new Set(suffixTypeUnitsToDelete.map(u => u.id));
+			affectedSuffixUnitIds.add(suffixUnit.id); // Also check the main suffix unit
```
huss, line 265: OED wants comments above the line of code and not at the end.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+
+			// Check meters using these units (as unitId or defaultGraphicUnit)
+			const affectedMetersList = Object.values(meterDataById).filter(meter =>
+				affectedSuffixUnitIds.has(meter.unitId) ||
+				(meter.defaultGraphicUnit !== null && affectedSuffixUnitIds.has(meter.defaultGraphicUnit))
```
huss, line 270: I'm not certain but I think .has returns false on null if null was not added to the set. Could you check if this condition is really needed?

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			);
+
+			// Check groups using these units (as defaultGraphicUnit)
+			const affectedGroupsList = Object.values(groupDataById).filter(group =>
+				group.defaultGraphicUnit !== null && affectedSuffixUnitIds.has(group.defaultGraphicUnit)
```
huss, line 275: See meter check comment.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -342,10 +460,12 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 							);
 						}
 					});
+					// Save twhich meters.groups need their default unit cleared
```
huss, line 463: twitch misspelled

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+				group.defaultGraphicUnit !== null && affectedSuffixUnitIds.has(group.defaultGraphicUnit)
+			);
+
+			// Display dependency warnings, affected meters and groups if any
+			if (affectedMetersList.length > 0 || affectedGroupsList.length > 0) {
+				msgElements.push(
+					<div key="suffix-dependencies-warning">
+						<span className="bold">{translate('conversion.delete.suffix.dependencies.warning')}</span>
+						{affectedMetersList.length > 0 && (
+							<div style={{ marginTop: '8px' }}>
```
huss, line 284: Two thoughts. First, OED tries to avoid lots of styling unless it is really needed. Is this essential? If you do want to do this then it probably should be in the central css styling files, esp. since it is reused.

src/client/app/translations/data.ts
```diff
@@ -92,6 +92,11 @@ const LocaleTranslationData = {
 		"conversion.delete.restricted": "This conversion cannot be deleted until the significant consequences of doing this are addressed.",
 		"conversion.delete.simulation.error": "There was an error simulating the deletion of this conversion. Please try again later.",
 		"conversion.delete.suffix.disable": "Deleting this suffix conversion will disable use of suffix",
+		"conversion.delete.suffix.conversions.to.delete": "The following conversions will also be deleted",
+		"conversion.delete.suffix.units.to.delete": "The following units will also be deleted",
+		"conversion.delete.suffix.dependencies.warning": "Warning: Some meters or groups depend on units that will be affected",
```
huss, line 97: This translation key along with conversion.delete.suffix.meters.affected, conversion.delete.suffix.groups.affected, conversion.delete.suffix.units.to.delete only seem to be in en. All the ones you added should be carefully checked.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			// List the conversions that will be cascade deleted
+			if (conversionsToDelete.length > 0) {
+				msgElements.push(
+					<div key="suffix-conversions-to-delete">
+						<span className="bold">{translate('conversion.delete.suffix.conversions.to.delete')}:</span>
+						<ul>
+							{conversionsToDelete.map((c, idx) => {
+								const s = unitDataById[c.sourceId]?.name || unitDataById[c.sourceId]?.identifier || c.sourceId;
+								const d = unitDataById[c.destinationId]?.name || unitDataById[c.destinationId]?.identifier || c.destinationId;
+								const bidirectional = c.bidirectional ? ' ↔ ' : ' → ';
```
huss, line 330: Please see src/client/app/utils/conversionArrow.ts for standard way to do this.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -223,10 +235,114 @@ export default function EditConversionModalComponent(props: EditConversionModalC
...
+			}
+
+			// List the conversions that will be cascade deleted
+			if (conversionsToDelete.length > 0) {
+				msgElements.push(
+					<div key="suffix-conversions-to-delete">
+						<span className="bold">{translate('conversion.delete.suffix.conversions.to.delete')}:</span>
+						<ul>
+							{conversionsToDelete.map((c, idx) => {
+								const s = unitDataById[c.sourceId]?.name || unitDataById[c.sourceId]?.identifier || c.sourceId;
```
huss, line 328: I don't like single letter variable names outside specific cases. Could these be refactored to be better names?

Also, I'm unclear on why you have the or (||) chain for the item identifier. I don't think names can be empty (DB constraint) so I don't think it is necessary. Also, OED never shows an id to the user (unless OED makes a mistake).

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -249,6 +365,7 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 					cancel = true;
 				} else {
 					// Group meters by lostUnits
+					// Display all meters that are loosing the same set of units
```
huss, line 368: loosing -> losing

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -277,6 +394,7 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 					});
 
 					// Group non-orphaned groups by lostUnits
+					// Display all groups loosing the same set of units
```
huss, line 397: see loosing above

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -365,6 +485,7 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 			setDeleteConfirmationMessage(msgElements);
 			handleCancelModalOpen();
 		} else {
+			// Confirm deletion if no blcoking issues
```
huss, line 488: blcoking

src/server/test/db/suffixUnitDeletionTests.js
```diff
@@ -0,0 +1,264 @@
...
+		const child = new Unit(undefined, 'Child of test-suffix', 'Child', Unit.unitRepresentType.QUANTITY,
+			1000, Unit.unitType.SUFFIX, '', Unit.displayableType.ALL, true, 'suffix unit created by OED');
+		await child.insert(conn);
+
+		const conversion = new Conversion(parent.id, child.id, false, 1, 0, 'test conversion');
+		await conversion.insert(conn);
+
+		await removeAdditionalConversionsAndUnits(parent, conn);
+
+		const childRow = await conn.oneOrNone('SELECT * FROM units WHERE id = $1', [child.id]);
```
huss, line 32: src/server/models/Unit.js has getById that probably can and should be used. Also, conversion has getBySourceDestination. Similar in other tests.

src/server/test/db/suffixUnitDeletionTests.js
```diff
@@ -0,0 +1,264 @@
...
+		expect(childRow).to.equal(null);
+
+		const conversionAfter = await Conversion.getBySourceDestination(parent.id, child.id, conn);
+		expect(conversionAfter).to.equal(null);
+	});
+
+	mocha.it('clears a meter\'s unit_id instead of blocking deletion', async () => {
+		const conn = testDB.getConnection();
+
+		const parent = new Unit(undefined, 'Parent2', 'Parent2', Unit.unitRepresentType.QUANTITY,
```
huss, line 42: FYI. The DB is reset before each test so this renaming of items for each test really isn't needed unless I'm missing something.

src/server/test/db/suffixUnitDeletionTests.js
```diff
@@ -0,0 +1,264 @@
...
+const { mocha, expect, testDB } = require('../common');
+const Unit = require('../../models/Unit');
+const Meter = require('../../models/Meter');
+const Group = require('../../models/Group');
+const Point = require('../../models/Point');
+const Conversion = require('../../models/Conversion');
+const gps = new Point(90, 45);
+const { removeAdditionalConversionsAndUnits } = require('../../services/graph/handleSuffixUnits');
+
+mocha.describe('Suffix unit cascade deletion', () => {
```
huss, line 14: Just noting that src/server/util/insertData.js has utility functions that allow for structured data to insert units, conversion, etc. into the DB. These are guaranteed to work in the future so are safer and probably would make these tests simpler to write/understand.

src/server/routes/units.js
```diff
@@ -219,20 +220,95 @@ router.post('/delete', adminAuthMiddleware('delete units'), async (req, res) =>
...
-			// Don't worry about checking if the unit already exists
-			// Just try to delete it to save the extra database call, since the database will return an error anyway if the row does not exist
-			await Unit.delete(unitId, conn);
+			const unit = await Unit.getById(req.body.id, conn);
+			if (!unit) {
+				failure(res, HTTP_CODES.NOT_FOUND, 'Unit not found');
+				return;
+			}
+			// Check for dependencies before deletion
+			const { checkUnitDependencies } = require('../services/graph/checkUnitDependencies');
```
huss, line 234: OED wants all requires at the top.

src/server/models/Conversion.js
```diff
@@ -96,6 +96,7 @@ class Conversion {
 			destination: destination
 		});
 	}
+
```
huss, line 99: Please remove the new blank line since not desired and the only change in this file.

src/server/routes/conversions.js
```diff
@@ -180,7 +183,85 @@ router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, re
...
+					log.warn(`Conversion deletion blocked due to dependencies: ${errorMsg}`);
+					failure(res, 400, errorMsg);
+					return;
+				}
+			}
+
+			// Perform all operations in a single transaction for atomicity
 			await conn.tx(async t => {
+				// Lock the units to prevent concurrent modifications
+				await t.one('SELECT * FROM units WHERE id = $1 FOR UPDATE', [sourceId]);
```
huss, line 236: Per another comment, this file (and units) has a number of raw SQL requests. This should probably be addressed.

src/server/routes/conversions.js
```diff
@@ -180,7 +183,85 @@ router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, re
 		const { sourceId, destinationId, meterIds = [], groupIds = [] } = req.body;
 		const conn = getConnection();
 		try {
+			// Get the source and destination units for the conversion (before transaction for validation)
```
huss, line 186: Could you help me understand why this is cleaning up items even after the simulation that I think happened before this request. Is it to be safe? I have not tried to figure this out so I appreciate your thoughts.

src/server/routes/conversions.js
```diff
@@ -192,10 +273,45 @@ router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, re
...
+							WHERE (c.source_id IN ($1, $2) OR c.destination_id IN ($1, $2))
+						))
+						AND NOT EXISTS (
+							SELECT 1 FROM conversions c 
+							WHERE (c.source_id = u.id OR c.destination_id = u.id)
+						)
+						LIMIT 100
+					`, [sourceId, destinationId]);
+					if (orphanedUnits.length > 0) {
+						log.warn(`Found ${orphanedUnits.length} potentially orphaned suffix units after cleanup: ${orphanedUnits.map(u => `${u.id} (${u.name})`).join(', ')}`);
```
huss, line 302: Would it be better to also return all message to the admin/user? What do you think?

src/server/routes/conversions.js
```diff
@@ -192,10 +273,45 @@ router.post('/delete', adminAuthMiddleware('delete conversions'), async (req, re
...
+						// Log metrics for monitoring
+						log.info(`Conversion deletion metrics: sourceId=${sourceId}, destId=${destinationId}, orphanedUnits=${orphanedUnits.length}`);
+					}
+				} catch (err) {
+					// Non-critical check, log but don't fail
+					log.warn(`Error checking for orphaned units: ${err}`);
```
huss, line 308: Why would these fail? Can you help me understand why it is not a more important issue?

src/server/routes/units.js
```diff
@@ -219,20 +220,95 @@ router.post('/delete', adminAuthMiddleware('delete units'), async (req, res) =>
...
-	} else {
+		return;
+	}
 		const conn = getConnection();
 		const unitId = req.body.id;
 		try {
-			// Don't worry about checking if the unit already exists
-			// Just try to delete it to save the extra database call, since the database will return an error anyway if the row does not exist
-			await Unit.delete(unitId, conn);
+			const unit = await Unit.getById(req.body.id, conn);
```
huss, line 228: See conversions.js as similar questions apply here.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -399,14 +521,16 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 		// Closes the warning modal
 		// Do not call the handler function because we do not want to open the parent modal
 		setShowDeleteConfirmationModal(false);
-
+		// Bundle the conversions, meters, and groups to delete
```
huss, line 524: First, what do you mean by "cleat"? Second, would it be possible/safe to use the returned list to from the simulate to the delete on the server? I do worry if it becomes out-of-date but it does mean the ones warned about are the ones acted on.

huss (changes requested):
@BunnyBea83 First, I want to apologize for how long this review took (see Discord msg on developer channel on OED server). I want to thank you for this contribution to OED. It is a larger effort that did many aspects (changes, tests, etc.) and esp. nice given this is your first contribution to OED.

I have tried to make a good first pass on this PR. Overall, it works well and the code is fine. I've made a number of comments for you to consider. Some are to get your thoughts and we can discuss those before changes are made, as appropriate. Given this and the scope of the PR, I am probably going to need to do another pass on aspects of the work and will delay extensive testing until it is closer to finalized (I'm assuming changes will be needed). This is normal on a PR of this size. Please let me know if anything is not clear or you have thoughts/ideas.

Below are some comments that seemed best to place here:

- The description says it  Partly Addresses issue 1448. Are there aspects that you feel are not completed in that issue if this PR is merged?

- I tried to find the old git commit record for the previous PRs but did not readily find it. Am I missing it? If not, I'm not sure you may have caused this. If it is gone then maybe the description should list the previous developers in an appropriate way so it is easy for others to see they worked on this previously. I can do this if you prefer and agree.

- I had an unusual suffix unit in my setup so I tried to delete it. The visualization shows:

Image

When I went to the conversion and tried to delete it it failed (note lack of failure reason):

Image

and during the process it also a gave warning about what it was going to do:

Image

The terminal where OED was started had these messages:
```text
web-1       | [INFO@2026-09-08T17:12:23.859+00:00] Suffix-related unit 43 is used in conversion deletion as source. Cleaning up related conversions and units.
web-1       | 
web-1       | [INFO@2026-09-08T17:12:23.864+00:00] Suffix-related unit 44 is used in conversion deletion as destination. Cleaning up related conversions and units.
web-1       | 
web-1       | [ERROR@2026-09-08T17:12:23.864+00:00] Error while deleting conversion and updating meters/groups: QueryResultError {
web-1       |     code: queryResultErrorCode.noData
web-1       |     message: "No data returned from the query."
web-1       |     received: 0
web-1       |     query: "SELECT * FROM units WHERE id = 44;"
web-1       | }
web-1       | Stacktrace: 
web-1       | QueryResultError: No data returned from the query.
web-1       |     at Query.<anonymous> (/usr/src/app/node_modules/pg-promise/lib/query.js:217:41)
web-1       |     at Query.handleReadyForQuery (/usr/src/app/node_modules/pg-promise/node_modules/pg/lib/query.js:137:14)
web-1       |     at Client._handleReadyForQuery (/usr/src/app/node_modules/pg-promise/node_modules/pg/lib/client.js:312:19)
web-1       |     at Connection.emit (node:events:514:28)
web-1       |     at /usr/src/app/node_modules/pg-promise/node_modules/pg/lib/connection.js:117:12
web-1       |     at Parser.parse (/usr/src/app/node_modules/pg-protocol/dist/parser.js:36:17)
web-1       |     at Socket.<anonymous> (/usr/src/app/node_modules/pg-protocol/dist/index.js:11:42)
web-1       |     at Socket.emit (node:events:514:28)
web-1       |     at addChunk (node:internal/streams/readable:324:12)
web-1       |     at readableAddChunk (node:internal/streams/readable:297:9)
```

The web console had an uncaught exception:

Image

I have not tried to trace this down. When I deleted the other suffix conversion (to 100 w bulb) it worked fine and cleaned up the other one too.
- Small point but could space be added before the final delete statement at the end of the conversion delete modal so it is separated. I might also make it bold to stand out more.

Image

- If I try to delete the standard OED test conversion of kg CO2 -> kg it has in the warning (meter one is similar):

Image

The admin is not used to seeing the connection to kg and Metric ton as these are artifacts of OED's processing of the suffix unit. This is seen in the unit visualization on the lower analyzed graphic. I don't know if it is easy to remove these or somehow label them/improve the message so it is clear for an admin. I'm happy to discuss possibilities.
  -  The delete seemed to work fine when done in this case. I'm noting that the link to kg CO2 for the meters stays and in the analyzed unit graphic it is shown as orphaned (as it is). This is a suffix unit and I welcome thoughts on if/how there might be another/better way to deal with it. At a minimum, the help system should note this if it stays.
- Also relating to deleting the standard OED test conversion of kg CO2 -> kg, It has:
```
Affected meters:
    "Electric Utility kWh in MTon CO₂"
```
A lot of other meters are listed below that have affected so I think it may be confusing to the admin. Is it possible to clarify how they differ in the text?

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: These will be removed as they are not used in the current codebase.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: Braces and an explicit return statement have been added.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: When manually creating a unit OED currently allows admins to define unit or a units suffix unit as " ". However, there appears to be no issues when I attempt to delete a " " unit or any conversions related to it. I agree that in a bigger picture a utility function would be useful for cases to check for undefined or empty unit labels. At this current point in time, my code has been adjusted to reflect consistency with the unit modal checks.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: Else if has been reduced to one line.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: Clear commentry on functionality has been included.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: Thank you for the feedback/insight. The code has been adjusted to first filter the conversions, then find the units to delete based on the found conversions.I've also added clear commentary on the functionality.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: The location of the comment has been moved.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: The condition wasn't needed. Thank you for pointing that out. The manual check for null has been removed.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I've removed the unneeded manual check for null also present here.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I moved the styling into the shared CSS class folder as opposed to the inline styling. Due to the complexity of information that can result from one conversion removal, the styling made it easier for a user to parse the infromation visually. Adding the class syling keeps the reused code consistant and adds a visual seperator between groups.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I've altered the single letter variables to use more descriptive names.

The or (||) chain has also been modified to return only the unit name. I confirmed that the database has constraints preventing unit names from being empty or NULL, so the additional fallbacks were unnecessary. I also removed the ID fallback since OED should not display internal IDs to users.

This section of code was originally authored by hmhngx. While I cannot concretely say why the fallback chain was implemented, I assume it was intended as a fallback measure to ensure that some form of a unit label would be displayed to the user if the unit name was unavailable.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: Thank you for the resource. I've implemented the standard way.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I've corrected the spelling error.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I've corrected the spelling error.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I've corrected the spelling error.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I've corrected the spelling error.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: I corrected the spelling error. I kept the phrasing as “modal” and changed the comment to use “process” instead of “route to.” I kept “modal” in the comment because this line of code is specifically referring to processing the UI modal.

I changed the comment to:

// Process the appropriate modal based on the cancel conditions set above

src/client/app/translations/data.ts
BunnyBea83: The translation keys have been sorted in alphabetical order, and the French and Spanish sections now reflect the same keys. I am not fluent in either language, and it appears that `\u{26A1}` is used to indicate that a key still needs translation. I’ve added `\u{26A1}` to the corresponding French and Spanish sections to indicate that those entries still need to be translated.

I can use a translator to translate the sections I added, but I can’t guarantee the accuracy of those translations, so I felt it would be better to leave them marked for review by someone fluent in the language.

src/server/models/Conversion.js
BunnyBea83: The blank line has been removed.

src/server/routes/conversions.js
BunnyBea83: I've removed the unused import.

BunnyBea83:
>  The description says it  Partly Addresses issue 1448. Are there aspects that you feel are not completed in that issue if this PR is merged?

Issue 1448 feels completed in terms of successfully removing chains of auto-created units. However, I listed the PR as partially addressing the issue because it currently cannot delete multi-hop paths, as described in the “Limitation” section of the PR. This limitation may be addressed as I make the changes requested during this PR review.

> I tried to find the old git commit record for the previous PRs but did not readily find it. Am I missing it? If not, I'm not sure you may have caused this. If it is gone then maybe the description should list the previous developers in an appropriate way so it is easy for others to see they worked on this previously. I can do this if you prefer and agree.

I was unable to successfully merge the previous PR due to the merge conflicts it was presenting. Because of those conflicts, I manually incorporated the code from that PR into my branch and used the same commit messages that were present in the original PR. I would appreciate any advice on the most appropriate way to credit the contributors for their previous work.

The following commits contain the features from hmhngx's PR that I manually implemented from PR #1561:

- d879267
- b3b1b34
- 4cb3205
- ff4f60e
- e41845c

Hmhngx mentioned that their work was built off of PR #1470, so I referenced both PRs in this PR to keep the overlapping work documented in one place and to credit the contributors who previously worked on this issue. The remaining commits contain my own implementations and changes.

> When I went to the conversion and tried to delete it it failed (note lack of failure reason):

This failure occurs in lines 243 to 252 of `conversions.js`. When `removeAdditionalConversionsAndUnits` is called on `sourceInTx`, it recursively deletes connected suffix units and their conversions. `removeAdditionalConversionsAndUnits` is then called on `destInTx`. If the destination unit was already deleted during the cleanup initiated by `sourceInTx`, the subsequent call attempts to operate on a unit that no longer exists, resulting in the error.

I will address this by handling this edge case so that a unit that has already been deleted during the source cleanup is not processed again during the destination cleanup.

> The admin is not used to seeing the connection to kg and Metric ton as these are artifacts of OED's processing of the suffix unit. This is seen in the unit visualization on the lower analyzed graphic. I don't know if it is easy to remove these or somehow label them/improve the message so it is clear for an admin.

I'm not sure these can be removed outright because they're reachable through the conversion chain and used for real unit-compatibility calculations. However, I agree that the messaging wasn't clear enough as to why they're relevant to an admin who didn't set the connection themselves. 

I've now added labeling instead where OED created units are now tagged in the affected meters/groups lists. This way the admin can tell which units they set up directly vs internal artifacts. I also restructured the surrounding lists so meters/groups sharing the same lost units are grouped together more clearly. Each group now gets its own heading with a bulleted list of lost units underneath, and groupings are separated by a light divider. I recognize that the style of formatting is slightly different compared to other UI notifications. Let me know if this addresses your concern or if you have something more specific in mind.

> The delete seemed to work fine when done in this case. I'm noting that the link to kg CO2 for the meters stays and in the analyzed unit graphic it is shown as orphaned (as it is). This is a suffix unit and I welcome thoughts on if/how there might be another/better way to deal with it. At a minimum, the help system should note this if it stays.

Thanks for flagging this. Based on the structure of the code the real underlying conversion `kg CO2` is never actually deleted by this feature, only the derived suffix units/conversions are. `removeAdditionalConversionsAndUnits` resets the parent's displayable back to visible at the end of its cleanup regardless of why it was called. So the children are removed, `kg CO2` is left visible and still connected to its never deleted conversion (e.g. electric_utility -> kg CO2), and the next graph rebuild regenerates the derived units from it. Which is why it shows as 'orphaned' in the visualization.

One direction I'm considering is blocking the deletion of a derived suffix conversion directly, and instead redirecting the admin to the action that resolves the issue. Either deleting the real underlying conversion, or clearing the unit's suffix field via `/edit`. The catch is that the real conversion might not be visible in the UI once the parent unit is hidden, so I'd want to confirm that's reachable before committing in that direction. Open to your thoughts.


>`[src/server/routes/conversions.js] comment on line 186` 
Could you help me understand why this is cleaning up items even after the simulation that I think happened before this request. Is it to be safe? I have not tried to figure this out so I appreciate your thoughts.

Good question. Yes, it's intentional for safety. `/simulate-delete` and `/delete` are separate requests with no shared transaction. There's a real time gap between when the simulation runs (when the pop up opens) and when the admin actually confirms. In that window, the database could change (e.g. another session editing units, conversions, etc.) so the simulation results could be stale by the time the admin confirms the delete. Re-checking and re-computing inside the actual delete transaction guarantees the cleanup is correct against the live state at commit time, rather than acting on a possibly outdated preview.

That said, I think the delete functionality can use a split solution as opposed to a blanket solution since not everything carries the same risk of stale:

-  **The Cascade Delete Logic** : The functionality should keep re-verifying independetly as it does currently. This is the part where stale data could cause problems since it actually removes data (e.g. a unit is classified to a meter between simulation and confirm).
- **Meter and Group IDs**: These could continue to use the meter and group IDs supplied by the client from the simulation results. These operations clear references rather than deleting the meters or groups themselves, so they have a different risk profile. This is also how the current code behaves: `/delete` uses the `meterIds` and `groupIds` sent by the client, which were derived from the` /simulate-delete` response rather than recomputed server-side during `/delete`.

A more thorough approach, worth considering separately, would be to re-run the simulation when the admin confirms the deletion and compare the results against what was originally shown. If they differ, the updated results could be displayed to the admin and require them to reconfirm before proceeding.

This would close the staleness gap more completely than either trusting the original simulation or silently re-verifying during deletion. However, I think this would be a larger change than what is currently included in the PR, so I wanted to raise it as a possible future approach rather than implement it without first confirming that it is desired.

You raised related versions of the same question on the `units.js` and `EditConversionModalComponent.tsx` threads as well. I think the same reasoning applies to both, but I wanted to consolidate the core discussion here first.

src/client/app/components/conversion/EditConversionModalComponent.tsx
```diff
@@ -365,6 +485,7 @@ export default function EditConversionModalComponent(props: EditConversionModalC
 			setDeleteConfirmationMessage(msgElements);
 			handleCancelModalOpen();
 		} else {
+			// Confirm deletion if no blcoking issues
 			msgElements.push(
```
BunnyBea83, line 535: A line break has been added before the final delete statement and it is also presented in bold.

src/server/routes/units.js
BunnyBea83: The require has been moved to the top.

src/client/app/components/conversion/EditConversionModalComponent.tsx
BunnyBea83: Yes — this is actually already what happens for meterIds/groupIds specifically. /delete uses the list sent from the client (computed in checkState() from the /simulate-delete response) directly, rather than recomputing it server-side. I think that's a reasonable tradeoff here: these fields only get nulled out (not removed), so if the list is slightly stale, worst case is a minor cosmetic gap an admin can clear manually and not a data integrity risk.

The remaining delete logic is different though, as it re-verifies independently rather than trusting the simulation's list, since acting on stale data there could cause a real correctness problem.

Full reasoning on this split is in my consolidated reply here: https://github.com/OpenEnergyDashboard/OED/pull/1693#issuecomment-5629105470

src/server/routes/conversions.js
BunnyBea83: Yes, it's intentional, to be safe. Rather than trusting what the simulation produced, this check re-verifies against the live database so it can hard-block the delete if a real dependency appeared after the simulation ran (e.g. a meter newly assigned to this unit in the gap between simulate and confirm).

Full reasoning on the split between what re-verifies vs. what's safe to trust from the simulation is in my consolidated reply here: https://github.com/OpenEnergyDashboard/OED/pull/1693#issuecomment-5629105470

src/server/routes/conversions.js
BunnyBea83: A lock_by_id.sql file has been added, replicating the same locking behavior as the raw SQL previously used here. clear_default_graphic_unit.sql files were also added under the meter/ and group/ SQL directories. The raw SQL on lines 269 and 273 has been replaced with calls to Meter.clearDefaultGraphicUnit/Group.clearDefaultGraphicUnit, and the row locks now use Unit.lockById.

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: Meter, Group, and Conversion are now in use.

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: The raw SQL queires have been converted into sql files and placed in their designated files.

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: Verbose return details have been added.

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: hmhngx originally authored that portion of code. At the time, their implementation of the code set the units `displayable` to none as opposed to fully deleting the unit. The `canSafelyHide` functions purpose was to classify if a `displayable` could be set to none based on a units dependancy to meters or groups. This function was no longer in use, and has since been removed.

src/server/services/graph/checkUnitDependencies.js
```diff
@@ -0,0 +1,132 @@
...
+	// Suffix units created by OED can be hidden if no meters/groups depend on them
+	// (conversions are expected and will be cleaned up)
+	return deps.meters.length === 0 && deps.groups.length === 0;
+}
+
+/**
+ * Gets detailed dependency information for UI warnings.
+ * @param {number} unitId The unit ID to check
+ * @param {*} conn Database connection
+ * @returns {Promise<Object>} Detailed dependency info
```
BunnyBea83: Added a verbose return description. Single letter variables were also converted to distinct labels.

src/server/routes/conversions.js
BunnyBea83: This code was authored by hmhngx. While I cannot say exactly why it was implemented, I can hypothesize that it was added to flag a possible side effect of the code's operation in the case of an error. The orphan check is a read-only diagnostic query that would probably only fail in the case of a connection issue or a timeout error. I don't think a failure here should block an entire delete that has already succeeded, since this check exists to flag a possible side effect, not to validate the delete itself.

That said, if a bug were to surface in the SQL query itself (e.g. a future schema change breaking a column it relies on), it would fail the same way every time, silently, and a blocking error here wouldn't have caught that either. I've made the error message more explicit in the case that something arises during the check.

src/server/routes/conversions.js
BunnyBea83: Yes, I think thats important for the admin/user to know when units may become orphaned due to conversion deletion. I adjusted the backend detection of orhpaned units to be logged to the admin, and implemented orphaned unit detection for the front end simulation as well.

src/server/services/graph/handleSuffixUnits.js
BunnyBea83: The unused function has since been removed as it is no longer in use.

src/server/services/graph/handleSuffixUnits.js
BunnyBea83: Thank you for the feedback. I've fixed the placement of the JSDoc/const. As for limit consistancy, both are circular dependancy safety nets, but they have different depths as groupings can be moderately deeper than suffix chains. Most autogenerated chains appear to be one level deep, so 10 is used as a generous ceiling for the specific case.

src/server/services/graph/handleSuffixUnits.js
BunnyBea83: Thank you for the catch. The doccument has been reformatted.

src/server/routes/units.js
BunnyBea83: Similar reasoning to conversions.js, there's a small window between the pre-check and the actual delete. The re-read guarantees the delete acts on the live state of the data, rather than operating on data that could be stale from a moment earlier.

All raw SQL in this route has also been addressed. This particular case involved the cik deletion, coinciding with the cik discussion on the `checkUnitDependencies.js` thread, and matching the implementation already in `conversions.js`, the same targeted cik delete was replaced with a single `redoCik()` call after the whole transaction has completed. This call only happens once, after the transaction has closed, rather than per-unit inside the recursive cleanup, to avoid redundant recalculation on deeper suffix chains.

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: Good catch, and I think you're right. This loop deletes every conversion referencing the unit unconditionally, without checking whether the other side is an auto-created unit that would become orphaned as a result(the 0TestUnit → kg of 0TestUnit → gallon of 0TestUnit case). This is the same limitation flagged in my PR description on multi-hop generated units. 
The deletion is safe in that it won't corrupt data or violate constraints, but it can leave a downstream auto-created unit orphaned. Ill add a TODO comment pointing out the limitation explicitly.

src/server/services/graph/checkUnitDependencies.js
BunnyBea83: I agree that the targeted delete alone isn't sufficient. It wouldn't correctly handle an orphaned unit since it only removes rows that directly reference the deleted unit's ID, not rows for other unit pairs whose shortest path happened to route through it. 

I've kept a minimal targeted delete (Cik.deleteByUnitId) before the unit's row is removed to satisfy the foreign key constraint on cik, not as a correctness fix. This way the delete transaction doesn't fail outright. The actual fix is a single `redoCik()` call after the full delete transaction completes. I didn't want to call it once per unit inside the recursive cleanup because it would mean redundant recalculations on deeper suffix chains. Calling it once after delete completion gives one correct rebuild of the full graph.

One clarification: `redoCik` keeps the Cik lookup table consistent, but that's separate from a unit's row still existing in the `units` table with nothing referencing it. I've added a check (`Unit.findOrphanedUnits`, run after the delete completes, plus a matching pre-delete check in the simulation) that detects and reports this case to the admin. It doesn't prevent the orphaning, just surfaces it -- the TODO above about a full path re-walk to actually prevent it still applies.

For reference, `redoCik()` was added to both `routes/conversions.js` and `routes/units.js`, each called once after their respective delete transactions complete.

src/server/services/graph/handleSuffixUnits.js
BunnyBea83: Thank you for the advice. I've added a check to determine if a destination unit has connections beyond the current conversion to be deleted, logging as a warnign when found. In this case the check surfaces the situation server-side rather than a silent removal. Separately, I've also added admin facing pre and post delete checks (`potentiallyOrphanedUnits` in the simulation and `Unit.findOrphanedUnits` after the delete) that detect and report orphaned units directly to the admin. I'm interested in your input on whether OED should skip a deletion outright if it has other connections or would orphan a unit, VS the current approach of a warning/report but still proceeding.

src/server/services/conversionSimulation.js
```diff
@@ -10,6 +10,54 @@ const Unit = require('../models/Unit');
...
+ * Simulates what conversions and units would be removed when deleting a conversion involving suffix units.
+ * This mirrors the logic in removeAdditionalConversionsAndUnits but without actually deleting.
+ * @param {*} suffixUnit The suffix unit to check
+ * @param {*} allConversions All conversions in the system
+ * @param {*} allUnits All units in the system
+ * @returns {Object} Object with arrays of conversionIds and unitIds that would be removed
+ */
+function simulateSuffixUnitCleanup(suffixUnit, allConversions, allUnits) {
+	const conversionsToRemove = [];
+	const unitsToHide = [];
```
BunnyBea83: Former functionality hid the units instead of deleting them. Identifiers were altered to include `delete` in the name as opposed to `hide`.

src/server/services/conversionSimulation.js
```diff
@@ -20,34 +68,62 @@ async function simulateDeleteConversion({ sourceId, destinationId }, conn) {
...
+		conversionsToRemove.push(...cleanup.conversionsToRemove);
+	}
+	
+	// Remove duplicates (in case both source and dest are suffix units and share conversions)
+	const uniqueConversionsToRemove = Array.from(
+		new Map(conversionsToRemove.map(c => [`${c.sourceId}-${c.destinationId}`, c])).values()
 	);
+	
+	// 4. Remove all affected conversions from the simulation
+	const newConversions = allConversions.filter(c => {
```
BunnyBea83, line 104: The code formerly only notfied the user of directly affected conversions and units being deleted. It there were additional conversions attatched to units being deleted the user wouldn't be notified of their removal. The code now accounts for those cases.

src/server/services/graph/handleSuffixUnits.js
BunnyBea83: I've expanded the comment to explain the mechanism. A missing unit occurs when a bidirectional conversion is stored as two rows, both resolving to the same unit, and one gets deleted while the other's lookup is still executing.

Recomputing relatedConversions each iteration narrows the window of a missing unit occurring since only one conversion would be acted upon per iteration before re-querying. I've implemented the recomputation using the `while (true)` loop. However, it doesn't fully eliminate the occurrence of a missing unit because `Promise.all` could still hit two rows resolving the same unit. Given that the code now logs and handles the missing units gracefully, I don't think this residual case needs reporting as an error because this behavior is expected to happen. This leads me to lean more toward leaving it as a warning rather than something that needs to be surfaced further.

src/server/services/graph/handleSuffixUnits.js
BunnyBea83: No, OED was not guaranteed to be in a valid state if an exception occured. The `catch` block would log and `break` instead of re-throwing. If a failure occured part way then a partially completed cascade could be left committed to the database. I've changed it to a `throw err` instead, which should propogate through the recursion and trigger a full rollback.

src/server/test/db/suffixUnitDeletionTests.js
BunnyBea83: Thank you for letting me know about the utility functions. I rewrote the tests to use `insertUnits`,`insertConversions`, and `insertGroups` instead of making the objects directly. Since the `insertMeters` appeared to require a file or data array, I kept the meter test on a direct constructor as my tests only needed a meter with a given unitID. `insertStandardUnits` didn't appear to have typeOfUnit === SUFFIX or non-empty suffix feilds so I kept my parent/child implementation and condensed their duplicated calls into one helper function.

src/server/test/db/suffixUnitDeletionTests.js
BunnyBea83: Thank you for letting me know. I've removed the numbered naming and reused the same simple names across tests since each one gets a clean DB.

src/server/test/db/suffixUnitDeletionTests.js
BunnyBea83: Thank you for the other utility function. I've created a small helper function that throws rather than returning null when nothing matches. I also continued to use `getBySourceDestination` for conversion checks.

PR #1694, opened 2026-08-07
Unit readings test cleanup

### LINKED ISSUE #1662: Update unit readings route tests to newer route parameter testing style
## Description

While working on issue #1575, we reviewed `src/server/test/routes/unitReadingsRouteTests.js`. This file appears to be older than the newer `ParamsTest.js` route test files.

It directly calls validation methods and also checks some correct-result behavior for readings helpers. However, OED has been moving toward testing route behavior through HTTP route tests instead of directly testing validation functions.

This may need a future cleanup to better align the unit readings tests with the newer route parameter testing approach.

## Possible work

* Review `src/server/test/routes/unitReadingsRouteTests.js`.
* Compare its coverage with:

  * `src/server/test/routes/readingsParamsTest.js`
  * `src/server/test/routes/compareReadingsParamsTest.js`
* Identify validation tests that are already covered by newer route parameter tests.
* Identify any missing parameter coverage that should be added to route tests.
* Decide whether correct-result tests should remain, be moved, or be removed if covered elsewhere.
* Consider future changes related to moving away from `:id` syntax and using params instead.

## Context

This came up during the route test cleanup work in #1575. The file does not follow the newer `ParamsTest.js` naming/style and may need non-trivial updates, so it is better tracked as future work rather than handled as a simple duplicate-test cleanup.

Better-Code-Saul-E:
# Description

Migrated `src/server/test/routes/unitReadingsRouteTests.js` to match the newer `ParamsTest.js` route parameter testing style and renamed the file to `unitReadingsParamsTest.js`.

- Renamed the test file to `unitReadingsParamsTest.js` for naming consistency.
- Updated validation checks to use standard validation helper.
- Refactored HTTP request queries to include required parameter like 'readingInterval' and UTC time intervals.
- Cleanedup unsused imports and dead code.

Fixes #1662

## Type of change

## Checklist

## Limitations

N/A


huss:
@Better-Code-Saul-E Thank you for another contribution. It will be about a week before I can review this work. I didn't know if you saw this but the tests and/or checks failed.  FYI.

Better-Code-Saul-E:
Found the bug! I added an await when creating the test meter.

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -4,58 +4,55 @@
...
+						'1,',
+						',1',
+						'1,,2',
+						'1;2',
+						'1.5',
+						'-1',
+						'1 2',
+					],
+					query: valid_query,
+					expectedStatuses: [HTTP_CODES.OK, HTTP_CODES.NOT_FOUND, HTTP_CODES.INTERNAL_SERVER_ERROR]
```
huss: It took me a little while to figure out why this worked and what was wrong. If you look at validatePathIdSegmentPatterns() that is called via this call, it expects expectedStatus (no pluralization). What was happening is the default value is HTTP_CODES.BAD_REQUEST so it all worked. Given this, this should be ``expectedStatus: [HTTP_CODES.BAD_REQUEST]`` to fix  key name but also to limit to only the desired value of return code.

Having said that, there really is a larger issue here. expectValidNumericIdInPath and expectValidCommaSeparatedIds expects the key of expectedStatuses but the other 7 in the file want expectedStatus. Making it even more confusing, there is a variable in 5 functions (``const expectedStatuses``) so the name overlaps. I'm open on how to fix this but I think the passed key should be consistent and maybe the more commonly used one (expectedStatus).

Also, src/server/test/routes/readingsParamsTest.js & src/server/test/routes/mapsParamsTest.js do the same plural usage as here. Once the util functions are fixed up they need to be revisited and maybe sweep the code base to be sure all are correct. I'm also unsure if any other places have bad/extra codes and maybe OED needs an issue to carefully check.

So, do you want to extend this PR to fix the primary issue, do a follow on work or leave this to others?

huss (changes requested):
@Better-Code-Saul-E First, I'm sorry this review took so long (see OED Discord server in developer channel).

Thank you for another contribution. I also appreciate you figuring out the test failure even though it was outside your original work. Maybe the description and/or title should indicate that it is included in this PR. Depending on what is decided for my comment it probably needs other changes.

I only made one comment because I think it is going to change the code. While my quick review found everything seems okay in the big picture, I'm waiting for the comment to get resolved before I carefully check the entire file. Let me know if anything is not clear or you have thoughts/questions.

src/server/test/routes/unitReadingsParamsTests.js
Better-Code-Saul-E: Im currently working on another issue. When I finish I will attempt at the follow on!

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -79,33 +77,58 @@ mocha.describe('unit readings routes', () => {
...
 					1: [
 						{ reading: 1, min: 1, max: 1, startTimestamp: timeInterval.startTimestamp.valueOf(), endTimestamp: timeInterval.endTimestamp.valueOf() }
 					]
 				};
-				expect(response).to.deep.equal(expectedResponse);
+				expect(res.body).to.deep.equal(expectedResponse);
 			});
 		});
 	});
+
```
huss, line 101: Could the second blank line be removed.

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -79,33 +77,58 @@ mocha.describe('unit readings routes', () => {
...
-				expect(validationResult).to.equal(true);
+
+			mocha.it('validates when the time interval is valid', async () => {
+				await expectValidCommaSeparatedIds({
+					baseEndpoint: BAR_METERS_ENDPOINT,
+					validValues: ['1', '12', '1,2,3'],
+					query: valid_query
+				})
 			});
 		});
```
huss, line 134: Not your doing but could the second blank line be removed.

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -122,45 +145,70 @@ mocha.describe('unit readings routes', () => {
 			});
 
 			mocha.it('returns bar readings correctly when called correctly', async () => {
-				const timeInterval = new TimeInterval(moment('2017-01-01'), moment('2017-01-02'));
-
+				const timeString = '2017-01-01T00:00:00.000Z_2017-01-02T00:00:00.000Z';
+                const timeInterval = TimeInterval.fromString(timeString);
```
huss, line 149: This line is space indented but should be tab. Not sure if others do this but in VSC can convert indentation to tabs with a command.

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -70,7 +67,8 @@ mocha.describe('unit readings routes', () => {
 			mocha.it('returns line readings correctly when called correctly', async function () {
 				// The moments in these tests all involve TimeInterval that converts to UTC
 				// and not the DB so okay to use local timezone.
-				const timeInterval = new TimeInterval(moment('2017-01-01'), moment('2017-01-02'));
+				const timeString = '2017-01-01T00:00:00.000Z_2017-01-02T00:00:00.000Z';
+                const timeInterval = TimeInterval.fromString(timeString);
 
 				// getMeterLineReadings is called by meterLineReadings. This makes it appear the result is what is given here.
 				readingsStub = sinon.stub(Reading, 'getMeterLineReadings');
```
huss, line 74: This isn't your doing but this seems to be the only file that does a sinon.stub. Do you think there is a reason to use it and/or do these tests? I think it is mocking up a response to be sure it is correct. However, these are param tests and not ones to check the returned values. In this case, src/server/test/web/readingsLineMeterBasic.js seems to do the basic tests and other files verify the values. Thus, I'm thinking this one is redundant and not a param test so it should be removed. What do you think? If yes, could you check the others to see if they are needed or can do the same change.

Having said that, the tests should still try valid and invalid params. Other routes use standard testing functions to do that. I'm hoping the same methods can be used here. This will require more test code to cover these routes for those cases. I can help with defining the values if you want that.

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -122,45 +145,70 @@ mocha.describe('unit readings routes', () => {
...
 					1: [
 						{ reading: 1, startTimestamp: timeInterval.startTimestamp.valueOf(), endTimestamp: timeInterval.endTimestamp.valueOf() }
 					]
 				};
 
-				expect(response).to.deep.equal(expectedResponse);
+				expect(res.body).to.deep.equal(expectedResponse);
 			});
 		});
 	});
```
huss, line 181: Please remove second blank link. There is at least one more. I won't label any more that should be done.

src/server/test/routes/unitReadingsParamsTests.js
```diff
@@ -122,45 +145,70 @@ mocha.describe('unit readings routes', () => {
...
 					]
 				});
-				const response = await meterBarReadings([1], 99, 1, timeInterval);
+
+				const bar_test_query = {
+					timeInterval: timeString,
+					graphicUnitId: '99',
+					barWidthDays: '1'
+				};
+
```
huss, line 164: Some tests have blank lines and some don't. OED prefers not unless code block separation is needed. Here it sets the values and then uses them in a query and then checks the response. I personally prefer no blank lines in this case. This is a very small point and you can fix up if you want (or not).

src/server/test/routes/unitReadingsParamsTests.js
huss: > So, do you want to extend this PR to fix the primary issue, do a follow on work or leave this to others?

I see you fixed the one route but the others that are outside your original test code are not done. Should this become a new issue or will this PR address it?

huss (changes requested):
@Better-Code-Saul-E Thank you for the update to address my previous comment. I have now looked at the code more carefully. Overall, it works well. I made several trivial comments about formatting. I also made one comment on how the routes are tested. This one will require a little thought and changes to address if that is decided. I also asked a clarifying question on my previous comment. Please let me know if you have any thoughts/questions or something is not clear.

Issue #1695, opened 2026-08-10
Add Swagger/OpenAPI documentation for REST API


neilcabanilla:
## Background:
Currently, this project does not have Swagger/OpenAPI documentation for REST API.

Swagger/OpenAPI creates a standard way to describe API endpoints. Adding Swagger/OpenAPI documentation would be beneficial because it would make the API easier to understand and use, especially for new contributors who may not be familiar with the project yet.

## Motivation:
Adding Swagger/OpenAPI could help with:

- Show all available API endpoints
- Reduce the need to look through source code to understand how the API works
- Make API documentation easier to organize and maintain 
- Provide examples how to call it using curl
- Allows develops to view and test API 

## Proposed Approached:
My teammate (@kujoon226 ) and I did research on Swagger/OpenAPI on what it is and how it could be supported in the project. We also created a draft PR with a basic implementation.

At first, we ran into an issue when installing swagger-jsdoc (JSDoc) because npm was installing the latest version, which was not compatible with the Node.js version used by the project. To solve this, we installed JSDoc version 6.2.8, which worked with the current environment.

After setting up JSDoc, we also installed swagger-ui-express (Swagger UI Express) so the API documentation could be displayed interactively in the browser. This allows developers to view the available API endpoints and test them directly from the Swagger page.

This approach could be maintainable because documentation is kept close to the API code, which makes it easier to update when the endpoint changes. For example, if a developer changes something in an endpoint, they can easily update it because it's in the same area (above of the API). This helps documentation to be in sync with the API, while reducing the chance of it becoming outdated

The draft PR is still in work in progress and meant to be used as a starting point for discussion. We are open for feedback, suggestions, and other ideas on how this could be improved

## Pending Questions
Can CI/CD be used to help keep the Swagger/OpenAPI documentation up to date?

Our Draft PR #1696

Johnsonjahnifah:
Hello Neil, I am a student who is looking to contribute to the project here and U wanted to do user research to better understand  and identify any gaps in the project.This issue seems intriguing and i would like to know your experience with OED API. If you would be willing to have a short conversation about this and the challenges you are experiencing i would greatly appreciate it.

Mr25thbaam:
[...]

PR #1696, opened 2026-08-10
Add Swagger/OpenAPI documentation for REST API

kujoon226:
# Description

This PR adds an initial Swagger/OpenAPI implementation for OED's REST API. It adds Swagger UI at /api-docs and documents the GET /api/conversions endpoint as an example. This is a prototype for maintainer feedback before documenting additional endpoints.

Partly Addresses #1695 

## Type of change

- [x ] This change requires a documentation update

## Checklist

- [ x] I have followed the [OED pull request](https://openenergydashboard.org/developer/pr/) ideas
- [ x] I have removed text in ( ) from the issue request
- [ x] You acknowledge that every person contributing to this work has signed the [OED Contributing License Agreement](https://openenergydashboard.org/developer/cla/) and each author is listed in the Description section.

## Limitations

Currently only the GET `/api/conversions` endpoint is documented as an example. Additional API endpoints still need Swagger/OpenAPI documentation. `swagger-jsdoc` is currently pinned to version 6.2.8 for compatibility with OED's current Node environment.

### DIFF
--- a/src/server/swagger.js
+++ b/src/server/swagger.js
@@ -1,0 +1,20 @@
+const swaggerJsdoc = require('swagger-jsdoc');
+
+const options = {
+	definition: {
+		openapi: '3.0.0',
+		info: {
+			title: 'OED API',
+			version: '1.0.0',
+			description: 'OpenEnergyDashboard API documentation'
+		},
+		servers: [
+			{
+				url: '/'
+			}
+		]
+	},
+	apis: ['./src/server/routes/*.js']
+};
+
+module.exports = swaggerJsdoc(options);
--- a/src/server/swagger.js
+++ b/src/server/swagger.js
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 const swaggerJsdoc = require('swagger-jsdoc');
 
 const options = {
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,8 @@
 				"redux": "~4.2.1",
 				"redux-thunk": "~2.4.0",
 				"serve-favicon": "~2.5.0",
+				"swagger-jsdoc": "~6.2.8",
+				"swagger-ui-express": "~5.0.1",
 				"usehooks-ts": "~3.1.0",
 				"xml2js": "~0.5.0"
 			},
@@ -96,6 +98,46 @@
 			"dev": true,
 			"engines": {
 				"node": ">=0.10.0"
+			}
+		},
+		"node_modules/@apidevtools/json-schema-ref-parser": {
+			"version": "9.1.2",
+			"resolved": "https://registry.npmjs.org/@apidevtools/json-schema-ref-parser/-/json-schema-ref-parser-9.1.2.tgz",
+			"integrity": "sha512-r1w81DpR+KyRWd3f+rk6TNqMgedmAxZP5v5KWlXQWlgMUUtyEJch0DKEci1SorPMiSeM8XPl7MZ3miJ60JIpQg==",
+			"dependencies": {
+				"@jsdevtools/ono": "^7.1.3",
+				"@types/json-schema": "^7.0.6",
+				"call-me-maybe": "^1.0.1",
+				"js-yaml": "^4.1.0"
+			}
+		},
+		"node_modules/@apidevtools/openapi-schemas": {
+			"version": "2.1.0",
+			"resolved": "https://registry.npmjs.org/@apidevtools/openapi-schemas/-/openapi-schemas-2.1.0.tgz",
+			"integrity": "sha512-Zc1AlqrJlX3SlpupFGpiLi2EbteyP7fXmUOGup6/DnkRgjP9bgMM/ag+n91rsv0U1Gpz0H3VILA/o3bW7Ua6BQ==",
+			"engines": {
+				"node": ">=10"
+			}
+		},
+		"node_modules/@apidevtools/swagger-methods": {
+			"version": "3.0.2",
+			"resolved": "https://registry.npmjs.org/@apidevtools/swagger-methods/-/swagger-methods-3.0.2.tgz",
+			"integrity": "sha512-QAkD5kK2b1WfjDS/UQn/qQkbwF31uqRjPTrsCs5ZG9BQGAkjwvqGFjjPqAuzac/IYzpPtRzjCP1WrTuAIjMrXg=="
+		},
+		"node_modules/@apidevtools/swagger-parser": {
+			"version": "10.0.3",
+			"resolved": "https://registry.npmjs.org/@apidevtools/swagger-parser/-/swagger-parser-10.0.3.tgz",
+			"integrity": "sha512-sNiLY51vZOmSPFZA5TF35KZ2HbgYklQnTSDnkghamzLb3EkNtcQnrBQEj5AOCxHpTtXpqMCRM1CrmV2rG6nw4g==",
+			"dependencies": {
+				"@apidevtools/json-schema-ref-parser": "^9.0.6",
+				"@apidevtools/openapi-schemas": "^2.0.4",
+				"@apidevtools/swagger-methods": "^3.0.2",
+				"@jsdevtools/ono": "^7.1.3",
+				"call-me-maybe": "^1.0.1",
+				"z-schema": "^5.0.1"
+			},
+			"peerDependencies": {
+				"openapi-types": ">=7"
 			}
 		},
 		"node_modules/@babel/code-frame": {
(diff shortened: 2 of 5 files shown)


huss:
@kujoon226 I've only looked at this some. I had a few questions:

- Did you consider alternatives to swagger?
- I had trouble getting the bundle sizes for the new software. Do you happen to know this? Can it be limited to only for developments?

neilcabanilla:
> @kujoon226 I've only looked at this some. I had a few questions:
> 
> * Did you consider alternatives to swagger?
> * I had trouble getting the bundle sizes for the new software. Do you happen to know this? Can it be limited to only for developments?

Hello @huss,

We didn't really look into other alternatives and were mainly focused on understanding swagger, but we will try to look into other alternatives such as Postman. 

For the bundle sizes, did you mean the size of the new packages added ( swagger-jsdoc, swagger-ui-express, swagger-ui-dist) ? The installed sizes for these is 13.7 MB. 

We also looked into limiting Swagger to developmentl. Since the project already uses OED_PRODUCTION environment variable, we could do a condition check in the app.js when calling Swagger (api/docs). If OED_PRODUCTION is yes, then it prevents Swagger documentation route from being accessed in production.

However, the packages would still need to be installed (for production) because OED's installation uses npm ci, which includes dev dependencies. We can look further into ways to avoid installing Swagger packages in production as well.

huss:
> We didn't really look into other alternatives and were mainly focused on understanding swagger, but we will try to look into other alternatives such as Postman.

Thanks. That would be nice. Providing what you looked at, your analysis and any thoughts/conclusions would be nice.

> For the bundle sizes, did you mean the size of the new packages added ( swagger-jsdoc, swagger-ui-express, swagger-ui-dist) ? The installed sizes for these is 13.7 MB.

Yes, OED wants to limit the bundle size, esp. what is sent to the client (web browser) since that happens on each usage. If it stays on the server then it is not a big deal.

> We also looked into limiting Swagger to developmentl. Since the project already uses OED_PRODUCTION environment variable, we could do a condition check in the app.js when calling Swagger (api/docs). If OED_PRODUCTION is yes, then it prevents Swagger documentation route from being accessed in production.

I'm open to thoughts on this. My first thought is that only developers would want to see this but I'm open to other ideas.

I'm also interested in if this needs to be part of the running OED or would be effective via another mechanism. For example, OED has documentation help pages on our website. Now, those would be available in a different way/place so I'm interested in people's thoughts.

I think the reality is that almost none of the OED API/route points are ever used outside OED called code. Thus, I think this is mostly for developers and not for other machines or outside people to figure out how to make OED API calls. I don't know if that impacts your thoughts.

Also, as I have not used this before, how does it relate to the route parameter testing that has and is being brought online? I would think they should align but this is much easier for someone to see/understand.

> However, the packages would still need to be installed (for production) because OED's installation uses npm ci, which includes dev dependencies. We can look further into ways to avoid installing Swagger packages in production as well.

OED wants to use `npm ci` to make sure that we control the actual version of all packages. I'm open to ideas on how to limit the dev dependencies so they are only installed on development environments (as appropriate). I have not looked at this in any detail so I don't recall if this is already done or not (you indicate not).

Thanks for looking at this and I welcome any ideas/thoughts.

neilcabanilla:
Hello @huss,

> I'm open to thoughts on this. My first thought is that only developers would want to see this but I'm open to other ideas.
> 
> I'm also interested in if this needs to be part of the running OED or would be effective via another mechanism. For example, OED has documentation help pages on our website. Now, those would be available in a different way/place so I'm interested in people's thoughts.
> 
> I think the reality is that almost none of the OED API/route points are ever used outside OED called code. Thus, I think this is mostly for developers and not for other machines or outside people to figure out how to make OED API calls. I don't know if that impacts your thoughts.

Yes, this is primarily for developers, especially new developers coming into the project. The plan is to make the Swagger page only available in development/test environments, and not part of OED's production. 

> Also, as I have not used this before, how does it relate to the route parameter testing that has and is being brought online? I would think they should align but this is much easier for someone to see/understand.

Swagger lets developers manually test the API routes by sending real requests. It shows whether the request succeeded or failed based on the HTTP response code, and it also displays the actual response data. The route parameter tests are automated, while Swagger gives developers a visual way to test and understand the same routes. Ideally, both should reflect the same expected behavior.

> OED wants to use npm ci to make sure that we control the actual version of all packages. I'm open to ideas on how to limit the dev dependencies so they are only installed on development environments (as appropriate). I have not looked at this in any detail so I don't recall if this is already done or not (you indicate not).

Our thought was to keep the Swagger packages as dev dependencies and only enable Swagger in the development/test environment.

In development/test, we would use npm ci, which installs Swagger along with the other development dependencies. While, for production, we could use something like npm ci --omit=dev, which would skip installing Swagger and the other dev dependencies

We would also only enable the /api-docs route in development/test by doing a condition statement, so Swagger would not be available in the production version of OED. Doing this keeps Swagger only in development.

huss:
I think the plan you laid out sounds fine. I look forward to seeing it.

src/server/swagger.js
```diff
@@ -0,0 +1,24 @@
...
+		servers: [
+			{
+				url: '/'
+			}
+		]
+	},
+	apis: ['./src/server/routes/*.js']
+};
+
+module.exports = swaggerJsdoc(options);
```
huss, line 24: Can you help me understand how this file/export is used in the code?

huss (commented):
@kujoon226 Thank you for this contribution and work. I realize this is a draft PR but I thought I would take a look and give any thoughts.

- @neilcabanilla opened the issue associated with this PR. I wanted to check if they are also a contributor to this work.
- I made one comment to help me understand one part of the code.
- Now I'll do my big picture thoughts. I tried using http://localhost:3000/api-docs/ and it works fine. I agree it is nice. Whenever OED is considering adding a new feature, it needs to consider the cost/benefit tradeoff. Thus, my thinking is this:
  - The documentation/@openapi info needed for each route is non-trivial so it will take time to create and then it needs to be kept up-to-date.
  - The information being added duplicates a lot of what the route and parameter checks indicate. Having said that, it is in a clearer form and it is much easier to access than looking at the code.
  - The feature to probe the route is nice and the formatting of the return values is very easy to read. What I'm wondering is how often a developer will do this. My guess is it isn't common (I don't do it very often). Furthermore, I think it is likely for more advanced developers who would know other mechanisms (curl and other UI tools). I do see that the person doing it another way needs to determine the proper values to send. There are a number of ways to do that: client/server routes, test cases, using the web page while inspecting and several available tools. 
  - The sample works well for the get route. Some routes, such as the edit ones, require information that is specific to each running instance so a sample could not be run successfully. That would require getting the values via the ideas in the last bullet and/or probing the database.
  - OED developers should, in general, be using the Redux Toolkit to make requests of the server. Normally it is set up as part of creating a new endpoint and then simply used. The codebase should have sample usage if it already exists. If not, then the new one would not yet have the Swagger info and would need to be set up.

Given this, I wanted to get your thinking on making the effort to do this across the OED codebase. I hope I appreciate what you did and what it might add. I also suspect you have ideas on usage/maintenance/etc. that I have not considered. I just want to have a big picture discussion/consideration of the benefits vs the cost of doing this. I welcome any and all thoughts on this and know I am still open to moving this forward if it seems appropriate.

src/server/swagger.js
kujoon226: Hello @huss , Sorry for the late reponse. I am working on this feature with @neilcabanilla. 
Regarding your question about how swagger.js and its export are used:

apis: ['./src/server/routes/*.js'] tells Swagger to look through all the .js files inside the routes folder for API documentation( checks routes that has  @openapi included). Then module.exports = swaggerJsdoc(options) takes the Swagger documentation generated from those files and exports it so other parts of the application can use it. 
Route files → Swagger reads the documentation → module.exports shares it → Swagger UI displays it in the browser.

For the broader questions, I agree with the points you've mentioned. The main question is whether Swagger is worth the time needed to create and maintain the documentation.

We are also looking into automatically generating some of the OpenAPI information from the existing routes and validation rules. If that works well, it could reduce the amount of extra work and help keep the documentation up to date. Some parts, such as clear descriptions and examples, may still need to be written by hand.

I agree that Swagger would support the Redux Toolkit workflow rather than replace it. It could still be useful for learning about the API, checking requests and responses, testing, and debugging.

One benefit I see for OED is making the API easier to discover and understand, especially for developers who are new to the project. Experienced developers may already use curl, tests, the web page, or inspect the client/server code to understand how a route works. Swagger would not replace those methods, but it could provide one place to quickly see the available routes, required parameters, and expected responses without having to first find the relevant code or tests. Whether this benefit is enough to justify documenting and maintaining all of OED's routes is something we still need to evaluate.

For your point about edit routes requiring information specific to each running instance, I wanted to make sure I understand correctly. Are you referring to things like authentication tokens for protected routes or instance-specific values such as an existing group or meter ID, or both?

My suggestion is to test automatic generation with a few different routes, including both GET and edit routes. We can then see how accurate and useful the results are and how much manual work is still needed. Based on that, we can decide whether documenting the entire codebase is worthwhile.

huss: @neilcabanilla @kujoon226 

> Whether this benefit is enough to justify documenting and maintaining all of OED's routes is something we still need to evaluate.

I agree and you ideas seem okay with me - see below.

> For your point about edit routes requiring information specific to each running instance, I wanted to make sure I understand correctly. Are you referring to things like authentication tokens for protected routes or instance-specific values such as an existing group or meter ID, or both?

I assume potential users could be told that they need to be appropriately logged in to test controlled routes. It might also be worth considering if more route specific documentation could easily be created to show in Swagger so it is clear on the access level.

The harder question may be existing IDs, etc. that would only be known before the route is tried. Making if more complex is the fact that you need to look at Redux state or probe the DB to get IDs so it will require more sophistication on the part of the person doing it. Again, documentation can help but it probably raises the bar unless someone has another way.

> My suggestion is to test automatic generation with a few different routes, including both GET and edit routes. We can then see how accurate and useful the results are and how much manual work is still needed. Based on that, we can decide whether documenting the entire codebase is worthwhile.

If you want to see about this and then discuss with OED about moving forward more generally then that seems like a good plan. I appreciate your effort.

PR #1702, opened 2026-08-27
Dev-1655: Add Separate Credentials for OED Test Database

### LINKED ISSUE #1655: use separate DB testing password

While src/server/test/common.js looks for OED_DB_TEST_PASSWORD to use first, it isn't set in any configuration files so it defaults to the same password as the normal one for the regular OED database. While OED testing should only use the test database, having the same password would mean it might work by accident if something goes wrong and it could modify the real, non-test data.

OED should implement the OED_DB_TEST_PASSWORD and set it to a different value. Since it is only for testing, it can be a set value unlike the regular ones.

### Additional context

Care needs to be taken as this password is also used in the GitHub workflow. (It is also in the Travis files but maybe those should go as they have not been used in a long time.) Verifying it works for developer, production and GitHub testing is important.

AdrienSch:
# Description

Add separate credentials for OED test user. 

The documentation has been modified accordingly in https://github.com/OpenEnergyDashboard/OpenEnergyDashboard.github.io/pull/62 .

Fixes #1655 

## Type of change

## Checklist

## Limitations

- None


AdrienSch:
@huss If I remove .travis.yml, should I change the credentials that still refer to Travis?

huss:
@AdrienSch Thank you for this contribution. Here are my initial thoughts:

> If I remove .travis.yml, should I change the credentials that still refer to Travis?

First, I looked at .travis.yml and it is out-of-date with the current OED setup. Thus, it either needs to be updated or removed at this point. I concur that removing it makes sense since it being out-of-date indicates that no one uses it. Second, are you referring to .github/workflows/build.yml that has environment variables with values that contain travis? If so, changing them to something else would be nice. Not sure the best choice but this is specific to GitHub at this time so either that or something less specific.

Also, the PR description  #[1655]. If you think this finishes that issue then it should be without the [ ] so it will automatically close the issue if the PR is merged. That would also link the issue to this PR for others to see.

Finally, I see this is a draft PR. If you want me to look at anything or think about something else then please let me know.

containers/database/init.sql
```diff
@@ -5,7 +5,14 @@
...
 DROP DATABASE IF EXISTS oed_testing;
-CREATE DATABASE oed_testing WITH OWNER oed;
+CREATE DATABASE oed_testing WITH OWNER oedtestuser;
+
+REVOKE CONNECT ON DATABASE oed FROM PUBLIC;
+GRANT CONNECT ON DATABASE oed TO oed;
 
 ALTER ROLE oed SUPERUSER;
+
+\c oed_testing
```
huss, line 17: Can you help me understand why the DB creation files that have (probably too many) creations of the btree_gist extension do not cover this? Would this replace the need for them?

containers/database/init.sql
```diff
@@ -5,7 +5,14 @@
 CREATE USER oed WITH SUPERUSER PASSWORD 'opened';
 DROP DATABASE IF EXISTS oed;
 CREATE DATABASE oed WITH OWNER oed;
+CREATE USER oedtestuser WITH NOSUPERUSER NOCREATEDB NOCREATEROLE PASSWORD 'oedtestpassword';
```
huss, line 8: I was thinking this may be a security issue because a non-secure account is always created. It does have restrictions so any access should not create significant problems but any unnecessary access should be avoided. PR #1554 (merged) tried to do the initial fix for this. It probably is a good idea to see if the appropriate parts can be applied here.

Also, when I run ``npm run test`` on my local machine I get:
```test
Baselines
    1) "before each" hook for "can be saved and retrieved"


  0 passing (9ms)
  1 failing

  1) "before each" hook for "can be saved and retrieved":
     error: password authentication failed for user "oedtestuser"
```

I don't think this has anything to do with this test but will happen for all tests. I tried wiping out Postgres so it reinstalled (to make sure init.sql would run) but it still did not work in the same way. I have not tried to figure out what is up.

This also raises that current sites will need a migration plan with OED provided code to change the DB to support these changes while preserving their current data. OED has a standard setup for DB changes (assuming it can be used for this).

containers/database/init.sql
```diff
@@ -5,7 +5,14 @@
...
 DROP DATABASE IF EXISTS oed_testing;
-CREATE DATABASE oed_testing WITH OWNER oed;
+CREATE DATABASE oed_testing WITH OWNER oedtestuser;
+
+REVOKE CONNECT ON DATABASE oed FROM PUBLIC;
```
huss, line 12: I realize this file had no comments but it would be nice to add some (at least for what you added) to explain why steps are being taken.

containers/database/nuke.sql
```diff
@@ -6,4 +6,10 @@ DROP DATABASE IF EXISTS oed;
 DROP DATABASE IF EXISTS oed_testing;
 
 CREATE DATABASE oed WITH OWNER oed;
-CREATE DATABASE oed_testing WITH OWNER oed;
+CREATE DATABASE oed_testing WITH OWNER oedtestuser;
+
+REVOKE CONNECT ON DATABASE oed FROM PUBLIC;
```
huss, line 11: Same comment as init on why/commenting. I'd also like to better understand why the DB destruction code is doing all these steps.

.github/workflows/build.yml
```diff
@@ -9,16 +9,18 @@ on:
...
     env:
-      OED_DB_USER: test
-      OED_DB_PASSWORD: travisTest
-      OED_DB_DATABASE: travis_ci_dummy
-      OED_DB_TEST_DATABASE: travis_ci_test
+      OED_DB_USER: oed
+      OED_DB_PASSWORD: opened
+      OED_DB_DATABASE: oed
+      OED_DB_TEST_DATABASE: oed_testing
+      OED_DB_TEST_USER: oedtestuser
```
huss, line 16: My comment to secure the DB probably won't work with these hardcoded values unless special steps are taken. I'm not sure how to address this. Could special values be used on GitHub in the DB setup? Can the .env values be accessed to get them? It isn't actually a security issue on GitHub but I was concerned about everywhere else.

.github/workflows/build.yml
```diff
@@ -9,16 +9,18 @@ on:
 jobs:
   run-checks-tests:
     env:
-      OED_DB_USER: test
-      OED_DB_PASSWORD: travisTest
-      OED_DB_DATABASE: travis_ci_dummy
-      OED_DB_TEST_DATABASE: travis_ci_test
+      OED_DB_USER: oed
```
huss, line 12: Do you know why the testing setup needs to know the standard passwords since I would hope it never accesses the other parts of the DB?

client.js
```diff
@@ -12,21 +12,26 @@ const pgclient = new Client({
 	database: process.env.OED_DB_TEST_DATABASE
 });
 
-pgclient.connect();
+/* New Code for setting up the Test Database! */
```
huss, line 15: I think comments should not refer to new as they become obsolete quickly.

This is causing a JSDoc warning since it isn't actual JSDoc. I hope it is not hard to make it proper JSDoc as OED is moving to have that across the code base.

client.js
```diff
@@ -12,21 +12,26 @@ const pgclient = new Client({
...
+		`CREATE USER ${process.env.OED_DB_TEST_USER} WITH NOSUPERUSER NOCREATEDB NOCREATEROLE ` +
+		`PASSWORD '${process.env.OED_DB_TEST_PASSWORD}'`
+	);
 
-pgclient.query(alterUser, (err, res) => {
-	if (err) throw err
-});
+	await pgclient.query(`ALTER DATABASE ${process.env.OED_DB_TEST_DATABASE} OWNER TO ${process.env.OED_DB_TEST_USER}`);
+
+	await pgclient.query('CREATE EXTENSION IF NOT EXISTS btree_gist');
```
huss, line 29: The usual question about btree_gist.

client.js
```diff
@@ -12,21 +12,26 @@ const pgclient = new Client({
 	database: process.env.OED_DB_TEST_DATABASE
 });
 
-pgclient.connect();
+/* New Code for setting up the Test Database! */
+async function setupDatabase() {
```
huss, line 16: A few thoughts on this code:

1. Can you help me understand why you changed this to be this function. It may be to centralize the error but I want to know for sure.
2. When an error happens it now terminates the process. Can you explain why you are doing this way versus the old way?
3. Adding comments would help.

docker-compose.yml
```diff
@@ -126,6 +126,8 @@ services:
             # This user is used when the standard OED testing is run. It is left in this Docker
             # config file in case sites want to run the test suite.
             OED_DB_TEST_DATABASE: oed_testing
+            OED_DB_TEST_USER: oedtestuser
```
huss, line 129: Is this needed here because the tests run in the the Docker web container? I also noticed they are not in the DB section.

huss (changes requested):
@AdrienSch Thank you for another contribution to OED. As discussed in a Docker msg on the OED server/developer channel, I am sorry this took so long to review. I have made some comments where some are to help me understand the changes. Please let me know if anything is not clear and your thoughts/questions.

Issue #790, opened 2022-09-26
need for non-admin to see some info on non-displayable meters


huss:
### Describe the bug

src/client/app/components/ChartDataSelectComponent.tsx calls metersInGroup in src/client/app/utils/determineCompatibleUnits.ts to determine what units are compatible with a group. If an included the meter is not displayable, it is not in Redux state so the code fails because it expects all meters to be in the Redux state if needed by that type of user. The issue is the group needs the meter so the needed access is indirect. OED already decided to allow groups to be graphed even if it contains meters (or groups) that are hidden.

OED needs to fix this and the obvious idea is to load some info on all meters into Redux state (at least units). This may causes issues since the current Redux state is used to determine used meters in a number of places in OED so including ones not displayable to the user may cause issues. The files that may be impacted by this work is (from a modest analysis that may be lacking):
1. The two files above.
2. src/client/app/components/MeterDropDownComponent.tsx uses the props state of meters (may be okay if prop state is okay).
3. src/client/app/containers/MeterDropdownContainer.ts
4. src/client/app/containers/groups/DatasourceBoxContainer.ts
5. src/client/app/components/meters/MetersDetailComponent.tsx

Some files that also deal with non-displayable meters are:
1. src/client/app/components/groups/GroupViewComponent.tsx
2. src/client/app/containers/MultiCompareChartContainer.ts

The current code uses:
1. src/client/app/utils/api/MetersApi.ts
2. src/server/routes/meters.js
3. src/server/models/Meter.js
4. src/client/app/actions/meters.ts

### To Reproduce

1. Create a meter that has displayable of false.
6. Create a displayable group that includes this meter
7. Log out as admin and refresh the web page
8. Select any meter (or probably group) and the web browser console will show an error.

### Expected behavior

OED should not fail if an underlying meter of a group is not displayable.

### Additional context

It might be a good idea to check if non-displayable groups can cause a similar issue.

PR #1703, opened 2026-08-27
Require authentication for readings of non-displayable meters

anupamme:
# Description

Require authentication for readings of non-displayable meters.

`GET /api/readings/line/raw/meter/:meter_id` and `GET /api/readings/line/count/meters/:meter_ids`
returned reading data and reading counts for a client-supplied meter id without checking the meter's
`displayable` flag, so an unauthenticated request could read or count readings for a meter the site
owner had marked hidden. This adds the same check already used by the single-meter route in
`src/server/routes/meters.js` to both readings routes, and returns 404 rather than 500 for a meter id
that does not exist.

Group readings routes are deliberately left alone. Per #790, OED has already decided that a group may
be graphed even when it contains hidden meters, so gating those on member visibility would change
intended behavior.

No client change was needed. The only callers of these two routes are
`src/client/app/redux/thunks/exportThunk.ts` (via `metersApi.lineReadingsCount` and
`metersApi.rawLineReadings`), and both pass only meters the user already has in Redux state, which
for a user who is not logged in contains displayable meters only. Raw export therefore behaves as
before.

Author: @anupamme, the sole person contributing to this work, who has signed the OED CLA.

Disclosure on tooling: this work was drafted with AI assistance (Claude) and then reviewed, corrected
and tested by me. Commit d85096a9f carries a `Co-Authored-By: Claude` trailer; that records tool
assistance and is not a second person under the CLA. I have left the trailer in place rather than
rewriting history, so the record stays accurate. I understand OED's AI policy is still being written
and that it may rule this contribution out; I would rather be transparent about it than not.

Partly Addresses #1704

I have used "Partly Addresses" rather than "Fixes" because the issue describes the pattern generally
and this pull request only changes the two routes named above, so the issue should stay open.

## Type of change

## Checklist

## Limitations

- `src/server/routes/meters.js` performs the equivalent visibility check but responds with
  `BAD_REQUEST` (400) where this pull request uses `FORBIDDEN` (403). That inconsistency predates this
  work and is not addressed here, since changing an unrelated route seemed out of scope. I am happy to
  align them if OED would prefer a single status code.
- Other meter-facing readings routes were reviewed for the same pattern. Anything not already public
  has been reported through the private channel in `SECURITY.md` rather than described here, so it is
  not restated in this pull request or in the linked issue.
- The automated checks have not run on this pull request. Per the developer pages, tests do not run
  automatically for a first-time contributor until the project authorizes them. Locally,
  `npm run check` is clean and `npm run test` passes.


huss:
@anupamme Thank you for this contribution. Here are my thoughts:

- OED appreciates you acknowledging that AI generated this code. OED is in the process of completing an AI policy and will publicly post it when ready. It will not allow for solely generated AI code (or a majority of the code) along with other requirements. As such, this PR does not seem to be acceptable under this policy.
- The PR description does not conform to OED standards as publicly posted on our website.
- While this may be an issue, the fix is limited to a single route without consideration to other places it may occur. OED would prefer PRs that fully address these types of issues.
- While not wrong, OED prefers that security issues be reported per our security policy to allow for fixes before publicly revealed.

Please let me know any thoughts. You will need to fix up this PR to conform to OED standards and address any comments unless OED agrees they are not valid for it to move forward. Given its nature, it will be closed on 30 August 2026 if OED does not hear from you.

huss:
This is to note that I received an email notice of a comment on this PR but when I looked it was not there. Usually this means the author of the comment removed it. If something is missing that is desired then please post it again.

anupamme:
Yes, there was an earlier comment, but it was misleading, so I deleted it.

Thanks for the detailed feedback. Here's where things stand:

- Reframing the claim: this PR hardens the public/restricted data boundary for non-displayable meters on the readings routes, rather than claiming a general IDOR fix. Updated the description accordingly.
- Coverage: extended the fix to the sibling /line/count/meters/:meter_ids endpoint, which had the same gap (no displayable check), so the fix isn't limited to a single route. meters.js's own displayable check was left as-is (pre-existing, different status code) since changing an unrelated route wasn't requested.
- Tests: added coverage for anonymous+displayable, anonymous+non-displayable, authenticated+non-displayable, and nonexistent-meter cases on both routes.
- Nonexistent meter: now returns 404 instead of silently 500ing.
- Disclosure process / AI policy: understood; happy to follow whatever process you'd prefer here going forward.

huss:
I sometimes provide basic thoughts on a PR when it comes in even if it has not been properly submitted to show OED's willingness to accept work and to be supportive. However, OED does have a policy that we don't review work that does not meet the basic requirements for a PR. This PR does not meet those requirements (see the [developer pages](https://openenergydashboard.org/developer/developer/), esp. information on the CLA & PR template). This was done because the project sometimes spent time discussing a PR that could not be accepted due to missing items that made it impossible to merge. Thus, once that is done, OED will then comment on your thoughts and the content of this PR. Note completing these steps does not mean OED will accept this work but it would mean it is eligible to discuss and consider. Please let me know if you have any questions/thoughts.

anupamme:
Thanks — that's fair, and the reasoning makes sense. I had not read the developer pages carefully
enough before submitting. I've now gone through them and fixed the submission itself:

- **CLA.** Signed via the form on the CLA page. The Description now names me as the sole person
  contributing, as the checklist item asks.
- **Issue.** There wasn't one; I've opened #1704 and linked it as "Partly Addresses" rather
  than "Fixes", since it describes the pattern more broadly than this PR changes.
- **PR template.** The description has been rebuilt to follow `pull_request_template.md`. I removed
  the scanner-output table and the vendor footer that were in the original body — neither belonged in
  an OED pull request, and the file/line reference in that table had gone stale anyway.
- **Title.** Fixed; the old one was truncated and mentioned groups, which this PR doesn't touch.
- **AI assistance.** Disclosed explicitly in the Description. The second commit carries a
  `Co-Authored-By: Claude` trailer, which records tool assistance rather than a second person under
  the CLA. I've deliberately left it in place instead of rewriting history, so the record stays
  accurate. I understand the policy in progress may still rule this out.
- **Documentation.** Added the `@param`/`@returns` tags the code guidelines ask for on the one new
  function.

On your earlier points about content, briefly, so it's on the record for whenever you get to it:

- **Other places it occurs.** You were right that one route wasn't the whole picture. I extended this
  PR to `/line/count/meters/:meter_ids`, which had the same gap. I also reviewed the rest of the
  meter-facing readings routes and found more instances — those are not public, so following your
  fourth point I sent them through the contact form under "Security" rather than writing them into
  this PR or the issue. Happy to discuss there, or here if you'd rather.
- **Group routes** are deliberately untouched, since #790 records that OED already decided a group may
  be graphed when it contains hidden meters.
- **Tests.** Mocha/Chai cases in `readingsParamsTest.js` for anonymous access to a displayable meter,
  anonymous access to a hidden meter, authenticated access to a hidden meter, and a nonexistent meter
  id, on both routes, plus the mixed-list case on the count route.
- **No client change needed.** The only callers are in `exportThunk.ts`, and they pass only meters
  already in Redux state, which for a logged-out user is displayable meters only — so raw export
  behaves as before.

One thing I can't fix from my side: the automated checks have never run on this PR, which I gather is
the first-time-contributor authorization. Locally `npm run check` is clean and `npm run test` passes.

Given the AI policy, I'd rather ask than assume: would you prefer the remaining routes handled in a
follow-up from me, or by an OED team member? I'm fine either way, and I'd understand if OED closes
this one on policy grounds — the information is with you regardless, which was the point.

src/server/routes/readings.js
```diff
@@ -19,6 +21,34 @@ const { success, failure } = require('./response');
 
 const router = express.Router();
 
+/**
+ * Fetches a meter by ID and enforces that non-displayable meters are only visible to
+ * authenticated requests. If the meter does not exist or the request is not allowed to
+ * see it, the response is sent here and null is returned so the caller stops processing.
+ * @param meterID {int} the id of the meter to fetch
+ * @param conn the database connection to use
```
huss: When possible, OED prefers the conn to be the last argument.

src/server/routes/readings.js
```diff
@@ -19,6 +21,34 @@ const { success, failure } = require('./response');
...
+ * @param req the Express request, used to check for a valid authentication token
+ * @param res the Express response, which is sent on failure
+ * @returns {Meter} the meter, or null if a failure response was already sent
+ */
+async function getAuthorizedMeter(meterID, conn, req, res) {
+	let meter;
+	try {
+		meter = await Meter.getByID(meterID, conn);
+	} catch (err) {
```
huss: OED has not done this in other places. If an id is provided that does not exist then it simply returns no data for that id. Given this should only happen from a programming error or request outside OED, I think OED will stick with that way unless you have reasons to do it otherwise.

src/server/routes/readings.js
```diff
@@ -19,6 +21,34 @@ const { success, failure } = require('./response');
...
+		if (err instanceof pgPromise.errors.QueryResultError && err.code === pgPromise.errors.queryResultErrorCode.noData) {
+			failure(res, HTTP_CODES.NOT_FOUND);
+			return null;
+		}
+		throw err;
+	}
+	if (!meter.displayable && !req.hasValidAuthToken) {
+		failure(res, HTTP_CODES.FORBIDDEN);
```
huss: Usually a failure message includes a message that does not give any sensitive information. Also, I think having the original route function return errors is somewhat nicer when that can readily be done. Please see my overall comment for other thoughts.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -120,6 +122,57 @@ mocha.describe('Readings Route Parameter Validation', () => {
 
 	const rawReadingsValidQuery = { timeInterval: READINGS_LINE_TIME_INTERVAL };
 
+	/**
+	 * Logs in as given user and returns the generated auth token.
+	 * The User object is expected to contain a plain-text password to use for login.
+	 * @param user The user to log in as.
+	 * @returns The auth token generated by the login route.
+	 */
+	async function getTokenForUser(user) {
```
huss, line 131: It would be good to look at PR #1691 on standardizing how login/out works in tests. It is still under review at this time. Later I realized that this is moved not modified code so unsure how to deal with this comment.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -120,6 +122,57 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+		const res = await chai.request(app)
+			.post('/api/loginLogout/login')
+			.send({
+				username: user.username,
+				password: user.password
+			});
+		expect(res).to.have.status(HTTP_CODES.OK);
```
huss, line 138: If this is for use in tests and not for testing then I'm less clear on why it has expect tests. If this is to protect against coding errors in a test on the user then normally OED just lets it fail or something else.  Having said this, others may feel having the test fail is appropriate. Later I realized that this is moved not modified code so unsure how to deal with this comment.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -120,6 +122,57 @@ mocha.describe('Readings Route Parameter Validation', () => {
 
 	const rawReadingsValidQuery = { timeInterval: READINGS_LINE_TIME_INTERVAL };
```
huss, line 124: I like the idea of having standard functions for doing certain user functions. If these go forward then OED should consider an issue to do it across the code base with centralized utility functions. Other methods/uses within OED should be investigated as part of that effort.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -368,6 +378,127 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+		 * within READINGS_LINE_TIME_INTERVAL. A real unit is assigned to the meter since
+		 * a meter with no unit assigned is forced to displayable=false on insert.
+		 * @param displayable Whether the created meter should be displayable.
+		 * @returns The inserted meter's id.
+		 */
+		async function createTestMeter(displayable) {
+			meterNameCounter += 1;
+			const gps = new Point(1, 1);
+			const start = moment.utc('2020-01-01T00:00:00Z');
+			const meter = new Meter(
```
huss: src/server/util/insertData.js has a function to insert meters. Using that would be nice since it will always work even after code changes. It is used in src/server/data/automatedTestingData.js (maybe elsewhere) and I hope it can be used but changes could be proposed if needed. OED has not used them everywhere but new tests should when possible.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -368,6 +378,127 @@ mocha.describe('Readings Route Parameter Validation', () => {
...
+		}
+
+		/**
+		 * Sends a request to the given endpoint as either an unauthenticated user or a
+		 * user with the given role, then verifies the response status.
+		 * @param endpoint The full readings endpoint path (base + meter id(s)) to request.
+		 * @param role The role to log in as, or null for an unauthenticated request.
+		 * @param expectedStatus The expected HTTP response status.
+		 */
+		async function expectMeterAccessStatus(endpoint, role, expectedStatus) {
```
huss: This is different from some other OED tests that use the before/after mocha functions to login/out. PR #1691 is an example.

huss:
This is to note that given the information provided with the PR and the lack of direction from OED, I've decided to move forward with the contribution.

For anyone else, OED hopes to update the PR template to include an AI disclosure. OED is working on a more general solution but for now that will be done. OED will require:

1. Disclosure of all AI tools used including version.
2. How each AI tool was used including giving the prompts used.
3. A certification that you have the copyright to the submitted work. Given recent court cases, this will require that less than 50% of the code and documentation came from AI. Less is even better. Every AI system you use must be checked that they provide you with copyright to this work that is compatible with the OED project MPL license.
4. You have personally reviewed all code and properly tested it. Thus, you are taking personal responsibility for the quality of the submission by guaranteeing you carefully reviewed every aspect in detail.
5. If more than one person is involved in a request, then each person on the team needs to certify and provide the information above. If you provide the above information then you must have personal knowledge that the other team members' provided you with the necessary information and you are faithfully providing or indicate the answers only apply for yourself or a given list of authors (by GitHub ID).

Other items may be added soon or in the future. It is also open to change given input, new information or changes in the AI sphere. I welcome any thoughts.

huss (changes requested):
@anupamme Thank you for this contribution and the update. Per a msg on the OED Discord server developer channel, I'm sorry this took a while to consider. I wanted it to get the time it deserves.

I've made some comments to consider. I also have a big picture thought that may change the structure so I did not do extensive testing of the current code. I wondered if there was a better, more centralized & efficient way to do this. This led me to the idea of modifying the DB access functions that any route uses that can return non-displayable info (as needed) to have a new parameter that indicates if it should return only displayable meters. The default would be true so non-displayable would not be enabled unless asked for. The route then does the authentication and sets the argument appropriately. The DB can include/exclude based on this (I'm hoping the SQL would be easy but that needs to be verified). This would guarantee access at the DB level and be restrictive by default. The DB would return no data for non-displayable meters as it currently does for nonexistent meter ids. This would change the response to be okay in this case and not forbidden as you propose. I think that is okay given it should only happen on a coding error or if an outside request is made. It also means displayable ones will have data returned. I have not thought through the details of doing this so I am putting it out for comment. If OED goes down this path then the current PR solution would be replaced with that methodology. Right now I'm inclined toward that idea or something comparable as it enforces a security concern at a higher level without real cost or risk. I welcome any and all thoughts.

Please let me know if anything is unclear or you have thoughts/ideas.

anupamme:
I've resolved most of the issues you raised. Do the code changes look okay now?

src/server/sql/reading/get_readings_by_meter_id_and_date_range.sql
```diff
@@ -11,4 +13,7 @@ FROM readings
 WHERE meter_id = ${meterID}
   AND start_timestamp >= COALESCE(${startDate}, '-infinity'::TIMESTAMP)
 	AND end_timestamp <= COALESCE(${endDate}, 'infinity'::TIMESTAMP)
+	AND (${requireDisplayable} = FALSE OR EXISTS (
```
huss: The thoughts here also apply to the modified count function:

- The logic seems more complex than needed. If the meter_id is limited to one desired then I think you can check where that meter is displayable. Since it will be false in the case needed I think it can be used directly.
- This is getting close to the point where a function makes sense to me. It isn't just taking the values directly in the table but ones to control logic. A function could have a default value of true for requireDisplayable which is desirable. I do see the model for readings is enforcing this but it is always good to have the DB do it too when easy. What do you think about this? Could you create the needed functions or would you want assistance. Note functions would mean doing a migration change too.

Having said all of this, if you look at src/server/routes/maps.js it has:

```js
		if (req.hasValidAuthToken && (await isTokenAuthorized(token, User.role.ADMIN))) {
			query = Map.getAll; // only admins can see disabled maps;
		} else {
			query = Map.getDisplayable;
		}
```

So, the JS in the route is controlling what is returned based on a maps displayable. This means two SQL functions but their logic is very simple. I think OED should be consistent in how displayable is handled. I am okay with the way maps does it and using that here or a function and changing maps. What would you prefer?

src/server/models/Reading.js
```diff
@@ -166,14 +166,19 @@ class Reading {
 
 	/**
 	 * Returns the count(number of rows) for a meter
-	 * @param meterID 
-	 * @param conn 
+	 * @param meterID
```
huss: I know you didn't create the uncommented parameters but they should have descriptions. OED is trying to make all functions have JSDoc. This also applies to the other modified function. Can you do this or should someone else?

src/server/routes/readings.js
```diff
@@ -55,9 +55,14 @@ router.get('/line/count/meters/:meter_ids', optionalAuthMiddleware, async (req,
 			const conn = getConnection();
 			meterIDs = req.params.meter_ids.split(',').map(s => parseInt(s));
 			timeInterval = TimeInterval.fromString(req.query.timeInterval);
+			// Non-displayable meters only contribute to the count for authenticated requests,
```
huss: I wondered about this so looked at other places in the code that use similar logic (search for hasValidAuthToken). The led me to these conclusions:

- src/server/routes/maps.js does it correctly. OED has multiple logins now so having a valid token is not enough. It also needs the roles desired (admin here).
- src/server/routes/meters.js is wrong. It does not validate the role which probably should be admin there. This is older code that probably was not updated when needed.
- This code needs to validate the role which should be admin here.
- I tried to quickly find all uses in this way but a careful check should be made.

Now, how to address this. src/server/routes/authenticator.js for optionalAuthMiddleware calls verifyActiveTokenAndGetUser that appears to return the role but it does not set it in the req, unlike the decoded token. I was thinking it should return the role to centralize where this is done and make it easy for routes to determine the role. What do you think? This is additional work so let me know if you want someone else to do it.

src/server/test/routes/readingsParamsTest.js
```diff
@@ -368,6 +378,159 @@ mocha.describe('Readings Route Parameter Validation', () => {
 		});
 	});
 
+	mocha.describe('Non-Displayable Meter Access Control', () => {
+		let conn;
```
huss, line 385: Comment comments about what these globally scoped to this function variables are for would be nice since they implicitly pass values between functions.

huss (changes requested):
@anupamme Thank you for the updated code that addressed my comments. Overall, they are good. One thing to note. Could you not resolve any comments. OED tracks what needs to be considered again based on unresolved comments. You are welcome to reply to a comment or put a thumbs up if you think you addressed it. OED will resolve the comment once it reviews the change and it seems fine.

I appreciate your continued efforts and changes. They continue to make me think about this, esp. since it is a security concern. I made a few more comments. Some are new ideas. For those you can give your thoughts and, if you think the idea has merit, whether you want to do it or have someone else do it. I welcome your thoughts/ideas and let me know if anything is not clear.

anupamme:
I've resolved the issues. Can you pls review again?

src/server/sql/reading/get_displayable_count_by_meter_id_and_date_range.sql
```diff
@@ -0,0 +1,14 @@
...
+-- Gets number of meter readings by id and date range. Returns 0 if the meter is not
+-- displayable, the same as if the meter id does not exist.
+SELECT
+  COUNT(*)
+FROM readings
+INNER JOIN meters ON meters.id = readings.meter_id
+WHERE readings.meter_id = ${meterID}
+  AND meters.displayable = TRUE
+  AND readings.start_timestamp >= COALESCE(${startDate}, '-infinity'::TIMESTAMP)
+	AND readings.end_timestamp <= COALESCE(${endDate}, 'infinity'::TIMESTAMP);
```
huss: I changed the code to mirror the non-displayable version and to simplify. Feel free to comment.

src/server/test/routes/readingsParamsTest.js
huss: I added a comment in an upcoming commit to address this.

src/server/test/routes/readingsParamsTest.js
huss: I created issue #1723 to address this idea.

src/server/routes/readings.js
huss: I opened issue #1724 to address since it was not done here.

huss (approved):
@anupamme Thank you for the updated code. I appreciate your willingness to modify the code even when earlier requests were later changed. Review and testing found it works as desired. I commented on some unresolved comments so they are now done or made into issues. Congratulations on your first accepted contribution to OED.

PR #1705, opened 2026-08-29
Issue1687: Created default meter reading gap validation constant

### LINKED ISSUE #1687: Share default meter reading gap validation constant between client and server
## Summary

The Admin Preferences page still has an inconsistency between the frontend and backend validation for the **Default Meter Reading Gap** setting.

The frontend currently allows values beyond the backend limit, which can result in users submitting values that are only rejected after the request reaches the server.

## Current behavior

- The frontend effectively treats the maximum Reading Gap as unlimited.
- The backend enforces a maximum value of `86400`.
- This creates inconsistent validation between the client and server.

## Expected behavior

The frontend and backend should share the same validation constant so both enforce the same maximum Reading Gap value.

## Proposed solution

Following the same pattern used for the file-size validation:

- Add a shared Reading Gap validation constant to `src/common/preferencesValidationConstants`.
- Replace the backend hardcoded value (`86400`) with the shared constant.
- Update the frontend validation, input limits, and validation messages to use the same shared constant.

This keeps the validation logic consistent and prevents the frontend from accepting values that the backend will reject.

aduques:
# Description

Fixes #1687

This issue was brought up while @GoodKimchi and I were working on #1677. @GoodKimchi   has identified that the max bounds for 'Default meter reading gap' was not also properly implemented, so the fix was requested in a new issue.

This PR introduces a shared maximum validation constant (MAX_METER_READING_GAP) of 86400 for the Default meter reading gap preference. The backend previously enforced this maximum, while the frontend treated the maximum as unlimited. As a result, users could submit values greater than 86400, only for the server to reject the request. The shared constant is now used by both the frontend and backend. The frontend marks values above 86400 as invalid, displays the correct range of 0-86400 as an in-line error message, and disables the Submit button when the value is outside that range. As intended, the user should not be able to get the opportunity to make the invalid request in the first place.

This PR builds on the shared preference-validation work in #1677, and should only be merged after it.

Authors: @aduques, @GoodKimchi 

# Screenshots

Screenshot 2026-08-29 at 3 47 47 PM

Screenshot 2026-08-29 at 3 48 17 PM

## Type of change

## Checklist

## Limitations

This PR depends on #1677 and should be merged after it. If any other changes are requested on #1677, it is likely that those changes need to be applied on this pull request.

The following changes have been manually tested on localhost and passed using 'npm run check' and 'npm run test'


aduques:
To avoid revisiting this issue, I wanted to look elsewhere in the Admin Settings Page to see if there is any other inconsistencies in the frontend and backend with min/max bounds. I have found the following:

- **Default meter maximum errors**: In frontend, uses MAX_ERRORS = 75 as max bounds, but in backend, the maximum is set to 10000.
- **Default help URL**: No frontend max bounds, but backend has maxLength: 500
- **displayTitle**: Frontend max bounds is maxLength={50} and backend has maxLength: SHORT_STRING_MAX_LENGTH (256)
- **warningFileSize**: backend is missing an equivalent check to the frontend's 'Number(localAdminPref.defaultWarningFileSize) > Number(localAdminPref.defaultFileSizeLimit)'
- **Meter reading frequency**: different checks are in place (maybe intentional?); Frontend checks if frequency is valid, whereas backend only has maxLength: SHORT_STRING_MAX_LENGTH
- **Meter Min/Max Date**: Only the frontend has the check that checks related to isSameOrAfter() and isSameOrBefore(). (This difference may be intentional, but just thought to point out)
- **Default Graph Type**: frontend does not have a maxLength but backend defaultChartToRender has a maxLength: SHORT_STRING_MAX_LENGTH
- **Language:** Frontend does not have a maxLength but backend defaultLanguage does (maybe intentional because frontend is a checkbox and no text input is needed ?)
- **Default Area Unit:** same as Graph and Language, backend has maxLength check
- **Default Time Zone**: backend has a maxLength check

In most cases, the backend is missing checks that the frontend has, but this might have been intentional. From my understanding, the additional checks in the frontend would prevent the request from even being considered as it would disable the 'Submit' button. 

However, it may also be the case that the inconsistencies here were overlooked at might want to be updated. It could also be the case that these additional checks are only possible in the frontend, but I have not looked into it too much yet.

The properties that are in the invalidFuncs should definitely be updated to have consistent checks as it determines if the Submit button is enabled or not. The other properties that were mention won't affect the visibility of the Submit button, but thought it be a good idea to mention.

src/client/app/components/admin/PreferencesComponent.tsx
```diff
@@ -229,11 +232,12 @@ export default function PreferencesComponent() {
 					value={localAdminPref.defaultMeterReadingGap}
 					onChange={e => makeLocalChanges('defaultMeterReadingGap', e.target.value)}
```
huss: This onChange and the one for defaultMeterMaximumErrors have the same issue that you did not introduce but relates to your change. The value submitted to makeLocalChanges should be ``Number(e.target.value)``. Without this change one cannot submit the value without error. Could you please fix that?

huss (changes requested):
Thanks to @aduques & @GoodKimchi for another contribution. As with your other PRs, I'm sorry this took so long to review. Review and testing found it worked well except for the one comment I made. It should be simple to address and then I think everything will be fine after that (the change worked for me). Please let me know if you have thoughts/questions.

src/client/app/components/admin/PreferencesComponent.tsx
aduques: I understand that the issue was that e.target.value was not typecasted correctly with Number. I have made the following changes:

      onChange={e => makeLocalChanges('defaultMeterReadingGap', Number(e.target.value))}
      onChange={e => makeLocalChanges('defaultMeterMaximumErrors', Number(e.target.value))}

huss (approved):
Thank you for the quick update to address my comment. Review and testing found it works great. Congratulations to @aduques & @GoodKimchi  on another accepted contribution to OED.

PR #1721, opened 2026-09-17
Reuse common docker-compose variables for database build args

### LINKED ISSUE #1654: use new common env variables in docker file

PR #1554 changed the docker config files to make them more secure. The docker-compose.yml file now has ``x-common-variables`` including ``OED_DB_PASSWORD``.  As indicated by a TODO in the file, this environment variable is repeated in the database service. This requires the person to change it in two places.

It would be better to have only the common copy and reuse it. Unfortunately, attempts to do this did not work. Hopefully someone else can find a solution. This may become less of an issue for sites using a secure vault as the value would presumably come from there but that is still not implemented.

### Additional context

@huss did this so anyone interested in what they did (if they can remember :-) can reach out to them.

jacobptnguyen:
# Description

PR #1554 added a shared block of variables (like the DB password) that both the database and web containers reuse at runtime, but the database's build step still had its own hardcoded copy of that same password. That meant changing it required editing it in two places, and it was easy to make them fall out of sync. This change makes the build step reuse the same shared block instead of the hardcoded copy, so the password only lives in one spot now. It also adds one unused line to the Dockerfile to quiet a harmless build warning caused by that reuse. Author: @jacobptnguyen

Fixes #1654

## Type of change

## Checklist

## Limitations

None


huss:
@jacobptnguyen Thank you for this PR. Until the box is checked about the CLA and it is done, OED cannot consider this PR. Please let me know if you have any questions.

jacobptnguyen:
Hi @huss, I updated the comment.

containers/database/Dockerfile
```diff
@@ -10,6 +10,9 @@ FROM postgres:15.3
 
 # Sent from the build for the database in the main docker-compose.yml file.
 ARG OED_DB_PASSWORD
+# Declared (but unused) so BuildKit doesn't warn that this build arg,
```
huss: I have not use BuildKit so wanted to ask. I tried commenting it out and it did not seem to cause any issues. Does OED need this to work correctly and when? I'm trying to figure out its need.

huss (commented):
@jacobptnguyen Thank you for your first contribution to OED. I tested the code in multiple scenarios (development and production modes, restart and without DB existing) and it worked fine. I did make one comment to get some information from you. I think it is ready to be merged once I understand the change to the Dockerfile. Please let me know you thoughts.

containers/database/Dockerfile
jacobptnguyen: Hi @huss, thanks for testing this! You're right, that line isn't needed. I added it to avoid a possible "build arg not consumed" warning, but I ran builds with and without it (--no-cache) and neither showed a warning, so I removed it. The PR now only changes docker-compose.yml.

huss (approved):
@jacobptnguyen Thank you for the updated code. Review and testing found it works as desired. Congratulations on your first accepted contribution to OED.

